PR #16, opened 2024-10-11
Introduce Initialization options that are passed to ServerSession

dsp-ant:
**Depends on:**
- #15 

There is currently no way for server implementations to pass capabilities, server name and server version to the protocol. We now introduce an `InitializationOption` object that can be passed. Users of this are free to use the new `get_capabilities` to fill capabilities based on the defined handlers. This will allow patterns such as 

```
app = Server()

@app.list_prompts()
async def list_my_prompts(...):
   ...

app.run(..., InitializationOptions(..., capabilities=app.get_capabilities()))
```

This is mostly an RFC. I am not sold on the specific approach but want to get something going so that we can correctly negotiate capabilities between our Zed implementation of the protocol and Python servers.

### DIFF
--- a/mcp_python/server/__init__.py
+++ b/mcp_python/server/__init__.py
@@ -32,6 +32,7 @@ from mcp_python.types import (
     ReadResourceResult,
     Resource,
     ResourceReference,
+    ServerCapabilities,
     ServerResult,
     SetLevelRequest,
     SubscribeRequest,
@@ -39,7 +40,6 @@ from mcp_python.types import (
 )
 
 logger = logging.getLogger(__name__)
-
 
 request_ctx: contextvars.ContextVar[RequestContext] = contextvars.ContextVar(
     "request_ctx"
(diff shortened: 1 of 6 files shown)


mcp_python/server/session.py
```diff
@@ -37,6 +38,12 @@ class InitializationState(Enum):
     Initialized = 3
 
 
+@dataclass
+class InitializationOptions:
```
jspahrsummers: Can we use a Pydantic model here for consistency?

mcp_python/server/session.py
```diff
@@ -37,6 +38,12 @@ class InitializationState(Enum):
...
+    server_name: str
+    server_version: str
+    capabilities: ServerCapabilities
```
jspahrsummers: Can we infer this by default, instead of always requiring it to be passed in?

mcp_python/server/__init__.py
```diff
@@ -276,13 +276,27 @@ async def handler(req: CompleteRequest):
 
         return decorator
 
+    def get_capabilities(self) -> ServerCapabilities:
```
jspahrsummers: Can we add a test for this?

mcp_python/server/__init__.py
```diff
@@ -276,13 +276,27 @@ async def handler(req: CompleteRequest):
...
+        """Convert existing handlers to a ServerCapabilities object."""
+        capabilities: dict[str, dict[str, Any] | None] = {
+            "prompts": {} if ListPromptsRequest in self.request_handlers else None,
+            "resources": {} if ListResourcesRequest in self.request_handlers else None,
+            "tools": {} if ListPromptsRequest in self.request_handlers else None,
+            "logging": {} if SetLevelRequest in self.request_handlers else None,
+        }
```
jspahrsummers: It seems like this code would be simpler if this is just passed in directly to the `ServerCapabilities` constructor. I guess the point is to do the `None` filtering, though—I wish we could figure out how to get Pydantic to do that

mcp_python/server/session.py
dsp-ant: I rather be explicit about it for now. I'll keep it in mind however.

mcp_python/server/__init__.py
dsp-ant: Okay I talked with Claude about this. There might be ways do this with pydantic, but we both prefer the more explicit approach. I added a helper function so people can pass it via `app.create_initialization_option()`. If you feel strongly about it, i am happy to change it.

PR #19, opened 2024-10-21
Upgrade to protocol version 2024-10-07

jspahrsummers:
Necessary to use the Python SDK with the latest versions of the inspector and example servers.

### DIFF
--- a/mcp_python/server/__init__.py
+++ b/mcp_python/server/__init__.py
@@ -162,7 +162,7 @@ class Server:
             async def handler(_: Any):
                 resources = await func()
                 return ServerResult(
-                    ListResourcesResult(resources=resources, resourceTemplates=None)
+                    ListResourcesResult(resources=resources)
                 )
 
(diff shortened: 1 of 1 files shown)


mcp_python/types.py
```diff
@@ -191,7 +210,7 @@ class ServerCapabilities(BaseModel):
 class InitializeRequestParams(RequestParams):
     """Parameters for the initialize request."""
 
-    protocolVersion: Literal[1]
+    protocolVersion: str | int
```
jspahrsummers, line 213: Lenient version parsing for now.

mcp_python/types.py
```diff
@@ -211,7 +230,7 @@ class InitializeRequest(Request):
 class InitializeResult(Result):
     """After receiving an initialize request from the client, the server sends this."""
 
-    protocolVersion: Literal[1]
+    protocolVersion: str | int
```
jspahrsummers, line 233: Lenient version parsing for now.

PR #25, opened 2024-11-04
Add in-memory transport

nick-merrill:
## Goal

Support running an MCP server in the same process as the client, while preserving MCP abstractions.

## Details

1. **(core change)** Adds a new `memory` transport module that enables in-process client-server communication.
This includes:
   - `create_client_server_memory_streams()` to create bidirectional memory streams
   - `create_connected_server_and_client_session()` to establish an in-process client-server connection

3. (minor) Enhances error handling and timeout support:
   - Adds configurable read timeouts to sessions via `read_timeout_seconds` parameter
   - Improves exception handling in the server with a new `raise_exceptions` flag to control whether errors are returned to clients or raised directly
   - Ensures proper cleanup of request context tokens in error cases

4. (minor) Makes server improvements:
   - Adds built-in ping handler support


mcp_python/server/__init__.py
```diff
@@ -360,20 +374,24 @@ async def run(
...
                                 except Exception as err:
+                                    if raise_exceptions:
+                                        raise err
                                     response = ErrorData(
                                         code=0, message=str(err), data=None
                                     )
+                                finally:
+                                    # Reset the global state after we are done
+                                    if token is not None:
+                                        request_ctx.reset(token)
```
nick-merrill, line 386: I may be wrong, but the original implementation looked like it might not reset the context, which I worry could leak context.

mcp_python/server/memory.py
```diff
@@ -0,0 +1,83 @@
...
+        server_to_client_send,
+    ):
+        yield client_streams, server_streams
+
+
+@asynccontextmanager
+async def create_connected_server_and_client_session(
+    server: Server,
+) -> AsyncGenerator[ClientSession, None]:
+    """Creates a ServerSession that is connected to the `server`."""
```
jspahrsummers:
```suggestion
    """Creates a ClientSession that is connected to the `server`."""
```

mcp_python/server/memory.py
```diff
@@ -0,0 +1,83 @@
...
+                server.run,
+                server_read,
+                server_write,
+                server.create_initialization_options(),
+            )
+
+            try:
+                # Client session could be created here using client_read and
+                # client_write This would allow testing the server with a client
+                # in the same process
```
jspahrsummers: Outdated comment?

mcp_python/client/session.py
```diff
@@ -36,8 +36,15 @@ def __init__(
         self,
         read_stream: MemoryObjectReceiveStream[JSONRPCMessage | Exception],
         write_stream: MemoryObjectSendStream[JSONRPCMessage],
+        read_timeout_seconds: int | float | None = None,
```
jspahrsummers: TIOLI: Should this be a `timedelta` to avoid unit errors entirely?

mcp_python/shared/session.py
```diff
@@ -147,7 +150,22 @@ async def send_request(
...
         await self._write_stream.send(JSONRPCMessage(jsonrpc_request))
 
-        response_or_error = await response_stream_reader.receive()
+        try:
+            with anyio.fail_after(self._read_timeout_seconds):
+                response_or_error = await response_stream_reader.receive()
+        except TimeoutError:
+            raise McpError(
+                ErrorData(
+                    code=408,
```
jspahrsummers: I guess this is inspired by HTTP 408? Can we put it into a named constant somewhere?

uv.lock
```diff
@@ -163,7 +163,7 @@ wheels = [
 
 [[package]]
 name = "mcp-python"
-version = "0.3.0.dev0"
+version = "0.4.0.dev0"
```
jspahrsummers: Generally minor version bumps should be more significant for pre-1.0, IMO:
```suggestion
version = "0.3.1.dev0"
```

jspahrsummers (approved):
successkid

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "mcp-python"
-version = "0.4.0.dev"
+version = "0.3.1.dev"
```
nick-merrill: @jspahrsummers it looks like the version was already bumped to `0.4.0dev`, so should I just leave it at that version, with the expectation being that these changes would be included with the next release?  Or should I bump to `0.4.1dev`?

nick-merrill: going with `0.5.0dev` https://anthropic.slack.com/archives/C07E1154H0E/p1730847903663459?thread_ts=1730845200.906129&cid=C07E1154H0E

jakhyiakutler25:
The default `PingRequest` handler in `Server.__init__` is a welcome addition — previously you would get an unhandled-request error if a client probed liveness before manually registering a ping handler. The `read_timeout_seconds` param on `ClientSession` is also something I needed for a data pipeline I am building where tool calls can run long and were eating the default anyio timeout. One small note: the `logger.info → logger.debug` change in `create_initialization_options` is good noise reduction, but worth calling out explicitly in the PR description so reviewers do not mistake it for a behavior change.

PR #27, opened 2024-11-06
Inherit environment variables deemed safe by default

jspahrsummers:
Equivalent to https://github.com/modelcontextprotocol/typescript-sdk/pull/36.

### DIFF
--- a/mcp_python/client/stdio.py
+++ b/mcp_python/client/stdio.py
@@ -10,7 +10,6 @@ from pydantic import BaseModel, Field
 
 from mcp_python.types import JSONRPCMessage
 
-
 # Environment variables to inherit by default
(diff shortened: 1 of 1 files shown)


mcp_python/client/stdio.py
```diff
@@ -9,6 +10,36 @@
...
+    env: dict[str, str] = {}
+
+    for key in DEFAULT_INHERITED_ENV_VARS:
+        value = os.environ.get(key)
+        if value is None:
+            continue
+
+        if value.startswith("()"):
+            # Skip functions, which are a security risk
+            continue
```
dsp-ant, line 37: There are more ways of defining functions.

jspahrsummers: This is copied from the behavior of `sudo`

PR #30, opened 2024-11-06
feat: add structured capability types

dsp-ant:
Replace generic capability dictionaries with structured types for prompts,
resources, tools, and roots. This improves type safety and makes capability
features like listChanged and subscribe more explicit in the protocol.


mcp_python/types.py
```diff
@@ -184,28 +184,70 @@ class Implementation(BaseModel):
...
+    """Whether the client supports notifications for changes to the roots list."""
+    model_config = ConfigDict(extra="allow")
+
+
 class ClientCapabilities(BaseModel):
     """Capabilities a client may support."""
 
     experimental: dict[str, dict[str, Any]] | None = None
     """Experimental, non-standard capabilities that the client supports."""
     sampling: dict[str, Any] | None = None
```
jspahrsummers: Should this be an empty `BaseModel`, like `LoggingCapability`?

mcp_python/types.py
```diff
@@ -184,28 +184,70 @@ class Implementation(BaseModel):
...
+
+    listChanged: bool | None = None
+    """Whether this server supports notifications for changes to the tool list."""
     model_config = ConfigDict(extra="allow")
 
 
+class LoggingCapability(BaseModel):
+    """Capability for logging operations."""
+
+    pass
```
jspahrsummers:
```suggestion
    model_config = ConfigDict(extra="allow")
```

mcp_python/types.py
dsp-ant: Yes. Good catch

PR #48, opened 2024-11-12
Example server to expose a fetch tool

dsp-ant:


examples/README.md
```diff
@@ -0,0 +1,7 @@
+# Python SDK Examples
+
+This folders aims to provide simple examples of using the Python SDK. Please refer to the
+[example servers repository](https://github.com/modelcontextprotocol/example-servers)
```
jspahrsummers:
```suggestion
[example-servers repository](https://github.com/modelcontextprotocol/example-servers)
```

examples/README.md
```diff
@@ -0,0 +1,7 @@
...
+for real-world servers.
+
+* **simple-tool**: provides a simple fetch tool
```
jspahrsummers: Is `simple-prompt` missing from this list? Or should we just get rid of the list?

If we keep it, I think we should hyperlink to the actual folders.

examples/servers/simple-tool/mcp_simple_tool/__init__.py
```diff
@@ -0,0 +1 @@
+
```
jspahrsummers, line 1: Missing README for this server

examples/servers/simple-tool/mcp_simple_tool/server.py
```diff
@@ -0,0 +1,99 @@
...
+import click
+import httpx
+import mcp.types as types
+from mcp.server import Server
+
+
+async def fetch_website(
+    url: str,
+) -> list[types.TextContent | types.ImageContent | types.EmbeddedResource]:
+    async with httpx.AsyncClient(follow_redirects=True) as client:
```
jspahrsummers: I think we should give this a dummy user-agent so requests are traceable to anyone using this code

examples/README.md
dsp-ant: I am just gonna get rid of it. I think I'll forget to keep it in sync.

examples/servers/simple-tool/mcp_simple_tool/__init__.py
dsp-ant: good catch

PR #62, opened 2024-11-22
Handle MCPErrors separately

dsp-ant:
- **When any handler raises an McpError, we pass it as a response**
- **Correct comment**

### DIFF
--- a/src/mcp/server/__init__.py
+++ b/src/mcp/server/__init__.py
@@ -332,8 +332,8 @@ class Server:
         read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
         write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
         initialization_options: InitializationOptions,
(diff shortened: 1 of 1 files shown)


src/mcp/server/__init__.py
```diff
@@ -331,8 +332,8 @@ async def run(
         read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
         write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
         initialization_options: InitializationOptions,
-        # When True, exceptions are returned as messages to the client.
-        # When False, exceptions are raised, which will cause the server to shut down
+        # When False, exceptions are returned as messages to the client.
+        # When True, exceptions are raised, which will cause the server to shut down
```
nick-merrill, line 336: 🤦 I would love to have Claude linting for this kind of thing. Or I should just learn not to write so many comments 😆

PR #66, opened 2024-11-25
Fix first Python demo so it works without further modifications

simonw:
You can't do `async with` ... outside of an `async def` function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,11 @@ Connections between clients and servers are established through transports like 
 MCP servers follow a decorator approach to register handlers for MCP primitives like resources, prompts, and tools. The goal is to provide a simple interface for exposing capabilities to LLM clients.
 
 ```python
+# /// script
+# dependencies = [
+#   "mcp"
+# ]
+# ///
 from mcp.server import Server, NotificationOptions
 from mcp.server.models import InitializationOptions
 import mcp.server.stdio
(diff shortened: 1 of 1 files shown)


simonw:
After applying this fix you can save the code as `/tmp/demo.py` and configure it in the inspector like this:

CleanShot 2024-11-25 at 08 45 09@2x

simonw:
The inline script dependencies trick here also means you don't even have to install `mcp` separately. See: https://simonwillison.net/2024/Nov/24/follow-theirs/

README.md
```diff
@@ -76,6 +76,11 @@ Connections between clients and servers are established through transports like
 MCP servers follow a decorator approach to register handlers for MCP primitives like resources, prompts, and tools. The goal is to provide a simple interface for exposing capabilities to LLM clients.
 
 ```python
+# /// script
+# dependencies = [
+#   "mcp"
+# ]
+# ///
```
simonw, line 83: The inline script dependencies trick here also means you don't even have to install `mcp` separately. See: https://simonwillison.net/2024/Nov/24/follow-theirs/

dsp-ant (approved):
Thank you so much @simonw !

README.md
dsp-ant: TIL. Thank you

PR #69, opened 2024-11-26
Fix experimental capabilities typing

tybug:
Companion pull to https://github.com/modelcontextprotocol/specification/pull/65. I'm not as familiar with typescript, sorry, so I'll leave updating that library to someone else 😄


jspahrsummers:
Thank you for the suggested fix!

Per https://github.com/modelcontextprotocol/specification/pull/65#issuecomment-2499560063, this is actually intentional here, so I'll go ahead and close this out. Really appreciate you taking the time to write it up!

Issue #87, opened 2024-12-03
Question: version in `create_initialization_options` is the pkg_ver for "mcp"


restlessronin:
This is more in the nature of a request for clarification.

The sample code in the SDK here:

https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/servers/simple-prompt/mcp_simple_prompt/server.py

shows the use of `create_initialization_options` to set the initialization options. However this function is setting the version to 'pkg_ver("mcp")' which seems to indicate that it is the protocol version that is being indicated.

In the sqlite example, the version is manually set to 0.1.0, so it won't use the protocol version. Is this a mistake?

Is the version meant to record the version of whatever the actual server program is? or is it mean to be the protocol version.

dsp-ant:
This version is pure informational and for debugging purposes. The protocol version is exchanged differently and hardcoded into the current version of the SDK.

restlessronin:
If it's for user visible debugging, perhaps the create_initialization_options could take the version as a parameter? it could be set to the pkg_ver of the server, that's more useful in the Claude Desktop UX than the the mcp version, which is relatively static. it's a small change, but i can submit the PR if that's of interest.

restlessronin:
@dsp-ant here it is: https://github.com/modelcontextprotocol/python-sdk/pull/90

PR #90, opened 2024-12-05
feat: add version string parameter to 'create_initialization_options'

restlessronin:
delta for issue #87 

I have added a version parameter with default 'None', so existing code will work as is. I have changed docs to reflect the version string.

Ran all tests and checks. some tests are giving warnings, but they seem unrelated to this change.

### DIFF
--- a/src/mcp/server/__init__.py
+++ b/src/mcp/server/__init__.py
@@ -114,6 +114,7 @@ class Server:
 
     def create_initialization_options(
         self,
+        version: str | None = None,
         notification_options: NotificationOptions | None = None,
         experimental_capabilities: dict[str, dict[str, Any]] | None = None,
     ) -> InitializationOptions:
@@ -133,7 +134,7 @@ class Server:
 
         return InitializationOptions(
             server_name=self.name,
-            server_version=pkg_version("mcp"),
+            server_version=version if version else pkg_version("mcp"),
             capabilities=self.get_capabilities(
                 notification_options or NotificationOptions(),
                 experimental_capabilities or {},
--- a/src/mcp/server/__init__.py
+++ b/src/mcp/server/__init__.py
@@ -101,8 +101,9 @@ class NotificationOptions:
 
 
 class Server:
-    def __init__(self, name: str):
+    def __init__(self, name: str, version: str | None = None):
         self.name = name
+        self.version = version
         self.request_handlers: dict[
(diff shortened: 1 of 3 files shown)


src/mcp/server/sse.py
```diff
@@ -19,8 +19,9 @@ async def handle_sse(request):
         async with sse.connect_sse(
             request.scope, request.receive, request._send
         ) as streams:
+            # Pass user visible version string, egs. 0.1.0
             await app.run(
-                streams[0], streams[1], app.create_initialization_options()
+                streams[0], streams[1], app.create_initialization_options("0.1.0")
```
dsp-ant: passing a constant here doesn't feel right. Is there a way for someone to correctly pass the version? Maybe we need to pass the version upon creation of `Server(...)`.

Note that the way it's implemented, it's not backwards compatible and would require us to wait for SDK 2.0 to merge.

src/mcp/server/__init__.py
```diff
@@ -114,6 +114,7 @@ def __init__(self, name: str):
 
     def create_initialization_options(
         self,
+        version: str | None = None,
         notification_options: NotificationOptions | None = None,
         experimental_capabilities: dict[str, dict[str, Any]] | None = None,
```
dsp-ant: Due to the change in ordering, this is not backwards compatible.

dsp-ant (changes requested):

src/mcp/server/sse.py
restlessronin: Yes. Passing the version string when the server is created is probably better.

src/mcp/server/__init__.py
restlessronin: In that case, perhaps the server __init__ would be a better place to make the change

```
```

since it could then be the last parameter?

```
    def __init__(self, name: str, version: str = None):
```

dsp-ant: That sounds good to me.

restlessronin: Do you have a preference for

```
    def __init__(self, name: str, version: str | None = None):
```
vs
```
    def __init__(self, name: str, version: str  = ""):
```

restlessronin: I picked the first for now.

dsp-ant (approved):

restlessronin:
🥳🎉🍾

PR #106, opened 2024-12-18
feat: Integrate FastMCP

dsp-ant:
## Summary
- Integrate FastMCP by Jeremiah Lowin into the official MCP Python SDK. FastMCP provides a high-level, Pythonic interface for building MCP servers.
- We are moving fastmcp in as `mcp.server.fastmcp` and expose the `FastMCP` class directly as `mcp.server.FastMCP`.
- We move all of the existing sdk into `mcp.server.lowlevel` but expose classes still via `mcp.server` to allow for backwards compatibilty. There is a low risk of BC breakage if people rely on re-exports particular imports from the current mcp.server.
- We move in the fastmcp cli under mcp.cli and install it as an optional package `mcp[cli]`. The exeutable is now is called `mcp`

## Compatibility
The change should be BC compatible. If not than it's a bug. We will release this as a minor version update, `1.2.0`.  In version `2.0.0` of the SDK, FastMCP will likely be renamed to `Server` and exposed as such directly from `mcp.server`, while the current SDK implementation will no longer be exposed via `mcp.server`, but only via `mcp.server.lowlevel`.

## Modules Naming
I am not set on the names of the mdoules. Feel free to modify or suggest changes as you see fit.

## Test plan
- [X] Test FastMCP examples
- [X] Test FastMCP tests


Sillocan:
This personally caused a ton of confusion with fastmcp in https://github.com/jlowin/fastmcp which is not compatible. You may want to add docs to indicate that these are in no way the same package.

Issue #103, opened 2024-12-17
Bug: request_meta is not available in server.request_context


phact:
**Describe the bug**
This makes it impossible to build a server with progress notifications

**To Reproduce**
do a tool_call for example:

{
  "method": "tools/call",
  "params": {
    "name": "6a41a978-0a31-4eaf-9d0c-431a5f784bc9",
    "arguments": {
      "input_value": "3*2"
    },
    "_meta": {
      "progressToken": 0
    }
  }
}

server.request_context.meta is None

**Expected behavior**
meta should include the progressToken

It's not getting assigned here https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/__init__.py#L447

dsp-ant:
Good catch. Do you mind having a go at create a PR? Happy to review.

PR #108, opened 2024-12-21
Fix Bug: request_meta is not available in server.request_context #103

dsp-ant:
## Summary
- Fixes **Bug: request_meta is not available in server.request_context #103**
- Pydantic fields starting with underscore are hidden/private and won't get deserialized. As a result, the `_meta` field was never available.

## Test plan
- Run pytest tests/server/test_request_context.py

### DIFF
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -102,9 +102,9 @@ class SseServerTransport:
         self._read_stream_writers[session_id] = read_stream_writer
         logger.debug(f"Created new session with ID: {session_id}")
 
-        sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream(
-            0, dict[str, Any]
-        )
+        sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[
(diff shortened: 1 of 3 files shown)


src/mcp/server/__init__.py
```diff
@@ -76,7 +76,7 @@ async def main():
 import mcp.types as types
 from mcp.server.models import InitializationOptions
 from mcp.server.session import ServerSession
-from mcp.server.stdio import stdio_server as stdio_server
+from mcp.server.stdio import stdio_server as stdio_erver
```
phact: typo?

phact (commented):
aside from typo LGTM +1

src/mcp/server/__init__.py
phact:
```diff
-from mcp.server.stdio import stdio_server as stdio_erver
+from mcp.server.stdio import stdio_server as stdio_server

phact:
Hi folks, curious when this will hit a released version. Thanks!

PR #112, opened 2024-12-24
Resolving Standard Input Encoding Issues: Wrapping sys.stdin with UTF-8

blackwhite084:
This change ensures that the standard input stream (`sys.stdin`) is read with UTF-8 encoding by re-wrapping it using `io.TextIOWrapper`. This addresses potential encoding issues where the default system encoding might not be UTF-8 (e.g., GBK on some systems), leading to incorrect character interpretation.

## Motivation and Context

In certain environments, the default encoding for `sys.stdin` might be something other than UTF-8 (like GBK). When the application expects UTF-8 encoded input, this discrepancy can lead to `UnicodeDecodeError` or incorrect interpretation of characters. This change ensures that regardless of the system's default locale, the input stream is treated as UTF-8, which is a more universal and recommended encoding for modern applications. This fixes a potential bug where the application might fail or behave unexpectedly when receiving non-ASCII characters through standard input in such environments.

## How Has This Been Tested?

This change has been tested by:
- Manually testing with input containing non-ASCII characters (e.g., 中文) in an environment where the default locale is set to GBK.
- Verifying that the application correctly reads and processes these characters without encoding errors.
- Confirming that the change does not negatively impact environments where the default locale is already UTF-8.

Ideally, more comprehensive testing would involve setting up CI jobs with different locales to ensure consistent behavior across various environments.

## Breaking Changes

No, this is a non-breaking change. It addresses a potential issue with encoding and makes the application more robust. Users do not need to update their code or configurations.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The decision to re-wrap `sys.stdin` with `io.TextIOWrapper` was made to ensure consistent UTF-8 encoding without modifying the underlying file descriptor or relying on environment variables. This approach is generally considered a safe and effective way to handle encoding issues with standard input in Python. It's important to note that the input source *should* ideally be sending UTF-8 encoded data for this fix to be fully effective. This change ensures that the application *interprets* the input as UTF-8.
```

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -29,6 +29,7 @@ import mcp.types as types
 
 import io
 sys.stdin = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8')
+sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
 
 
 @asynccontextmanager
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -27,6 +27,9 @@ from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStre
 
 import mcp.types as types
 
+import io
+sys.stdin = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8')
+
 
 @asynccontextmanager
 async def stdio_server(


dsp-ant:
This looks good. Thank you. I am curious if we can add a test for this. Also do you mind rebasing it on top of the v1.1.x branch since this is a bugfix?

te-chan:
It appears that other SDKs use UTF-8 for reading and writing via STDIO.

**TS SDK**  
[Read](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/stdio.ts#L23)  
[Write](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/stdio.ts#L192)  
[Default write encoding](https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback)

**Kotlin SDK**  
[Read](https://github.com/modelcontextprotocol/kotlin-sdk/blob/main/src/main/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt#L36)  
[Default read encoding](https://kotlinlang.org/api/kotlinx-io/kotlinx-io-core/kotlinx.io/read-string.html)  
[Write](https://github.com/modelcontextprotocol/kotlin-sdk/blob/main/src/main/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StdioClientTransport.kt#L51)

> In addition I am wondering if we need to also make sure that we write from the client side.

The Python SDK client already accepts an encoding parameter and uses UTF-8 by default. [Code](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/stdio.py#L69)  
However, if we decide to fix it to UTF-8, removing that option might help avoid confusion.

blackwhite084:
This PR was converted to #130

PR #120, opened 2024-12-29
add text encoding params to STDIO client

### LINKED ISSUE #110: Client side JSON decode error when server does not use UTF-8
**Describe the bug**
If the server sends a JSON RPC message that contains invalid characters it is possible to crash the client STDOUT reader.

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 113: invalid start byte
```

There is no error handling in place for this case, and the default policy throws a traceback which can crash the client application completely.

**To Reproduce**

I wrote an example client/server that can reproduce the error 100% of the time. Please see https://github.com/SecretiveShell/MCP-SDK-JSON-DECODE-ERROR

The important part is:

```python
    return [
        types.TextContent(
            type="text",
            text=b"This is a test with a problematic character: \x92".decode("windows-1252"),
        )
    ]
```

**Expected behavior**
Ideally it should be possible to pass through a value to describe how to handle the error. The `TextRecieveStream` line that throws the error has a parameter for this already which should be exposed to the user.

```python
    :param errors: handling scheme for decoding errors (defaults to ``strict``; see the
        `codecs module documentation`_ for a comprehensive list of options)
```

**Additional context**

I tested a small modification to the code that seems to have mitigated the issue for my application:

```python
                buffer = ""
                async for chunk in TextReceiveStream(process.stdout, errors="ignore"):
                    lines = (buffer + chunk).split("\n")
                    buffer = lines.pop()
```

Since the erroneous characters are generally punctuation I noticed no impact on performance from just ignoring them.

SecretiveShell:
This PR adds text encoding parameters to the STDIO Client. This allows the SDK user to configure how the text encoder will respond to errors and set the encoding used.

## Motivation and Context
Fixes #110

This fixes an issue I was encountering in some of my programs when handling text from the internet.

## How Has This Been Tested?
All of the included unit tests are passing with this (via `uv run pytest`)
Styling was checked via `ruff check`
I modified the example reproduction in #110 with an implementation of the new features (https://github.com/SecretiveShell/MCP-SDK-JSON-DECODE-ERROR/tree/test-fix)

## Breaking Changes

None, all default values remain the same.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This was discussed in discord, and it was suggested that it would be beneficial to add extra test cases for this specific error. Currently this PR does not include this

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -77,7 +77,8 @@ class StdioServerParameters(BaseModel):
     """
     The text encoding error handler.
 
-    See https://docs.python.org/3/library/codecs.html#codec-base-classes for explanations of possible values
+    See https://docs.python.org/3/library/codecs.html#codec-base-classes for
+    explanations of possible values
     """
 
 
@@ -108,7 +109,11 @@ async def stdio_client(server: StdioServerParameters):
         try:
             async with read_stream_writer:
                 buffer = ""
-                async for chunk in TextReceiveStream(process.stdout, encoding=server.encoding, errors=server.encoding_error_handler):
+                async for chunk in TextReceiveStream(
+                    process.stdout,
+                    encoding=server.encoding,
+                    errors=server.encoding_error_handler,
+                ):
                     lines = (buffer + chunk).split("\n")
                     buffer = lines.pop()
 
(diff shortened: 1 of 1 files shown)


dsp-ant (approved):
Thank you!

PR #165, opened 2025-01-23
feat: update types to 2024-11-05 schema

dsp-ant:
## Summary
- Update types to match 2024-11-05 schema
- Add new types for request cancellation and annotations
- Add size field to Resource type for context window estimation

## Test plan
- [ ] Run type checker
- [ ] Verify all existing code compiles with new types
- [ ] Test resource size reporting

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -1,4 +1,4 @@
-from typing import Any, Generic, Literal, TypeVar
+from typing import Annotated, Any, Generic, Literal, TypeVar
 
 from pydantic import BaseModel, ConfigDict, Field, FileUrl, RootModel
 from pydantic.networks import AnyUrl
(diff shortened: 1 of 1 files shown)


src/mcp/types.py
```diff
@@ -344,6 +343,12 @@ class ListResourcesRequest(PaginatedRequest):
     params: RequestParams | None = None
 
 
+class Annotations(BaseModel):
+    audience: list[Role] | None = None
+    priority: float | None = None
```
jerome3o-anthropic: [nit/non-blocking] the spec list min/max values for this (https://github.com/modelcontextprotocol/specification/blob/363e36b3fabad8ec6253e468a74c8f7fc972e539/schema/2024-11-05/schema.json#L18)

We could add pydantic level validation for this

jerome3o-anthropic (approved):

jerome3o-anthropic (approved):
LGTM! (one non-blocking nit, happy to leave it, or to have it in a follow up)

src/mcp/types.py
dsp-ant: Great point. Totally didnt remember that this was a thing. Thank you.

PR #168, opened 2025-01-23
draft: add transport parameter to `dev` command so that server starts in correct transport mode and passes URL to the inspector

Guro:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/cli/cli.py
+++ b/src/mcp/cli/cli.py
@@ -269,11 +269,16 @@ def dev(
 
         # Run the MCP Inspector command with shell=True on Windows
         shell = sys.platform == "win32"
+
+        env = dict(os.environ.items())
+        if transport == Transport.SSE:
+            env["SSE_HOSTPORT"] = "0.0.0.0:8000"
+
         process = subprocess.run(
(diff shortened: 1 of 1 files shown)


src/mcp/cli/cli.py
```diff
@@ -250,11 +269,16 @@ def dev(
 
         # Run the MCP Inspector command with shell=True on Windows
         shell = sys.platform == "win32"
+
+        env = dict(os.environ.items()) # Convert to list of tuples for env update
+        if transport == Transport.SSE:
+            env["SSE_HOSTPORT"] = "0.0.0.0:8000"
```
dsp-ant, line 275: We shouldn't listen by default on every connection, but only localhost, which we should designate as localhost:8000 and use the system mechanism to translate it to socketaddresses, so we cover both ipv4 and ipv6

dsp-ant (changes requested):
Thank you. Makes sense, but I am requesting changes since I don't want us to listen by default on all ips. Probably we want some sort of --address or --listen parameter that can be specified optionally when transport is SSE, and that defaults to localhost:8000

Guro:
Closing for now since i don't have time to finish this, will get back to it later

PR #170, opened 2025-01-24
fix: respect resource mime type in responses

dsp-ant:
## Summary
* Fix resource MIME type handling to ensure types are properly preserved in responses
* Add comprehensive test coverage for both FastMCP and low-level server implementations

Notable this is a BC breaking change due to the new return values in `read_resource()` on FastMCP.

## Test plan
- [x] Verify MIME type preservation for text resources
- [x] Verify MIME type preservation for binary resources
- [x] Test both FastMCP and low-level server implementations
- [x] Ensure proper error handling for invalid cases

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,16 +25,31 @@ This document contains critical information about working with this codebase. Fo
    - New features require tests
    - Bug fixes require regression tests
 
-4. Version Control
-   - Commit messages: conventional format (fix:, feat:)
-   - PR scope: minimal, focused changes
-   - PR requirements: description, test plan
-   - Always include issue numbers
-   - Quote handling:
-     ```bash
-     git commit -am "\"fix: message\""
-     gh pr create --title "\"title\"" --body "\"body\""
-     ```
+- For commits fixing bugs or adding features based on user reports add:
+  ```bash
+  git commit --trailer "Reported-by:<name>"
+  ```
+  Where `<name>` is the name of the user.
+
+- For commits related to a Github issue, add
+  ```bash
+  git commit --trailer "Github-Issue:<number>"
+  ```
+- NEVER ever mention a `co-authored-by` or similar aspects. In particular, never
+  mention the tool used to create the commit message or PR.
+
+## Pull Requests
+
+- Create a detailed message of what changed. Focus on the high level description of
+  the problem it tries to solve, and how it is solved. Don't go into the specifics of the
+  code unless it adds clarity.
+
+- Always add `jerome3o-anthropic` and `jspahrsummers` as reviewer.
+
+- NEVER ever mention a `co-authored-by` or similar aspects. In particular, never
+  mention the tool used to create the commit message or PR.
+
+## Python Tools
 
 ## Code Formatting
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,7 +33,7 @@ This document contains critical information about working with this codebase. Fo
 
 - For commits related to a Github issue, add
   ```bash
-  git commit --trailer "Github-Issue:<number>"
+  git commit --trailer "Github-Issue:#<number>"
   ```
 - NEVER ever mention a `co-authored-by` or similar aspects. In particular, never
   mention the tool used to create the commit message or PR.
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -197,14 +197,16 @@ class FastMCP:
             for template in templates
         ]
 
-    async def read_resource(self, uri: AnyUrl | str) -> str | bytes:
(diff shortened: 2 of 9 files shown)


CLAUDE.md
```diff
@@ -25,16 +25,31 @@ This document contains critical information about working with this codebase. Fo
...
-     ```
+- For commits fixing bugs or adding features based on user reports add:
+  ```bash
+  git commit --trailer "Reported-by:<name>"
+  ```
+  Where `<name>` is the name of the user.
+
+- For commits related to a Github issue, add
+  ```bash
+  git commit --trailer "Github-Issue:#<number>"
```
jspahrsummers, line 36:
```suggestion
  git commit --trailer "GitHub-Issue:#<number>"
```

tests/server/fastmcp/servers/test_file_server.py
```diff
@@ -88,7 +88,9 @@ async def test_list_resources(mcp: FastMCP):
 
 @pytest.mark.anyio
 async def test_read_resource_dir(mcp: FastMCP):
-    files = await mcp.read_resource("dir://test_dir")
+    files, mime_type = await mcp.read_resource("dir://test_dir")
+    assert mime_type == "text/plain"
```
jspahrsummers: I'm not sure where the code under test is, but this MIME type is wrong. `inode/directory` should be used for directories.

jspahrsummers: Also, why isn't the URL scheme still `file://`?

src/mcp/server/lowlevel/server.py
```diff
@@ -252,32 +252,55 @@ async def handler(_: Any):
         return decorator
 
     def read_resource(self):
-        def decorator(func: Callable[[AnyUrl], Awaitable[str | bytes]]):
+        def decorator(
+            func: Callable[[AnyUrl], Awaitable[str | bytes | tuple[str | bytes, str]]],
```
jspahrsummers: I have two concerns:
1. This doesn't support multiple results (diverges from [the spec](https://github.com/modelcontextprotocol/specification/blob/363e36b3fabad8ec6253e468a74c8f7fc972e539/schema/2024-11-05/schema.ts#L362))
2. `tuple[str | bytes, str]` seems ripe for errors. Should we wrap this into a class? Or just use `ReadResourceResult` directly?

src/mcp/server/fastmcp/server.py
```diff
@@ -197,14 +197,16 @@ async def list_resource_templates(self) -> list[MCPResourceTemplate]:
             for template in templates
         ]
 
-    async def read_resource(self, uri: AnyUrl | str) -> str | bytes:
+    async def read_resource(self, uri: AnyUrl | str) -> tuple[str | bytes, str]:
```
jspahrsummers: This return type isn't explained anywhere.

src/mcp/server/fastmcp/server.py
```diff
@@ -606,7 +608,7 @@ async def report_progress(
             progress_token=progress_token, progress=progress, total=total
         )
 
-    async def read_resource(self, uri: str | AnyUrl) -> str | bytes:
+    async def read_resource(self, uri: str | AnyUrl) -> tuple[str | bytes, str]:
```
jspahrsummers: The second value in this return type isn't explained anywhere.

jspahrsummers:
I think we should [make this follow the spec more closely](https://github.com/modelcontextprotocol/python-sdk/pull/170#discussion_r1930786198) if we're going to break backwards compatibility anyway.

tests/server/fastmcp/servers/test_file_server.py
dsp-ant: This is the 'default' case where no mime type is specified. We somehow need to define a standard one and text/plain is that.

src/mcp/server/lowlevel/server.py
dsp-ant: This is already a higher level interface than the spec. I played around with returning ReadResourceResult , but it feels pretty clunky to create in the handler. I can wrap it into a different type, that makes it easier to explain.

dsp-ant: Okay, I asked claude about it and we agree that adding a new wrapper class is best here. I dont want to expose ReadResourceResult as its cumbersome to use due it being either a TextContents or a BytesContents that usually require introspection into the actual return type.

jspahrsummers (changes requested):
What about multiple result values, per my comment?

dsp-ant:
> What about multiple result values, per my comment?
 
I thought about it, but honestly, i want to converge to one return value so people dont have to match on the return value. I chose the current appraoch as I want to keep lowlevel fairly BC compatible until v2.0.0 but I am okay changing the higher level FastMCP return value around, since its a bit more faster moving for 1.3.0.

jspahrsummers (approved):

PR #180, opened 2025-01-30
Relax pydantic, pydantic-settings, and uvicorn requirements

jeremydanielfox:
## Motivation and Context
This PR relaxes version constraints for several dependencies to improve compatibility, without changing functionality.

- `pydantic`: Changed from `>=2.10.1,<3.0.0` to `>=2.7.2,<3.0.0`
- `pydantic-settings`: Changed from `>=2.6.1` to `>=2.5.2`
- `uvicorn`: Changed from `>=0.30` to `>=0.23.1`

In addition to changing the version requirements, the main change in this PR is to fix a test that previously had been failing at pydantic 2.7.2, which had prompted @dsp-ant to upgrade the pydantic version. The relevant difference pre and post 2.7.2 pydantic was just in the JSON schema representation format. This fix was to make the test handle both pre-2.7.2 and post-2.7.2 pydantic JSON schema formats. Both schemas mean the same thing.

## How Has This Been Tested?

I tested this by running the test suite with both pydantic 2.7.2, and 2.10.1, to confirm the tests passed for both. For uvicorn and pydantic-settings, I examined the change log and the usage of these libraries in the code to make sure there were no breaking changes.

## Breaking Changes

There are no breaking changes here.

## Types of changes

Relaxes dependency requirements, fixes a test that failed at an earlier dependency.

## Checklist

- [x ] My code follows the repository's style guidelines
- [x ] I have added or updated documentation as needed

### DIFF
--- a/tests/server/fastmcp/test_func_metadata.py
+++ b/tests/server/fastmcp/test_func_metadata.py
@@ -237,26 +237,23 @@ async def test_lambda_function():
 def test_complex_function_json_schema():
     """Test JSON schema generation for complex function arguments.
     
-    Note: This test accepts two equivalent JSON Schema formats for models with defaults:
-    1. Pre-pydantic 2.7.2:
+    Note: Different versions of pydantic output slightly different
+    JSON Schema formats for model fields with defaults. The format changed in 2.9.0:
+
+    1. Before 2.9.0:
+       {
+         "allOf": [{"$ref": "#/$defs/Model"}],
+         "default": {}
+       }
+       
+    2. Since 2.9.0:
        {
          "$ref": "#/$defs/Model",
          "default": {}
        }
        
-    2. Pydantic 2.7.2+:
-       {
-         "allOf": [
-           {
-             "$ref": "#/$defs/Model"
-           }
-         ],
-         "default": {}
-       }
-       
-    Both formats are valid JSON Schema and represent the same validation rules.
-    The newer format using allOf is more correct according to the JSON Schema spec
-    as it properly composes the reference with additional properties.
+    Both formats are valid and functionally equivalent. This test accepts either format
+    to ensure compatibility across our supported pydantic versions.
     
     This change in format does not affect runtime behavior since:
     1. Both schemas validate the same way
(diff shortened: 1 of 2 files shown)


jerome3o-anthropic (approved):
This seems okay to me - for reference I also dug into the change logs for uvicorn and didn't see any issues.

tests/server/fastmcp/test_func_metadata.py
```diff
@@ -235,8 +235,48 @@ async def check_call(args):
...
-    assert meta.arg_model.model_json_schema() == {
+    actual_schema = meta.arg_model.model_json_schema()
+    
+    # Create a copy of the actual schema to normalize
+    normalized_schema = actual_schema.copy()
+    
+    # Normalize the my_model_a_with_default field to handle both pydantic formats
+    if 'allOf' in actual_schema['properties']['my_model_a_with_default']:
+        normalized_schema['properties']['my_model_a_with_default'] = {
+            '$ref': '#/$defs/SomeInputModelA',
```
jerome3o-anthropic, line 272: Seems odd to me that you've relaxed down to 2.7.2, but in order to get this test passing you cast it to the "Pre-pydantic 2.7.2" version.

I suspect either your comment is wrong, or I'm misunderstanding something

jerome3o-anthropic (changes requested):
Sorry - I just realized that there is an inconsistency in my understanding here, is the comment correct? / what version of pydantic were you using when you needed the test change to make it pass?

tests/server/fastmcp/test_func_metadata.py
jeremydanielfox: Sorry, the comment was wrong. I researched exactly when this change was made and updated the comment to reflect it.

The change happened in Pydantic 2.9.0. [Here are the release notes for that change](https://docs.pydantic.dev/latest/changelog/#v290-2024-09-05). You can [see the PR here](https://github.com/pydantic/pydantic/pull/10029/files) that made the change.

jeremydanielfox: To confirm that this works, I ran the test with pydantic 2.7.2 (the new version we are relaxing to) and 2.10.1 (the previously required lower bound) and confirmed the test passes with both.

jerome3o-anthropic (approved):

Issue #141, opened 2025-01-07
Resource Templates of FastMCP server not visible/usable with python-sdk based client


wanderingnature:
**Describe the bug**
Implement a client using the python-sdk.
Implement a FastMCP server exposing resource templates
python-sdk based client cannot display resource templates exposed by FastMCP server...

After fixing https://github.com/modelcontextprotocol/python-sdk/issues/129 this bug is exposed when using a client based on python-sdk and a server based on FastMCP.

**To Reproduce**
### Implement simple MCP sever exposing a resource template:

```python
#easy_echo.py
"""
FastMCP Echo Server
Implements resource template
"""

from mcp.server.fastmcp import FastMCP

\# Create server
mcp = FastMCP("Easy Echo Server")

@mcp.resource("template://resource/{text}/{more_text}", description="resource template")
def resource_template(text: str, more_text: str) -> str:
    """function name becomes the name"""
    return f"text: {text}, more_text {more_text}"

```

## Implement a simple MCP client to connect to the server and return the resource template

```python
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client

\# Create server parameters for stdio connection
server_params = StdioServerParameters(
    command=".venv/bin/uv",  # Executable
    args=["run", "--with", "mcp", "mcp", "run", "examples/fastmcp/easy_echo.py"],  # Properly split args
    env={"PATH": ".venv/bin"}  # Use dictionary for env variables
)


async def run():
    async with stdio_client(server_params) as (read, write):
        async with ClientSession(read, write) as session:
            # Initialize the connection
            await session.initialize()


            \# List available resource templates
            resource_templates = await session.list_resource_templates()
            print(f"{resource_templates=}") # raw list_resource_templates() result
            \# Extract relevant data
            templates = resource_templates.resourceTemplates
            print("\nAvailable Resource Templates:")
            for template in templates:
                print(f"🔹 Name: {template.name}")
                print(f"   🔗 URI: {template.uriTemplate}")
                print(f"   📜 Description: {template.description}")
                print()

            \# Read a resource template
            resource = await session.read_resource("template://resource/hello_world/more_text")
            print(f"{resource=}") # raw read_resource() results
            print(f"   🔗 URI: {resource.contents[0].uri}")
            print(f"   📜 Result: {resource.contents[0].text}")
            print()


if __name__ == "__main__":
    import asyncio
    asyncio.run(run())
```
### run the client test:
```bash
[...]

PR #186, opened 2025-02-03
fix: standardize resource response format

dsp-ant:
Github-Issue:#141

### DIFF
--- a/tests/issues/test_141_resource_templates.py
+++ b/tests/issues/test_141_resource_templates.py
@@ -1,0 +1,118 @@
+import pytest
+from pydantic import AnyUrl
+
+from mcp.server.fastmcp import FastMCP
+from mcp.shared.memory import (
+    create_connected_server_and_client_session as client_session,
+)
+from mcp.types import (
+    ListResourceTemplatesResult,
+    TextResourceContents,
+)
+
+
+@pytest.mark.anyio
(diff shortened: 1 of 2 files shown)


tests/issues/test_141_resource_templates.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Test case 1: Template with multiple parameters
+    @mcp.resource("resource://users/{user_id}/posts/{post_id}")
+    def get_user_post(user_id: str, post_id: str) -> str:
+        return f"Post {post_id} by user {user_id}"
+
+    # Test case 2: Template with optional parameter (should fail)
+    with pytest.raises(ValueError, match="Mismatch between URI parameters"):
+
+        @mcp.resource("resource://users/{user_id}/profile")
```
jerome3o-anthropic, line 27: [q] is this because it should be optional in the template also (i.e. as a query param)?

jerome3o-anthropic (approved):
lgtm! (one non blocking qn)

tests/issues/test_141_resource_templates.py
dsp-ant: Yes, it should match the numbedr of arguments. If you dont want an argument it shouldnt be specified in the resource.

PR #197, opened 2025-02-09
Enhance resource template to support values with slashes (Fix #159)

### LINKED ISSUE #159: Resource template parameters do not support values with slashes
**Describe the bug**

## Overview
Resource template values containing path separators (slashes) do not work.

### Simple Static Resources
```python
@mcp.resource("resource://my-resource")
def get_data() -> dict[str, Any]:
    return {
        "contents": [{
            "uri": "resource://my-resource",
            "mimeType": "text/plain",
            "text": "Hello, world!"
        }]
    }
```
This works reliably and shows up in the Inspector UI.

### Parameter 
```python
@mcp.resource("resource://{value}")
def get_content(value: str) -> dict[str, Any]:
    # This works for: resource://
    # But fails for: resourse://value/with/slashes
    ...
```

**To Reproduce**

### 1. Template Pattern with Regex
```python
@mcp.resource("resource://{value:.*}")
def get_content(value: str) -> dict[str, Any]:
    # Error: Mismatch between URI parameters set() and function parameters {'permalink'}
```

### 2. Generic Catch-all Pattern
```python
@mcp.resource("memory://*")
def get_content(uri: str) -> dict[str, Any]:
    # Error: Mismatch between URI parameters set() and function parameters {'uri'}
```

### 3. Path Parameter Approach
```python
@mcp.resource("resource://content/{path}")
def get_content(path: str) -> dict[str, Any]:
    # Works for simple paths but fails with slashes in {path}
```

### 4. Base Path with Query Parameters
```python
@mcp.resource("resource://content")
def get_content(path: str = None) -> dict[str, Any]:
    # Error: Mismatch between URI parameters set() and function parameters {'path'}
```

**Expected behavior**

Resource Templates with slashes in the value should be passed to the decorated function.

g-votte:
This PR enhances resource templates so that they can handle values containing slashes.

### Motivation
This PR fixes the bug reported in #159 where the following example raises an error for paths containing slashes such as `resource://content/path/with/slashes`.

```python

mcp = FastMCP("My App")

@mcp.resource("resource://content/{path}")
def get_content(path: str) -> str:
    return f"Path: {path}"
```

Currently, this raises `mcp.shared.exceptions.McpError: Unknown resource: resource://content/path/with/slashes.`

Screenshot from MCP inspector:
image

### Analysis
The root cause is in `ResourceTemplate.matches()` which explicitly excludes slashes from the regex matching:
https://github.com/modelcontextprotocol/python-sdk/blob/f10665db4c2f676da1131617ad67715952258712/src/mcp/server/fastmcp/resources/templates.py#L58

### Content of Fix
- Modified the regex pattern to allow slashes in parameter values
- Added test cases in `tests/server/fastmcp/resources/test_resource_template.py`
- Removed test case in `tests/issues/test_141_resource_templates.py` which expected errors for slash-containing values

## How Has This Been Tested?

- Unit testing: Added a test case in `tests/server/fastmcp/resources/test_resource_template.py`
- Integration testing: Verified the example from the Motivation section works correctly in MCP inspector

image

## Breaking Changes

While this change primarily enhances functionality without breaking existing behavior, there might be edge cases where applications rely on the current error behavior for paths containing slashes. However, such cases are expected to be rare.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

While #159 marks this as a "bug," there is an existing test case suggesting this might have been intentional behavior:
https://github.com/modelcontextprotocol/python-sdk/blob/f10665db4c2f676da1131617ad67715952258712/tests/issues/test_141_resource_templates.py#L61-L64

I would appreciate maintainers' input on the intended behavior. If rejecting slash-containing values is the desired design, please close this PR.

### DIFF
--- a/src/mcp/server/fastmcp/resources/templates.py
+++ b/src/mcp/server/fastmcp/resources/templates.py
@@ -55,7 +55,7 @@ class ResourceTemplate(BaseModel):
     def matches(self, uri: str) -> dict[str, Any] | None:
         """Check if URI matches template and extract parameters."""
         # Convert template to regex pattern
-        pattern = self.uri_template.replace("{", "(?P<").replace("}", ">[^/]+)")
+        pattern = self.uri_template.replace("{", "(?P<").replace("}", ">.+)")
         match = re.match(f"^{pattern}$", uri)
         if match:
             return match.groupdict()
--- a/tests/issues/test_141_resource_templates.py
+++ b/tests/issues/test_141_resource_templates.py
@@ -58,11 +58,6 @@ async def test_resource_template_edge_cases():
     with pytest.raises(ValueError, match="Unknown resource"):
         await mcp.read_resource("resource://users/123/posts")  # Missing post_id
 
-    with pytest.raises(ValueError, match="Unknown resource"):
-        await mcp.read_resource(
-            "resource://users/123/posts/456/extra"
-        )  # Extra path component
-
 
 @pytest.mark.anyio
 async def test_resource_template_client_interaction():
--- a/tests/server/fastmcp/resources/test_resource_template.py
+++ b/tests/server/fastmcp/resources/test_resource_template.py
@@ -46,6 +46,25 @@ class TestResourceTemplate:
         assert template.matches("test://foo") is None
         assert template.matches("other://foo/123") is None
 
+    def test_template_matches_slashes_in_value(self):
+        """Test matching URIs against a template, the value containing slashes."""
+
+        def my_func(path: str) -> dict:
+            return {"path": path}
+
+        template = ResourceTemplate.from_function(
+            fn=my_func,
+            uri_template="test://content/{path}",
+            name="test",
+        )
+
+        # Valid match
+        params = template.matches("test://content/path/with/slashes")
+        assert params == {"path": "path/with/slashes"}
+
+        # No match
+        assert template.matches("test://content/") is None
+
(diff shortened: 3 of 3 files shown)


ICeZer0:
I can't get any resource params to show up in my MCP Inspector, not sure if its related to the above.

This works fine.
```python
@mcp.resource("config://app")
def get_config() -> str:
    """Static configuration data"""
    return "App configuration here"
```

Once I add an input param it does not show up as a listed resource
```python
@mcp.resource("users://{hello}/message")
def get_config(hello: str) -> str:
    """dynamic user message"""
    return "Message here {hello}"
```

dsp-ant:
> ## Additional context
> While #159 marks this as a "bug," there is an existing test case suggesting this might have been intentional behavior:
> 
> https://github.com/modelcontextprotocol/python-sdk/blob/f10665db4c2f676da1131617ad67715952258712/tests/issues/test_141_resource_templates.py#L61-L64
> 
> I would appreciate maintainers' input on the intended behavior. If rejecting slash-containing values is the desired design, please close this PR.

Thank you for working on this and the detailed description. After more time thinking about this, I think the current behaviour is desired. If we want to support paths we should go down the route of FastAPI which offers `{var:path}` convertor.

I'll close this PR, and sorry for you spending the time without it being merged. I think the detailed description made it clear how to best think about this for me. Thank you

PR #198, opened 2025-02-10
Force stdin/stdout encoding to UTF-8

rnortman:
The character encoding of the stdin/stdout streams in Python is platform- dependent. On Windows it will be something weird, like CP437 or CP1252, depending on the locale. This change ensures that no matter the platform, UTF-8 is used.

## Breaking Changes

None known.

## Types of changes
- [ ] Documentation update

## Checklist
- [x] New and existing tests pass locally (but see below)

Existing tests fail on Windows because of backslashes vs forward slashes in path names before my change. The same tests still fail on Windows with my change. Tests pass on Linux.

## Additional context

This enforces that stdin/stdout are UTF-8-encoded independent of platform; otherwise it's a grab-bag what you get on Windows. The JSON spec requires UTF-8, UTF-16, or UTF-32 encoding, and UTF-8 is by far the most common, so this seems safe.

A proper "fix" would be for the MCP specification to either specify an exact encoding or have a handshake where encoding is established before any other messages are exchanged. (I would lean toward requiring UTF-8.)

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -20,6 +20,7 @@ Example usage:
 
 import sys
 from contextlib import asynccontextmanager
+from io import TextIOWrapper
 
 import anyio
 import anyio.lowlevel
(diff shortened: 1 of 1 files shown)


dsp-ant:
Thank you. I think this is a pretty valid concern. In fact, I believe we should start enforcing UTF-8 in the spec. Here is my draft: https://github.com/modelcontextprotocol/specification/pull/162. 

I will wait for the discussion on the draft pr before merging this.

dsp-ant (approved):

Issue #201, opened 2025-02-10
Client hangs when ctx.info() logging with pure Python client


sheffler:
**Describe the bug**
When a tool includes client-side logging with, the client seems to hang.
     The following tool works ok with Claude Desktop and Inspector, but does not complete with a Python client.

```
@mcp.tool()
async def simple_tool_with_logging(x:float, y:float, ctx:Context) -> str:
    await ctx.info(f"Processing Simple Tool")
    logger.debug("IN SIMPLE_TOOL")
    await ctx.report_progress(1, 2)
    return x*y
```

**To Reproduce**
Will check in a PR with test and proposed fix.

**Expected behavior**
I would expect the tool to run to completion.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: OSX

**Smartphone (please complete the following information):**

**Additional context**

samuelcolvin:
I've spent most of my day held up by this bug, see https://github.com/pydantic/pydantic-ai/pull/1140#issuecomment-2727454409

On `main` the problem is at

https://github.com/modelcontextprotocol/python-sdk/blob/c8978681f6b4903163325237dae2cb43ab2fbdf5/src/mcp/shared/session.py#L344-L346

If I change these lines to

```py
if not isinstance(notification.root, LoggingMessageNotification):
    await self._incoming_message_stream_writer.send(
        notification
    )
```
Everything works correctly, obviously that's not a proper solution, but it demonstrates the problem.

In case it helps, to reproduce this problem, run

```bash
npx @pydantic/mcp-run-python sse
```

(see https://github.com/pydantic/pydantic-ai/pull/1140 for context)

Then run

```py
import logging
from logging import basicConfig

from mcp import ClientSession, StdioServerParameters
from mcp.client.sse import sse_client
from mcp.client.stdio import stdio_client
from devtools import debug

import logfire


logfire.configure()
logfire.instrument_httpx(capture_all=True)
basicConfig(handlers=[logfire.LogfireLoggingHandler()], level=logging.DEBUG)
# disable httpx logs
logging.getLogger('httpcore.http11').setLevel(logging.CRITICAL)
logging.getLogger('httpx').setLevel(logging.CRITICAL)
logging.getLogger('urllib3.connectionpool').setLevel(logging.CRITICAL)

code = """
import numpy
a = numpy.array([1, 2, 3])
print(a)
a
"""


async def call_tools(session: ClientSession):
    await session.initialize()
[...]

PR #202, opened 2025-02-10
End to end logging

sheffler:
This addresses Issue #201 and provides a test and a fix.

## Motivation and Context
An MCP server may include logging messages (notifications/message) by using the ctx.{info,debug}() call.
Currently it seems that Claude Desktop and Inspector operate properly when an MCP server sends these
notifications, even if those clients do not do anything with them.

In the process of writing a Pure-Python client using the example given in simple-chatbot as a starting point,
if logging notifications are sent, they cause the client to hang.

There is an included fix in the shared/session.py file.  I believe the acknowledgement of the received notification is causing a deadlock.   Try running the test without the fix to observe server hanging.

The goal is to simply log the received notifications in the Client, and you can see that in the NotificationLoggingClientSession

## How Has This Been Tested?
The included test illustrates the pattern.

## Breaking Changes
possibly

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -324,9 +324,9 @@ class BaseSession(
                                 await self._in_flight[cancelled_id].cancel()
                         else:
                             await self._received_notification(notification)
-                            await self._incoming_message_stream_writer.send(
-                                notification
-                            )
+#                            await self._incoming_message_stream_writer.send(
+#                                notification
+#                            )
                     except Exception as e:
                         # For other validation errors, log and continue
                         logging.warning(
--- a/tests/mcp_stdio_client.py
+++ b/tests/mcp_stdio_client.py
@@ -1,0 +1,136 @@
+from mcp import ClientSession, ListToolsResult, StdioServerParameters
+from mcp.client.stdio import stdio_client
+from mcp.types import CallToolResult
+from mcp import Tool as MCPTool
+
+from contextlib import AsyncExitStack
+from typing import Any
+import asyncio
+
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+
+class NotificationLoggingClientSession(ClientSession):
+
+    def __init__(self, read_stream, write_stream):
+        print(f"NOTIFICATION LOGGING CLIENT SESSION")
+        super().__init__(read_stream, write_stream)
+
+    # override base session to log incoming notifications
+    async def _received_notification(self, notification):
+        print(f"NOTIFICATION:{notification}")
+        print(f"NOTIFICATION-END")
+        
+    async def send_progress_notification(self, progress_token, progress, total):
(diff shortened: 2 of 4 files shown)


src/mcp/shared/session.py
```diff
@@ -324,9 +324,9 @@ async def _receive_loop(self) -> None:
                                 await self._in_flight[cancelled_id].cancel()
                         else:
                             await self._received_notification(notification)
-                            await self._incoming_message_stream_writer.send(
-                                notification
-                            )
+#                            await self._incoming_message_stream_writer.send(
+#                                notification
+#                            )
```
dsp-ant, line 329: That seems to be a debugging left over?

tests/mcp_stdio_client.py
```diff
@@ -0,0 +1,136 @@
+from mcp import ClientSession, ListToolsResult, StdioServerParameters
+from mcp.client.stdio import stdio_client
+from mcp.types import CallToolResult
+from mcp import Tool as MCPTool
+
+from contextlib import AsyncExitStack
+from typing import Any
+import asyncio
```
dsp-ant, line 8: We generally don't use asyncio but anyio.

dsp-ant (changes requested):
In gneeral I am not too keen on end to end tests that involve multiple files but are disguised as unit tests. We should see if we can make this test work within one file by using memory streams, etc.

src/mcp/shared/session.py
sheffler: These are the lines that are present in shared/session that I believe cause the deadlock.  If you uncomment the lines and run the end-to-end test, it will not complete (it will hang).  If you comment the lines out, the test runs to completion.

sheffler:
Agreed, a more directed unit test might be better going forward after the issue is understood.   The scenario in this end-to-end test illustrates the deadlock with enough context to reproduce the issue described in #201 when run.

dsp-ant:
> Agreed, a more directed unit test might be better going forward after the issue is understood. The scenario in this end-to-end test illustrates the deadlock with enough context to reproduce the issue described in #201 when run.

That makes sense. Can we move this over to DRAFT status until it's ready to be properly reviewed?

sheffler:
This PR should probably be viewed as a bug-recreation rather than code that should be merged.  The end-to-end test does a good job at illustrating the deadlock, I believe.

dsp-ant:
> This PR should probably be viewed as a bug-recreation rather than code that should be merged. The end-to-end test does a good job at illustrating the deadlock, I believe.

That makes sense. Thank you. Let me take a look!

dsp-ant:
Closing this out in favor of #325. Thank you so much for writing this up, it was a great help in finding and fixing the issue.

PR #203, opened 2025-02-11
feat: add lifespan context manager support

dsp-ant:
## Summary
- Add comprehensive lifespan support to both Server and FastMCP classes
- Enable type-safe resource initialization and cleanup
- Pass context from startup to request handlers
- Add support for async startup/shutdown operations

## Test plan
- Run pytest to verify tests for both Server and FastMCP
- Test startup, shutdown, and context passing functionality
- Verify compatibility with existing code

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,9 @@ The [Model Context Protocol (MCP)](https://modelcontextprotocol.io) lets you bui
 The FastMCP server is your core interface to the MCP protocol. It handles connection management, protocol compliance, and message routing:
 
 ```python
+# Add lifespan support for startup/shutdown with strong typing
+from dataclasses import dataclass
+from typing import AsyncIterator
 from mcp.server.fastmcp import FastMCP
 
 # Create a named server
@@ -136,14 +139,17 @@ mcp = FastMCP("My App")
 # Specify dependencies for deployment and development
 mcp = FastMCP("My App", dependencies=["pandas", "numpy"])
 
-# Add lifespan support for startup/shutdown
+@dataclass
+class AppContext:
+    db: Database  # Replace with your actual DB type
+
 @asynccontextmanager
-async def app_lifespan(server: FastMCP) -> AsyncIterator[dict]:
-    """Manage application lifecycle"""
+async def app_lifespan(server: FastMCP) -> AsyncIterator[AppContext]:
+    """Manage application lifecycle with type-safe context"""
     try:
         # Initialize on startup
         await db.connect()
-        yield {"db": db}
+        yield AppContext(db=db)
     finally:
         # Cleanup on shutdown
         await db.disconnect()
@@ -151,7 +157,7 @@ async def app_lifespan(server: FastMCP) -> AsyncIterator[dict]:
 # Pass lifespan to server
 mcp = FastMCP("My App", lifespan=app_lifespan)
 
-# Access lifespan context in tools
+# Access type-safe lifespan context in tools
 @mcp.tool()
 def query_db(ctx: Context) -> str:
     """Tool that uses initialized resources"""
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -135,6 +135,28 @@ mcp = FastMCP("My App")
 
 # Specify dependencies for deployment and development
 mcp = FastMCP("My App", dependencies=["pandas", "numpy"])
+
+# Add lifespan support for startup/shutdown
+@asynccontextmanager
+async def app_lifespan(server: FastMCP) -> AsyncIterator[dict]:
```
jspahrsummers: Is it possible for this to be typed more usefully than `dict`? So, e.g., accessing `request_context.lifespan_context` ends up well typed later on.

README.md
```diff
@@ -334,7 +356,39 @@ def query_data(sql: str) -> str:
...
+async def query_db(name: str, arguments: dict) -> list:
+    ctx = server.request_context
+    db = ctx.lifespan_context["db"]
+    return await db.query(arguments["query"])
+```
+
+The lifespan API provides:
+- A way to initialize resources when the server starts and clean them up when it stops
+- Access to initialized resources through the request context in handlers
+- Support for both low-level Server and FastMCP classes
```
jspahrsummers: This bullet isn't very useful in this section of the README

README.md
jspahrsummers: Reading the code below, it seems like it is actually strongly typed, I guess? Using `dict` in the example might be a bit misleading. Maybe we could at least use a `TypedDict`?

src/mcp/server/fastmcp/server.py
```diff
@@ -85,13 +98,36 @@ class Settings(BaseSettings):
         description="List of dependencies to install in the server environment",
     )
 
+    lifespan: (
+        Callable[["FastMCP"], AbstractAsyncContextManager[LifespanResultT]] | None
+    ) = Field(None, description="Lifespan contexte manager")
```
jspahrsummers:
```suggestion
    ) = Field(None, description="Lifespan context manager")
```

src/mcp/server/lowlevel/server.py
```diff
@@ -446,9 +470,14 @@ async def run(
...
-            async with ServerSession(
-                read_stream, write_stream, initialization_options
-            ) as session:
+            from contextlib import AsyncExitStack
+
+            async with AsyncExitStack() as stack:
+                lifespan_context = await stack.enter_async_context(self.lifespan(self))
+                session = await stack.enter_async_context(
+                    ServerSession(read_stream, write_stream, initialization_options)
+                )
```
jspahrsummers, line 479: Is this different from doing `async with lifespan_context, session`?

src/mcp/server/lowlevel/server.py
```diff
@@ -460,21 +489,28 @@ async def run(
...
+                            warning.category.__name__,
+                            warning.message,
                         )
 
     async def _handle_request(
         self,
         message: RequestResponder,
         req: Any,
         session: ServerSession,
+        lifespan_context: object,
```
jspahrsummers:
```suggestion
        lifespan_context: LifespanResultT,
```

src/mcp/shared/context.py
```diff
@@ -12,3 +12,4 @@ class RequestContext(Generic[SessionT]):
     request_id: RequestId
     meta: RequestParams.Meta | None
     session: SessionT
+    lifespan_context: Any
```
jspahrsummers: Doesn't this destroy the strong typing?

jspahrsummers (approved):
Mostly questions/confusions from me, overall looks good!

src/mcp/server/lowlevel/server.py
dsp-ant: The AsyncExitStack() version provides more granular control over multiple context managers. If an error occurs during the second enter_async_context, the first one will still be properly cleaned up.

Chatted more with Claude about it, we came to the. conclusion that AsyncExitStack is better here.

Issue #213, opened 2025-02-17
default environment variables are not set if you provide your own variables in StdioServerParameters


peterj:
**Describe the bug**
I am using this through Autogen, but it seems like the root cause is in the `stdio.py` from mcp. Here's an example of how an mcp tools gets defined:

```python
brave_params = StdioServerParams(
  command="npx",
  args=["@modelcontextprotocol/server-brave-search"],
  env={"BRAVE_API_KEY": "someapikey"},
)

tools = await mcp_server_tools(brave_params)
```

When executing this, I get the error saying that `npx` is not found. If I remove the API key environment variable, there's no error. It looks like when the user provides their own env, the default environment variables are not set (path being one of them) and that's why I am hitting the error. 

Here's the specific line:
https://github.com/modelcontextprotocol/python-sdk/blob/5480f67cf956961d4c507a55587cea4a816b78ec/src/mcp/client/stdio.py#L100-L104

I'd suggest when the env is provided, the variables get merged together with the default variables.

dsp-ant:
The intended use here is similar to how Python's subprocess and similar API's work. By default there is a form of inheritance of env variables, but if you specify `env=`, then the callers is responsible for merging. This allows callers to define if they want a merged env or an isolated one.

The workable approach in your example is to merge get_default_environment() or os.environ with your keys and then pass them to `StdioServerParams`

I'll close this for now since the implemented behaviour is intended.

PR #214, opened 2025-02-17
merge custom set env variables with default ones

peterj:
## Motivation and Context
Fixes https://github.com/modelcontextprotocol/python-sdk/issues/213 -- the default env variables are not set if custom ones are provided.

## How Has This Been Tested?
unit test

## Breaking Changes
There is a behavior change where extra environment variables (default ones) will be set if users are providing their own.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -99,7 +99,7 @@ async def stdio_client(server: StdioServerParameters):
 
     process = await anyio.open_process(
         [server.command, *server.args],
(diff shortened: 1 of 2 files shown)


dsp-ant:
The intended use here is similar to how Python's subprocess and similar API's work. By default there is a form of inheritance of env variables, but if you specify `env=`, then the callers is responsible for merging. This allows callers to define if they want a merged env or an isolated one.

I'll close this for now since the implemented behaviour is intended.

PR #216, opened 2025-02-19
enable http headers in sse in the context

abdullahkady:
## Motivation and Context
In the server, it seems like a common case to need to access the raw HTTP headers (think a reverse proxy header, or some custom analytics headers).

Example use case, where we have two services that can possibly talk to the MCP server. The servers identify themselves through the `X-Client` HTTP header.

```python

mcp = FastMCP("test_w_headers", port=3001, debug=True)


@mcp.tool()
def add_sales_lead(lead: str, ctx: Context) -> str:
    """
	Adds a new sales lead. Tracks which app was the originator of 
	this request based on the X-Client custom header
	"""
	sales.add_lead(lead, tracking_attrs={"src": ctx.headers.get("X-Client")}
    return "Lead added successfully"

```

## How Has This Been Tested?
They were just tested manually locally

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
I'm not very familiar with the code. The `try/except` TODO should be done with proper type checking, but I just wanted to make it work to show the concept first.

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -526,6 +526,11 @@ class Server(Generic[LifespanResultT]):
             logger.debug(f"Dispatching request of type {type(req).__name__}")
 
             token = None
+            headers = {}
+            try:
+                headers = message.request.root.headers  # type: ignore
+            except Exception:
+                pass
             try:
                 # Set our global state that can be retrieved via
                 # app.get_request_context()
@@ -535,6 +540,7 @@ class Server(Generic[LifespanResultT]):
                         message.request_meta,
                         session,
                         lifespan_context,
+                        headers,
                     )
                 )
                 response = await handler(req)
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -528,6 +528,7 @@ class Server(Generic[LifespanResultT]):
             token = None
             headers = {}
             try:
+                # TODO: This try/catch and ignoring the type is wrong.
                 headers = message.request.root.headers  # type: ignore
             except Exception:
                 pass
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -118,6 +118,7 @@ class JSONRPCRequest(Request):
     jsonrpc: Literal["2.0"]
     id: RequestId
     params: dict[str, Any] | None = None
+    headers: dict[str, str] | None = None
 
 
 class JSONRPCNotification(Notification):
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -160,7 +160,8 @@ class SseServerTransport:
         logger.debug(f"Received JSON: {json}")
 
         try:
-            message = types.JSONRPCMessage.model_validate(json)
+            message_with_headers = {**json, "headers": dict(request.headers)}
+            message = types.JSONRPCMessage.model_validate(message_with_headers)
             logger.debug(f"Validated client message: {message}")
         except ValidationError as err:
             logger.error(f"Failed to parse message: {err}")
--- a/src/mcp/shared/context.py
+++ b/src/mcp/shared/context.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Generic, TypeVar
 
 from mcp.shared.session import BaseSession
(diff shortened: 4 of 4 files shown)


src/mcp/types.py
```diff
@@ -118,6 +118,7 @@ class JSONRPCRequest(Request):
     jsonrpc: Literal["2.0"]
     id: RequestId
     params: dict[str, Any] | None = None
+    headers: dict[str, str] | None = None
```
dsp-ant: I don't think we would want to add headers to the types. The types are a direct translation of the [schema.ts](https://github.com/modelcontextprotocol/specification/blob/main/schema/2024-11-05/schema.ts) in the specification. Adding headers here would not be compatible with JSON-RPC in it's most used form.

src/mcp/server/lowlevel/server.py
```diff
@@ -526,6 +526,12 @@ async def _handle_request(
             logger.debug(f"Dispatching request of type {type(req).__name__}")
 
             token = None
+            headers = {}
+            try:
+                # TODO: This try/catch and ignoring the type is wrong.
+                headers = message.request.root.headers  # type: ignore
+            except Exception:
+                pass
```
dsp-ant: I assume we most likely want headers to be extracted from the transport protocol itself. This means providing ways within a session context to access to original request headers for the HTTP Post in the case of SSE. For STDIO this would be empty. If we ever come to the conclusion that headers are absolutely necessary, I think we likely would move to a STDIO transport description that is akin to the header-based JSON-RPC that LSP is doing, which is of the form:
```
Some-Header: Some-Value
Some-Header: Some-Value
{"jsonrpc": "2.0", ...}
```

dsp-ant (changes requested):
I think the general idea to give access to headers in the context is great.  I think we should focus on exposing the headers from the transport protocol (HTTP Headers in HTTP+SSE transport and _no_ headers in STDIO) inside the context.

src/mcp/types.py
abdullahkady: @dsp-ant Thank you! That's fair. I was trying to initially do the change such that only the `sse` file is the one with changes, but here's my problem, and a suggestion for an approach to fix it. Please correct me if my understanding of the code is not complete.

- The HTTP request is only available in transport layer (`sse.py`)
- The only way to pass stuff down from the transport layer to the server implementation is going to be [through the streams](https://github.com/abdullahkady/python-sdk/blob/144db31f492677e966582a33a0cbfd1dcd50a407/src/mcp/server/sse.py#L176), which have `JSONRPCMessage` type
- We don't want to extend/change the `JSONRPCMessage` type because it is part of the schema (and honestly it makes sense, since it only applies to certain transports)

To solve this, here are my other 2 suggestions: 
- Inject the metadata (one of which is optional headers) in the [`params`](https://github.com/abdullahkady/python-sdk/blob/b6e45768f2a9764a4f91e4c1d19d7cb0272daeca/src/mcp/types.py#L121) of the `JSONRPCMessage`
- Change the streams so that they now hold a new type (a container type with the `JSONRPCMessage + meta`), where the `meta` can be anything that is transport-specific?

Let me know what you think.

dsp-ant:
Thanks for doing that. I really like the idea and just working through an alternative option. Will have something soon.

src/mcp/types.py
abdullahkady: @dsp-ant hey, any chance you have given this some thought and have any guidance for the approach? I would love to update the PR

gdoctor:
Adding my input here in case anyone has the same use case as me.

I am building an MCP Client. I need to be able to set different headers on a per-request basis. In my use-case, it's to differentiate users interacting with my client, among other things. Right now, you can only set headers when creating a new ClientSession, but I want to maintain semi-persistent ClientSession(s). This PR seems to address this in it's current state, while also making it easier to consume headers on the Server side. I understand there is some discussion on the best way to do this still, but the need is there for all us MCP Client creators.

vivek-athina:
@dsp-ant  Any update on this ? It will be really useful if this can be achieved out of the box.

ZAKLLL:
support this pr ,and it will be useful in sse/streamhttp server,like env config in stdio ,i think http metadata is also necessary in  sse mcp ，otherwise , the client cannot determine custom config expect request param ,it not suitable , of course ,it should be declared clearly in docs that header metadata just available in sse/streamhttp mcp server!

sidey:
I don't think this PR is a good solution.since context was implemented by contextvar,if some one wants to retrieve a head from the request.he can do it in a similar way.
eg. set a contextvar in a starlette middleware and then retrieve it anywhere.
I wonder what's your solution about this feature @dsp-ant

PR #239, opened 2025-03-03
refactor: reorganize message handling for better type safety and clarity

dsp-ant:
## Summary
- Reorganizes message handling in the codebase by introducing MessageFrame (previously ParsedMessage) and improving type definitions
- Enhances code organization by moving stream type definitions to more appropriate locations

## Test plan
- Ensure all existing tests pass
- Verify type checking completes without errors
- Check that renamed components maintain backward compatibility

GitHub-Issue:#201


tests/client/test_session.py
```diff
@@ -22,10 +22,10 @@
 @pytest.mark.anyio
 async def test_client_session_initialize():
     client_to_server_send, client_to_server_receive = anyio.create_memory_object_stream[
-        JSONRPCMessage
+        MessageFrame[None]
```
bhosmer-ant: Should it be `MessageFrame[NoneType]`?

src/mcp/types.py
```diff
@@ -180,6 +180,21 @@ class JSONRPCMessage(
     pass
 
 
+RawT = TypeVar("RawT")
+
+
+class MessageFrame(BaseModel, Generic[RawT]):
```
bhosmer-ant, line 186: Don't know how fastidious we are about this in general, but fwiw the other classes here are docstringed up pretty thoroughly (and the point of `raw` in particular might be worth noting here)

src/mcp/types.py
```diff
@@ -180,6 +180,21 @@ class JSONRPCMessage(
...
+    root: JSONRPCMessage
+    raw: RawT | None = None
```
bhosmer-ant, line 208: Looks like all the uses of `MessageFrame` pass `raw=None` explicitly - personally seems nice to require that but maybe there are ergo concerns I'm not thinking of? (Minor side benefit to removing the default would be avoiding the usual union-vs-sum confusion when `RawT = T | None`...)

bhosmer-ant (dismissed):
Couple minor observations inline but LGTM!

Kludex:
This broke a lot types 😢

src/mcp/shared/session.py
```diff
@@ -22,12 +22,18 @@
...
     RequestParams,
     ServerNotification,
     ServerRequest,
     ServerResult,
 )
 
+ReadStream = MemoryObjectReceiveStream[MessageFrame | Exception]
+ReadStreamWriter = MemoryObjectSendStream[MessageFrame | Exception]
+WriteStream = MemoryObjectSendStream[MessageFrame]
+WriteStreamReader = MemoryObjectReceiveStream[MessageFrame]
```
Kludex, line 35: You can't do this if `MessageFrame` is a generic. It means that all those new types have unkown parameters to the generic type.

src/mcp/types.py
```diff
@@ -180,6 +180,49 @@ class JSONRPCMessage(
...
+        This method allows for consistent serialization of the parsed message.
+        """
+        return self.message.model_dump(*args, **kwargs)
+
+    def model_dump_json(self, *args, **kwargs):
+        """
+        Dumps the model to a JSON string, delegating to the root JSONRPCMessage.
+        This method provides a convenient way to serialize the parsed message to JSON.
+        """
+        return self.message.model_dump_json(*args, **kwargs)
```
Kludex, line 223: This is not type safe. You are hiding the parameters with args and kwargs.

Kludex:
What's this `ParsedMessage` you mention?

src/mcp/client/session.py
```diff
@@ -59,8 +58,8 @@ class ClientSession(
 ):
     def __init__(
         self,
-        read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
-        write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
+        read_stream: ReadStream,
+        write_stream: WriteStream,
```
Kludex, line 62: This is not really the end type, since `ReadStream` and `WriteStream` are generics, something is missing here.

I believe it's easier to understand the code if we don't hide that those streams are anyio objects. Makes it easier to go around the code.

PR #245, opened 2025-03-06
Use proper generic for Context

Kludex:
Now you can use the Context with a proper lifespan state.

I suggest running a type checker in this project. I can see a lot of typing issues using Pyright on strict mode.

```py
from typing import AsyncIterator

from mcp.server.fastmcp.server import Context, FastMCP


@dataclass
class State:
    count: int


@asynccontextmanager
async def lifespan(mcp: FastMCP) -> AsyncIterator[State]:
    yield State(count=0)


mcp = FastMCP()


@mcp.tool()
def my_tool(x: int, ctx: Context[State]) -> int:
    count = ctx.request_context.lifespan_context.count
    return count


mcp.run()
```

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -37,7 +37,7 @@ from mcp.server.lowlevel.server import (
 from mcp.server.session import ServerSession
 from mcp.server.sse import SseServerTransport
 from mcp.server.stdio import stdio_server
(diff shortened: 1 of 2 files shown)


dsp-ant (approved):

dsp-ant:
yup, we know that it's not checking in strict mode. It's on the very long to do list. Would love to get some help !

PR #249, opened 2025-03-08
Add hot reload functionality

TheRobotCarlson:


Kludex:
Did you test this PR? I don't think this will reload, given the app is passed as an ASGI application object. It only works when it's passed as string.

TheRobotCarlson:
@Kludex I had monkey-patched that in my tests. I'll close out this PR until I can come back to address that.

Edison-A-N:
Consider trying this: https://gist.github.com/Edison-A-N/f4fa399c456525f5d821aa0ff9932ee2

PR #255, opened 2025-03-10
Add support for serverside oauth

praboud-ant:
## Motivation and Context
Implements https://spec.modelcontextprotocol.io/specification/draft/basic/authorization/ from the specification.

## How Has This Been Tested?
I've added some integration tests as part of this branch. I've also implemented a new MCP server (outside of this repo), and integrated it against the new parts of the the SDK, to play around with the interface a bit and see how it feels.

Tested with inspector: 

image

## Breaking Changes
No - this is strictly additive.

## Types of changes

- [ ] Documentation update

## Checklist


Kludex:
If review wanted, happy to review it.

src/mcp/server/auth/json_response.py
```diff
@@ -0,0 +1,6 @@
+from typing import Any
+from starlette.responses import JSONResponse
+
+class PydanticJSONResponse(JSONResponse):
+    def render(self, content: Any) -> bytes:
+        return content.model_dump_json(exclude_none=True).encode("utf-8")
```
jerome3o-anthropic: What is this for?

dsp-ant:
@Kludex its work in progress as far as I undersatnd but feel free to take a review pass.

src/mcp/server/auth/errors.py
```diff
@@ -0,0 +1,135 @@
+"""
+OAuth error classes for MCP authorization.
+
+Corresponds to TypeScript file: src/server/auth/errors.ts
+"""
+
+from typing import Dict, Optional, Any
```
dsp-ant: We don't use Dict or Optional, but the Python 3.10 equivalents of `dict`, `X | None`, etc. I will not mention it on every lien i spot, but I think it's basically all over the place.

src/mcp/server/auth/errors.py
```diff
@@ -0,0 +1,135 @@
...
+    
+    Corresponds to OAuthError in src/server/auth/errors.ts
+    """
+    error_code: str = "server_error"
+    
+    def __init__(self, message: str):
+        super().__init__(message)
+        self.message = message
+        
+    def to_response_object(self) -> Dict[str, str]:
```
dsp-ant: This has a specific shape and shrouldn't be generically typed, but typed to the exact shape we expect to be returned either with a Pydantic model or a TypedDict.

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,155 @@
...
+
+def validate_scope(requested_scope: str | None, client: OAuthClientInformationFull) -> list[str] | None:
+    if requested_scope is None:
+        return None
+    requested_scopes = requested_scope.split(" ")
+    allowed_scopes = [] if client.scope is None else client.scope.split(" ")
+    for scope in requested_scopes:
+        if scope not in allowed_scopes:
+            raise InvalidRequestError(f"Client was not registered with scope {scope}")
+    return requested_scopes
```
dsp-ant: This feels like a method on OAuthClientInformationFull or the underlying scope model, given we exclusively operate on client.scope here.

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,155 @@
...
+            raise InvalidRequestError(
+                f"Redirect URI '{auth_request.redirect_uri}' not registered for client"
+            )
+        return auth_request.redirect_uri
+    elif len(client.redirect_uris) == 1:
+        return client.redirect_uris[0]
+    else:
+        raise InvalidRequestError("redirect_uri must be specified when client has multiple registered URIs")
+
+def create_authorization_handler(provider: OAuthServerProvider) -> Callable:
```
dsp-ant: This pattern feels like a 1:1 translation of TypeScript code that doesn't match the rest of the SDK.

We generally use more OOP approaches in the SDK, so an alternative could be:

```
class Handler(protocol):
  async def handle(self, request: Request) -> Respnse:
    ...

@dataclass
class AuthorizationHandler(Handler):
  provider: OAuthServerProvider
  def handle(request: Request) -> Response:
    ...
```

Either way we go, we at leat want to strongly type the return values instead of just returning a Callable.

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,155 @@
...
+        scopes = validate_scope(auth_request.scope, client)
+        
+        auth_params = AuthorizationParams(
+            state=auth_request.state,
+            scopes=scopes,
+            code_challenge=auth_request.code_challenge,
+            redirect_uri=redirect_uri,
+        )
+        
+        response = RedirectResponse(url="", status_code=302, headers={"Cache-Control": "no-store"})
```
dsp-ant: We should probalby not redirect without a URI?

src/mcp/server/auth/handlers/register.py
```diff
@@ -0,0 +1,109 @@
...
+                client_uri=client_metadata.client_uri,
+                logo_uri=client_metadata.logo_uri,
+                scope=client_metadata.scope,
+                contacts=client_metadata.contacts,
+                tos_uri=client_metadata.tos_uri,
+                policy_uri=client_metadata.policy_uri,
+                jwks_uri=client_metadata.jwks_uri,
+                jwks=client_metadata.jwks,
+                software_id=client_metadata.software_id,
+                software_version=client_metadata.software_version,
```
dsp-ant: This can probably be written as 
```
          return OAuthClientInformationFull(
              # Auto-generated fields
              client_id=client_id,
              client_id_issued_at=client_id_issued_at,
              client_secret=client_secret,
              client_secret_expires_at=client_secret_expires_at,
              # Inherit all metadata fields
              **client_metadata.model_dump()
          )
```

dsp-ant: This can probably be written as

src/mcp/server/auth/middleware/bearer_auth.py
```diff
@@ -0,0 +1,100 @@
...
+    async def authenticate(self, conn: HTTPConnection):
+
+        if "Authorization" not in conn.headers:
+            return None
+            
+        auth_header = conn.headers["Authorization"]
+        if not auth_header.startswith("Bearer "):
+            return None
+            
+        token = auth_header[7:]  # Remove "Bearer " prefix
```
dsp-ant: I think this can be simplified by
```suggestion
        auth_header = conn.headers.get("Authorization")
        if not auth_header or not auth_header.startswith("Bearer "):
            return None

        token = auth_header[7:]  # Remove "Bearer " prefix
```

src/mcp/server/auth/middleware/bearer_auth.py
```diff
@@ -0,0 +1,100 @@
...
+        
+        try:
+            # Validate the token with the provider
+            auth_info = await self.provider.verify_access_token(token)
+
+            if auth_info.expires_at and auth_info.expires_at < int(time.time()):
```
dsp-ant: We should use datetime comparisions and not rely on timestamps.

src/mcp/shared/auth.py
```diff
@@ -0,0 +1,123 @@
...
+class OAuthClientInformation(BaseModel):
+    """
+    RFC 7591 OAuth 2.0 Dynamic Client Registration client information.
+
+    Corresponds to OAuthClientInformationSchema in src/shared/auth.ts
+    """
+    client_id: str
+    client_secret: Optional[str] = None
+    client_id_issued_at: Optional[int] = None
+    client_secret_expires_at: Optional[int] = None
```
dsp-ant: Probably should be parsed as a Timedelta

src/mcp/shared/auth.py
```diff
@@ -0,0 +1,123 @@
...
+
+class OAuthTokens(BaseModel):
+    """
+    OAuth 2.1 token response.
+
+    Corresponds to OAuthTokensSchema in src/shared/auth.ts
+    """
+    access_token: str
+    token_type: str
+    expires_in: Optional[int] = None
```
dsp-ant: Also here, probably watn to parse as a timedelta.

src/mcp/shared/auth.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class OAuthTokens(BaseModel):
+    """
+    OAuth 2.1 token response.
+
+    Corresponds to OAuthTokensSchema in src/shared/auth.ts
+    """
+    access_token: str
+    token_type: str
```
dsp-ant: Should we enumerate the potential types here to ensure pydnatic validation?

tests/server/fastmcp/auth/streaming_asgi_transport.py
```diff
@@ -0,0 +1,197 @@
...
+
+import anyio
+import anyio.streams.memory
+from anyio.abc import TaskStatus
+import httpx
+from httpx._transports.asgi import ASGIResponseStream
+from httpx._transports.base import AsyncBaseTransport
+from httpx._models import Request, Response
+from httpx._types import AsyncByteStream
+import asyncio
```
dsp-ant: we don't use asyncio for testing.

dsp-ant (changes requested):

src/mcp/server/auth/errors.py
jerome3o-anthropic: done in 84bfe75e140b47265ec0ed317250bef1afc1ea2a

src/mcp/server/auth/errors.py
jerome3o-anthropic: done in 84bfe75e140b47265ec0ed317250bef1afc1ea2a

jerome3o-anthropic (commented):
Working through the changes now

src/mcp/server/auth/router.py
```diff
@@ -0,0 +1,176 @@
...
+            AUTHORIZATION_PATH, 
+            endpoint=create_authorization_handler(provider),
+            methods=["GET", "POST"]
+        ),
+        Route(
+            TOKEN_PATH, 
+            endpoint=create_token_handler(provider, client_authenticator),
+            methods=["POST"]
+        )
+    ])
```
Kludex: I think those can be customized, can't they?

src/mcp/server/auth/handlers/authorize.py
jerome3o-anthropic: [opinion] after hacking around in here, I actually think we should be thinking about `OAuthClientInformationFull` as a type similar to the ones in mcp.types, and it would be odd to add methods there

src/mcp/server/auth/handlers/authorize.py
jerome3o-anthropic: I've implemented proper type signatures for the functions. I think this is okay for now

src/mcp/server/auth/handlers/register.py
jerome3o-anthropic: done

src/mcp/shared/auth.py
praboud-ant: A `timedelta`, or a `datetime`? Don't we want an absolute time here?

src/mcp/server/auth/handlers/register.py
praboud-ant: The reason I did it this way is that it helps pyright out - pyright can't otherwise figure out what fields are provided through `client_metadata`, and can't warn you if one is missing.

src/mcp/server/auth/json_response.py
praboud-ant: Lemme add a comment - this uses Pydantic's JSON serialization, since the stock `json` doesn't know how to serialize eg: `AnyHttpUrl`

Kludex:
The auth part on the client should probably be implemented using the httpx.Auth class.

praboud-ant:
@dsp-ant this is ready for another look

src/mcp/server/auth/middleware/auth_context.py
```diff
@@ -0,0 +1,57 @@
...
+    Get the access token from the current context.
+
+    Returns:
+        The access token if an authenticated user is available, None otherwise.
+    """
+    auth_user = auth_context_var.get()
+    return auth_user.access_token if auth_user else None
+
+
+class AuthContextMiddleware(BaseHTTPMiddleware):
```
Kludex: If I don't see this PR anymore, please don't use BaseHTTPMiddleware. Use a pure ASGI middleware.

Kludex (dismissed):

pyproject.toml
```diff
@@ -101,8 +100,12 @@ mcp = { workspace = true }
...
     "error",
+    # this is a long-standing issue with fastmcp, which is just now being exercised by tests
+    "ignore:Unclosed:ResourceWarning",
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
+    # this is a problem in starlette
+    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
```
Kludex: You are using an old version of Starlette. We should bump it.

pyproject.toml
```diff
@@ -101,8 +100,12 @@ mcp = { workspace = true }
 xfail_strict = true
 filterwarnings = [
     "error",
+    # this is a long-standing issue with fastmcp, which is just now being exercised by tests
+    "ignore:Unclosed:ResourceWarning",
```
Kludex: Please remove this. I can help find out the unclosed resources. Is it only reproducible in this PR?

src/mcp/server/streaming_asgi_transport.py
```diff
@@ -0,0 +1,205 @@
...
+import anyio
+import anyio.abc
+import anyio.streams.memory
+from httpx._models import Request, Response
+from httpx._transports.base import AsyncBaseTransport
+from httpx._types import AsyncByteStream
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+
+class StreamingASGITransport(AsyncBaseTransport):
```
Kludex, line 23: Is this a copy of https://github.com/encode/httpx/pull/3059 ? I didn't check the code. Can you please reference so we can remove the workaround when merged?

pyproject.toml
praboud-ant: This is reproducible in the PR only because the PR adds integration tests to previously untested flows. I poked at the test failure for a bit, without much success - my preference would be to leave this as-is, and fix the resource leak in tests in a separate PR.

pyproject.toml
praboud-ant: Similar to the above - this is a problem predating this PR. It's only popping up here because we previously didn't have integration tests which exercised this flow in the tests. This PR is already pretty big, and I'd prefer to address the starlette bump in a followup.

src/mcp/server/streaming_asgi_transport.py
praboud-ant: It's not - any similarity between the two is accidental convergence. This is a good flag, though - we should try swapping these out when https://github.com/encode/httpx/pull/3059 merges.

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,258 @@
...
+                client_id = best_effort_extract_string("client_id", params)
+                client = client_id and await self.provider.get_client(client_id)
+            if redirect_uri is None and client:
+                # make last-ditch effort to load the redirect uri
+                if params is not None and "redirect_uri" not in params:
+                    raw_redirect_uri = None
+                else:
+                    raw_redirect_uri = AnyHttpUrlModel.model_validate(
+                        best_effort_extract_string("redirect_uri", params)
+                    ).root
```
dsp-ant: This validation can also fail. How do we deal with this?

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,258 @@
...
+                try:
+                    redirect_uri = client.validate_redirect_uri(raw_redirect_uri)
+                except (ValidationError, InvalidRedirectUriError):
+                    pass
```
dsp-ant: Is this intended that we don't throw an error in this case? If so we probably want to make a comment as to why. Future me will likely wonder about this.

regismesquita:
Hi @praboud-ant with the release of the new specification is this PR still progressing? I am looking to develop some stuff that relies on the new specs with oauth and it is unclear to me the timeline of dropping the SSE and supporting the new stream with OAUTH.

src/mcp/server/fastmcp/server.py
```diff
@@ -455,6 +483,29 @@ def decorator(func: AnyFunction) -> AnyFunction:
 
         return decorator
 
+    def custom_route(
+        self,
+        path: str,
+        methods: list[str],
+        name: str | None = None,
+        include_in_schema: bool = True,
+    ):
```
bhosmer-ant, line 508: I think it's worth adding a docstring comment here, to match the other FastMCP decorators. FWIW here's Claude's suggestion (I also added it to [this commit](https://github.com/modelcontextprotocol/python-sdk/commit/b685fcc71266496045d5ecae3540fe17bcdf5a61)):
```suggestion
    ):
        """Decorator to register a custom HTTP route on the FastMCP server.

        Allows adding arbitrary HTTP endpoints outside the standard MCP protocol,
        which can be useful for OAuth callbacks, health checks, or admin APIs.
        The handler function must be an async function that accepts a Starlette
        Request and returns a Response.

        Args:
            path: URL path for the route (e.g., "/oauth/callback")
            methods: List of HTTP methods to support (e.g., ["GET", "POST"])
            name: Optional name for the route (used for URL reversing)
            include_in_schema: Whether to include in OpenAPI schema, defaults to True

        Example:
            @server.custom_route("/health", methods=["GET"])
            async def health_check(request: Request) -> Response:
                return JSONResponse({"status": "ok"})
        """        
```

src/mcp/server/auth/handlers/authorize.py
```diff
@@ -0,0 +1,258 @@
...
+            # Catch-all for unexpected errors
+            logger.exception(
+                "Unexpected error in authorization_handler", exc_info=validation_error
+            )
+            return await error_response(
+                error="server_error", error_description="An unexpected error occurred"
+            )
+
+
+def create_error_redirect(
```
bhosmer-ant: looks like this is unused - maybe superceded by `construct_redirect_uri`?

src/mcp/server/auth/provider.py
```diff
@@ -0,0 +1,289 @@
...
+        +--------+     +------------+     +-------------------+
+                            |   ^                  |
+        +------------+      |   |                  |
+        |            |      |   |    Redirect      |
+        |redirect_uri|<-----+   +------------------+
+        |            |
+        +------------+
+
+        Implementations will need to define another handler on the MCP server return
+        flow to perform the second redirect, and generates and stores an authorization
```
bhosmer-ant: nit: "generate and store"

src/mcp/server/fastmcp/server.py
```diff
@@ -465,11 +516,8 @@ async def run_stdio_async(self) -> None:
             )
 
     async def run_sse_async(self) -> None:
-        """Run the server using SSE transport."""
```
bhosmer-ant: nit: maybe restore this comment?

pyproject.toml
```diff
@@ -110,8 +110,12 @@ mcp = { workspace = true }
...
```
bhosmer-ant: Fixed the problem and removed the warning suppression in [this commit](https://github.com/modelcontextprotocol/python-sdk/commit/b685fcc71266496045d5ecae3540fe17bcdf5a61)

pyproject.toml
```diff
@@ -110,8 +110,12 @@ mcp = { workspace = true }
...
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
+    # this is a problem in starlette
+    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
```
bhosmer-ant: Upgraded starlette and removed this warning suppression in [this commit](https://github.com/modelcontextprotocol/python-sdk/commit/b685fcc71266496045d5ecae3540fe17bcdf5a61)

bhosmer-ant (commented):
LGTM (Disclaimer: I'm new to OAuth - studied the PR pretty hard, but I'm definitely relying on more knowledgeable folks to spot anything subtle.) 

Left a few comments inline, and put some small changes (fixes for the resource leaks, starlette upgrade, docstring for `create_route`) in [this commit](https://github.com/modelcontextprotocol/python-sdk/commit/b685fcc71266496045d5ecae3540fe17bcdf5a61). 

I'll hold off approving pending those changes (if you agree) and a look from @jerome3o-anthropic, but thank you for this substantial piece of work!

dsp-ant (commented):
Just adding some ocmments that i had pending.

praboud-ant:
Unfortunately, it looks like https://github.com/modelcontextprotocol/modelcontextprotocol/pull/284 is going to significantly change how auth works for MCP servers, and I don't think ~any of the stuff I've implemented here is going to make sense in the new spec. I don't think it makes sense to merge this / I think merging this is actively harmful - anybody that starts to integrate with this auth implementation is going to have to rip out that integration & redo it once the new spec lands. I'm going to close this PR; if it turns out that any of the code here is salvageable, we can get it out of git history.

praboud-ant:
After discussing this further offline, it probably is worth moving forwards with this PR, since the new spec still contemplates MCP servers optionally implementing an OAuth authorization server: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/284/files#diff-961c36b26a515b33b6cda2907f55dc53504999d9775e090f789d310617f45efaR290 - it's just discouraged if there exists an existing authorization server which supports RFC8707.

src/mcp/server/auth/handlers/authorize.py
praboud-ant: Good catch - lemme move the model_validate inside the try/catch.

praboud-ant:
@dsp-ant / @bhosmer-ant - this is ready for another look

praboud-ant:
@dsp-ant / @bhosmer-ant bumping this again

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "httpx>=0.27",
     "httpx-sse>=0.4",
     "pydantic>=2.7.2,<3.0.0",
-    "starlette>=0.27",
+    "starlette>=0.46",
```
ihrpr: this should not be upgraded

src/mcp/server/sse.py
```diff
@@ -120,17 +120,15 @@ async def sse_writer():
                         }
                     )
 
-        # Ensure all streams are properly closed
-        async with read_stream, write_stream, read_stream_writer, sse_stream_reader:
```
bhosmer-ant: ooc was the outer `with` unnecessary or was it actively messing things up?

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -993,165 +987,6 @@ async def test_client_registration_invalid_grant_type(
         )
 
 
-class TestFastMCPWithAuth:
```
bhosmer-ant: was this the test that leaked and made the other ones hang?

bhosmer-ant (approved):
New changes LGTM (and things look relatively undisturbed from the last time it was  looked over in some depth) - based on that, tests passing, and your e2e check passing, I'm good with landing it. (would ofc benefit from @jerome3o-anthropic also looking before you push the button, mod available time)

tests/server/fastmcp/auth/test_auth_integration.py
ihrpr: yes, basically this :
```
async with aconnect_sse(
                test_client, "GET", "/sse", headers={"Authorization": authorization}
            ) as event_source:
```

src/mcp/server/sse.py
ihrpr: I just reverted all the changes to see transport, returning response was not needed at all.

The fun part for the top `async with` was that it was masking "bad test" and sse tests were just hanging. When reverted all the sse changes notices that some other test randomly fails with " ResourceWarning: Unclosed <MemoryObjectReceiveStream at 106c0dd20>"

ihrpr: I think it's a nice update, I would prefer to have it in a separate PR and test separately. There are a bunch of open PRs addressing different parts of memory leaks, maybe we can just bundle all of them together

README.md
```diff
@@ -309,6 +309,33 @@ async def long_task(files: list[str], ctx: Context) -> str:
...
 
+### Authentication
+
+Authentication can be used by servers that want to expose tools accessing protected resources.
+
+`mcp.server.auth` implements an OAuth 2.0 server interface, which servers can use by
+providing an implementation of the `OAuthServerProvider` protocol.
+
+```
+mcp = FastMCP("My App",
```
jerome3o-anthropic, line 320: [non blocking] wonder if its worth noting that this wont scale beyond one instance

jerome3o-anthropic (approved):
Thanks for pushing this over the line @ihrpr - it's awesome to see it

I've looked over this, not with a fine-toothed comb but for the general shapes, and it LGTM!

ihrpr (approved):

README.md
ihrpr: I think that's something we can add to the example

PR #257, opened 2025-03-11
Drop `AbstractAsyncContextManager` for proper type hint to `BaseSession`

Kludex:
You can see that `session` assumes `ClientSession` type now.

```py


async def main():
    async with ClientSession() as session:
        await session.initialize()
```

The real problem is that `AbstractAsyncContextManager` is a generic, so it expects some type as an argument. We can't really add `BaseClient` as that type, because the return type from the `ClientSession` `__aexit__` is not `BaseClient`, it's actually `ClientSession`.

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -8,6 +8,7 @@ import anyio.lowlevel
 import httpx
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from pydantic import BaseModel
+from typing_extensions import Self
 
 from mcp.shared.exceptions import McpError
 from mcp.types import (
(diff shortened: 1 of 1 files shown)


src/mcp/shared/session.py
```diff
@@ -8,6 +8,7 @@
 import httpx
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from pydantic import BaseModel
+from typing_extensions import Self
```
dsp-ant, line 11: Isn't this in typing as well?

dsp-ant (approved):

src/mcp/shared/session.py
dsp-ant: my bad, it's in typing in 3.11 not in 3.10

PR #266, opened 2025-03-12
Fix more types

Kludex:


Kludex:
I've fixed those types in https://github.com/modelcontextprotocol/python-sdk/pull/271

PR #267, opened 2025-03-12
Close unclosed resources in the whole project

Kludex:
I'll be using this branch on PydanticAI. I can't add tests otherwise.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,5 +96,5 @@ filterwarnings = [
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes from read_resource is deprecated. Use Iterable[ReadResourceContents] instead.:mcp.shared.exceptions.McpError"
+    "ignore:Returning str or bytes from read_resource is deprecated. Use Iterable[ReadResourceContents] instead.:DeprecationWarning"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,5 +94,6 @@ xfail_strict = true
 filterwarnings = [
     "error",
     # This should be fixed on Uvicorn's side.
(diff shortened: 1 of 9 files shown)


src/mcp/shared/session.py
```diff
@@ -245,11 +245,14 @@ async def send_request(
...
-            raise McpError(response_or_error.error)
         else:
-            return result_type.model_validate(response_or_error.result)
+            if isinstance(response_or_error, JSONRPCError):
+                raise McpError(response_or_error.error)
+            else:
+                return result_type.model_validate(response_or_error.result)
+        finally:
+            await response_stream.aclose()
+            await response_stream_reader.aclose()
```
jerome3o-anthropic: Seems like we'd also like to delete the entry in `self._response_streams` also

jerome3o-anthropic (approved):
This looks good, thank you

src/mcp/shared/session.py
jerome3o-anthropic: optional - I can get this later if needed

jerome3o-anthropic:
Looks like there are a bunch of check failures. I'm not sure if these are even related though

dsp-ant:
This leaves one pytest error which might be legit

pyproject.toml
```diff
@@ -67,6 +67,9 @@ packages = ["src/mcp"]
 include = ["src/mcp", "tests"]
 venvPath = "."
 venv = ".venv"
+strict = [
+    "src/mcp/server/fastmcp/tools/base.py",
```
Kludex, line 71: More should be added here.

pyproject.toml
```diff
@@ -85,3 +88,10 @@ members = ["examples/servers/*"]
 
 [tool.uv.sources]
 mcp = { workspace = true }
+
+# TODO(Marcelo): This should be enabled!!! There are a lot of resource warnings.
```
Kludex:
```suggestion
```

dsp-ant (approved):

PR #270, opened 2025-03-12
Add support for Linux configuration path in get_claude_config_path

emsi:
## Motivation and Context
There's a community version of Claude desktop at:
https://github.com/emsi/claude-desktop (originally forked from https://github.com/aaddrick/claude-desktop-debian).

Unfortunately current version of MCP SDK does not support Linux yet there is nothing in the docs that indicate that and the `uv run mcp install server.py` command exits with a message "Claude app not found", while it's not actually looking for Claude Desktop app on Linux at all.
This patch addresses that and enables the SDK to install MCP servers in Claude desktop app on Linux.

## Types of changes

### DIFF
--- a/src/mcp/cli/claude.py
+++ b/src/mcp/cli/claude.py
@@ -17,6 +17,8 @@ def get_claude_config_path() -> Path | None:
         path = Path(Path.home(), "AppData", "Roaming", "Claude")
     elif sys.platform == "darwin":
         path = Path(Path.home(), "Library", "Application Support", "Claude")
+    elif sys.platform.startswith("linux"):
+        path = Path(Path.home(), ".config", "Claude")
     else:
         return None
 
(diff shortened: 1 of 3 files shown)


dsp-ant (dismissed):
Happy to accept this as it's a good 80/20 solution, but to be correct, we should probably correctly respect XDG env variables and fallback to home for linux.

emsi:
> Happy to accept this as it's a good 80/20 solution, but to be correct, we should probably correctly respect XDG env variables and fallback to home for linux.

This https://github.com/modelcontextprotocol/python-sdk/pull/270/commits/d73cf9b8e8fe053397c86c8180634810a389fc21 should honor `XDG_CONFIG_HOME`

dsp-ant (dismissed):
Love it! Thank you

PR #246, opened 2025-03-07
enable custom path of sse server

elsejj:
## Motivation and Context

For an access server that might proxy multiple MCP servers, although it's possible to customize the path for each MCP using `Server`, it's too cumbersome compared to `FastMCP`. Consider adding a custom path setting option to FastMCP.

## How Has This Been Tested?

yes

## Breaking Changes

No,  default value is as same as before.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

```python
# setup a sse server on custom path `/mcp1/sse`
FastMCP("Demo", 
  sse_path = "/mcp1/sse",
  message_path="/mcp1/message/"
)

# setup a sse server on default path `/sse`
FastMCP("Demo")
```

PR #278, opened 2025-03-13
feat: add url_prefix support to FastMCP SSE

pridkett:
Previously the paths for SSE were hardcoded to /sse and /messages this made it hard to run behind an API gateway.  Requests to the /sse endpoint, even if proxied properly, would return /messages for the location of the /messages endpoint even if it should've been /api/v1/my_mcp_server/messages. This brings in a new parameter to FastMCP, url_prefix, that allows you set a prefix of something such as /api/v1/my_mcp_server and have the serve function properly. This can be passed either on the FastMCP constructor or with the environment variable FASTMCP_URL_PREFIX.

DCO-1.1 Signed-off-by: Patrick Wagstrom <160672+pridkett@users.noreply.github.com>

## Motivation and Context
There are situations when you may want something like an API Gateway to front multiple MCP servers, in this case they can't all serve the endpoints of `/sse` and `/messages`, but unfortunately, those were hardcoded. An because of the way that the connection is kept open on `/sse`, you can't always use body rewrite features inside of the API gateway to change the URLs.

## How Has This Been Tested?
I created, with the assistance of Claude, `tests/server/fastmcp/test_url_prefix.py` which covers most of the tests. I've also tested this locally with my API gateway and observed that it works.

## Breaking Changes
None observed

## Types of changes
- [ ] Documentation update

## Checklist

On the last one, if the other environment variables like `FASTMCP_HOST` and `FASTMCP_PORT` had a spot in the documentation, I'd add `FASTMCP_URL_PREFIX` next to them, but I didn't see where that might be.

## Additional context
Without this, it still is possible to run multiple MCP servers behind an API gateway, but you need to run them in `stdio` mode and use [Supergateway](https://github.com/supercorp-ai/supergateway) with the `--ssePath` and `--messagePath` flags set.

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -70,6 +70,7 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     # HTTP settings
     host: str = "0.0.0.0"
     port: int = 8000
+    url_prefix: str = ""
 
     # resource settings
     warn_on_duplicate_resources: bool = True
@@ -464,7 +465,7 @@ class FastMCP:
         from starlette.applications import Starlette
         from starlette.routing import Mount, Route
 
-        sse = SseServerTransport("/messages/")
+        sse = SseServerTransport(f"{self.settings.url_prefix}/messages/")
 
         async def handle_sse(request):
             async with sse.connect_sse(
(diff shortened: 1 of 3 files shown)


pridkett:
This is pretty similar to #246 - I'll let the maintainers figure out if we should do this with `url_prefix` or use `sse_path`/`message_path` as in #246.

PR #284, opened 2025-03-14
fix: enable HTTP redirect handling in SSE client

eastlondoner:
Updated the httpx AsyncClient to explicitly set follow_redirects=True to handle redirect responses from remote MCP servers. This fixes an issue where connections to servers that return redirects (like yo-mcp.com) would fail.

## Motivation and Context
When connecting to certain MCP servers (like yo-mcp.com) that respond with HTTP redirects, the client would fail to establish a connection. This is because the redirect wasn't being followed automatically by the SSE client. The TypeScript client handles this correctly, but the Python client needed an explicit configuration.

## How Has This Been Tested?
- Added a test script (`scripts/test_redirect.py`) that connects to https://yo-mcp.com/mcp/7NIUz0cvfStY6-wSsVyY3 and verifies the redirect handling works correctly
- Ran the script which confirmed successful connection with the server, including following the redirect and completing session initialization
- Verified in logs that the client correctly follows the HTTP 302 redirect

## Breaking Changes
None. This is a non-breaking bug fix that enables better compatibility with remote servers.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
The issue was simple - httpx has `follow_redirects=True` by default, but for some reason the redirect wasn't being followed by the SSE client. Adding the explicit parameter resolves the issue. The test script shows how to connect to a remote MCP server and can be used as a reference for other developers.

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -43,7 +43,7 @@ async def sse_client(
     async with anyio.create_task_group() as tg:
         try:
             logger.info(f"Connecting to SSE endpoint: {remove_request_params(url)}")
-            async with httpx.AsyncClient(headers=headers) as client:
+            async with httpx.AsyncClient(headers=headers, follow_redirects=True) as client:
                 async with aconnect_sse(
                     client,
                     "GET",


src/mcp/client/sse.py
```diff
@@ -43,7 +43,7 @@ async def sse_client(
     async with anyio.create_task_group() as tg:
         try:
             logger.info(f"Connecting to SSE endpoint: {remove_request_params(url)}")
-            async with httpx.AsyncClient(headers=headers) as client:
+            async with httpx.AsyncClient(headers=headers, follow_redirects=True) as client:
```
gruckion, line 46: LGTM

gruckion (commented):
Looks good, is there going to be another option needed? Shouldn’t we enable a way to provide a set of options to inject in?

allenporter:
There are lint errors that should be fixed.

My take is I don't think this PR should add more options.

src/mcp/client/sse.py
dsp-ant: We have multiple spots where we create an AsyncClient. We should introduce a function to create a stnadard AsyncClient with respective options.

dsp-ant (changes requested):
Thanks for looking into this. Happy to accept once the lint errors are fixed.

Note that I think it would be preferable if we had a function somewhere that creates an AsynClient object with whatever options we choose to be the default for the MCP SDK.

Kludex:
I think we can pass the client, instead of using a factory.

dsp-ant:
> I think we can pass the client, instead of using a factory.

that works too, but we need to make sure we have a central creation point to not have to sync options across client object creations.

ihrpr:
applying suggested changes in https://github.com/modelcontextprotocol/python-sdk/pull/655

PR #285, opened 2025-03-14
Strict types on the client side

Kludex:


src/mcp/client/session.py
```diff
@@ -76,18 +76,12 @@ def __init__(
...
-        )
-        roots = (
-            types.RootsCapability(
-                # TODO: Should this be based on whether we
-                # _will_ send notifications, or only whether
-                # they're supported?
-                listChanged=True,
-            )
-            if self._list_roots_callback is not None
-            else None
```
Kludex, line 90: The `None` was never being used, given that in `__init__` they have a default value.

src/mcp/shared/version.py
```diff
@@ -1,3 +1,3 @@
 from mcp.types import LATEST_PROTOCOL_VERSION
 
-SUPPORTED_PROTOCOL_VERSIONS = [1, LATEST_PROTOCOL_VERSION]
+SUPPORTED_PROTOCOL_VERSIONS: tuple[int, str] = (1, LATEST_PROTOCOL_VERSION)
```
dsp-ant, line 3: We can actually make this a list of strings ony land remove the old 1. That predates official release and is long gone.

dsp-ant (approved):

PR #287, opened 2025-03-14
Avoid double JSON encoding/decoding

Kludex:


src/mcp/server/sse.py
```diff
@@ -156,11 +156,11 @@ async def handle_post_message(
             response = Response("Could not find session", status_code=404)
             return await response(scope, receive, send)
 
-        json = await request.json()
```
Kludex, line 159: The `json()` calls the `json` module, we can just use the body, and Pydantic's `model_validate_json`. It's faster.

dsp-ant (approved):
TIL!

samuelcolvin:
nice!

Issue #269, opened 2025-03-12
README is full of errors.


emsi:
README says to run:
`uv add "mcp[cli]"`

If anyone will follow this advice she'll get the error:
> error: Requirement name `mcp` matches project name `mcp`, but self-dependencies are not permitted without the `--dev` or `--optional` flags. If your project name (`mcp`) is shadowing that of a third-party dependency, consider renaming the project.

The instruction is completely wrong.
One needs to first create its own project. Initialize it with `uv init` and only then can use the `uv add` command. But even then the subsequent calls of `mcp` won't work because `uv add` does not install `mcp[cli]` as a command. The best way to run mcp cli via `uv` is `uv run mcp`.

In general the README.md in current state can be a source of confusion.
Someone has already reported this #111 but it was not addressed.

dsp-ant:
Do you mind taking a stab at improving the README?

tadeodonegana:
I have just opened a PR addresing this issue #288

Issue #111, opened 2024-12-24
Requirement name `mcp` matches project name `mcp`


SHTechBoBo:
When I run the first step `uv add "mcp[cli]"`, 

it said: 
```error: Requirement name `mcp` matches project name `mcp`, but self-dependencies are not permitted without the `--dev` or `--optional` flags. If your project name (`mcp`) is shadowing that of a third-party dependency, consider renaming the project.```

I guess something wrong in `pyproject.toml`, but I don't know how to solve it.

alexjacobs08:
This is a genuine issue though no?  If you pull the project and follow the installation instructions you get an error right off the bat

tadeodonegana:
I have just opened a PR addressing issue #288.

jinghuan-Chen:
Try:
uv add "mcp[cli]" --dev

PR #288, opened 2025-03-14
docs: Update installation instructions for clarity

tadeodonegana:
Update installation instructions in README to acknowledge issues #269 and #111

Used the [official docs](https://modelcontextprotocol.io/quickstart/server) to enhance the existing example and make it more straightforward.

Demonstration of the updated installation flow:

https://github.com/user-attachments/assets/50985341-626d-431d-a5c4-03f01a470bc5

## Motivation and Context
Issues #269 and #111.

## Types of changes

- [x] Documentation update

## Checklist


13rac1 (commented):
This is far more clear, thank you! _Note: I do not have access to approve._

Rendered View: https://github.com/modelcontextprotocol/python-sdk/blob/d9e27a89/README.md#installation

README.md
```diff
@@ -69,23 +69,38 @@ The Model Context Protocol allows applications to provide context for LLMs in a
...
 
 ```bash
-uv add "mcp[cli]"
+# Create a new directory for your project
+uv init my-mcp-project
+cd my-mcp-project
+
+# Create virtual environment and activate it
+uv venv
+source .venv/bin/activate  # On Windows use: .venv\Scripts\activate
```
taha-yassine:
```suggestion
```
There's no need to explicitly create and activate the virtual environment, `uv` takes care of that implicitly. See https://docs.astral.sh/uv/concepts/projects/layout/#the-project-environment

jihun-im-open: I guess we need to create `.venv` to isolate packages

README.md
```diff
@@ -69,23 +69,38 @@ The Model Context Protocol allows applications to provide context for LLMs in a
...
+
+# Install MCP with CLI tools and the httpx package
+uv add "mcp[cli]" httpx
```
Kludex: Why would you need httpx? It's a dependency of mcp already.

dsp-ant (changes requested):
I appreciate the effort. I don't think we need additional instructions to create a project. If we want to go down the path, we should skip the explicit creation of the .venv, since `uv run` will manage this for us, and I don't want to confuse people with additional instructions about venv management.

README.md
tadeodonegana: Thank you for your comment, @taha-yassine. I added it because the official [MCP docs](https://modelcontextprotocol.io/quickstart/server) explicitly mention the virtual environment. It might be a good idea to review those docs as well.

I have removed the explicit creation of the virtual environment in this PR in order to proceed. Thanks!

README.md
tadeodonegana: Thank you for your comment, @Kludex.

I originally added it because the official [MCP docs](https://modelcontextprotocol.io/quickstart/server) explicitly mention it—likely due to the example provided in the documentation.

Since, as you mentioned, the dependency isn’t needed to run the python-sdk, I’ve removed it from this PR. Gracias!

tadeodonegana:
I have applied the requested changes. @dsp-ant, I would appreciate another review of these changes to see if we can move forward. Thanks!

PR #289, opened 2025-03-14
fix: Watch startup errors for stdio client

mattzh72:
## Summary
Improves `stdio_client` error handling to prevent indefinite hangs when subprocess failures occur, particularly when wrapping MCP servers with a synchronous interface. Adds structured error propagation to make failures catchable instead of failing silently inside `anyio`.

This is a minimal fix, but more long term, it would be great if either:
- MCP can have a native synchronous Python client (happy to contribute this)
- MCP can establish a handshake protocol with the server

## Motivation and Context
When integrating MCP servers with a **sync interface**, failures in `anyio.open_process()` or unexpected subprocess exits could result in the process hanging indefinitely instead of raising an exception. Previously, errors inside `anyio`'s process handling were not properly surfaced, making debugging difficult. 

This PR ensures:
- **Immediate errors** (missing binaries, bad arguments) raise `RuntimeError` instead of hanging.
- **Background process failures** (e.g., subprocess exits early) are surfaced via a structured exception.
- **Minimal startup wait (`startup_wait_time`)** allows for quick failure detection instead of silent hangs.

This is particularly useful when **MCP servers are used in synchronous mode**, where an uncatchable error would otherwise stall execution.

## How Has This Been Tested?
- Added and refined tests:
  - **Happy path**: Ensures valid subprocesses work as expected.
  - **Spawn failure**: Verifies missing binaries raise `RuntimeError` immediately.
  - **Nonzero exit**: Ensures subprocess errors propagate correctly instead of hanging.
- Ran tests locally with `pytest` and `uv run pytest`.
- Verified behavior in a real MCP client with a sync wrapper.

## Breaking Changes
- No API-breaking changes, but:
  - Subprocess failures that previously resulted in silent hangs now raise structured `RuntimeError` exceptions.
  - Callers relying on silent failures may need to adjust their error handling.

Note that the field defaults to 0.0, so should not have any effect unless explicitly set. 

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
- **Introduces `startup_wait_time`** (default `0.0`) to allow detection of immediate failures without forcing all cases to wait.
- **Improves test coverage** to prevent regression.
- **Avoids `ExceptionGroup` issues** by ensuring a single `RuntimeError` is raised when subprocess failures occur.


mattzh72:
We encountered this bad behavior while writing a synchronous MCP client. The uncatchable error stalls our server event loop.

cc @dsp-ant

mattzh72:
cc @Kludex @allenporter for suggestions here

ihrpr:

Similar to https://github.com/modelcontextprotocol/python-sdk/pull/434, the issue should be resolved with #333 

```
except OSError as exc:
```
is a good addition, would be great to add it as a separate PR 🙏

PR #295, opened 2025-03-16
Improve readability of CLI help option

Kludex:
You can try `mcp run --help` locally.

Now it looks like this:

```bash
❯ mcp run --help
                                                                                                                                                                                 
 Usage: mcp run [OPTIONS] FILE_SPEC                                                                                                                                              
                                                                                                                                                                                 
 Run a MCP server.                                                                                                                                                               
 The server can be specified in two ways:                                                                                                                                        
 1. Module approach: server.py - runs the module directly, expecting a server.run() call.                                                                                        
 2. Import approach: server.py:app - imports and runs the specified server object.                                                                                               
                                                                                                                                                                                 
 Note: This command runs the server directly. You are responsible for ensuring all dependencies are available.                                                                   
 For dependency management, use `mcp install` or `mcp dev` instead.                            
```

### DIFF
--- a/src/mcp/cli/cli.py
+++ b/src/mcp/cli/cli.py
@@ -294,15 +294,14 @@ def run(
 ) -> None:
     """Run a MCP server.
 
-    The server can be specified in two ways:
-    1. Module approach: server.py - runs the module directly, expecting a server.run()
-       call
-    2. Import approach: server.py:app - imports and runs the specified server object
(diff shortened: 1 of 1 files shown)


Kludex:
Btw, the import approach should have been `server:app` instead of adding the `.py` in it, as `uvicorn` does.

dsp-ant (approved):

PR #296, opened 2025-03-16
show_server_info

barnuri:
## Types of changes

- [ ] Documentation update

## Checklist


src/mcp/server/fastmcp/server.py
```diff
@@ -65,7 +65,7 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
 
     # Server settings
     debug: bool = False
-    log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "ERROR"
+    log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "INFO"
```
Kludex: I think this was changed on the last release. I find more useful to have INFO as default instead of ERROR - matching also Uvicorn that runs underneath on SSE.

As per the rest of this PR. I don't think we should include `show_server_info`, but we should make `FastMCP` settings type safe, so it's easier to configure.

barnuri: kk will move it

barnuri: why not moving transport too ?

Kludex: I didn't get what you are asking.

barnuri: move transport: Literal["stdio", "sse"]
to server config instead func param of run

barnuri:
@Kludex can you review ?

src/mcp/server/fastmcp/server.py
barnuri: resolved

dsp-ant (changes requested):
@Kludex any opinions? I personally understand why you would want this, but i think it would be better suited in your application as either a call before starting the server or as an extension to the FastMCP server class. I appreciate the effort but would politely decline adding this.

Kludex:
I think it's more useful in the stdio case. Since nothing happens when you start the application.

But on the HTTP, it will be a bit noisy...

If it's in debug level then I think it's fine.

barnuri:
ok i will change to debug

barnuri:
@Kludex @dsp-ant moved it to helper method and class, I think its very useful, for example in my code when im using sse i will add a route that return this object in some route for example /info (like in rest api have swagger json doc)
the idea of it is to have something simple without creating a client that can be easily visible
same for using stdio and then i will use the debug log

barnuri:
> @Kludex @dsp-ant moved it to helper method and class, I think its very useful, for example in my code when im using sse i will add a route that return this object in some route for example /info (like in rest api have swagger json doc) the idea of it is to have something simple without creating a client that can be easily visible same for using stdio and then i will use the debug log i hope you would like it :)

reping

barnuri:
@dsp-ant @Kludex can you tell me what do you think ?

barnuri:
@jerome3o-anthropic can you review ?

src/mcp/server/fastmcp/server.py
```diff
@@ -154,10 +157,47 @@ def run(self, transport: Literal["stdio", "sse"] = "stdio") -> None:
         if transport not in TRANSPORTS.__args__:  # type: ignore
             raise ValueError(f"Unknown transport: {transport}")
 
+        async def run(): await self._run(transport)
```
ihrpr: why do we need this? we can call `anyio.run` with parameters

ihrpr (changes requested):

Sorry for the delays in reviews. Suggested a few small changes as we should be good to go.

src/mcp/types.py
```diff
@@ -1128,3 +1128,28 @@ class ServerResult(
...
+    instructions: str | None
+    tools: list[Tool]
+    prompts: list[Prompt]
+    resources: list[Resource]
+    resource_templates: list[ResourceTemplate]
+
+    @property
+    def assets(self) -> dict[
+        str, list[Tool] | list[Prompt] | list[Resource] | list[ResourceTemplate]
+    ]:
```
Kludex: You can have a `TypedDict` for this.

src/mcp/server/fastmcp/server.py
Kludex: I don't think this is valid Python... 🤔

src/mcp/server/fastmcp/server.py
```diff
@@ -84,6 +85,8 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     # prompt settings
     warn_on_duplicate_prompts: bool = True
 
+    show_server_info: bool = True
```
ihrpr:
```suggestion
    show_server_info: bool = Flase
```

src/mcp/server/fastmcp/server.py
```diff
@@ -154,10 +157,47 @@ def run(self, transport: Literal["stdio", "sse"] = "stdio") -> None:
         if transport not in TRANSPORTS.__args__:  # type: ignore
             raise ValueError(f"Unknown transport: {transport}")
 
+        async def run(): await self._run(transport)
+        anyio.run(run)
```
ihrpr:
```suggestion
        anyio.run(self._run, transport)
```

src/mcp/server/fastmcp/server.py
barnuri: fixed

src/mcp/server/fastmcp/server.py
barnuri: change to False

dsp-ant:
I am closing this out since we haven't made much progress. I think this is mostly on us for screwing up reviewing this properly. Sorry @barnuri . I think in general we are still interested in the feeature, but mindful that we shouldn't add the underlying types to src/mcp/types.py since that file reflects the official protocol. This is an SDK feature only. If you are interested in redoing this ontop of the current state of the repo, please do so and ping me either via the MCP Contributor discord or on the PR and I will get to it.

Again, I am very sorry that we didn't review this in proper time and made the progress very slow.

PR #306, opened 2025-03-17
refactor: remove StdioServerParameters class

QuantGeekDev:
client/stdio and client/sse use two different patterns for initialization, this PR tries to standardize it after discussions with @Kludex  by removing StdioServerParameters

## Motivation and Context
Continuation to the discussion here:  https://github.com/modelcontextprotocol/python-sdk/pull/300

## How Has This Been Tested?
Testing with Claude Desktop's config

## Breaking Changes
This is a breaking change to how Stdio is initialized, it is no longer done via StdioServerParameters

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,15 +521,8 @@ if __name__ == "__main__":
 The SDK provides a high-level client interface for connecting to MCP servers:
 
 ```python
-from mcp import ClientSession, StdioServerParameters, types
+from mcp import ClientSession, types
 from mcp.client.stdio import stdio_client
-
-# Create server parameters for stdio connection
-server_params = StdioServerParameters(
-    command="python",  # Executable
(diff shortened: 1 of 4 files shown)


Kludex:
I don't think we want to add breaking changes, so we'd support both, and add a deprecation warning when using `settings` parameter.

QuantGeekDev:
> I don't think we want to add breaking changes, so we'd support both, and add a deprecation warning when using `settings` parameter.

Got it, that makes sense. I'll add implementation and test cases for this

QuantGeekDev:
@Kludex Do you think it makes sense to add this, or are we going to see an overhaul here with the changes from the new spec? I can get this in over the weekend

Kludex:
Can this bot not close PRs?

PR #325, opened 2025-03-19
Fix #201: Move incoming message stream from BaseSession to ServerSession

### LINKED ISSUE #201: Client hangs when ctx.info() logging with pure Python client

dsp-ant:
## Summary
- Move the incoming message stream from BaseSession to ServerSession where it's actually needed
- Add message_handler callback to ClientSession for direct handling of incoming messages
- Simplify client code by removing need for separate receive loop tasks

## Test plan
- All existing tests pass
- Added message_handler to ClientSession for optional direct message handling
- Refactored tests to use the new message handler pattern instead of listening to incoming message stream

Fixes #201

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,0 +1,53 @@
+import logging
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.sse import sse_client
+from mcp.client.stdio import stdio_client
+
+# disable httpx logs
+logging.basicConfig(level=logging.DEBUG)
+logging.getLogger("httpcore.http11").setLevel(logging.CRITICAL)
+logging.getLogger("httpx").setLevel(logging.CRITICAL)
+logging.getLogger("urllib3.connectionpool").setLevel(logging.CRITICAL)
+
+code = """
+import numpy
+a = numpy.array([1, 2, 3])
+print(a)
+a
+"""
+
+
+async def call_tools(session: ClientSession):
+    # await session.initialize()
+    await session.set_logging_level("debug")
+    tools = await session.list_tools()
+    print(f"{tools=}")
+    result = await session.call_tool("run_python_code", {"python_code": code})
+    print(f"{result=}")
+
+
+async def sse():
+    async with sse_client("http://localhost:3001/sse") as (read, write):
+        async with ClientSession(read, write) as session:
+            await call_tools(session)
+
+
+async def stdio():
+    server_params = StdioServerParameters(
+        command="npx",
+        args=[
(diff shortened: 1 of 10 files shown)


test.py
Kludex, line 1: delete?

dsp-ant: whoopsy, done.

sheffler:
Good explanation, and it makes sense.

This is the kind of subtle error that can be illustrated in an end-to-end test.  Without a clear specification as to the behavior of the client, and whether it reads incoming_messages, then an implementation is free to do what it wants.  In this case, the implementation I showed didn't read them, and there was the deadlock.

I think the callback pattern will work nicely.

j-z10:
We've also encountered this issue, and have to create a background task to consume the `session.incoming_messages` at the client side . Hope this can be merged quickly.

Kludex (commented):
What's the difference between `_logging_callback` and `_message_handler`? Seems those concepts are similar?

dsp-ant:
> What's the difference between `_logging_callback` and `_message_handler`? Seems those concepts are similar?

one is specific to logging, one handles all messages, incase you want do run custom messages or just intercept all messages

tanhaipeng:
very good, hope to update the version soon

jkawamoto:
After commit 568cbd1a66eaa0aa8a2ceca7f5c84719d9210871, my end-to-end test #294 is getting stuck:
https://github.com/modelcontextprotocol/python-sdk/actions/runs/14043734979.

(It was working correctly with commit 9ae4df85fbab97bf476ddd160b766ca4c208cd13:
https://github.com/modelcontextprotocol/python-sdk/actions/runs/14001749703)

I suspect something might be missing in commit 568cbd1a66eaa0aa8a2ceca7f5c84719d9210871. 

#361 fixes my hanging issue so far.

PR #333, opened 2025-03-20
send errors to pending requests if server closes

### LINKED ISSUE #332: Client requests hang forever if server dies
**Describe the bug**
If the stdio streams close before responding to a request (i.e. the server dies), the pending request waits forever (or for the read timeout) instead of erroring.

**To Reproduce**
1. `await session.initialize()` or some other request
2. in the server request handler, exit instead of responding to the request
3. on the client, the `await` never resolves so the client hangs indefinitely

**Expected behavior**
The typescript SDK makes the pending requests return a `-32000` "Connection closed" in this case

https://github.com/modelcontextprotocol/typescript-sdk/blob/0fa2397174eba309b54575294d56754c52b13a65/src/shared/protocol.ts#L260-L263

mroch:
## Motivation and Context
Fixes #332 

The client blocks indefinitely if the server streams close before responding to a pending request.

## How Has This Been Tested?
Added a test. Without this fix, the test fails because the request blocks forever.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist


ihrpr (dismissed):
Thank you!! 

Would be great to add 

```
self._response_streams.clear() 
```
before merging.

ihrpr (approved):

PR #336, opened 2025-03-20
refactor: Make types.py strictly typechecked.

dsp-ant:
We are now making the types module typecheck in strict mode.
This is mostly achieved by passing the correct types to the
generic type variables.

We ran into one specific issue with `JSONRPCRequest` and
`JSONRPCNotification`. Both are generic classes that take a
`dict[str, Any]` as params and just a plain string as method.

However, since the TypeVar for `RequestT` and `NotificationT` are
bound to `RequestParams` and `NotificationParams` respectively
we get into a type issue.

There are two ways of solving this:
1. Widen the bound by allowing explicitly for `dict[str, Any]`
2. Make JSONRPCRequest and JSONRPCNotificaiton not part of the
   type hierarchy with Request and Notification roots.

It felt most naturally to keep JSONRPCRequest/JSONRPCNotification
part of the type hierarchy and allow for general passing of
dict[str, Any].

This now typechecks.

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -64,8 +64,10 @@ class NotificationParams(BaseModel):
     """
 
 
-RequestParamsT = TypeVar("RequestParamsT", bound=RequestParams)
-NotificationParamsT = TypeVar("NotificationParamsT", bound=NotificationParams)
+RequestParamsT = TypeVar("RequestParamsT", bound=RequestParams | dict[str, Any] | None)
+NotificationParamsT = TypeVar(
+    "NotificationParamsT", bound=NotificationParams | dict[str, Any] | None
+)
 MethodT = TypeVar("MethodT", bound=str)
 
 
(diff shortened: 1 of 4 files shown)


src/mcp/types.py
```diff
@@ -64,8 +64,10 @@ class Meta(BaseModel):
     """
 
 
-RequestParamsT = TypeVar("RequestParamsT", bound=RequestParams)
-NotificationParamsT = TypeVar("NotificationParamsT", bound=NotificationParams)
+RequestParamsT = TypeVar("RequestParamsT", bound=RequestParams | dict[str, Any] | None)
+NotificationParamsT = TypeVar(
+    "NotificationParamsT", bound=NotificationParams | dict[str, Any] | None
+)
```
Kludex, line 70: This is a bit weird tho... I'd like to spend some minutes checking if we can be a bit stricter here.

Kludex: I'll do it tomorrow.

PR #350, opened 2025-03-23
fix: override input with json.loads only when output value is valid

omerc7:
Tool input values in pre_parse_json should only be transformed or cast after the output of json.loads has been validated via Pydantic.

## Motivation and Context
The input to the tool may undergo slight transformations in func_metadata.py::pre_parse_json.
For example, a field expected to be a string might be mistakenly cast to a different type. If the input arguments are {"field_name": "5"}, pre_parse_json will incorrectly cast the string "5" to the integer 5, resulting in an error when the field is expected to remain a string.

## How Has This Been Tested?
Added num_annotated_with_str to test_func_metadata.py
Tested on local MCP server

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist


ihrpr (changes requested):

looks like typecheck and tests are failing

ihrpr:
Closing this PR as it's stale now. Please re-open when changes are applied and this change is still needed.

Issue #357, opened 2025-03-24
`@mcp.tool()` Fails to Find Context Argument if its Type `Context` Includes Type Variables


jkawamoto:
**Describe the bug**
The code does not work because the type annotation for the `ctx` argument is not simply `Context`; it includes type variables `ServerSessionT` and `AppContext`.

```python
@mcp.tool()
def query_db(ctx: Context[ServerSessionT, AppContext]) -> str:
    """Tool that uses initialized resources"""
    db = ctx.request_context.lifespan_context.db
    return db.query()
```

The MCP server assumes that `ctx` will be passed by clients, as its implementation only checks whether the annotation is exactly `Context`. 

You can see this limitation in the following code:

https://github.com/modelcontextprotocol/python-sdk/blob/9ae4df85fbab97bf476ddd160b766ca4c208cd13/src/mcp/server/fastmcp/tools/base.py#L53-L58

The necessity of adding type variables to `Context` is explained in issue #xxx.


**To Reproduce**
1. Apply the following patch:

```patch
diff --git a/examples/fastmcp/simple_echo.py b/examples/fastmcp/simple_echo.py                                                                                                                                                                                                                                                                                                    
index c261526..3069b7e 100644                                                                                                                                                                                                                                                                                                                                                     
--- a/examples/fastmcp/simple_echo.py
+++ b/examples/fastmcp/simple_echo.py
@@ -1,14 +1,15 @@
 """                                                                                                                                                                                                                                                                                                                                                                              
 FastMCP Echo Server                                                                                                                                                                                                                                                                                                                                                              
 """                                                                                                                                                                                                                                                                                                                                                                              
+from typing import Any                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                  
-from mcp.server.fastmcp import FastMCP                                                                                                                                                                                                                                                                                                                                           
+from mcp.server.fastmcp import FastMCP, Context                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                  
 # Create server                                                                                                                                                                                                                                                                                                                                                                  
 mcp = FastMCP("Echo Server")                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                  
 @mcp.tool()                                                                                                                                                                                                                                                                                                                                                                      
[...]

PR #358, opened 2025-03-24
Fix `@mcp.tool()` Failing to Detect Context Argument When Type Variables Are Used in `Context`

jkawamoto:
## Motivation and Context

When defining tools to safely access the app context (lifespan context) using `Context`, it is necessary to include type variables.

However, the current implementation of `@mcp.tool()` fails to detect the context argument when type variables are specified (as detailed in #357).

This PR addresses and resolves this issue.

## How Has This Been Tested?

The following commands were executed locally to verify the fix:
```
uv sync --frozen --all-extras --dev
uv run --frozen ruff check .
uv run --frozen pytest
```

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist


Kludex (changes requested):
I don't think this solves the issue. Add a test, and you'll see.

jkawamoto:
Actually, it passes the current tests and also works on my local server.

The cause of #357 is that it only checks if `param.annotation` is `Context`.
When type variables are specialized, the resulting class is a subclass of the generic class. 
Also, a class is a subclass of itself.
So, we just need to use `issubclass(param.annotation, Context)` instead of `param.annotation is Context`.

This part is for handling cases where `param.annotation` is `typing.Literal` or similar, btw:
```python
if get_origin(param.annotation) is not None:
    continue
```

Kludex:
I'm asking to add a test where we use the generic params for `Context`. The context needs to be passed to the function.

This should work for:

```py
context: Context[SessionParam]  # missing lifespan, since it has a default
context: Context[SessionParam, LifespanParam]  # both
context: Context[SessionParam, LifespanParam] | None = None  # union
context: Annotated[Context[Session, LifespanParam], "metadata"]  # Not necessary for now
```

jkawamoto:
I've added a test for the second case.

Regarding the first case:
```python
Context[SessionParam]
```
This is not only useless but also invalid. It raises the following exception:
```aiignore
Too few parameters for <class 'mcp.server.fastmcp.server.Context'>; actual 1, expected 2
```

About the third case: I don't believe this situation should occur. If a tool requests a context, it should always be provided.

Kludex (approved):

Kludex (approved):

dsp-ant (approved):

PR #362, opened 2025-03-24
fix: support custom path when use sse server

Midicon:
Support custom path when use sse server

## Motivation and Context
I think when I use the SSE Server, it should support customizing the prefix path. Example: if sse's url is http://localhost:8000/some/path/to/sse, it should request http://localhost:8000/some/path/to/messages when sending messages

## How Has This Been Tested?
I've tested in my application

## Breaking Changes
Not necessary

## Types of changes

- [ ] Documentation update

## Checklist


dsp-ant:
You should be able to configure this by using `message_path`

Midicon:
@dsp-ant Modifying the message_path will also modify the router of Starlette, which will lead to the error: "Routed paths must start with '/'".

PR #368, opened 2025-03-25
Fix #355: Fix type error with lifespan context

### LINKED ISSUE #355: Type Errors in Sample Code
**Describe the bug**
I attempted to use `AppContext` in my project by utilizing the sample code provided in the `README.md`. However, it raised the following type error:
```
error: Cannot access attribute "db" for class "None"
```

**To Reproduce**
Below is the sample code taken from the `README.md`.

```python
from collections.abc import AsyncIterator

class Database:   # Replace with your actual DB type
    @classmethod
    async def connect(cls):
        return cls()

    async def disconnect(self):
        pass

    def query(self):
        return "Hello, World!"


# Create a named server
mcp = FastMCP("My App")


@dataclass
class AppContext:
    db: Database


@asynccontextmanager
async def app_lifespan(server: FastMCP) -> AsyncIterator[AppContext]:
    """Manage application lifecycle with type-safe context"""
    # Initialize on startup
    db = await Database.connect()
    try:
        yield AppContext(db=db)
    finally:
        # Cleanup on shutdown
        await db.disconnect()


# Pass lifespan to server
mcp = FastMCP("My App", lifespan=app_lifespan)


# Access type-safe lifespan context in tools
@mcp.tool()
def query_db(ctx: Context) -> str:
    """Tool that uses initialized resources"""
    db = ctx.request_context.lifespan_context.db
    return db.query()
```

Store the code above in a file named `readme.py` and run the following command:
```
uv run pyright readme.py
```

**Expected behavior**
The type check should pass without any errors.

 - OS: macOS 15.3.2
 - Browser N/A
 - Version N/A

dsp-ant:
## Summary
Fix type error when using strongly typed custom lifespan contexts with FastMCP.

- The  TypeVar was incorrectly defined with 
- This change removes the default, allowing proper type propagation
- Fixed type annotations in relevant places to maintain type safety
- Added test case demonstrating the fixed functionality


src/mcp/shared/context.py
```diff
@@ -7,7 +7,7 @@
 from mcp.types import RequestId, RequestParams
 
 SessionT = TypeVar("SessionT", bound=BaseSession[Any, Any, Any, Any, Any])
-LifespanContextT = TypeVar("LifespanContextT", default=None)
+LifespanContextT = TypeVar("LifespanContextT")
```
Kludex, line 10: But doesn't it makes sense to default to `None` on `FastMCP` context?

dsp-ant: FastMCP sets the default lifespan which yields and empty dictionary. There is actually never a None. This one defaults to Any, if `default` is not set, which I think is correct.

Kludex: Hmm... Then the default should be:

```py
class EmptyDict(TypedDict): ...
```

Kludex:
This PR doesn't fix the typing tho

Kludex:
🤔 

Screenshot 2025-03-26 at 15 23 56

Issue #359, opened 2025-03-24
stdio_client not starting in windows: getting NotImplementedError


JexinSam:
anyio.open_process in stdio_client not working in windows. It raises NotImplementedError.

```python
async with stdio_client(server_params) as (read, write):
        async with ClientSession(
            read, write, sampling_callback=handle_sampling_message
        ) as session:
            # Initialize the connection
            await session.initialize()
```

It gives out error:

```
   async with stdio_client(server_param) as (read, write):
  File "C:\Python311\Lib\contextlib.py", line 204, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\mcp\client\stdio.py", line 104, in stdio_client
    process = await anyio.open_process(
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\anyio\_core\_subprocesses.py", line 190, in open_process
    return await get_async_backend().open_process(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\anyio\_backends\_asyncio.py", line 2561, in open_process
    process = await asyncio.create_subprocess_exec(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\subprocess.py", line 218, in create_subprocess_exec
    transport, protocol = await loop.subprocess_exec(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\base_events.py", line 1680, in subprocess_exec
    transport = await self._make_subprocess_transport(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\asyncio\base_events.py", line 502, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
```

Kludex:
Hmmm... I think we need to choose the spawn mode somewhere...

saqadri:
@dsp-ant I had several users report issues with mcp-agent on Windows at least in part due to this. I got around to spending several hours to fix this, and figured I'd contribute the fix back upstream (or downstream? 🙃 ). #372 should cover most cases.

isakulaksiz:
#369

[...]

PR #372, opened 2025-03-26
Fixes to stdio_client to support Windows more robustly

saqadri:
This change should fix most cases of #359, with issues spawning stdio MCP servers on Windows.

## Motivation and Context
There is a somewhat longstanding issue with `stdio_cleint` that it doesn't work on Windows. 
Example:

Servers running with "npx" have problems because of the way "npx" command is handled in Windows (it is typically stored as a ps1 script):

```
mcp:
  servers:
    todoist:
      command: "npx"
      args: [ "@abhiz123/todoist-mcp-server", "-y"]
    brave-search:
      command: "npx"
      args: ["@modelcontextprotocol/server-brave-search", "-y"]
    fetch:
      command: "uvx"
      args: ["mcp-server-fetch"]
```

Replacing "npx" with "npx.cmd" usually mitigates this, but that makes the command platform-specific.
This change adds some of that handling in the stdio_client itself, and adds helpers for processing the command and spawning a process in platform-specific ways.

## How Has This Been Tested?
Tested on Windows VM in Azure using all [mcp-agent examples](https://github.com/lastmile-ai/mcp-agent/tree/main/examples). All of them use stdio_client, and all of them now work.

Note: for using with streamlit, I had to add an additional change which I didn't add to this PR, since it may not belong to the SDK:
```
asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
```

In addition:
* All tests pass
* some pyright failures in repo, but unrelated to this change

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Open to feedback on how to make this more elegant.

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -190,7 +190,7 @@ def _get_executable_command(command: str) -> str:
         command: Base command (e.g., 'uvx', 'npx')
 
     Returns:
-        List[str]: Platform-appropriate command
+        str: Platform-appropriate command
     """
 
     try:
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -168,10 +168,14 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
         try:
             yield read_stream, write_stream
         finally:
-            # Clean up process
+            # Clean up process to prevent any dangling orphaned processes
             try:
                 process.terminate()
                 if sys.platform == "win32":
+                    # On Windows, terminating a process with process.terminate() doesn't 
+                    # always guarantee immediate process termination. 
+                    # So we give it 2s to exit, or we call process.kill() 
+                    # which sends a SIGKILL equivalent signal.
                     try:
                         with anyio.fail_after(2.0):
                             await process.wait()
@@ -199,20 +203,21 @@ def _get_executable_command(command: str) -> str:
         else:
             # For Windows, we need more sophisticated path resolution
             # First check if command exists in PATH as-is
-            command_path = shutil.which(command)
-            if command_path:
+            if command_path:= shutil.which(command):
                 return command_path
 
             # Check for Windows-specific extensions
             for ext in [".cmd", ".bat", ".exe", ".ps1"]:
                 ext_version = f"{command}{ext}"
-                ext_path = shutil.which(ext_version)
-                if ext_path:
+                if ext_path:= shutil.which(ext_version):
                     return ext_path
 
             # For regular commands or if we couldn't find special versions
             return command
     except Exception:
+        # If anything goes wrong, just return the original command
+        # shutil.which() could raise exceptions if there are permission 
+        # issues accessing directories in PATH
         return command
 
 
@@ -245,12 +250,13 @@ async def _create_platform_compatible_process(
 
             return process
         except Exception:
-            # Don't raise, let's try to create the process using the default method
-            process = None
-
-    # Default method for creating the process
-    process = await anyio.open_process(
-        [command, *args], env=env, stderr=errlog, cwd=cwd
-    )
+            # Don't raise, let's try to create the process without creation flags   
+            process = await anyio.open_process(
+                [command, *args], env=env, stderr=errlog, cwd=cwd
+            )
+    else:
+        process = await anyio.open_process(
+            [command, *args], env=env, stderr=errlog, cwd=cwd
+        )
 
     return process
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -220,7 +220,7 @@ async def _create_platform_compatible_process(
     command: str,
     args: list[str],
     env: dict[str, str] | None = None,
-    errlog: int | TextIO = subprocess.PIPE,
+    errlog: TextIO = sys.stderr,
     cwd: Path | str | None = None,
 ):
     """
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -1,4 +1,6 @@
 import os
+import shutil
+import subprocess
 import sys
 from contextlib import asynccontextmanager
 from pathlib import Path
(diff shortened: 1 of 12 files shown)


src/mcp/server/fastmcp/server.py
```diff
@@ -652,9 +652,9 @@ async def read_resource(self, uri: str | AnyUrl) -> Iterable[ReadResourceContent
         Returns:
             The resource content as either text or bytes
         """
-        assert self._fastmcp is not None, (
-            "Context is not available outside of a request"
-        )
+        assert (
```
saqadri: Formatter

tests/shared/test_sse.py
```diff
@@ -138,9 +138,7 @@ def server(server_port: int) -> Generator[None, None, None]:
             time.sleep(0.1)
             attempt += 1
     else:
-        raise RuntimeError(
-            f"Server failed to start after {max_attempts} attempts"
-        )
+        raise RuntimeError(f"Server failed to start after {max_attempts} attempts")
```
saqadri: Formatter

tests/shared/test_ws.py
```diff
@@ -134,9 +134,7 @@ def server(server_port: int) -> Generator[None, None, None]:
             time.sleep(0.1)
             attempt += 1
     else:
-        raise RuntimeError(
-            f"Server failed to start after {max_attempts} attempts"
-        )
+        raise RuntimeError(f"Server failed to start after {max_attempts} attempts")
```
saqadri: Formatter

src/mcp/client/stdio.py
```diff
@@ -159,4 +165,92 @@ async def stdin_writer():
...
+
+            return process
+        except Exception:
+            # Don't raise, let's try to create the process using the default method
+            process = None
+
+    # Default method for creating the process
+    process = await anyio.open_process(
+        [command, *args], env=env, stderr=errlog, cwd=cwd
+    )
```
dsp-ant: We use if/else patterns everywhere else, we should do that here too.

src/mcp/client/stdio.py
```diff
@@ -159,4 +165,92 @@ async def stdin_writer():
...
+            # Clean up process
+            try:
+                process.terminate()
+                if sys.platform == "win32":
+                    try:
+                        with anyio.fail_after(2.0):
+                            await process.wait()
+                    except TimeoutError:
+                        # Force kill if it doesn't terminate
+                        process.kill()
```
dsp-ant: Why is this needed for windows?

src/mcp/client/stdio.py
```diff
@@ -159,4 +165,92 @@ async def stdin_writer():
...
+
+    try:
+        if sys.platform != "win32":
+            return command
+        else:
+            # For Windows, we need more sophisticated path resolution
+            # First check if command exists in PATH as-is
+            command_path = shutil.which(command)
+            if command_path:
+                return command_path
```
dsp-ant: I prefer the more modern
```suggestion
            if command_path := shutil.which(command):
                return command_path
```

src/mcp/client/stdio.py
```diff
@@ -159,4 +165,92 @@ async def stdin_writer():
...
+
+            # Check for Windows-specific extensions
+            for ext in [".cmd", ".bat", ".exe", ".ps1"]:
+                ext_version = f"{command}{ext}"
+                ext_path = shutil.which(ext_version)
+                if ext_path:
+                    return ext_path
```
dsp-ant:
```suggestion
                if ext_path := shutil.which(ext_version):
                    return ext_path
```

src/mcp/client/stdio.py
```diff
@@ -159,4 +165,92 @@ async def stdin_writer():
...
+
+            # For regular commands or if we couldn't find special versions
+            return command
+    except Exception:
+        return command
```
dsp-ant: What exception could happen here? Can we add at least a comment or be more specific about the exception?

dsp-ant (approved):
I really appreciate the work to make windows work. I have too little understanding of windows to have a strong opinion. Some minor comments. Throwing it back at you, but happy to merge with the minor changes that I requested.

In general it might be worthwhile to start a `src/mcp/client/stdio/` folder and move `stdio.py` to `src/mcp/client/stdio/__init__.py` or whatever, so that we can have a `src/mcp/client/stdio/win32.py`, or even just have a helper `src/mcp/platform/win32/`.

src/mcp/client/stdio.py
saqadri: This helps prevent orphaned processes lingering on when the parent Python process exits.

On Windows, terminating a process with `process.terminate()` doesn't always guarantee immediate process termination. The Windows terminate() method sends a polite request to close (often using the `WM_CLOSE` message or similar signals), but the application may choose to handle this in different ways.

So here I decided to wait for an arbitrary amount of time (2s), and if we don't see the process exiting, we'll call `process.kill()` which sends a SIGKILL equivalent signal.

saqadri: I can add a summary of this to the code as a comment for posterity

src/mcp/client/stdio.py
saqadri: The main one I was worried about is any exceptions thrown by shutil.which(). I never hit this clause, so I can also remove it if you prefer

saqadri:
@dsp-ant thanks for the prompt review! I believe I've addressed everything

src/mcp/client/stdio/__init__.py
```diff
@@ -159,4 +169,48 @@ async def stdin_writer():
...
         tg.start_soon(stdin_writer)
-        yield read_stream, write_stream
+        try:
+            yield read_stream, write_stream
+        finally:
+            # Clean up process to prevent any dangling orphaned processes
+            if sys.platform == "win32":
+                await terminate_windows_process(process)
+            else:
+                process.terminate()
```
cristipufu, line 179: This still hangs on linux if the invoked process expects SIGINT, eg: https://github.com/modelcontextprotocol/servers/blob/main/src/everything/index.ts#L13

Shouldn't we try to force kill, same as we do on Windows?

cristipufu: @dsp-ant check this out please: https://github.com/modelcontextprotocol/python-sdk/pull/555

BenMawnMahlauNBTC: just as a note the fix in cristipufu's pr works on windows 11, the merged fix still hangs.

Issue #373, opened 2025-03-26
How to set / get `client_id`?


samirbajaj:
I'd like to be able to retrieve the `client_id` from inside a tool call on an MCP server. However, it is always `None`. What do I need to set in the MCP client in order to be able to retrieve it on the server?

The client code is basically taken from the documentation:
```
from mcp import ClientSession, StdioServerParameters, types
from mcp.client.stdio import stdio_client

server_params = StdioServerParameters(
    command="python",  # Executable
    args=["example_server.py"],  # Optional command line arguments
    env=None,  # Optional environment variables
)


async def run():
    async with stdio_client(server_params) as (read, write):
        async with ClientSession(
            read, write
        ) as session:
            # Initialize the connection
            await session.initialize()

            # Call a tool
            result = await session.call_tool("my_tool")
```

The tool function is as follows:

```
mcp = FastMCP("docint")

@mcp.tool()
async def my_tool(ctx: Context) -> Optional[str]:
    """
    Extracts client ID and request ID from the context.

    Returns:
        client ID and request ID from the context.
    """
    return f"{ctx.client_id} {ctx.request_id}" # <----- THIS
```

This prints `None 2` (`client_id` as `None`).

What am I missing?

Thank you.

samirbajaj:
Can one of the contributors or maintainers confirm if this is supposed to work, please?

Or is it broken / WIP as the above comment suggests?

Thank you.

maxisbey:
`ctx.client_id` isn't populated by the SDK: it just echoes a `client_id` key from the request's `_meta` (both on v2 `main` and the 1.x `FastMCP` `Context`), and clients don't send one by default, hence the `None`. To have it set, pass it per call: `await session.call_tool("my_tool", meta={"client_id": "my-app"})` (the `meta=` argument on `call_tool` has been there since 1.19 and is on v2's `Client`/`ClientSession` too). If what you're actually after is an authenticated client's identity, that's the OAuth `client_id` on the access token: `get_access_token().client_id`, covered in [The caller's identity](https://py.sdk.modelcontextprotocol.io/v2/run/authorization/#the-callers-identity).

<details><summary>Example (v2)</summary>

```python
import anyio

from mcp import Client
from mcp.server.mcpserver import Context, MCPServer

mcp = MCPServer("demo")


@mcp.tool()
def my_tool(ctx: Context) -> str:
    return f"client_id={ctx.client_id!r}"


async def main() -> None:
[...]

PR #380, opened 2025-03-27
Added support for raw request injection in RequestContext.

### LINKED ISSUE #195: Pass entire request object to handlers; add raw request to MCP base Request
I have an MCP server that sits in front of my API to allow LLMs to interact with it. My API requires an authorization header.

I have hacked a way to do this in my fork, but essentially the MCP client is able to pass through headers. Only need to use this for the `/sse` request. Currently, the handlers extract the arguments they need in the decorator. We could instead add a field to the base Request class called `raw_request` or `headers` if we just need that and then ensure this is added to the request object before passing it to the handler.

```python
# src/mcp/types.py
class Request(BaseModel, Generic[RequestParamsT, MethodT]):
    """Base class for JSON-RPC requests."""

    method: MethodT
    params: RequestParamsT
    headers: dict[str, Any] | None = None # <<<<<<<<

---------------------
# src/mcp/server/fastmcp/server.py
    def call_tool(self):
        def decorator(
            func: Callable[
                ...,
                Awaitable[
                    Sequence[
                        types.TextContent | types.ImageContent | types.EmbeddedResource
                    ]
                ],
            ],
        ):
            logger.debug("Registering handler for CallToolRequest")

            async def handler(req: types.CallToolRequest):
                try:
                    results = await func(req)  # <<<<<<<<<
                    return types.ServerResult(
                        types.CallToolResult(content=list(results), isError=False)
                    )
                except Exception as e:
                    return types.ServerResult(
                        types.CallToolResult(
                            content=[types.TextContent(type="text", text=str(e))],
                            isError=True,
                        )
                    )

            self.request_handlers[types.CallToolRequest] = handler
            return func

        return decorator

-----------------------------
# src/mcp/server/fastmcp/server.py
    async def run_sse_async(self, middleware: list[type] = []) -> None:
        """Run the server using SSE transport."""
        from starlette.routing import Mount, Route

        sse = SseServerTransport("/messages/")

        async def handle_sse(request):
            async with sse.connect_sse(
                request.scope, request.receive, request._send
            ) as streams:
                await self._mcp_server.run(
                    streams[0],
                    streams[1],
                    raw_request=request, # <<<<<<<<<<<<<<
                )

        starlette_app = Starlette(
            debug=self.settings.debug,
            routes=[
                Route("/sse", endpoint=handle_sse),
                Mount("/messages/", app=sse.handle_post_message),
            ],
        )

        config = uvicorn.Config(
            starlette_app,
            host=self.settings.host,
            port=self.settings.port,
            log_level=self.settings.log_level.lower(),
        )
        server = uvicorn.Server(config)
        await server.serve()

---------------------------
# src/mcp/server/lowlevel/server.py
    async def run(
        self,
        initialization_options: InitializationOptions,
        raw_request: Any | None = None, # <<<<<<<<<<<<<<<<<<<<<<<<<
        # When False, exceptions are returned as messages to the client.
        # When True, exceptions are raised, which will cause the server to shut down
        # but also make tracing exceptions much easier during testing and when using
        # in-process servers.
    ):
        with warnings.catch_warnings(record=True) as w:
            async with ServerSession(
                read_stream, write_stream, initialization_options
            ) as session:
                    logger.debug(f"Received message: {message}")

                    match message:
                        case (
                            RequestResponder(
                                request=types.ClientRequest(root=req)
                            ) as responder
                        ):
                            with responder:
                                if raw_request is not None:
                                    req.headers = raw_request.headers # <<<<<<<<<<<<<<<<
                                await self._handle_request(
                                    message, req, session, raise_exceptions
                                )
                        case types.ClientNotification(root=notify):
                            await self._handle_notification(notify)

                    for warning in w:
                        logger.info(
                            f"Warning: {warning.category.__name__}: {warning.message}"
                        )
```

and then use this like:

```python
# already supported on client
transport = await exit_stack.enter_async_context(
    sse_client(url, headers={"authorization": "..."})
)
```

```python
# on server
mcp_server = FastMCP("example", transport="sse")

async def handle_call_tool(
    self: FastMCP, req: types.CallToolRequest # <<<<<<<<<<<<<<<<<<
) -> Sequence[types.TextContent | types.ImageContent | types.EmbeddedResource]:
    headers = {}
    if "authorization" in req.headers:
        headers = {"Authorization": req.headers["authorization"]}
    # ...http client call to api or if MCP is served from the app itself, check the key
```

I know auth is a part of the 2025 H1 roadmap so this may be usurped already in terms of how things will be supported. This goes beyond auth headers though since it could be useful to have access to the raw request in total instead within the tool execution context.

ylassoued:
closes #195
Added support for injecting the raw request into the `RequestContext`.

## Motivation and Context
In real-life applications, authentication may rely on request headers. In the case where the MCP server needs to act on behalf of the client to retrieve resources from a remote server or DB, thus requiring to authenticate on behalf of the client, getting the request headers is essential.
With the solution implemented in this PR, you may access the raw request from any end point as follows:
```python
mcp = FastMCP()

@mcp.resource(uri="resource://hello")
async def get_greeting() -> str:
    ctx: Context[ServerSession, object] = mcp.get_context()
    raw_request: Request = ctx.request_context.request # Starlette Request object
    headers = raw_request.headers
    # do something with the headers
```

## Breaking Changes
The only breaking change is in typing. As a matter of fact the following generic classes have now an additional type argument `RequestT` (raw request type):
- `class RequestContext(Generic[SessionT, LifespanContextT, RequestT])`
- `class Server(Generic[LifespanResultT, RequestT])`

## Types of changes

- [x ] Documentation update

## Checklist

- [ x] My code follows the repository's style guidelines


ylassoued:
@dsp-ant, I am not sure how the reviewing process works here. I was not able to add reviewers to this PR. Will somebody eventually be assigned as reviewer to this PR?

georg-ort:
pls mere this it's really essential to develop professional remote mcp servers!

ylassoued:
> pls merge this it's really essential to develop professional remote mcp servers!

Thank you @georg-ort! I am hoping that someone will review it and approve it so that I may merge the PR.

georg-ort:
Well I have to thank you @ylassoued! I'm currently using your fork in order to progress since I do need authentication context in my tool calls or our product wouldn't work.

ylassoued:
Oh brilliant. Thank you @georg-ort for supporting this. I am glad somebody if finding it useful :-).

emilioflc:
Thank you so much for this, @ylassoued! This looks incredibly useful across a variety of contexts. I actually ran into the exact same situation where I needed authentication context in my tool calls and came across your PR. Really appreciate you putting this together. Fingers crossed it gets reviewed and merged soon!

ylassoued:
Thank you very much @emilioflc. My pleasure :-).

null-crafter:
is this good to merge? this pull request is very helpful to me.

albertocalderari:
Hey, we sort of would like to get this too as we ended up implementing the same thing

rem-lag:
I'd just spent some time poking around trying to figure out if this could be done with the sdk as is so I'm happy to see there's already a PR for it, I hope it can be merged in because this functionality would make my life a lot easier!

Thanks for putting the work in @ylassoued

wbeard:
Thank you for the contribution! I'm using your branch locally and see the headers I need. Would love to see this merged.

samirbajaj:
Can we please get a review on this PR? I would love to see it merged.

Thank you.

281743556:
When are you planning to launch? We urgently need this feature

kelvin1804:
Please merge. This feature is essential for all remote MCPs
Thanks for your contribution @ylassoued

JorgeRuizITCL:
Seems good, even implements generics for requests different than Starlette.

wenxuwan:
This pr looks like it injects the header of the first /s se request for all subsequent /message requests, but if the user needs the header of the subsequent requests what do they need to do, like tools/call and tools/list requests, is there a way for the user to get the header information?

toughnoah:
Please merge!!!  It is really useful for tools authorization!!

samirbajaj:
> > hi @ylassoued, This pr looks like it injects the header of the first /sse request for all subsequent /messages requests, but if the user needs the header of subsequent requests what do they need to do, such as tools/call and tools/list requests, is there any way for the user to get the header information? I want to carry tracer information in the header of each subsequent /messages related http request and get the related tracer information in tools.
> 
> I looked at the client-related code and found that there is now no parameter allowing the user to carry a header in subsequent requests 😭

I imagined that I'd associate a header value with the `client_id` that I can access on subsequent requests, but I can't seem to extract the `client_id` either ... https://github.com/modelcontextprotocol/python-sdk/issues/373

wenxuwan:
> tools/call and tools/list

All subsequent requests (e.g. tools/call and tools/list) are a separate http request, which should allow the developer to customise the header of each request, but the Python client doesn't have this ability right now ☹️

wenxuwan:
> > > hi @ylassoued, This pr looks like it injects the header of the first /sse request for all subsequent /messages requests, but if the user needs the header of subsequent requests what do they need to do, such as tools/call and tools/list requests, is there any way for the user to get the header information? I want to carry tracer information in the header of each subsequent /messages related http request and get the related tracer information in tools.
> > 
> > 
> > I looked at the client-related code and found that there is now no parameter allowing the user to carry a header in subsequent requests 😭
> 
> I imagined that I'd associate a header value with the `client_id` that I can access on subsequent requests, but I can't seem to extract the `client_id` either ... #373

All subsequent requests (e.g. tools/call and tools/list) are a separate http request, which should allow the developer to customise the header of each request, but the Python client doesn't have this ability right now ☹️

ylassoued:
Hi @wenxuwan, @samirbajaj, apologies for the delay in getting back to you. I have to admit, I have not fully understood your concern with carrying the request headers for all subsequent /messages requests. I assumed that when you initialised a client with headers, then it would inject the headers in every request. Am I wrong in assuming so?

wenxuwan:
> Hi @wenxuwan, @samirbajaj, apologies for the delay in getting back to you. I have to admit, I have not fully understood your concern with carrying the request headers for all subsequent /messages requests. I assumed that when you initialised a client with headers, then it would inject the headers in every request. Am I wrong in assuming so?

hi @ylassoued  ,thanks for the reply, what I mean is that apart from the header information in the initialisation phase, I would like to be able to inject some other custom header information, such as the tracer information, in subsequent requests (the trace_id and span_id should be different for each subsequent http request)

ylassoued:
> I would like to be able to inject some other custom header information, such as the tracer information, in subsequent requests (the trace_id and span_id should be different for each subsequent http request)

Thanks @wenxuwan! Any chance you could write some sample code to illustrate the use case? For example, by writing how ideally the user can make use of the header in the subsequent calls.

wenxuwan:
> > I would like to be able to inject some other custom header information, such as the tracer information, in subsequent requests (the trace_id and span_id should be different for each subsequent http request)
> 
> Thanks @wenxuwan! Any chance you could write some sample code to illustrate the use case? For example, by writing how ideally the user can make use of the header in the subsequent calls.



> > I would like to be able to inject some other custom header information, such as the tracer information, in subsequent requests (the trace_id and span_id should be different for each subsequent http request)
> 
> Thanks @wenxuwan! Any chance you could write some sample code to illustrate the use case? For example, by writing how ideally the user can make use of the header in the subsequent calls.

hi @ylassoued , It's not easy to write a demo because the current python sdk doesn't have this capability, but as @quitrk 
 said, each http request should have its own request_context, and currently your implementation is that all subsequent requests share the request_context of the /sse request, which may solve some of the problems (like the This may solve some of the problems (e.g., authentication of tools), but it's not really right. We should carry the authentication information in the request_context of subsequent requests instead of sharing the request_context of the /sse request.

ylassoued:
> I think the basic idea is that an MCP client could send different data (be it headers / different header values or whatever else) for the requests it submits for your tools / resources / prompts. Those headers could contain information unique to each request, like a trace_id as @wenxuwan mentioned. These could in turn be used on the MCP server side for observability (e.g use Grafana, CloudWatch etc to monitor your applications and data processes).
> 
> Unless I'm missing something, the solution in this PR doesn't account for that fact, and reuses the same initial request that was submitted when an MCP client connected to the /sse endpoint, which is also a bit misleading since you're basically caching it in the request_context, which is supposed to pertain to the actual current request.

OK, I see what you mean. Yes, I confirm that in this implementation, it is the initial request that is injected in the context. This was sufficient to solve the authentication issue as you mentioned it already. But I do understand your concern, and I believe now that this implementation is misleading. Instead of injecting the request, I should have simply injected the session-wide initial request headers. But, as you suggested, it would be cleaner to inject the current request for each call instead of injecting the initial one. I'll have to dig into the code again to figure this out... Thanks @wenxuwan, @samirbajaj, and @quitrk!

wenxuwan:
> > I think the basic idea is that an MCP client could send different data (be it headers / different header values or whatever else) for the requests it submits for your tools / resources / prompts. Those headers could contain information unique to each request, like a trace_id as @wenxuwan mentioned. These could in turn be used on the MCP server side for observability (e.g use Grafana, CloudWatch etc to monitor your applications and data processes).
> > Unless I'm missing something, the solution in this PR doesn't account for that fact, and reuses the same initial request that was submitted when an MCP client connected to the /sse endpoint, which is also a bit misleading since you're basically caching it in the request_context, which is supposed to pertain to the actual current request.
> 
> OK, I see what you mean. Yes, I confirm that in this implementation, it is the initial request that is injected in the context. This was sufficient to solve the authentication issue as you mentioned it already. But I do understand your concern, and I believe now that this implementation is misleading. Instead of injecting the request, I should have simply injected the session-wide initial request headers. But, as you suggested, it would be cleaner to inject the current request for each call instead of injecting the initial one. I'll have to dig into the code again to figure this out... Thanks @wenxuwan, @samirbajaj, and @quitrk!

hi @ylassoued ,Currently, all subsequent requests except /sse are handled inside mcp.server.sse.SseServerTransport.handle_post_message, which transforms the request and sends it via writer.send(message) to the mcp.shared.session.BaseSession._receive_loop. This is where the problem occurs, handle_post_message converts the http request to a message while processing the request, resulting in the header information of the http request being lost 😣. Therefore, we can try to stuff the request_context into the message at the time of conversion and make sure that subsequent conversions will carry the request_context.

ronald0samuel:
Can you add this to the stdio server as well? Need a way to get the context from client to server

jlowin:
This is super useful -- since I'm sure the low-level server is in flux with things like Streamable HTTP coming into focus, I've added support for accessing the current Starlette request in the standalone FastMCP repo, without modifying the low-level MCP server. Would welcome thoughts for enhancement there: https://github.com/jlowin/fastmcp/pull/302

ihrpr:
related https://github.com/modelcontextprotocol/python-sdk/pull/216

ronald0samuel:
> How are we supposed to pass in the correct headers from the mcp client

to reiterate, suppose i have a chatbot application which has an mcp client connecting to some remote mcp server. Suppose there are multiple users using this application, each user has an ID to identify themselves and the response from the server at any time depends on the which user has made a request. I don't want to give the user ID in the tool schema since its not upto the chatbot to keep track of user IDs, all i need is a way to pass custom metadata per tool call so that the server returns information pertaining to a particular user. This PR doesn't help with that since the headers are finalised while initialising the client and cannot be changed after app startup

ylassoued:
This is great new!!! Thank you very much @ihrpr!!!!! :-)

Issue #244, opened 2025-03-06
`@app.resource()` does not accept `ctx: Context`


Kludex:
I'm having this exception:

```bash
Traceback (most recent call last):
  File "/Users/marcelotryle/dev/pydantic/logfire-mcp/main.py", line 121, in <module>
    @mcp.resource("schema://records")
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marcelotryle/dev/pydantic/logfire-mcp/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/server.py", line 383, in decorator
    raise ValueError(
ValueError: Mismatch between URI parameters set() and function parameters {'ctx'}
```

When I have this resource:

```py
@mcp.resource("schema://records")
async def get_logfire_records_schema(ctx: Context) -> str: ...
```

It seems the `ctx` is not being taken in consideration.

cc @jlowin

luarss:
Not a fix, but a potential workaround:

```python
@mcp.resource("schema://records")
async def get_logfire_records_schema() -> str:
    ctx = mcp.get_context()
```

jlowin:
Sorry just seeing this - this is well-supported now in FastMCP v2 https://gofastmcp.com/servers/context#resources-and-templates

RubenCristino:
> Sorry just seeing this - this is well-supported now in FastMCP v2 https://gofastmcp.com/servers/context#resources-and-templates

'm experiencing the same issue as others:
mismatch between uri parameters set() and function parameters {'ctx'}

I'm currently using the following version of the fastmcp package:
Version: 2.10.6

Despite having the latest version, the error persists even when following the documented examples.

RubenCristino:
> I'm having this exception:
> 
> Traceback (most recent call last):
>   File "/Users/marcelotryle/dev/pydantic/logfire-mcp/main.py", line 121, in <module>
>     @mcp.resource("schema://records")
>      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/marcelotryle/dev/pydantic/logfire-mcp/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/server.py", line 383, in decorator
>     raise ValueError(
> ValueError: Mismatch between URI parameters set() and function parameters {'ctx'}
> When I have this resource:
> 
> @mcp.resource("schema://records")
> async def get_logfire_records_schema(ctx: Context) -> str: ...
[...]

PR #248, opened 2025-03-08
Enable context injection into resources

jonvet:
## Motivation and Context
According to the readme it should be possible to pass a `Context` to a resource (like it is for tools), but that's currently throwing an error (https://github.com/modelcontextprotocol/python-sdk/issues/244)

## How Has This Been Tested?
Added pytests, and confirmed with inspector that the below demo server works as intended:
```
from mcp.server.fastmcp import FastMCP, Context

mcp = FastMCP("Demo")


@mcp.resource("identification://{name}")
def get_ids(name: str, ctx: Context) -> str:
    request_id = ctx.request_id
    client_id = ctx.client_id
    return str(f"{request_id},{client_id}")
```
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
I tried to follow a similar structure to the way context is injected in tools

PR #413, opened 2025-04-02
fix: enables context injection into resources

fubuloubu:
rebase of https://github.com/modelcontextprotocol/python-sdk/pull/248 (copied below)

---

closes https://github.com/modelcontextprotocol/python-sdk/issues/244

## Motivation and Context

According to the readme it should be possible to pass a `Context` to a resource (like it is for tools), but that's currently throwing an error (https://github.com/modelcontextprotocol/python-sdk/issues/244)

## How Has This Been Tested?

How Has This Been Tested?

Added tests, and confirmed with inspector that the below demo server works as intended:

```py

mcp = FastMCP("Demo")


@mcp.resource("identification://{name}")
def get_ids(name: str, ctx: Context) -> str:
    request_id = ctx.request_id
    client_id = ctx.client_id
    return str(f"{request_id},{client_id}")
```

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

I tried to follow a similar structure to the way context is injected in tools


stuzero:
context injection into prompts would be great also

ChristianHinge:
It would be fantastic with this feature! Since most of my endpoints use context, I currently have to wrap them all as tools.

Kludex:
This PR doesn't seem to be enough. We can't read from the `ctx` on the resource yet...

sam-tombury:
There seems to be some complexity in how that interacts with the concept of concrete/direct resources vs resource templates. In particular, a resource function might accept _only_ a Context param (for example, if it lists all objects that a user can access). I think this PR would cause those to appear as nullary (0-arg) resource templates, which seem to be an edge case for clients (e.g. at first glance, they don't seem to work properly in the MCP Inspector).

Reading through the code, it appears that concrete resources are already lazy - so we could consider Context-only resources as concrete/direct and thread through the context in the [read](https://github.com/modelcontextprotocol/python-sdk/blob/10cf0f78a8ee8b5bc7152591cb5eb115ab2b6656/src/mcp/server/fastmcp/resources/base.py#L45-L48) method.

felixweinberger:
Requesting changes to put this back in @fubuloubu's queue to addres @Kludex' comments here.

> This PR doesn't seem to be enough. We can't read from the `ctx` on the resource yet... All the tests are receiving the `Context`, but they are actually not using it. 😅

felixweinberger (changes requested):
Requesting changes to put this back in @fubuloubu's queue to addres @Kludex' comments here.

> This PR doesn't seem to be enough. We can't read from the `ctx` on the resource yet... All the tests are receiving the `Context`, but they are actually not using it. 😅

felixweinberger:
This looks to be the same / similar issue as being tackled by #1406. Closing this one in favor of consolidating on the active discussion.

PR #386, opened 2025-03-27
Fix SSE client handling of nested path URLs

danilpavlov:
Fix SSE client failing when connecting to servers with nested paths (e.g., http://domain/path1/path2/sse). Previously, the client worked only with simple URLs like http://domain/sse.

The fix extracts the base path segment between domain and /sse endpoint using regex, then correctly constructs the response URL by preserving the path structure. This ensures proper communication with MCP servers hosted under non-root paths.

Without this fix, clients couldn't connect to SSE servers deployed under nested paths, as the endpoint URLs for posting messages were constructed incorrectly.

## Motivation and Context

I got problem while deploying my MCP server inside a Kubernetes cluster. The SSE client was only working with simple URLs like http://example-domain/sse, but failed to properly handle more complex paths such as http://example-domain/path1/path2/sse.

When the server was deployed under nested paths, the client couldn't establish a proper connection because it was incorrectly constructing the endpoint URLs for posting messages. This made it impossible to use MCP servers hosted under non-root paths, which is a common scenario in Kubernetes environments where services often run behind path-based routing.

## How Has This Been Tested?
I ve tested on version for local mcp server (http://localhost:8000/sse) and with special mask (http://domain/dummy/sse)

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #422, opened 2025-04-03
fix(sse): preserve base path when resolving message endpoint from SSE

sterankin:
## Problem

When using the `sse_client()` transport, the server sends the `endpoint` as `/messages`. This is resolved using `urljoin(url, sse.data)`, but this strips the full base path (e.g., `/mcp/weather-alerts`) and results in incorrect POSTs to the root `/messages`.

## Solution

This patch updates the `endpoint` resolution to correctly derive `uri_base` from the provided SSE URL by removing `/sse`, and then joining the relative path properly:

```python
uri_base = url.rsplit("/sse", 1)[0]
endpoint_url = urljoin(uri_base + "/", sse.data.lstrip("/"))
```

This ensures that if the client connects to:
https://example-mcp-server.com/mcp/weather-alerts/sse

The messages are correctly posted to:
https://example-mcp-server..com/mcp/weather-alerts/messages

instead of:
https://example-mcp-server..com/messages

Notes
This is backward-compatible and only affects the computed POST endpoint.

Patch tested successfully against MCP servers behind API gateways.

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -62,7 +62,8 @@ async def sse_client(
                                 match sse.event:
                                     case "endpoint":
                                         uri_base = url.rsplit("/sse", 1)[0]
-                                        endpoint_url = urljoin(uri_base + "/", sse.data.lstrip("/"))
+                                        endpoint_url = urljoin(uri_base + "/", 
+                                                               sse.data.lstrip("/"))
                                         logger.info(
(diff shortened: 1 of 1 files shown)


danilpavlov:
Hi, @sterankin! We are working on this problem in [this PR](https://github.com/modelcontextprotocol/python-sdk/pull/386)

Fell free to join us.

sterankin:
Closing as duplicate

PR #425, opened 2025-04-04
Fix setup instructions for Claud and MCP

evgenyneu:
Fixed the instructions, previous ones did not work. Tested on Mac :D.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ The Model Context Protocol allows applications to provide context for LLMs in a 
 
 ### Adding MCP to your python project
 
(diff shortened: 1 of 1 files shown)


ihrpr (changes requested):

I think the core issue is that we need to have an instruction to install mcp package.  mcp install server.py and mcp dev server.py should work directly if the mcp package is properly installed.

ihrpr:
Closing this PR, looks stale now. Please re-open if the changes are still needed and suggestions applied.

Issue #378, opened 2025-03-26
FastMCP should support all of RFC 6570


ciccolo-anthropic:
**Is your feature request related to a problem? Please describe.**
The MCP protocol [docs](https://github.com/modelcontextprotocol/docs/blob/main/docs/concepts/resources.mdx#resource-templates) state that resource templates are specified according to RFC 6570. But the MCP code uses a [regex](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/server.py#L369) that rejects some valid URI templates. In particular, it rejects templates with multiple optional params like `search://emails{?query,start,end}`, which would be useful to have.

**Describe the solution you'd like**
FastMCP accepts and correctly handles any valid RFC 6570 template.

**Describe alternatives you've considered**
Optional params can be made mandatory with the tool treating an empty string as if the param was not set. If the full spec is too hard to implement, incrementally supporting it might still be useful.

dsp-ant:
I think that's totally valid. I am not sure we have the time at the moment to work on it but feel free to send a PR.

codecracker007:
hey @ciccolo-anthropic  yes the full spec of 6570 will indeed be big do you have any priority or preference of the spec in mind on what can be added for now which can be used frequently Since some of the operators and expansions will be a bit annoying to work with when performing reverse matching on the URI template to  variable names

@dsp-ant do you guys plan on being full RFC 6570 spec complaint and support multiple levels of template expressions for now its just expansion with no operators right?

felixweinberger:
https://github.com/modelcontextprotocol/python-sdk/pull/427 should be addressing this

PR #427, opened 2025-04-04
Added RFC 6570 complaint form style query expansion as optional param… 

codecracker007:
# Add RFC 6570 support for optional parameters as form-style query expansions

## Motivation and Context

Adds optional parameter support with RFC 6570 compliance as form-style query expansions. This change addresses issue #378, .

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/resources/templates.py
+++ b/src/mcp/server/fastmcp/resources/templates.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import re
+import urllib.parse
 from collections.abc import Callable
 from typing import Any
 
@@ -16,7 +17,7 @@ class ResourceTemplate(BaseModel):
     """A template for dynamically creating resources."""
 
     uri_template: str = Field(
-        description="URI template with parameters (e.g. weather://{city}/current)"
+        description="URI template with parameters (e.g. weather://{city}/current{?units,format})"
     )
     name: str = Field(description="Name of the resource")
     description: str | None = Field(description="Description of what the resource does")
@@ -26,6 +27,14 @@ class ResourceTemplate(BaseModel):
     fn: Callable[..., Any] = Field(exclude=True)
     parameters: dict[str, Any] = Field(
         description="JSON schema for function parameters"
+    )
+    required_params: set[str] = Field(
+        default_factory=set,
+        description="Set of required parameters from the path component",
+    )
+    optional_params: set[str] = Field(
+        default_factory=set,
+        description="Set of optional parameters specified in the query component",
     )
 
     @classmethod
@@ -48,6 +57,41 @@ class ResourceTemplate(BaseModel):
         # ensure the arguments are properly cast
         fn = validate_call(fn)
 
+        # Extract required and optional parameters from function signature
+        required_params, optional_params = cls._analyze_function_params(fn)
+
+        # Extract path parameters from URI template
+        path_params: set[str] = set(
+            re.findall(r"{(\w+)}", re.sub(r"{(\?.+?)}", "", uri_template))
+        )
+
+        # Extract query parameters from the URI template if present
+        query_param_match = re.search(r"{(\?(?:\w+,)*\w+)}", uri_template)
+        query_params: set[str] = set()
+        if query_param_match:
+            # Extract query parameters from {?param1,param2,...} syntax
+            query_str = query_param_match.group(1)
+            query_params = set(
+                query_str[1:].split(",")
+            )  # Remove the leading '?' and split
+
+        # Validate path parameters match required function parameters
+        print(f"path_params: {path_params}")
+        print(f"required_params: {required_params}")
+        if path_params != required_params:
+            raise ValueError(
+                f"Mismatch between URI path parameters {path_params} "
+                f"and required function parameters {required_params}"
+            )
+
+        # Validate query parameters are a subset of optional function parameters
+        if not query_params.issubset(optional_params):
+            invalid_params: set[str] = query_params - optional_params
+            raise ValueError(
+                f"Query parameters {invalid_params} do not match optional "
+                f"function parameters {optional_params}"
+            )
+
         return cls(
             uri_template=uri_template,
             name=func_name,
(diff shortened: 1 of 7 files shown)


ihrpr (changes requested):

This PR makes good progress toward [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570) support by adding form-style query expansion. 

There are a few suggestions: 

 1. Type Conversion for Query Parameters. Query parameters from URLs are always strings, but the code doesn't convert them to match function parameter
  types.

  Example of the problem:
  ```
  @server.resource("resource://items/{category}{?limit}")
  def get_items(category: str, limit: int = 10) -> str:
      # When called with resource://items/books?limit=20
      # limit will be string "20" not int 20
      return f"Items in {category}, limit: {limit} (type: {type(limit)})"
```
  Suggested fix: Add type conversion based on function annotations in 
  
 
  2. Handle Edge Cases in Query Parsing. The current implementation might not handle edge cases properly:
  - Empty query values (?param=)
  - URL-encoded values (?name=John%20Doe)

  3. Add Tests for Type Conversion Scenarios

  4. Consider Documenting Query Parameter Behavior

**Comment by codecracker007** [comment]
@ihrpr thanks for the feedback

during the fn call with params              [result = self.fn(**params)](https://github.com/modelcontextprotocol/python-sdk/blob/f2f4dbdcbd30fd00ced777fd1f59d00624362c97/src/mcp/server/fastmcp/resources/templates.py#L73)

can i use [call_fn_with_arg_validation](url) https://github.com/modelcontextprotocol/python-sdk/blob/f2f4dbdcbd30fd00ced777fd1f59d00624362c97/src/mcp/server/fastmcp/utilities/func_metadata.py#L45 for type validation and conversion and should i fallback to default values for optional parameters if validation fails for any one of them

**Comment by codecracker007** [comment]
@ihrpr the pydantic validate_call already takes care of the compatible type conversion based on annontated parameters for query parameters and raises  validation error for incompatible types added a decorator which check if the validation error is raised by optional parameters and ignores them so they fall back to default values to keep it compliant

added relevant tests covering these cases

for the cases in query parsing the empty values and url encoded parameters are handled by parse_qs added relevant tests covering these

**Review by felixweinberger — CHANGES_REQUESTED** [review]
Apologies for the delay in coming back to this - there seem to be some failing tests & merge conflicts, would you be able to resolve?

I'll prioritize responding quickly if you have the bandwidth to update the PR here.

File: src/mcp/server/fastmcp/utilities/func_metadata.py [hunk 1]
```diff
@@ -521,3 +523,110 @@ def _convert_to_content(
         result = pydantic_core.to_json(result, fallback=str, indent=2).decode()
 
     return [TextContent(type="text", text=result)]
+
+
+def use_defaults_on_optional_validation_error(
```
felixweinberger, line 528: We should fail fast if an optional parameter was provided with a non-matching type.

I think we should completely remove this as well as the call to it in `templates.py`

src/mcp/server/fastmcp/server.py
```diff
@@ -514,15 +535,6 @@ def decorator(fn: AnyFunction) -> AnyFunction:
             has_func_params = bool(inspect.signature(fn).parameters)
 
             if has_uri_params or has_func_params:
-                # Validate that URI params match function params
-                uri_params = set(re.findall(r"{(\w+)}", uri))
-                func_params = set(inspect.signature(fn).parameters.keys())
-
-                if uri_params != func_params:
-                    raise ValueError(
-                        f"Mismatch between URI parameters {uri_params} and function parameters {func_params}"
```
felixweinberger, line 523: note: removing this validation makes sense because we now support optional params.

src/mcp/server/fastmcp/resources/templates.py
```diff
@@ -34,40 +46,127 @@ def from_function(
...
+        # Get schema from TypeAdapter using the original function for correct schema
+        parameters = TypeAdapter(original_fn).json_schema()
 
-        # ensure the arguments are properly cast
-        fn = validate_call(fn)
+        # First, apply pydantic's validation and coercion
+        validated_fn = validate_call(original_fn)
+
+        # Then, apply our decorator to handle default fallback for optional params
+        final_fn = use_defaults_on_optional_validation_error(validated_fn)
```
felixweinberger, line 61: remove, we should fail fast and not silently when invalid params are provided by clients.

src/mcp/server/fastmcp/resources/templates.py
```diff
@@ -34,40 +46,127 @@ def from_function(
...
+            # First add extracted parameters
+            fn_params = {
+                name: value
+                for name, value in params.items()
+                if name in self.required_params or name in self.optional_params
+            }
+
+            # self.fn is now multiply-decorated:
+            # 1. validate_call for coercion/validation
+            # 2. our new decorator for default fallback on optional param validation err
```
felixweinberger, line 168: don't use a fallback, just fail if the optional params are supplied incorrectly

tests/issues/test_141_resource_templates.py
```diff
@@ -112,3 +146,113 @@ def get_user_profile(user_id: str) -> str:
 
         with pytest.raises(Exception):  # Specific exception type may vary
             await session.read_resource(AnyUrl("resource://users/123/invalid"))  # Invalid template
+
+
```
felixweinberger, line 150: thank you for adding significant test coverage - we'll want to update these to catch expected validation errors rather than the fallback logic.

src/mcp/server/fastmcp/resources/templates.py
```diff
@@ -34,40 +46,127 @@ def from_function(
         mime_type: str | None = None,
     ) -> ResourceTemplate:
         """Create a template from a function."""
-        func_name = name or fn.__name__
+        original_fn = fn
```
felixweinberger, line 49: `original_fn`, `fn`, `final_fn` - this gets quite confusing.

I think if we remove the `use_defaults_on_optional_validation_error` we shouldn't need any of these and can just stick to `fn`

Kludex:
Don't we want something more FastAPI-ish? Like `Annotated[type[T], Query()]`? 🤔

felixweinberger:
> Don't we want something more FastAPI-ish? Like `Annotated[type[T], Query()]`? 🤔

Something like this?

```python
from typing import Annotated

@mcp.resource("search://{category}/{id}")
def search(
    category: Annotated[str, Path()],
    id: Annotated[int, Path()],
    format: Annotated[str, Query()] = "json",
    limit: Annotated[int, Query()] = 10
):
    ...
```

Note: there's also https://github.com/modelcontextprotocol/python-sdk/pull/1439 which is touching the same area so we'd need some coordination cc: @beaterblank

Kludex:
> > Don't we want something more FastAPI-ish? Like `Annotated[type[T], Query()]`? 🤔
> 
> Something like this?
> 
> ```python
> from typing import Annotated
> 
> @mcp.resource("search://{category}/{id}")
> def search(
>     category: Annotated[str, Path()],
>     id: Annotated[int, Path()],
>     format: Annotated[str, Query()] = "json",
>     limit: Annotated[int, Query()] = 10
> ):
>     ...
> ```
> 
> Note: there's also #1439 which is touching the same area so we'd need some coordination cc: @beaterblank

For `category` and `id`, you wouldn't need it, but you could... Also, I think `Query`/`Path` would have the `default` parameter, so you don't need to add default in the parameters.

felixweinberger:
> > > Don't we want something more FastAPI-ish? Like `Annotated[type[T], Query()]`? 🤔
> > 
> > 
> > Something like this?
> > ```python
> > from typing import Annotated
> > 
> > @mcp.resource("search://{category}/{id}")
> > def search(
> >     category: Annotated[str, Path()],
> >     id: Annotated[int, Path()],
> >     format: Annotated[str, Query()] = "json",
> >     limit: Annotated[int, Query()] = 10
> > ):
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note: there's also #1439 which is touching the same area so we'd need some coordination cc: @beaterblank
> 
> For `category` and `id`, you wouldn't need it, but you could... Also, I think `Query`/`Path` would have the `default` parameter, so you don't need to add default in the parameters.

OK thanks, so more like this?

```python
from typing import Annotated

@mcp.resource("search://{category}/{id}")
def search(
    category: str,
    id: int,
    format: Annotated[str, Query(default="json")],
    limit: Annotated[int, Query(default=10)],
):
    ...
```

beaterblank:
> > Don't we want something more FastAPI-ish? Like `Annotated[type[T], Query()]`? 🤔
> 
> Something like this?
> 
> ```python
> from typing import Annotated
> 
> @mcp.resource("search://{category}/{id}")
> def search(
>     category: Annotated[str, Path()],
>     id: Annotated[int, Path()],
>     format: Annotated[str, Query()] = "json",
>     limit: Annotated[int, Query()] = 10
> ):
>     ...
> ```
> 
> Note: there's also #1439, which is touching the same area, so we'd need some coordination cc: @beaterblank

Yes, using `Annotated` sounds like a good approach, but we should ideally support both styles. 

~Also, a small nitpick, the classes should be named `PathParameter` and `QueryParameter`, since I spent a few minutes confused :)~ or maybe its best stick with how fast api does it.

I also noticed an issue in my PR: I’m currently inferring types only from the URI, but this should be done based on the function signature value's types instead. can reuse the code from this PR for that.

Additionally, since both features modify the same function in different ways, we should refactor the logic into smaller, more focused functions. That will make the codebase cleaner and easier to extend in the future.

At this point, I could either merge this PR into mine and resolve the conflicts properly, or do it the other way around.



merging this into #1439

beaterblank:
Can close this PR merged into #1439 .

felixweinberger:
Consolidating discussion on https://github.com/modelcontextprotocol/python-sdk/pull/1439

Issue #399, opened 2025-03-31
[Progress] ProgressNotification The message field is missing.


Zhaobudaoyuema:
hello. I noticed that the 3.26 version specification added a `message` field for progress updates. However, I couldn’t find it in the current 1.6.0 Python SDK. Could you please let me know when it will be supported?
![Image](https://github.com/user-attachments/assets/4cdaeb20-d670-4175-ba1c-a9c20cf988e9)

![Image](https://github.com/user-attachments/assets/62947905-d02f-4828-ae11-81363849f808)

aksheyd:
This would be my first PR but could I try adding this @dsp-ant? 

I'm thinking of adding the type to the class, adding it as an arg to send_progress_notification() in client/server/shared, editing shared/progress.py, and adding a test case.

PR #435, opened 2025-04-05
feat: add message to ProgressNotification

aksheyd:
Added an optional message field to ProgressNotificationParams according to [latest mcp spec](https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress/).

## Motivation and Context
Addressing #399 to add spec guidelines to sdk

## How Has This Been Tested?
Added test_progress_notification to test messages bidirectionally. I'm hesitant that covers all angles so I'd like if someone could look over the test case a little more.

## Breaking Changes
This did break test_176_progress_token, but it was over a schema check rather than implementation of the code (checking message=None since it's now an optional (default set to None) arg in fastmcp.server.report_progress().

Outside of test cases, no. Since the message field is optional, it shouldn't impact anything on release. 

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This is my first pull request to MCP (and honestly, to a codebase of this magnitude and quality). I’m incredibly excited to contribute, but also want to acknowledge that I might have missed some things. I’d really appreciate any extra scrutiny or tips reviewers can offer, especially around the test cases. Thanks in advance, and apologies if anything is off!

### DIFF
--- a/src/mcp/shared/progress.py
+++ b/src/mcp/shared/progress.py
@@ -47,7 +47,7 @@ class ProgressContext(
         self.current += amount
 
         await self.session.send_progress_notification(
-            self.progress_token, self.current, total=self.total, message=self.message
+            self.progress_token, self.current, total=self.total, message=message
         )
 
 
--- a/src/mcp/shared/progress.py
+++ b/src/mcp/shared/progress.py
@@ -78,7 +78,7 @@ def progress(
     if ctx.meta is None or ctx.meta.progressToken is None:
         raise ValueError("No progress token provided")
 
-    progress_ctx = ProgressContext(ctx.session, ctx.meta.progressToken, total, None)
+    progress_ctx = ProgressContext(ctx.session, ctx.meta.progressToken, total)
     try:
         yield progress_ctx
     finally:
--- a/src/mcp/shared/progress.py
+++ b/src/mcp/shared/progress.py
@@ -42,7 +42,6 @@ class ProgressContext(
     progress_token: ProgressToken
     total: float | None
     current: float = field(default=0.0, init=False)
-    message: str | None
 
     async def progress(self, amount: float, message: str | None = None) -> None:
         self.current += amount
--- a/src/mcp/shared/progress.py
+++ b/src/mcp/shared/progress.py
@@ -42,12 +42,13 @@ class ProgressContext(
     progress_token: ProgressToken
     total: float | None
     current: float = field(default=0.0, init=False)
+    message: str | None
 
     async def progress(self, amount: float) -> None:
         self.current += amount
 
         await self.session.send_progress_notification(
-            self.progress_token, self.current, total=self.total
+            self.progress_token, self.current, total=self.total, message=self.message
         )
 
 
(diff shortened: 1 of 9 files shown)


src/mcp/shared/progress.py
```diff
@@ -77,7 +78,7 @@ def progress(
     if ctx.meta is None or ctx.meta.progressToken is None:
         raise ValueError("No progress token provided")
 
-    progress_ctx = ProgressContext(ctx.session, ctx.meta.progressToken, total)
+    progress_ctx = ProgressContext(ctx.session, ctx.meta.progressToken, total, None)
```
ihrpr: Do we even need message in progress context?
```suggestion
    progress_ctx = ProgressContext(ctx.session, ctx.meta.progressToken, total)
```

src/mcp/shared/progress.py
```diff
@@ -42,12 +42,13 @@ class ProgressContext(
     progress_token: ProgressToken
     total: float | None
     current: float = field(default=0.0, init=False)
+    message: str | None
 
     async def progress(self, amount: float) -> None:
```
ihrpr:
```suggestion
    async def progress(self, amount: float, message: str | None = None) -> None:
```

src/mcp/shared/progress.py
```diff
@@ -42,12 +42,13 @@ class ProgressContext(
     progress_token: ProgressToken
     total: float | None
     current: float = field(default=0.0, init=False)
+    message: str | None
```
ihrpr:
```suggestion
```

src/mcp/shared/progress.py
```diff
@@ -42,12 +42,13 @@ class ProgressContext(
...
 
     async def progress(self, amount: float) -> None:
         self.current += amount
 
         await self.session.send_progress_notification(
-            self.progress_token, self.current, total=self.total
+            self.progress_token, self.current, total=self.total, message=self.message
```
ihrpr:
```suggestion
            self.progress_token, self.current, total=self.total, message=message
```

ihrpr (changes requested):
Thank you for adding the support for the message in Progress Notifications! 

Some comments: 
- Having the message parameter on the progress() method (not the context manager) would be more intuitive

It would look like
```python
async with progress(total=100) as p:
    await p.progress(10, message="Loading configuration...")
    await p.progress(30, message="Connecting to database...")
    await p.progress(40, message="Fetching data...")
    await p.progress(20, message="Processing results...")
```

would be great cover ProgressContext with tests as well

aksheyd:
That makes sense. I reviewed and committed your changes, and I'll add a test case for the ProgressContext as soon as possible.

aksheyd:
Added a test case for the progress context manager.

Looks like the 3.11 check failed on `test_188_concurrency.py` by fractions of a second. It passed locally for me, so not sure how to best approach this. Would it make sense to trigger a re-run with a blank commit?

tests/shared/test_progress_notifications.py
```diff
@@ -0,0 +1,358 @@
...
+                    raise ValueError("Server sending back incorrect progressToken")
+
+                # Send progress notifications
+                await serv_sesh.send_progress_notification(
+                    progress_token=progressToken,
+                    progress=0.25,
+                    total=1.0,
+                    message="Server progress 25%",
+                )
+                await anyio.sleep(0.2)
```
ihrpr: I don't think we need those sleep, do we?

ihrpr (approved):
thank you!

would be great to remove sleep in tests before merging

tests/shared/test_progress_notifications.py
aksheyd: Yea, we only really need the sleeps before closing the task groups, so I removed the unnecessary ones around the sending of the notifications, as they are already awaited.

ihrpr (approved):
Thank you!

PR #440, opened 2025-04-07
Fix: Use absolute path to uv executable in Claude Desktop config

arcAman07:
When calling 'mcp install', the Claude Desktop config is updated with just
'uv' as the command rather than the full path to the executable. This can
cause Claude Desktop to fail when it can't find uv in its PATH.

This change locates the absolute path of the uv executable and uses that
in the configuration, with a fallback to the plain 'uv' command if not found.

## Motivation and Context
This bug was causing significant frustration for developers using Claude Desktop integration with MCP servers. The problem was particularly difficult to diagnose for new users who might not understand why Claude Desktop couldn't find the uv command.
The fix ensures a more reliable developer experience by removing configuration-related obstacles that might block effective use of the MCP tools.

## How Has This Been Tested?
Yes tested it locally and it works properly.

## Breaking Changes
Nope

## Types of changes
- [✔ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist
- [ ✔] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [✔ ] My code follows the repository's style guidelines
- [✔ ] New and existing tests pass locally
- [ ✔] I have added appropriate error handling

### DIFF
--- a/src/mcp/cli/claude.py
+++ b/src/mcp/cli/claude.py
@@ -5,6 +5,7 @@ import os
 import sys
 from pathlib import Path
 from typing import Any
+import shutil
 
 from mcp.server.fastmcp.utilities.logging import get_logger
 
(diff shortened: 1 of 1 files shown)


ihrpr (approved):
thank you

PR #455, opened 2025-04-08
Add optional headers argument to websocket transport

andrewmjc:
This is a small PR to expose the additional_headers param on ws_connect through a new 'headers' argument in the websocket transport

## Motivation and Context

This enables users to pass through custom headers that the MCP server might require for auth, remote customization, version headers, or otherwise. Currently, the SSE transport supports settings headers, but the WS transport does not.

I am aware that the JS Websocket implementation only allows setting the Sec-WebSocket-Protocol header, so adding headers to all SDKs might be tricky, but I think it's better to expose the ability to set these headers where we can.

## How Has This Been Tested?

Will work the exact same as the current WS transport. The receiving server can now view the headers sent in the handshake request.

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/websocket.py
+++ b/src/mcp/client/websocket.py
@@ -50,7 +50,9 @@ async def websocket_client(
     write_stream, write_stream_reader = anyio.create_memory_object_stream(0)
 
     # Connect using websockets, requesting the "mcp" subprotocol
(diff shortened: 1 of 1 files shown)


ihrpr:
Closing this PR, looks stale now. Please re-open if the changes are still needed and suggestions applied.

Issue #444, opened 2025-04-07
Support JSON-RPC batching


jspahrsummers:

john0312:
Hello, I'm interested in implementing this, please let me know if someone else has been assigned to this, if not, I'll gradually open PR on this in the next few days if that's acceptable.

spacewander:
Excuse me, which PR implements the batch api?

Issue #445, opened 2025-04-07
Add support for tool annotations


jspahrsummers:

aksheyd:
Hi! I'd love to help implement tool annotation support based on the spec update. Totally understand if it's already being worked on or if someone more experienced is needed – just wanted to express interest!

PR #456, opened 2025-04-08
Upgrade types.py to 2025-03-26 protocol version

john0312:
This PR upgrades types.py to the 2025-03-26 protocol version.

## Motivation and Context
This is needed to support batch JSON RPC #444 

## How Has This Been Tested?
All tests and type check passed.

## Breaking Changes
None

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -29,7 +29,7 @@ for reference.
   not separate types in the schema.
 """
 
-LATEST_PROTOCOL_VERSION = "2024-11-05"
+LATEST_PROTOCOL_VERSION = "2025-03-26"
 
 ProgressToken = str | int
 Cursor = str
@@ -131,6 +131,12 @@ class JSONRPCNotification(Notification[dict[str, Any] | None, str]):
     params: dict[str, Any] | None = None
 
 
+class JSONRPCBatchRequest(RootModel[list[JSONRPCRequest | JSONRPCNotification]]):
+    """A JSON-RPC batch request"""
+
+    pass
+
+
 class JSONRPCResponse(BaseModel):
     """A successful (non-error) response to a request."""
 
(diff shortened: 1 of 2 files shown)


john0312:
The support for tool annotations #445 will probably benefit from this PR as well.

HeetVekariya:
@john0312 PR seems good.

- We do need to add tests for this annotation functionality same as added in the [Swift SDK](https://github.com/modelcontextprotocol/swift-sdk/pull/47) by @mattt .
- I would like to add tests for it, as in when the PR got merged. If you are working on it, we can collaborate for the same.

john0312:
@HeetVekariya I'm not a maintainer, so it is not my call to make on who gets to work on tool annotation tests. In fact, I'm not even certain if my PR will get reviewed/merged.

src/mcp/types.py
```diff
@@ -705,15 +748,77 @@ class ListToolsRequest(PaginatedRequest[RequestParams | None, Literal["tools/lis
     params: RequestParams | None = None
 
 
+class ToolAnnotations(BaseModel):
+    """
```
aksheyd, line 752: To set the optional bool default values, I think you would do:

```python3
    readOnlyHint: bool = False
    destructiveHint: bool = True
    idempotentHint: bool = False
    openWorldHint: bool = True
```

john0312: Looking through the rest of the code, the convention seems to be to set the default to None instead of False/True? I think the default convention makes sense because we want to tell if the other side specified True/False or left it absent.

Note that this PR probably conflicts with a later PR #482, in which the default is left as None as well.

Not sure if you've permission to merge, but if you do, it looks like PR #482 is likely to be merged, so maybe I'll leave the ToolAnnotations out so we can merge the rest?

eavanvalkenburg:
any updates on this PR, would love to get the spec up to the latest!

bracesproul:
Following up on this. Any idea when this will get merged & released?

john0312:
Unfortunately I'm unable to secure a code review from repository owners and thus unable to merge this. If anyone is able to persuade any repository owner to review and merge this, I'm would love to promptly and expeditiously get this merged.

Also, I believe some of the changes here has already been landed in various other PRs such as:
https://github.com/modelcontextprotocol/python-sdk/commit/1a330ac672c9237fe8f463cacb599b663f8926ad#diff-9b043ee327a542607e73b38ebb19c77049650efbe3a1f9d5891b0f8af09599ae

So maybe check if what you wanted is already in there.

PR #459, opened 2025-04-08
Add message queue for SSE messages POST endpoint

akash329d:
## Motivation and Context
This PR decouples session state from individual server instances by introducing a message dispatch abstraction. The previous implementation required client messages to be handled by the same server that initialized the session. Now, servers poll a shared message dispatch (with in-memory or Redis backend) to receive client messages, allowing any server instance to handle POST requests for any session to the message endpoint. 

## How Has This Been Tested?
All unit tests are passing, tested redis implementation locally. Added unit / integration tests as well for redis implementation.

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist


src/mcp/server/fastmcp/server.py
```diff
@@ -479,7 +484,29 @@ async def run_sse_async(self) -> None:
 
     def sse_app(self) -> Starlette:
         """Return an instance of the SSE server app."""
-        sse = SseServerTransport(self.settings.message_path)
+        message_queue = None
+        if self.settings.message_queue == "redis":
```
praboud-ant: The configuration should allow you to provide an instance of a message queue class, rather than picking from an enum of options provided by the SDK. Integrators should be able to use any implementation they want, as long as it conforms to the interface defined by the SDK.

We should still provide the in-memory & Redis implementations as nice out-of-the-box options, but they shouldn't be privileged in any particular way.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "anyio>=4.5",
     "httpx>=0.27",
     "httpx-sse>=0.4",
-    "pydantic>=2.7.2,<3.0.0",
+    "pydantic>=2.7.2,<=2.10.1",
```
praboud-ant: What's this constraint for?

uv.lock
```diff
@@ -1,10 +1,6 @@
 version = 1
-revision = 1
```
praboud-ant, line 2: There are a lot of seemingly unrelated changes in the `uv.lock` - are these necessary? Ideally, the diff here should just be to add the new dep on Redis.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,126 @@
...
+        self, session_id: UUID, timeout: float = 0.1
+    ) -> types.JSONRPCMessage | Exception | None:
+        """Get the next message for the specified session."""
+        # Check if session exists
+        if not await self.session_exists(session_id):
+            return None
+
+        # Pop from the left side of the list (queue)
+        # Use BLPOP with timeout to avoid busy waiting
+        result = await self._redis.blpop([self._session_queue_key(session_id)], timeout)  # type: ignore[attr-defined]
```
praboud-ant: I'm confused about this comment - doesn't using a timeout here mean that we do busy wait? This implementation will blpop, wait 100ms in redis, sleep for 10ms in the client, then retry.

More generally - redis pub/sub is probably a better fit for this, because we can avoid needing to have one redis connection + periodic polls per SSE request.

pyproject.toml
akash329d: The current tests fail with a Pydantic version above this, so its a valid constraint.

README.md
```diff
@@ -385,6 +385,30 @@ app.router.routes.append(Host('mcp.acme.corp', app=mcp.sse_app()))
 
 For more information on mounting applications in Starlette, see the [Starlette documentation](https://www.starlette.io/routing/#submounting-routes).
 
+#### Message Queue Options
+
+By default, the SSE server uses an in-memory message queue for incoming POST messages. For production deployments or distributed scenarios, you can use Redis:
```
praboud-ant: I think it's worth explicitly saying that anything that implements the MessageQueue protocol is fine; we've just provided the redis implementation as a default / as a convenience.

src/mcp/server/message_queue/base.py
```diff
@@ -0,0 +1,103 @@
...
+            session_id: The UUID of the session this message is for
+            message: The message to publish
+
+        Returns:
+            bool: True if message was published, False if session not found
+        """
+        ...
+
+    @asynccontextmanager
+    async def active_for_request(self, session_id: UUID, callback: MessageCallback):
```
praboud-ant: I think I'd call this `subscribe` or something; the current naming is not super clear about what this does.

src/mcp/server/message_queue/base.py
```diff
@@ -0,0 +1,103 @@
...
+
+import mcp.types as types
+
+logger = logging.getLogger(__name__)
+
+MessageCallback = Callable[[types.JSONRPCMessage | Exception], Awaitable[None]]
+
+
+@runtime_checkable
+class MessageQueue(Protocol):
```
praboud-ant: A naming / clarity nit - at this point, this has morphed from a queue to a dispatch system. The `InMemoryQueue` doesn't actually queue anywhere - all the queueing is handled by the caller in `sse.py`. I think `s/Queue/Dispatch` would make the intent a lot easier to understand.

src/mcp/server/sse.py
```diff
@@ -96,7 +101,12 @@ async def connect_sse(self, scope: Scope, receive: Receive, send: Send):
 
         session_id = uuid4()
         session_uri = f"{quote(self._endpoint)}?session_id={session_id.hex}"
-        self._read_stream_writers[session_id] = read_stream_writer
+
+        async def message_callback(message: types.JSONRPCMessage | Exception) -> None:
```
praboud-ant: Right now, this will block if the MCP server implementation doesn't consume `read_stream` quickly enough. For the in-memory dispatch implementation, this is basically fine, because that just causes the corresponding `POST /message` call to block. However, in the Redis implementation, it will block anything else from getting consumed from the Redis pubsub stream.

Fundamentally, something has to have a sized buffer (with size>0; the channels we have here are size=0, so messages block immediately), and writes to that buffer must fail once that buffer fills up. Moreover, we need to signal error to the client when this happens, and probably shut down. I'm not 100% sure on what the right way to signal that error to the client is. If the message is a request, we can return an error response... but if it's a notification, or a response, it's not clear to me how to signal to the client that something went wrong. `LoggingMessageNotification` is the closest thing I can find in the spec: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts. That doesn't seem like quite what this was meant for, though. I guess we could send an error log notification, then cancel the taskgroup / close the connection, but it feels a little bit wrong.

@jerome3o-anthropic: do you have thoughts on how the server can best signal to the client "ahh, fatal error, shutting down"?

As for how to implement this; I think what would make the most sense to me is to have the `read_stream` buffer be given some size (say, 32, but that's totally arbitrary), make the `message_callback` be a synchronous function, and call `read_stream_writer.send_nowait()`. If it errors out, we can then push another message to the SSE stream writer. (This should also be with nowait; we'll need to add a buffer to the SSE channel to accommodate this. If the nowait call fails, we should just silently move on - the client is probably jammed up somehow, and isn't going to get our message anyway.)We should just skip sending an error to the client if the client is backed up reading messages.) Once that's done, we can cancel the TG (though we'll need to split the TG used to send the SSE stream from the one that wraps the `yield` from this function.

This logic is a bit fiddly, so it's probably best to implement this here, rather than in the message dispatch we expect clients might implement variants of.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+        logger.debug(f"Initialized Redis message queue with URL: {redis_url}")
+
+    def _session_channel(self, session_id: UUID) -> str:
+        """Get the Redis channel for a session."""
+        return f"{self._prefix}session:{session_id.hex}"
+
+    @asynccontextmanager
+    async def active_for_request(self, session_id: UUID, callback: MessageCallback):
+        """Request-scoped context manager that ensures the listener task is running."""
+        await self._redis.sadd(self._active_sessions_key, session_id.hex)
```
praboud-ant: I think it's somewhat preferable to use separate keys for tracking the active sessions, rather than one big set value, so the set doesn't grow unboundedly.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+        channel = self._session_channel(session_id)
+        await self._pubsub.subscribe(channel)  # type: ignore
+
+        logger.debug(f"Registered session {session_id} in Redis with callback")
+        async with anyio.create_task_group() as tg:
+            tg.start_soon(self._listen_for_messages)
+            try:
+                yield
+            finally:
+                tg.cancel_scope.cancel()
```
praboud-ant: One kinda squirrelly part of this is that it's extremely important that the cancel doesn't happen after we've read a message off the Redis connection, otherwise the data will just get entirely lost. Right now, I think there's a failure mode here where one SSE stream being closed at the wrong moment will cause messages from another SSE stream to be dropped.

Looking through the Redis implementation, there are definitely multiple `async` calls within `get_message`, so it's possible for the bytes to be read off the connection, then the context to be cancelled, and the message dropped.

I think the way we have to structure this is either with multiple connections, or by wrapping the `get_message()` + dispatch calls in a `shield=True` scope: https://anyio.readthedocs.io/en/stable/cancellation.html#shielding. We'd then need to set a timeout on the `get_message` call, and just silently retry if the read times out. (This essentially gives us known safe points that we're OK being cancelled at.)

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+                    continue
+
+                data: str = cast(str, message["data"])
+                msg: None | types.JSONRPCMessage | Exception = None
+                try:
+                    json_data = json.loads(data)
+                    if isinstance(json_data, dict):
+                        json_dict: dict[str, Any] = json_data
+                        if json_dict.get("_exception", False):
+                            msg = Exception(
```
praboud-ant: This mangles the type of the exception, which we should probably avoid.

It occurs to me that the only type of exception we pass this way is a pydantic `ValidationError`, for invalid messages. I wonder whether it'd be simpler to just pass the underlying (invalid) `str` message through the dispatch in these cases, and get the error back when we call `model_validate_json`.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+                                f"{json_dict['type']}: {json_dict['message']}"
+                            )
+                        else:
+                            msg = types.JSONRPCMessage.model_validate_json(data)
+
+                    if msg and session_id in self._callbacks:
+                        await self._callbacks[session_id](msg)
```
praboud-ant: We should probably log if there's no matching callback

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+                    ignore_subscribe_messages=True,
+                    timeout=None,  # type: ignore
+                )
+                if message is None:
+                    continue
+
+                # Extract session ID from channel name
+                channel: str = cast(str, message["channel"])
+                if not channel.startswith(self._prefix):
+                    continue
```
praboud-ant: We should probably log in this case also

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+
+                session_hex = channel.split(":")[-1]
```
praboud-ant: This is slightly paranoid, but this validation doesn't necessarily check that the channel name is formatted as exactly `f"{self._prefix}session{id}"`, as we expect from `_session_channel`. I'd prefer something which will fail if the format is not exactly as we expect.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,143 @@
...
+                    session_id = UUID(hex=session_hex)
+                except ValueError:
+                    logger.error(f"Invalid session channel: {channel}")
+                    continue
+
+                data: str = cast(str, message["data"])
+                msg: None | types.JSONRPCMessage | Exception = None
+                try:
+                    json_data = json.loads(data)
+                    if isinstance(json_data, dict):
```
praboud-ant: We should log if `json_data` is not a dict; if you end up implementing my suggestion below re: passing through the underlying invalid message, I think this logic can go away entirely & be replaced with just `model_validate_json`.

src/mcp/server/message_queue/base.py
```diff
@@ -0,0 +1,103 @@
...
+class InMemoryMessageQueue:
+    """Default in-memory implementation of the MessageQueue interface.
+
+    This implementation immediately calls registered callbacks when messages
+    are received.
+    """
+
+    def __init__(self) -> None:
+        self._callbacks: dict[UUID, MessageCallback] = {}
+        self._active_sessions: set[UUID] = set()
```
praboud-ant: `_callbacks` & `_active_sessions` duplicate the same state - I think we can just get rid of the `_active_sessions` variable.

praboud-ant (commented):
ptal @akash329d

src/mcp/server/sse.py
akash329d: Is the observed functionality here different between the Redis implementation vs InMemory? 

Wouldn't we block at `message_callback` in both cases? Or are you saying this is fine in the InMemory case because we are also blocking the `POST` response? Not sure what guarantees there are in the SSE protocol for how fast the response needs to be after the `POST`.

src/mcp/server/message_queue/redis.py
akash329d: Nice idea, yeah. That makes it cleaner 👍

src/mcp/server/message_queue/redis.py
akash329d: It seems like the max # of redis keys & the max number of members in a set are equal, so I think there should be no difference in practice? https://redis.io/docs/latest/develop/data-types/sets/

nickste:
Just a quick note to say thanks for this PR! We hadn't been able to deploy the MCP server behind a load balancer because of the session state coupling. I've deployed this PR behind an AWS NLB and ECS cluster and confirmed it has fixed the session issue for us.

nickste:
FWIW - I've noticed that I'm now getting this error in logs when using your latest commit:
```
Could not find session for ID: [session-id]
```
If I instead use commit `7cabcea26a02d02b3a2f964c45f70681a5dd6ded` I don't get the error. When trying the latest commit, I did change naming from `RedisMessageQueue` to `RedisMessageDispatch`. Let me know if I can be helpful with any other debugging.

praboud-ant:
@nickste huh! It's awesome to get feedback on this PR, though I feel like I should caution that the PR is still being actively worked on, and probably isn't ready for production use yet.

re: the `Could not find session for ID` bug above - are you seeing this for all requests, or is it seemingly a random subset?

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,254 @@
...
+        logger.debug(
+            f"Message {message_id} published to Redis channel for session {session_id}"
+        )
+        return message_id
+
+    async def publish_message_sync(
+        self,
+        session_id: UUID,
+        message: types.JSONRPCMessage | str,
+        timeout: float = 120.0,
```
akash329d: Might need to think about a reasonable timeout for this. In the default non-redis implementation its effectively infinite.

src/mcp/server/sse.py
akash329d: I think I have a potential solution here; we have a higher level task group take care of the handler logic (which is separate from the listener task group). So once the listener task group completes; we still allow all the handlers to finish out. This allows us to follow the same behavior as the original in-memory implementation with an essentially 'infinite' sized buffer. 

Let me know what you think; alternatively we can go to a fixed size stream_buffer and throw errors back to the message POST if we exceed that. Initially attempted this approach, but there wasn't a straightforward way to plumb through the ACKs on handler completion.

dschonholtz:
Hey i know this isn't done, but just want to let you know I have a full team of folks excited for this one. It's blocking our adoption internally.
Keep up the great work! Excited to see this one go in

akash329d:
Talked offline with @praboud-ant. We discovered that the current implementation is non-blocking, and responds to message POSTs immediately (without waiting on stream write). I adjusted to go back to the previous implementation without ACKs, so we now immediately respond to the POST.

src/mcp/server/message_queue/redis.py
```diff
@@ -0,0 +1,167 @@
...
+                    data: str = cast(str, message["data"])
+                    try:
+                        if session_id in self._task_groups:
+                            self._task_groups[session_id].start_soon(
+                                self._handle_message, session_id, data
+                            )
+                        else:
+                            logger.warning(
+                                f"Message dropped: no task group for session: {session_id}"
+                            )
```
akash329d: Adjusted to push the message handling onto the task group for that given session. This is so that we don't block the message dispatch listener on stream writing in the callback (since that is blocking). 

This should be fine, as that task group will only be cancelled if the SSE connection for the corresponding session disconnects (at which point we'd be fine cancelling the stream write since there is nothing to respond to).

src/mcp/server/message_queue/redis.py
praboud-ant: Huh, I guess that seems fine to me, then.

src/mcp/server/sse.py
praboud-ant: Yeah, this is more elegant. I'm a little wary of arbitrarily deeply queueing tasks, but I think in practice it's probably fine. We can always come back to this later if becomes a problem in practice.

praboud-ant (commented):
This is looking great! All of the code looks logically right to me - I think the one remaining thing that we should probably do is to add some test coverage for the Redis dispatch code. This logic is fairly complicated, and having to test it manually isn't ideal. [FakeRedis](https://fakeredis.readthedocs.io/en/latest/) is a pretty good way to write these sorts of tests, so we don't actually have to run a separate Redis container.

akash329d:
Thanks for reporting the bug! I did some debugging and I think I have the root cause figured out.

When you shutdown the MCP server and restart it, MCP Inspector attempts to auto-reconnect the SSE session on a cadence. When this happens, it initiates a new SSE connection (getting a new session ID), but it doesn't perform a re-initialization of the SSE connection.

In the in-memory (non-Redis) implementation, this causes an error when you try to do a tool call/query: `Received request before initialization was complete.`

However, in the Redis implementation there's a race condition. The same RuntimeError on initialization completion triggers the `finally` clause in the `subscribe` context manager, which cancels the task group. This ends up swallowing that exception and instead we get the `TypeError: 'NoneType' object is not callable exception` (which happens because we're not returning anything from the connect_sse method).

The fact that we haven't noticed this before also exposes a separate bug: with the current connect_sse method, it never actually exits. Even when the EventSourceResponse task closes on SSE disconnect, the request just stays active because we have no mechanism to kill it.

That's what I've found so far - will try to have a fix out soon.

For a quick explainer on the Exception swallowing, here's a repro which I think explains it pretty well:
```python
import anyio

async def failing_task():
    await anyio.sleep(0.5) 
    print("Task: raising exception...")
    raise ValueError("Exception from spawned task!")

async def main():    
        try:
            tg.start_soon(failing_task)
        finally:
            tg.cancel_scope.cancel()
            await anyio.sleep(10)
            print("In finally block - cancelling task group...")
    
    print("Done!")

if __name__ == "__main__":
    anyio.run(main)
```

The result of this is the program completing successfully, but not printing the `"In finally block - cancelling task group..."`. With my prior understanding I thought that print would happen 😅 .

praboud-ant (approved):

ihrpr:
Reverting in https://github.com/modelcontextprotocol/python-sdk/pull/649 
- In spec version 2025-03-26, Streamable HTTP transport supersedes the HTTP+SSE transport from version 2024-11-05
- We're bringing the Python SDK to the new spec version (2025-03-26), which means focusing on Streamable HTTP rather than enhancing the legacy transport, unless it's a fix
- Adding Redis integration to SSE creates maintenance burden for a transport being superseded. It would be still useful to have an MCP Server with Redis but as an example so users can reference it.

PR #482, opened 2025-04-10
Add ToolAnnotations support in FastMCP and lowlevel servers

bhosmer-ant:
## Summary
- Add proper support for ToolAnnotations in FastMCP and lowlevel servers
- Update test_tool_annotations_in_fastmcp to access ToolAnnotations fields as properties rather than dictionary keys
- Add test_lowlevel_server_tool_annotations to verify Tool annotations functionality in lowlevel server
- Update server.py to convert dictionary annotations to ToolAnnotations objects

## Test plan
- Run all tests with `uv run pytest` to verify changes work correctly
- Verify tool annotations are properly accessed and validated in both server implementations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dev = [
     "pytest-flakefinder>=1.1.0",
     "pytest-xdist>=3.6.1",
     "pytest-examples>=0.0.14",
+    "pytest-pretty>=1.2.0",
 ]
 docs = [
     "mkdocs>=1.6.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,7 +114,5 @@ filterwarnings = [
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
-    # this is a problem in starlette 0.27, which we're currently pinned to
-    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "starlette>=0.27",
     "sse-starlette>=1.6.1",
     "pydantic-settings>=2.5.2",
-    "uvicorn>=0.23.1",
+    "uvicorn>=0.23.1; sys_platform != 'emscripten'",
 ]
 
 [project.optional-dependencies]
@@ -89,8 +89,8 @@ venv = ".venv"
 strict = ["src/mcp/**/*.py"]
 
 [tool.ruff.lint]
-select = ["E", "F", "I", "UP"]
-ignore = []
+select = ["C4", "E", "F", "I", "PERF", "UP"]
+ignore = ["PERF203"]
 
 [tool.ruff]
 line-length = 88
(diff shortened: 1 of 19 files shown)


src/mcp/server/fastmcp/server.py
```diff
@@ -172,12 +172,17 @@ def _setup_handlers(self) -> None:
 
     async def list_tools(self) -> list[MCPTool]:
         """List all available tools."""
+        from mcp.types import ToolAnnotations
```
ihrpr: this needs to be moved to the top to imports

src/mcp/server/fastmcp/tools/base.py
```diff
@@ -30,6 +30,9 @@ class Tool(BaseModel):
     context_kwarg: str | None = Field(
         None, description="Name of the kwarg that should receive context"
     )
+    annotations: dict[str, Any] | None = Field(
```
ihrpr: Can we use ToolAnnotations directly here, why do we need dictionary? I think it would be a better user experience:
- Users would get proper type checking when providing annotations
- Better IDE support
- Users don't need to reference documentation to know which exactly field is available

```suggestion
    annotations: ToolAnnotations | None = Field(
```

ihrpr (changes requested):
Just to check why we need dictionary and can be have annotations typed for users?

src/mcp/server/fastmcp/tools/base.py
bhosmer-ant: [voice of Claude] You're absolutely right! I'll use ToolAnnotations directly here, that would be a better user experience. 

idk who's above blaming the LLM but it aint me 😬 😬 😬

pyproject.toml
```diff
@@ -89,8 +89,8 @@ venv = ".venv"
 strict = ["src/mcp/**/*.py"]
 
 [tool.ruff.lint]
-select = ["E", "F", "I", "UP"]
-ignore = []
+select = ["C4", "E", "F", "I", "PERF", "UP"]
```
ihrpr: Same here, is this intentional?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "starlette>=0.27",
     "sse-starlette>=1.6.1",
     "pydantic-settings>=2.5.2",
-    "uvicorn>=0.23.1",
+    "uvicorn>=0.23.1; sys_platform != 'emscripten'",
```
ihrpr: is this intentional?

bhosmer-ant: argh no these are strays, will remove when I merge `main` here in a sec

bhosmer-ant: ...or rather, they'll remove themselves from the diff - these changes are in `main` I think

bhosmer-ant:
@ihrpr ok requested changes should be in (really good change, thanks a bunch) and up to date w/main. 

Only change to config is the warning suppression - still a mystery why everybody isn't seeing the test failures with starlette 0.27 and no suppression, but I'm happy to take it out of this PR if you think that's cleaner.

src/mcp/server/fastmcp/tools/base.py
```diff
@@ -38,9 +42,10 @@ def from_function(
         name: str | None = None,
         description: str | None = None,
         context_kwarg: str | None = None,
+        annotations: ToolAnnotations | None = None,
     ) -> Tool:
         """Create a Tool from a function."""
-        from mcp.server.fastmcp import Context
+        from mcp.server.fastmcp.server import Context
```
ihrpr, line 48: what do we need to change this? I see the problem, but why do we even need it here?
on top of the file there is 

```

if TYPE_CHECKING:
```

ihrpr:
> @ihrpr ok requested changes should be in (really good change, thanks a bunch) and up to date w/main.
> 
> Only change to config is the warning suppression - still a mystery why everybody isn't seeing the test failures with starlette 0.27 and no suppression, but I'm happy to take it out of this PR if you think that's cleaner.

yes please, let's take it out before merging. Separately let's look why there is that warning. I checked out this PR and run all tests without the warning suppression and all works fine (CI should also give us a signal)

ihrpr (approved):
Let's remove warning suppression before merging

src/mcp/server/fastmcp/tools/base.py
bhosmer-ant: There's this a little further down in the function:
```
    if issubclass(param.annotation, Context):
```
(and having it up at the top introduces a circular dependency)

bhosmer-ant:
@ihrpr ok killed the warning suppression, can I ask you to reapprove, sorry 😬

ihrpr (approved):
Looks good!

Issue #391, opened 2025-03-29
Proactor Windows warning on cleanup when using stdio_client


S4mpl3r:
**Describe the bug**  
When connecting to a sdio MCP server on Windows, a warning is printed during the final cleanup. This does not crash the program but indicates that there might be a need for a more graceful shutdown of subprocess transports.

**To Reproduce**  
Run the following script:
   ```python
   import asyncio

   from mcp import ClientSession, StdioServerParameters, stdio_client


   async def main():
       server_params = StdioServerParameters(
           command="npx", args=["-y", "@playwright/mcp@latest"]
       ) # Or any other server

       async with stdio_client(server_params) as (read, write):
           async with ClientSession(read, write) as session:
               await session.initialize()
               print(await session.list_tools())
               await asyncio.sleep(2.0)


   if __name__ == "__main__":
       asyncio.run(main())
   ```
Running the above, we get a bunch of warnings and error messages at the end:
   ```
   Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x...>
   Traceback (most recent call last):
     ...
   ValueError: I/O operation on closed pipe
   ```

**Expected behavior**  
No warnings or errors should be printed during final cleanup. All subprocess transports and associated resources should be closed cleanly without attempting operations on closed pipes.

**Desktop:**  
- OS: Windows 10   
- Version: Python 3.11 (mcp v1.6.0)

t94j0:
I'm running into the same thing. From my attempt to debug, it looks like the `finally` branch of the app lifespan is never called

ashrobertsdragon:
This is an old asyncio bug with the Windows Proactor event loop [https://github.com/python/cpython/issues/83413#issue-1199028070](url). They thought they fixed it a couple years ago, but apparently there are a still a few cases where it still slips through. I can confirm it happening with Windows 11 and Python 3.13 as well.
Suppressing the ValueError when closing the session seems to do the trick. So, for your example, it would probably look like this:
```py
import asyncio
from contextlib import supress

from mcp import ClientSession, StdioServerParameters, stdio_client


async def main():
    server_params = StdioServerParameters(
        command="npx", args=["-y", "@playwright/mcp@latest"]
    ) # Or any other server

    with supress(ValueError):
        async with stdio_client(server_params) as (read, write):
            async with ClientSession(read, write) as session:
                await session.initialize()
                print(await session.list_tools())
                await asyncio.sleep(2.0)


if __name__ == "__main__":
[...]

PR #496, opened 2025-04-12
Fix process termination in stdio client

ashrobertsdragon:
**Github-Issue:** #391

- Added `tg.cancel_scope.cancel()` in the `finally` block of `stdio_client` to ensure that `stdin_writer` and `stdout_reader` tasks are properly cancelled  
- This ensures associated transports are closed before invoking `process.terminate()`  
- Removed the now-unnecessary `terminate_windows_process()` function

## Motivation and Context  
This change addresses a race condition specific to Windows' Proactor event loop, where ignored exceptions would occur during shutdown of the stdio client.  

The underlying problem was that subprocess transports (stdin/stdout) remained open when `process.terminate()` was called. On Windows, this could trigger a race during garbage collection and transport teardown, resulting in uncatchable exceptions.  

The previous workaround, `terminate_windows_process()`, attempted to mitigate this by waiting before forcefully terminating the process. However, on Windows, `kill()` is functionally identical to `terminate()`([https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.kill](url)), so the method provided no real benefit. Explicitly cancelling the stdio tasks ensures a clean and predictable shutdown.

## How Has This Been Tested?  
Tested on Windows 11 by repeatedly connecting and disconnecting multiple clients over stdio transport to multiple servers found both within the repository and using installed locally.  

The ignored exception could not be captured via standard exception handling (`try/except`, `sys.unraisablehook`, or pytest's `capsys` / `capfd`). As a result, an automated test was not feasible for this specific issue. All testing was conducted through repeated manual runs and verification of clean shutdown behavior.

## Breaking Changes  
No breaking changes are expected. The removed function, `terminate_windows_process()`, was internal and not designed for external use. Unless it was being used directly (which would be a misuse), no user code should be affected.

## Types of changes  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Documentation update

## Checklist  

## Additional context  
There are two failing tests that predate this change:  
- `test_messages_are_executed_concurrently` fails with a timing difference of less than 5ms  
- `test_desktop` fails due to assumptions about POSIX-style file paths  

No changes to documentation or error handling were necessary for this fix.

### DIFF
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -6,17 +6,16 @@ from typing import Literal, TextIO
 
 import anyio
 import anyio.lowlevel
-from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+from anyio.streams.memory import (
+    MemoryObjectReceiveStream,
+    MemoryObjectSendStream,
+)
 from anyio.streams.text import TextReceiveStream
 from pydantic import BaseModel, Field
 
 import mcp.types as types
 
-from .win32 import (
-    create_windows_process,
-    get_windows_executable_command,
-    terminate_windows_process,
-)
+from .win32 import create_windows_process, get_windows_executable_command
 
 # Environment variables to inherit by default
 DEFAULT_INHERITED_ENV_VARS = (
@@ -173,10 +172,9 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
             yield read_stream, write_stream
         finally:
             # Clean up process to prevent any dangling orphaned processes
-            if sys.platform == "win32":
-                await terminate_windows_process(process)
-            else:
-                process.terminate()
+            tg.cancel_scope.cancel()
+            process.terminate()
+            await process.wait()
 
 
 def _get_executable_command(command: str) -> str:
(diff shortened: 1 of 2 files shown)


ashrobertsdragon:
@jerome3o-anthropic @jspahrsummers This is my first pull request. I know am supposed to add the two of you as reviewers, but I haven't been able to figure out how to do that. I never got the gear icon that is shown in online tutorials

src/mcp/client/stdio/__init__.py
```diff
@@ -173,10 +172,9 @@ async def stdin_writer():
             yield read_stream, write_stream
         finally:
             # Clean up process to prevent any dangling orphaned processes
-            if sys.platform == "win32":
-                await terminate_windows_process(process)
-            else:
-                process.terminate()
+            tg.cancel_scope.cancel()
+            process.terminate()
```
cristipufu: This won't work if the process doesn't react to SIGTERM. We need a timeout and a fallback on SIGKILL.
Please test it with https://github.com/modelcontextprotocol/servers/blob/main/src/everything/index.ts#L13 (windows & unix)

cristipufu: Ref: https://github.com/modelcontextprotocol/python-sdk/pull/555

ihrpr:
Thank you for your contribution! 

Sorry for delays in reviews. There are a few PRs solving the same issue. Let's continue discussion in https://github.com/modelcontextprotocol/python-sdk/pull/555

PR #512, opened 2025-04-14
Python lint: Ruff rules for comprehensions and performance

cclauss:
Fix the Python lint issues raised by:
% `ruff check --select=C4,PERF --ignore=PERF203--statistics | sort -k2`
```
4	C408    	unnecessary-collection-call
1	C416    	unnecessary-comprehension
1	PERF102 	incorrect-dict-iterator
2	PERF401 	manual-list-comprehension
Found 8 errors.
No fixes available (6 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
## Motivation and Context

These proposed changes can make code easier to reason about or faster or both.

% `ruff rule PERF401`
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

Fix is sometimes available.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

Take care that if the original for-loop uses an assignment expression
as a conditional, such as `if match:=re.match("\d+","123")`, then
the corresponding comprehension must wrap the assignment
expression in parentheses to avoid a syntax error.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

https://docs.astral.sh/ruff

### DIFF
--- a/examples/clients/simple-chatbot/mcp_simple_chatbot/main.py
+++ b/examples/clients/simple-chatbot/mcp_simple_chatbot/main.py
@@ -122,8 +122,10 @@ class Server:
 
         for item in tools_response:
             if isinstance(item, tuple) and item[0] == "tools":
-                for tool in item[1]:
-                    tools.append(Tool(tool.name, tool.description, tool.inputSchema))
+                tools.extend(
+                    Tool(tool.name, tool.description, tool.inputSchema)
+                    for tool in item[1]
+                )
 
         return tools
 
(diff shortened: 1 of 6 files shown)


Kludex (approved):

Kludex:
Thanks :)

PR #515, opened 2025-04-15
Properly clean up response streams in BaseSession

### LINKED ISSUE #169: Leaking STDIO client: Unclosed <MemoryObjectReceiveStream>
**Describe the bug**

While running the MCP stdio client I noticed that my application was logging a warning from the `anyio` library, specifically related to unclosed streams.

When I dug in further I noticed that the `stdio` client does not close any of the streams after completing, see the [following link](https://github.com/modelcontextprotocol/python-sdk/blob/d64b0679b86b647682ebb0bd1b0d9ad96dfa0b1e/src/mcp/client/stdio.py#L153).

This is as opposed to the `sse_client` which [does close the streams.](https://github.com/modelcontextprotocol/python-sdk/blob/d64b0679b86b647682ebb0bd1b0d9ad96dfa0b1e/src/mcp/client/sse.py#L141-L142)

Here is a slightly scrubbed more detailed error:

```
/home/user/src/microsoft/autogen/python/.venv/lib/python3.12/site-packages/anyio/streams/memory.py:183: ResourceWarning: Unclosed <MemoryObjectReceiveStream at 7f5394255fa0>
  warnings.warn(
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/user/src/microsoft/autogen/python/.venv/lib/python3.12/site-packages/anyio/streams/memory.py:313: ResourceWarning: Unclosed <MemoryObjectSendStream at 7f5394256270>
  warnings.warn(
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/user/src/microsoft/autogen/python/.venv/lib/python3.12/site-packages/anyio/streams/memory.py:183: ResourceWarning: Unclosed <MemoryObjectReceiveStream at 7f5394256de0>
  warnings.warn(
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

The context here is that I am trying to add support for MCP to microsoft/autogen but this is most likely a blocker.

**To Reproduce**

As mentioned, I am running this code from another project, but this should be easily reproducible with a small example, I will try to do that and update if I find anything.

**Expected behavior**

No warnings should be logged as these streams should be properly closed

**Screenshots**

bhosmer-ant:
## Summary
- Ensures that response streams are properly closed and removed from tracking dictionary by wrapping request handling in a try/finally block
- Prevents memory leaks from response streams and their readers accumulating in the exit stack over time
- Adds a test to verify proper cleanup when exceptions occur during requests
- Adds a pytest warning filter needed for `starlette` 0.27 (pinned), without which tests fail

## Test plan
- Run the existing test suite
- Added a specific unit test that verifies stream cleanup during exceptions
- Manual testing with BaseSession and derived classes

May fix #169

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "starlette>=0.27",
     "sse-starlette>=1.6.1",
     "pydantic-settings>=2.5.2",
-    "uvicorn>=0.23.1",
+    "uvicorn>=0.23.1; sys_platform != 'emscripten'",
 ]
 
 [project.optional-dependencies]
@@ -89,8 +89,8 @@ venv = ".venv"
 strict = ["src/mcp/**/*.py"]
 
 [tool.ruff.lint]
-select = ["E", "F", "I", "UP"]
-ignore = []
+select = ["C4", "E", "F", "I", "PERF", "UP"]
+ignore = ["PERF203"]
 
 [tool.ruff]
 line-length = 88
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "httpx>=0.27",
     "httpx-sse>=0.4",
     "pydantic>=2.7.2,<3.0.0",
-    "starlette>=0.46.2",
+    "starlette>=0.27",
     "sse-starlette>=1.6.1",
     "pydantic-settings>=2.5.2",
     "uvicorn>=0.23.1",
@@ -53,7 +53,6 @@ dev = [
     "pytest-flakefinder>=1.1.0",
     "pytest-xdist>=3.6.1",
     "pytest-examples>=0.0.14",
-    "starlette>=0.46.2",
 ]
 docs = [
     "mkdocs>=1.6.1",
@@ -114,5 +113,7 @@ filterwarnings = [
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
+    # this is a problem in starlette 0.27, which we're currently pinned to
+    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,5 @@ filterwarnings = [
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
-    # this is a problem in starlette 0.27, which we're currently pinned to
-    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "httpx>=0.27",
     "httpx-sse>=0.4",
     "pydantic>=2.7.2,<3.0.0",
-    "starlette>=0.27",
+    "starlette>=0.46.2",
     "sse-starlette>=1.6.1",
     "pydantic-settings>=2.5.2",
     "uvicorn>=0.23.1",
@@ -53,6 +53,7 @@ dev = [
     "pytest-flakefinder>=1.1.0",
     "pytest-xdist>=3.6.1",
     "pytest-examples>=0.0.14",
+    "starlette>=0.46.2",
 ]
 docs = [
     "mkdocs>=1.6.1",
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "httpx>=0.27",
     "httpx-sse>=0.4",
     "pydantic>=2.7.2,<3.0.0",
-    "starlette>=0.27",
+    "starlette>=0.46.2",
```
Kludex: I was told by @dsp-ant that you don't want to do that in Anthropic because of internal packages or something?

pyproject.toml
```diff
@@ -53,6 +53,7 @@ dev = [
     "pytest-flakefinder>=1.1.0",
     "pytest-xdist>=3.6.1",
     "pytest-examples>=0.0.14",
+    "starlette>=0.46.2",
```
Kludex: Not needed in two places.
```suggestion
```

src/mcp/shared/session.py
```diff
@@ -230,42 +229,45 @@ async def send_request(
...
+
+            if isinstance(response_or_error, JSONRPCError):
+                raise McpError(response_or_error.error)
+            else:
+                return result_type.model_validate(response_or_error.result)
+
+        finally:
+            self._response_streams.pop(request_id, None)
+            await response_stream.aclose()
+            await response_stream_reader.aclose()
```
Kludex, line 275: Is this correct tho? Do you want to close the stream on every request? 🤔

pyproject.toml
bhosmer-ant: Ah ok, will check w/him - without the upgrade the new test gets the multipart warning we were seeing in #255

pyproject.toml
bhosmer-ant: Think it was giving me the multipart error without the two places but will verify (if we upgrade)

src/mcp/shared/session.py
bhosmer-ant: Well, we're [creating](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/shared/session.py#L228-L230) them on every request, so I think the old code was just piling 
up open streams until the end of the session (or getting unclosed resource errors if they got `__del__`d before the session's `__aexit__`, or if the latter never ran).

Please check my logic though - I'm new to this code. E.g. is there a reason we need to leave streams open after [getting the response](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/shared/session.py#L252)?

Kludex: Yeah, but the connection stays open for some time after you send back the response.

bhosmer-ant: Ah ok, sounds like this would need a more sophisticated solution. I'll move this to draft and come back to it. Thanks!

Kludex: By default, it stays alive for 5 secs (https://www.uvicorn.org/settings/#timeouts) fyi

bhosmer-ant: @Kludex just to clarify - these are in-memory streams, not network streams. (`_receive_loop` gets the network response and writes it to `response_stream`). There may still be a problem I'm not aware of with closing the streams immediately, but want to make sure you weren't worried about closing while _network_ connections were still open.

[edit: aha wrote this before seeing your fyi, so it sounds like you were worried about closing HTTPS streams, right? undrafting based on this assumption]

pyproject.toml
ihrpr: yeah, we can't upgrade it now. @bhosmer-ant do you know what exactly helps in the upgrade, maybe we can find another way?

ihrpr (changes requested):
can't upgrade starlette for now

pyproject.toml
bhosmer-ant: > yeah, we can't upgrade it now. @bhosmer-ant do you know what exactly helps in the upgrade, maybe we can find another way?

The current version of `starlette` gives me this error on every test when I run `uv run pytests`:
```
ERROR tests/client/test_config.py - PendingDeprecationWarning: Please use `import python_multipart` instead.
```

@praboud-ant ran into this in #255 and added the following to `filterwarnings` in `pyproject.toml`:
```
"ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
```
...unfortunately I suggested upgrading and he took me up on it, he'll need to back it out too. Sorry Peter :(

I can add the `filterwarning` here when I back out the upgrade, but I'm curious: are tests succeeding for you on `main`? They're all failing for me with this warning.

bhosmer-ant:
@ihrpr thanks for taking a look! Backed out the starlette bump, but as a result needed to add a warning suppression to get tests to pass (they're failing on `main` as well for me, not sure why they haven't been blowing up for you?)

pyproject.toml
```diff
@@ -113,5 +113,7 @@ filterwarnings = [
     # This should be fixed on Uvicorn's side.
     "ignore::DeprecationWarning:websockets",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
-    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel"
+    "ignore:Returning str or bytes.*:DeprecationWarning:mcp.server.lowlevel",
+    # this is a problem in starlette 0.27, which we're currently pinned to
+    "ignore:Please use `import python_multipart` instead.:PendingDeprecationWarning",
```
ihrpr: tried running test without this change, all works for me, lets' remove this

bhosmer-ant: done!

src/mcp/shared/session.py
```diff
@@ -187,7 +187,6 @@ def __init__(
         self._receive_notification_type = receive_notification_type
         self._session_read_timeout_seconds = read_timeout_seconds
         self._in_flight = {}
-
```
ihrpr, line 190: nit - is this intentional?

ihrpr (approved):
Looks good!

PR #525, opened 2025-04-15
Python lint: Ruff rules for pylint and code complexity

cclauss:
Add ruff rules for Pylint and McCabe cyclomatic code complexity.
* https://docs.astral.sh/ruff/rules/#pylint-pl
* https://docs.astral.sh/ruff/rules/#mccabe-c90

% `ruff check --select=C90,PL --statistics | sort -k2`
```
 5	C901   	[ ] complex-structure
 1	PLC0414	[ ] useless-import-alias
19	PLR0402	[*] manual-from-import
 1	PLR0911	[ ] too-many-return-statements
 1	PLR0912	[ ] too-many-branches
 5	PLR0913	[ ] too-many-arguments
 1	PLR0915	[ ] too-many-statements
26	PLR2004	[ ] magic-value-comparison
 1	PLW1510	[*] subprocess-run-without-check
[*] 20 fixable with the `--fix` option (1 hidden fix can be enabled with the `--unsafe-fixes` option).
Found 60 errors.
```

## Motivation and Context

Pylint checks for errors, enforces a coding standard, looks for code smells, and can suggest how code could be refactored.
McCabe detects functions with high complexity that are hard to understand and maintain.

The C90 and PLR091 rules enable the setting of upper limits on code complexity.  If a contributor wants to raise these limits, that can lead to useful discussion in code reviews about why complexity needs to rise and maintainability needs to go down.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,18 +102,36 @@ executionEnvironments = [
   { root = "examples/servers", reportUnusedFunction = false },
 ]
 
-[tool.ruff.lint]
-select = ["C4", "E", "F", "I", "PERF", "UP"]
-ignore = ["PERF203"]
-
 [tool.ruff]
 line-length = 120
 target-version = "py310"
 extend-exclude = ["README.md"]
 
+[tool.ruff.lint]
+select = [
+    "C4",    # flake8-comprehensions
+    "C90",   # mccabe
+    "E",     # pycodestyle
+    "F",     # pyflakes
+    "I",     # isort
+    "PERF",  # Perflint
+    "PL",    # Pylint
+    "UP",    # pyupgrade
+]
+ignore = ["PERF203", "PLC0415", "PLR0402"]
+mccabe.max-complexity = 24  # Default is 10
+
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]
 "tests/server/fastmcp/test_func_metadata.py" = ["E501"]
+"tests/shared/test_progress_notifications.py" = ["PLW0603"]
+
+[tool.ruff.lint.pylint]
+allow-magic-value-types = ["bytes", "float", "int", "str"]
+max-args = 23         # Default is 5
+max-branches = 23     # Default is 12
+max-returns = 13      # Default is 6
+max-statements = 102  # Default is 50
 
 [tool.uv.workspace]
 members = ["examples/servers/*", "examples/snippets"]
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -85,7 +85,6 @@ import mcp.types as types
 from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.models import InitializationOptions
 from mcp.server.session import ServerSession
-from mcp.server.stdio import stdio_server as stdio_server
 from mcp.shared.context import RequestContext
 from mcp.shared.exceptions import McpError
 from mcp.shared.message import ServerMessageMetadata, SessionMessage
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -182,9 +182,8 @@ class FastMCP(Generic[LifespanResultT]):
                 raise ValueError("Cannot specify both auth_server_provider and token_verifier")
             if not auth_server_provider and not token_verifier:
                 raise ValueError("Must specify either auth_server_provider or token_verifier when auth is enabled")
-        else:
-            if auth_server_provider or token_verifier:
-                raise ValueError("Cannot specify auth_server_provider or token_verifier without auth settings")
+        elif auth_server_provider or token_verifier:
+            raise ValueError("Cannot specify auth_server_provider or token_verifier without auth settings")
 
         self._auth_server_provider = auth_server_provider
         self._token_verifier = token_verifier
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -139,14 +139,14 @@ class FuncMetadata(BaseModel):
             if field_info.alias:
                 key_to_field_info[field_info.alias] = field_info
 
-        for data_key in data.keys():
+        for data_key, data_value in data.items():
             if data_key not in key_to_field_info:
                 continue
 
             field_info = key_to_field_info[data_key]
-            if isinstance(data[data_key], str) and field_info.annotation is not str:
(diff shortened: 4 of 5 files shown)


tests/client/test_config.py
```diff
@@ -44,7 +44,9 @@ def test_command_execution(mock_config_path: Path):
 
     test_args = [command] + args + ["--help"]
 
-    result = subprocess.run(test_args, capture_output=True, text=True, timeout=5)
+    result = subprocess.run(
```
ihrpr: this probably be better as a separate PR

ihrpr (changes requested):
Thank you for contributing to Python SDK!

1. Please can you provide rationale for the specific numerical thresholds chosen
2. Is implementing both C90 and PL linting rules simultaneously necessary?
3. Please remove the widespread import style changes which aren't central to the linting goals

The subprocess check=False change is great, but should be submitted as a separate PR focused on that specific fix.

tests/client/test_config.py
cclauss: Done in:
* #794

cclauss:
Please read the git commit when reviewing pull requests.
> Please can you provide rationale for the specific numerical thresholds chosen

>> The C90 and PLR091 rules enable the setting of upper limits on code complexity. If a contributor wants to raise these limits, that can lead to useful discussion in code reviews about why complexity needs to rise and maintainability needs to go down.

  The current code complexity determines the values set in this PR, with comments showing the recommended values.  To find complexity in the codebase, just lower any of these values by one and run `ruff check`.  Failure to review and merge this pr has forced an increase in `max-complexity`, `max-branches`, `max-returns`, and `max-statements`.   Putting C90 and PLR091 in the same PR makes sense because they both set upper limits on code complexity and maintainability.
```diff
- mccabe.max-complexity = 13  # Default is 10
- max-branches = 14    # Default is 12
- max-returns = 7      # Default is 6
- max-statements = 69  # Default is 50
+ mccabe.max-complexity = 24  # Default is 10
+ max-branches = 23    # Default is 12
+ max-returns = 13      # Default is 6
- max-statements = 99  # Default is 50
+ max-statements = 102  # Default is 50
```
I have the RUFF settings to ignore the pylint import rules [PLC0414](https://docs.astral.sh/ruff/rules/useless-import-alias) and [PLR0402](https://docs.astral.sh/ruff/rules/manual-from-import).

pyproject.toml
```diff
@@ -90,18 +90,26 @@ venvPath = "."
...
-[tool.ruff.lint]
-select = ["C4", "E", "F", "I", "PERF", "UP"]
-ignore = ["PERF203"]
-
 [tool.ruff]
 line-length = 88
 target-version = "py310"
 
+[tool.ruff.lint]
+select = ["C4", "C90", "E", "F", "I", "PERF", "PL", "UP"]
```
Kludex: I think we should have each rule in a single line and have the link as a comment inline.

Kludex: This is unrelated to this PR.

cclauss: Perhaps `pyproject-fmt` to pre-commit?

https://github.com/tox-dev/pyproject-fmt

src/mcp/server/lowlevel/server.py
```diff
@@ -83,7 +83,7 @@ async def main():
 from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.models import InitializationOptions
 from mcp.server.session import ServerSession
-from mcp.server.stdio import stdio_server as stdio_server
+from mcp.server.stdio import stdio_server as stdio_server  # noqa: PLC0414
```
felixweinberger: Why is `# noqa: PLC0414` necessary? Would prefer to avoid adding lint-ignores.

```
PLC0414 | useless-import-alias | Import alias does not rename original package

PLC0414	[useless-import-alias](https://docs.astral.sh/ruff/rules/useless-import-alias/)	Import alias does not rename original package
```

Could we instead just change this to:

```
from mcp.server.stdio import stdio_server
```

tests/shared/test_progress_notifications.py
```diff
@@ -41,7 +41,7 @@ async def run_server():
                 capabilities=server.get_capabilities(NotificationOptions(), {}),
             ),
         ) as server_session:
-            global serv_sesh
+            global serv_sesh  # noqa: PLW0603
```
felixweinberger: As above, would be great not to be adding #noqa statements. Can we improve the design instead?

felixweinberger (changes requested):
Apologies for the delay in coming back to this.

It would be great if we could address the issues introduced by the new lint rules instead of adding linter overrides.

Happy to come back to this quickly if you have the capacity to take a stab at this?

src/mcp/server/lowlevel/server.py
cclauss:
```suggestion
from mcp.server.stdio import stdio_server
```

pyproject.toml
```diff
@@ -102,18 +102,36 @@ executionEnvironments = [
...
+    "C4",    # flake8-comprehensions
+    "C90",   # mccabe
+    "E",     # pycodestyle
+    "F",     # pyflakes
+    "I",     # isort
+    "PERF",  # Perflint
+    "PL",    # Pylint
+    "UP",    # pyupgrade
+]
+ignore = ["PERF203", "PLC0415", "PLR0402"]
```
felixweinberger, line 121: "[PLC0415](https://docs.astral.sh/ruff/rules/import-outside-top-level/)" and "[PLR0402](https://docs.astral.sh/ruff/rules/manual-from-import/)" are newly added. 

I'm assuming that's because we've added to the selection of tools above causing new lint errors to appear that we previously were ignoring by omission?

If so I'm OK with this trade-off in the short term to have a bit more manageable complexity.

pyproject.toml
```diff
@@ -102,18 +102,36 @@ executionEnvironments = [
...
+mccabe.max-complexity = 24  # Default is 10
+
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]
 "tests/server/fastmcp/test_func_metadata.py" = ["E501"]
+"tests/shared/test_progress_notifications.py" = ["PLW0603"]
```
felixweinberger, line 127: Not my favorite solution, but assuming the same thing applies as above - we have this code already, we were just ignoring it by omission.

[PLW0603](https://docs.astral.sh/ruff/rules/global-statement/)

felixweinberger:
@cclauss thank you for the contribution and apologies again for the delay!

PR #540, opened 2025-04-19
Add mount_path support for proper SSE endpoint routing with multiple FastMCP servers

tim-watcha:
Add mount_path capability to FastMCP for proper SSE endpoint handling when mounted under sub-paths

## Motivation and Context
When multiple FastMCP servers are mounted under different sub-paths in a single Starlette application, the session URIs were being generated incorrectly. FastMCP would create session URIs that didn't include the mount path, causing clients to send messages to the wrong endpoint. This change fixes the issue by
adding a `mount_path` setting that properly normalizes URIs.

## How Has This Been Tested?
- Added unit tests for the path normalization function
- Added unit tests for SSE app creation with different mount paths
- Tested with a real application mounting multiple FastMCP servers under different paths
- Manually verified session URI generation and message routing

## Breaking Changes
None. This is backwards compatible with existing code. The new `mount_path` setting defaults to "/".

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
The issue occurs because the SseServerTransport class creates a session URI without considering that the server might be mounted under a sub-path. The fix adds a mount_path setting to FastMCP and implements a path normalization helper method that combines the mount path with the message path to ensure correct
URI generation.

This solution maintains clean separation between FastMCP and SseServerTransport without requiring changes to the SseServerTransport class itself.

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -144,11 +144,14 @@ class FastMCP:
     def instructions(self) -> str | None:
         return self._mcp_server.instructions
 
-    def run(self, transport: Literal["stdio", "sse"] = "stdio") -> None:
+    def run(
+        self, transport: Literal["stdio", "sse"] = "stdio", mount_path: str = ""
+    ) -> None:
         """Run the FastMCP server. Note this is a synchronous function.
 
         Args:
             transport: Transport protocol to use ("stdio" or "sse")
+            mount_path: Optional mount path for SSE transport
         """
         TRANSPORTS = Literal["stdio", "sse"]
         if transport not in TRANSPORTS.__args__:  # type: ignore
@@ -157,7 +160,7 @@ class FastMCP:
         if transport == "stdio":
             anyio.run(self.run_stdio_async)
         else:  # transport == "sse"
-            anyio.run(self.run_sse_async)
+            anyio.run(lambda: self.run_sse_async(mount_path))
 
     def _setup_handlers(self) -> None:
         """Set up core MCP protocol handlers."""
@@ -463,11 +466,11 @@ class FastMCP:
                 self._mcp_server.create_initialization_options(),
             )
 
-    async def run_sse_async(self) -> None:
+    async def run_sse_async(self, mount_path: str = "") -> None:
         """Run the server using SSE transport."""
         import uvicorn
 
-        starlette_app = self.sse_app()
+        starlette_app = self.sse_app(mount_path)
 
         config = uvicorn.Config(
             starlette_app,
@@ -504,8 +507,11 @@ class FastMCP:
         # Combine paths
         return mount_path + endpoint
 
-    def sse_app(self) -> Starlette:
+    def sse_app(self, mount_path: str = "") -> Starlette:
         """Return an instance of the SSE server app."""
+        # Update mount_path in settings if provided
+        if mount_path:
+            self.settings.mount_path = mount_path
         # Create normalized endpoint considering the mount path
         normalized_endpoint = self._normalize_path(
             self.settings.mount_path, self.settings.message_path
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -597,7 +597,7 @@ class FastMCP:
         # Combine paths
         return mount_path + endpoint
 
-    def sse_app(self, mount_path: str = "") -> Starlette:
+    def sse_app(self, mount_path: str | None = None) -> Starlette:
         """Return an instance of the SSE server app."""
         from starlette.middleware import Middleware
         from starlette.routing import Mount, Route
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -467,6 +467,7 @@ class FastMCP:
     async def run_sse_async(self) -> None:
         """Run the server using SSE transport."""
         import uvicorn
+
         starlette_app = self.sse_app()
 
         config = uvicorn.Config(
@@ -481,33 +482,35 @@ class FastMCP:
     def _normalize_path(self, mount_path: str, endpoint: str) -> str:
         """
         Combine mount path and endpoint to return a normalized path.
-        
+
         Args:
             mount_path: The mount path (e.g. "/github" or "/")
             endpoint: The endpoint path (e.g. "/messages/")
-            
+
         Returns:
             Normalized path (e.g. "/github/messages/")
         """
         # Special case: root path
         if mount_path == "/":
             return endpoint
-        
+
         # Remove trailing slash from mount path
         if mount_path.endswith("/"):
             mount_path = mount_path[:-1]
-        
+
         # Ensure endpoint starts with slash
         if not endpoint.startswith("/"):
             endpoint = "/" + endpoint
-        
+
         # Combine paths
         return mount_path + endpoint
-    
+
     def sse_app(self) -> Starlette:
         """Return an instance of the SSE server app."""
         # Create normalized endpoint considering the mount path
-        normalized_endpoint = self._normalize_path(self.settings.mount_path, self.settings.message_path)
+        normalized_endpoint = self._normalize_path(
+            self.settings.mount_path, self.settings.message_path
+        )
         sse = SseServerTransport(normalized_endpoint)
 
         async def handle_sse(request: Request) -> None:
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -556,7 +556,7 @@ class FastMCP:
                 self._mcp_server.create_initialization_options(),
             )
 
-    async def run_sse_async(self, mount_path: str = "") -> None:
+    async def run_sse_async(self, mount_path: str | None = None) -> None:
         """Run the server using SSE transport."""
         import uvicorn
 
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -180,7 +180,7 @@ class FastMCP:
         return self._mcp_server.instructions
 
     def run(
-        self, transport: Literal["stdio", "sse"] = "stdio", mount_path: str = ""
+        self, transport: Literal["stdio", "sse"] = "stdio", mount_path: str = None
     ) -> None:
         """Run the FastMCP server. Note this is a synchronous function.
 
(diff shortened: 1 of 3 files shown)


richardhundt:
I wish Starlette allowed some reflection on where a route was mounted. This kind of handshake could benefit from that, and I'm guessing the Starlette folks never thought it would be needed. That mount point duplication in settings just seems superfluous, but looking at the Starlette internals, I couldn't see a way of getting at those values either. Perhaps asking them nicely to add this capability would work.

tim-watcha:
> I wish Starlette allowed some reflection on where a route was mounted. This kind of handshake could benefit from that, and I'm guessing the Starlette folks never thought it would be needed. That mount point duplication in settings just seems superfluous, but looking at the Starlette internals, I couldn't see a way of getting at those values either. Perhaps asking them nicely to add this capability would work.
Here's the comment content in English:

@richardhundt I agree that Starlette's lack of reflection on mount points is a bit unfortunate. While we wait for that capability, I was thinking of adding a syntactic sugar method that could help avoid the duplication:

```python
def get_mount(self, mount_path: Optional[str] = None) -> Mount:
    """
    Configure mount_path and return a Mount object for Starlette routing.
    
    Args:
        path: The path to mount the FastMCP server at
        mount_path: Optional override for the mount_path setting (defaults to path)
    
    Returns:
        Starlette Mount object configured for this FastMCP server
    """
    # Set mount_path setting (use provided value or path argument)
    if mount_path:
        self.settings.mount_path = mount_path
    
    # Return configured Mount object
    return Mount(path, app=self.sse_app())
```

This would allow for more concise code:

```python
# Current approach
github_mcp.settings.mount_path = "/github"
routes = [Mount("/github", app=github_mcp.sse_app())]

# Proposed approach
routes = [github_mcp.get_mount("/github")]
```

What do you think? This could help prevent mount path mismatches and simplify the API.

richardhundt:
Looks like this will all be moot soon. The update specification reduces all communication to just a single endpoint supporting both SSE and transactional (request/response) style HTTP interactions (controlled by the Accept header). It complicates the clients a bit because they need to distinguish SSE streaming from transactional responses, but that's manageable and solves all of these routing issues.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

xorcus:
I just hit the same problem. Could you please consider allowing for the `base_path` to be supplied?
Example modification in `python-sdk/src/mcp/server/fastmcp/server.py`:
```python
    def sse_app(self, base_path: str = '') -> Starlette:
        """Return an instance of the SSE server app."""
        sse = SseServerTransport(f'{base_path}{self.settings.message_path}')
```

The original code looks like this:
```python
    def sse_app(self) -> Starlette:
        """Return an instance of the SSE server app."""
        sse = SseServerTransport(self.settings.message_path)

        async def handle_sse(request: Request) -> None:
            async with sse.connect_sse(
                request.scope,
                request.receive,
                request._send,  # type: ignore[reportPrivateUsage]
            ) as streams:
                await self._mcp_server.run(
                    streams[0],
                    streams[1],
                )

        return Starlette(
            debug=self.settings.debug,
            routes=[
                Route(self.settings.sse_path, endpoint=handle_sse),
                Mount(self.settings.message_path, app=sse.handle_post_message),
            ],
        )
```

Exposing the `base_path` as suggested would allow for mounting with `Starlette` as follows:
```python
    Mount('/mcp', mcp.sse_app(base_path='/mcp')),
```

It is a bit redundant, but it works.

tim-watcha:
I've implemented a solution for this issue by adding a `mount_path` parameter to the `sse_app` method, similar to what you suggested with `base_path`. This change gives
users multiple options for configuring the mount path:

1. Using `settings.mount_path` for persistent configuration
2. Passing the path directly to `sse_app(mount_path="...")` for ad-hoc mounting
3. Passing to `run(transport="sse", mount_path="...")` for direct execution

The implementation also properly handles path normalization through the existing `_normalize_path` method to ensure paths are correctly combined.

Your example of `Mount('/mcp', mcp.sse_app(base_path='/mcp'))` now works with our implementation as `Mount('/mcp', mcp.sse_app('/mcp'))`.

All tests have been updated and are passing, and the README has been expanded with examples showing the different approaches. @xorcus 

1bd89ff

dwayn (commented):
This looks great to me and provides a nice way to mount sse mcp servers on a path that is not just on the root path.

This is fundamentally needed to support multi-server gateway implementations. 

Even with http+sse transport being replaced by streamable http, there are many implementations that will likely continue to use sse for a while so being able to still support them in gateway implementations without awful hacks is kind of important.

At a minimum, if the preference is to not muddy the FastMCP settings for a transport that is going away, being able to inject mount path using sse_app(mount_path="/foo/bar") would be super helpful. Otherwise, any of us that are trying to build generalized gateways end up with hacky messes.

dwayn:
For anyone else wanting a fix for this problem, I wrote a middleware that fixes this.

Source here: https://github.com/dwayn/fastmcp-mount
It is also uploaded to PyPi: https://pypi.org/project/fastmcp-mount/

src/mcp/server/fastmcp/server.py
```diff
@@ -178,11 +179,14 @@ def name(self) -> str:
     def instructions(self) -> str | None:
         return self._mcp_server.instructions
 
-    def run(self, transport: Literal["stdio", "sse"] = "stdio") -> None:
+    def run(
+        self, transport: Literal["stdio", "sse"] = "stdio", mount_path: str = ""
```
ihrpr:
```suggestion
        self, transport: Literal["stdio", "sse"] = "stdio", mount_path: str = None
```

src/mcp/server/fastmcp/server.py
```diff
@@ -567,14 +571,49 @@ async def run_sse_async(self) -> None:
...
+            mount_path = mount_path[:-1]
+
+        # Ensure endpoint starts with slash
+        if not endpoint.startswith("/"):
+            endpoint = "/" + endpoint
+
+        # Combine paths
+        return mount_path + endpoint
+
+    def sse_app(self, mount_path: str = "") -> Starlette:
```
ihrpr:
```suggestion
    def sse_app(self, mount_path: str | None = None) -> Starlette:
```

src/mcp/server/fastmcp/server.py
```diff
@@ -567,14 +571,49 @@ async def run_sse_async(self) -> None:
...
         """Return an instance of the SSE server app."""
         from starlette.middleware import Middleware
         from starlette.routing import Mount, Route
 
+        # Update mount_path in settings if provided
+        if mount_path:
```
ihrpr:
```suggestion
        if mount_path is not None:
```

src/mcp/server/fastmcp/server.py
```diff
@@ -552,11 +556,11 @@ async def run_stdio_async(self) -> None:
                 self._mcp_server.create_initialization_options(),
             )
 
-    async def run_sse_async(self) -> None:
+    async def run_sse_async(self, mount_path: str = "") -> None:
```
ihrpr:
```suggestion
    async def run_sse_async(self, mount_path: str | None = None) -> None:
```

ihrpr (changes requested):
thank you! 

Just a few small things and it's good to merge, we'll also need to incorporate this into streamable http

ihrpr (approved):
thank you

elizabetht:
Hi @tim-watcha I am following your example as in README. But can't find the apps running at the different mount paths. If my endpoint was /cars/v1/sse, I am not seeing the route being found in this approach - getting 404 not found for /cars/v1/sse
```
if __name__ == "__main__":
  routes = []
  for endpoint, mcp_app in mcp_servers.items():
    routes.append(Mount(endpoint, mcp_app.sse_app(mount_path=endpoint)))

  main_app=Starlette(routes=routes)
  uvicorn.run(main_app, host="0.0.0.0", port=8000)
```

tim-watcha:
> Hi @tim-watcha I am following your example as in README. But can't find the apps running at the different mount paths. If my endpoint was /cars/v1/sse, I am not seeing the route being found in this approach - getting 404 not found for /cars/v1/sse
> 
> ```
> if __name__ == "__main__":
>   routes = []
>   for endpoint, mcp_app in mcp_servers.items():
>     routes.append(Mount(endpoint, mcp_app.sse_app(mount_path=endpoint)))
> 
>   main_app=Starlette(routes=routes)
>   uvicorn.run(main_app, host="0.0.0.0", port=8000)
> ```

1. Which client did you use?
2. If the endpoint was `/cars/v1/sse`, the client should connect via `/cars/v1/sse/sse` and `/cars/v1/sse/messages`

lalith-b:
@tim-watcha, so the url routing after this merge if it was set to `/cars/v1` it resolves to `/cars/v1/sse` and `/cars/v1/messages` will be handled as routes,

however, If there's no extra `/` towards the end of messages route|url I'm seeing a 307 redirect `/cars/v1/messages/` will solve the redirect, why is this happening does this solve the trailing slash and make it consistent with queryStrings for sessionIds ?

jlowin:
@dwayn good news - the strategy you describe was implemented in #659

dwayn:
> @dwayn good news - the strategy you describe was implemented in #659

@jlowin This is awesome news! I just checked out the PR and it looks great, this is the true fix that was needed for it.

I think the only other concern is to ensure that the same handling is applied to the Streamable HTTP server transport implementation to ensure that the URLs handed out for SSE connections get their dynamic prefixes as well. 
Not sure who is working on that implementation, but definitely worth pointing out to ensure it doesn't get missed, since this fix to SSE is coming in now, shortly before it is deprecated, and I would bet the transport implementation is probably close to finished.

mindcrime:
If I'm following this correctly, this helps people using FastMCP. But what about people using the Vanilla Python SDK directly? I'm seeing the same issue where if I'm running behind a proxy that adds a custom context path, the initial connection to /mypath/sse works, but then the client tries to followup by connecting to /messages, not /mypath/messages. Is there a way to make this work in the Vanilla SDK case?

felixweinberger:
Hi @tim-watcha! Coming to this PR because I'm looking into https://github.com/modelcontextprotocol/python-sdk/pull/1007 and https://github.com/modelcontextprotocol/python-sdk/pull/1000 which I think are both running into edges with the `mount_path` added as part of this PR.

I think one key confusion here is that the `mount_path` only seems to be applied to `messages` but not `sse` and `mcp` - is that intentional? 

If I'm understanding correctly, the mount_path is only actually used to modify the `/messages` path that's sent back to the client when using SSE - is that right?

Kludex:
I'm reverting this in V2.

Issue #443, opened 2025-04-07
Implement Streamable HTTP transport


jspahrsummers:
- server
- client
- state management
- resumability
- examples
- integration tests

**Update**

PRs are out, review in progress, once done, documentation will be added and transport will be released.

apryiomka:
We need this too, please. We are interested in HTTP request / response use case: https://github.com/modelcontextprotocol/specification/pull/206#issuecomment-2766559523. We have a turn based agent framework that cannot maintain long term TCP connection and is request / response based.

karanb192:
I'm picking this up. Will share a PR by the weekend.

Specification PR is merged: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/206/files
Server-side support has been added in TypeScript SDK: https://github.com/modelcontextprotocol/typescript-sdk/pull/266/files

@jspahrsummers , can you please assign this issue to me?

frankgu968:
Potentially dupe of #405 ?

hfawaz:
we need this please

rem-lag:
Might be nice to get #380 in around the same time this is implemented too

Jeetanshu18:
how much time it will take to get implemented

adelante4:
It would be nice if somebody posts an update. Thanks.

ihrpr:
> It would be nice if somebody posts an update. Thanks.

PRs for Server and Client transport are out, some are still pending the reviews. Once the review is done we'll merge and do the release - hopefully later this week - early next week.

theresakern:
Since all the PRs have been merged, when will the Release with the changes happen? :) v1.7.1 does not seem to contain the streamable http changes..

nightzjp:
我看到相应的代码更新了。但是1.7.1貌似没发布该功能

lucasalvarezlacasa:
Any ETA's of when this will be available?

hughlv:
Good job. I'm waiting for this feature. Which version of the SDK will this be included?

de-code:
Just to say thank you for this. Amazing work. 🎉 

(BTW It's part of the [v1.8.0 release](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v1.8.0))

lucasalvarezlacasa:
Amazing!!. Thank you so much for the work. 🚀

PR #553, opened 2025-04-21
StreamableHttp - Server transport with state management

ihrpr:
Adding support for [StreamableHTTP transport](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) transport for MCP servers, providing bidirectional communication over
  HTTP with streaming support.

### Features

  - Session management
  - Implementation for server termination (DELETE)
  - Includes support for JSON and SSE response modes
  - Handles Related request IDs for notifications like logging and progress
  - Example of a server with tool notifications

  
### Follow ups 
 - Suport stateless mode
 - Support GET SSE
 - Streamable Http Client
 

https://github.com/modelcontextprotocol/python-sdk/issues/443


examples/servers/simple-streamablehttp/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Features
+
+- Uses the StreamableHTTP transport for server-client communication
+- Task management with anyio task groups
+- Ability to send multiple notifications over time to the client
+- Proper resource cleanup and lifespan management
+
+## Usage
+
+Start the server on the default or custom port:
```
jerome3o-anthropic, line 15: probably worth mentioning that the endpoint is `/mcp` somewhere

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,646 @@
...
+            # If we're not using session IDs, return True
+            return True
+
+        # Get the session ID from the request headers
+        request_session_id = self._get_session_id(request)
+
+        # If no session ID provided but required, return error
+        if not request_session_id:
+            response = self._create_error_response(
+                "Bad Request: Missing session ID",
```
jerome3o-anthropic, line 546: I get this error when I run the example server:
```sh
uv run mcp-simple-streamablehttp --port 3000 --log-level DEBUG
```

Then 
```sh
curl http://localhost:3000/mcp/ -d '
{
    "jsonrpc": "2.0",
    "method": "initialize",
    "params": {
        "protocolVersion": "2025-03-26",
        "capabilities": {},
        "clientInfo": {
            "name": "curl-client",
            "version": "1.0.0"
        }
    }
}' \
    -X POST | \
    jq
```

output:
```json
{
  "jsonrpc": "2.0",
  "id": "server-error",
  "error": {
    "code": -32600,
    "message": "Bad Request: Missing session ID"
  }
}
```

jerome3o-anthropic: But this is before I'm able to get the session from the server to start with

jerome3o-anthropic: Ah - I was missing the id, thats why. it mistook the message as a notification, and then didn't trigger the `if is_initialization_request` path. That could have a better error, but idt it should block here

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
```diff
@@ -0,0 +1,196 @@
...
+) -> int:
+    # Configure logging
+    logging.basicConfig(
+        level=getattr(logging, log_level.upper()),
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    )
+
+    app = Server("mcp-streamable-http-demo")
+
+    @app.call_tool()
```
jerome3o-anthropic, line 70: [nit, also could be a follow up] Would be nice to use fastmcp here - i usually end up going:

```python
server = FastMCP(...)
...
server._mcp_server # <- to access the lower level mcp interface
```

But perhaps we should just make it easier to get to

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,646 @@
...
+        Yields:
+            Tuple of (read_stream, write_stream) for bidirectional communication
+        """
+
+        # Create the memory streams for this connection
+        read_stream: MemoryObjectReceiveStream[JSONRPCMessage | Exception]
+        read_stream_writer: MemoryObjectSendStream[JSONRPCMessage | Exception]
+
+        write_stream: MemoryObjectSendStream[JSONRPCMessage]
+        write_stream_reader: MemoryObjectReceiveStream[JSONRPCMessage]
```
jerome3o-anthropic: I don't think the initialization/typing here is needed when you pass in type vars to `anyio.create_memory_object_stream`

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,646 @@
...
+                try:
+                    async for message in write_stream_reader:
+                        # Determine which request stream(s) should receive this message
+                        target_request_id = None
+                        if isinstance(
+                            message.root, JSONRPCNotification | JSONRPCRequest
+                        ):
+                            # Extract related_request_id from params if it exists
+                            if (params := getattr(message.root, "params", None)) and (
+                                related_id := params.get("related_request_id")
```
jerome3o-anthropic: Are we expecting `related_request_id` in the MCP message `.params`? that seems odd to me - I don't see it referenced in the spec

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,646 @@
...
+                            ) is not None:
+                                target_request_id = str(related_id)
+                        else:
+                            target_request_id = str(message.root.id)
+
+                        # Send to the specific request stream if available
+                        if (
```
jerome3o-anthropic, line 615: should something happen in the `else` here, is that an error case?

src/mcp/shared/session.py
```diff
@@ -267,16 +267,21 @@ async def send_request(
...
+        self,
+        notification: SendNotificationT,
+        related_request_id: RequestId | None = None,
+    ) -> None:
         """
         Emits a notification, which is a one-way message that does not expect
         a response.
         """
+        if related_request_id is not None and notification.root.params is not None:
+            notification.root.params.related_request_id = related_request_id
```
jerome3o-anthropic: This feels odd to me - what's the story here? surely we don't want to be adding `related_request_id` to spec messages? (did I miss this being added somewhere?)

src/mcp/types.py
```diff
@@ -58,6 +58,7 @@ class Meta(BaseModel):
         model_config = ConfigDict(extra="allow")
 
     meta: Meta | None = Field(alias="_meta", default=None)
+    related_request_id: RequestId | None = None
```
jerome3o-anthropic: Yeah this seems off

jerome3o-anthropic (changes requested):
Concerns about `related_request_id` - seems like it's accidentally added. Otherwise looks good

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
ihrpr: I think I was quite confused here, we have two folders for servers: fastmcp and servers. I was planning to add fastmcp exaple as a follow up. 

fastmcp might also help with `related_request_id` clarity - it's hidden under log. Low level server does not have a place where we can nicely inject `related_request_id`

src/mcp/server/streamableHttp.py
ihrpr: Spec specifies it very loosely: 
> These messages SHOULD relate to the originating client request. These requests and notifications MAY be batched.

Basically when there is a POST request from a client and server initiates SSE to respond to it, we want that stream to have all related notification and eventually a response. To understand which response/notification is realted to the origianl request id we need some way to associate it. 
- for respose it's quite easy - we have an id
- notifications do not have ids, hence we need to have some way to make them related. 

the current implemetation provides a way to say that this logging notification is sent from "this tool". With fastmcp is't nicer as we just add it to `log` and if `log` was called withing a tool it will be automatically associated and client will receive all the stream notifications and responses in the correct SSE response

src/mcp/server/streamableHttp.py
ihrpr: yeah, good catch, it's not expected

ihrpr: oh, wait, it will be in SSE stream in #561 - anything that is not caught by requst will be sent to standalone SSE

src/mcp/shared/session.py
ihrpr: I guess we can introduce options, like we have in [typescript sdk](https://github.com/modelcontextprotocol/typescript-sdk/blob/c04842ec7ee2cb9bc11e8f146719aebf5f340881/src/shared/protocol.ts#L95)

ihrpr:
@jerome3o-anthropic, very good point on spec, somehow I completely ignored that fact it's part of the spec 🙈

What do you think about using [meta](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts#L56)  field to propagate `related_request_id`?

ihrpr:
> @jerome3o-anthropic, very good point on notification params, somehow I completely ignored that fact it's part of the spec 🙈
> 
> What do you think about using [meta](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.ts#L56) field to propagate `related_request_id`?

I think I'd like to change the approach to use clean separation: https://github.com/modelcontextprotocol/python-sdk/pull/590

jerome3o-anthropic (approved):
Approving with the assumption we're gonna tidy up the `related_request_id`/`_meta` stuff.

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
```diff
@@ -0,0 +1,201 @@
...
+from uuid import uuid4
+
+import anyio
+import click
+import mcp.types as types
+from mcp.server.lowlevel import Server
+from mcp.server.streamableHttp import (
+    MCP_SESSION_ID_HEADER,
+    StreamableHTTPServerTransport,
+)
```
dsp-ant, line 13: Let's call this `mcp.server.http.streamable`, or something similar. `streamableHttp` is not PEP 8 compliant.

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return [
+            types.TextContent(
+                type="text",
+                text=(
+                    f"Sent {count} notifications with {interval}s interval"
+                    f" for caller: {caller}"
+                ),
+            )
+        ]
```
dsp-ant, line 103: We never check the name of the tool that is called. We should.

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
```diff
@@ -0,0 +1,201 @@
...
+                task_group.start_soon(run_server)
+
+                # Handle the HTTP request and return the response
+                await http_transport.handle_request(scope, receive, send)
+        else:
+            response = Response(
+                "Bad Request: No valid session ID provided",
+                status_code=HTTPStatus.BAD_REQUEST,
+            )
+            await response(scope, receive, send)
```
dsp-ant, line 186: This feels awfully complex as a lowlevel API, particularly if every MCP Server using the lowlevel API would have to go and use something like this. I wonder if we can provide a more high-level abstraction for this, that makes it more managable.

src/mcp/server/streamableHttp.py
dsp-ant, line 1: As mentioned above, let's rename this.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+    JSONRPCNotification,
+    JSONRPCRequest,
+    JSONRPCResponse,
+    RequestId,
+)
+
+logger = logging.getLogger(__name__)
+
+# Maximum size for incoming messages
+MAXIMUM_MESSAGE_SIZE = 4 * 1024 * 1024  # 4MB
```
dsp-ant, line 43: Why did we choose this?

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+            response_message.model_dump_json(by_alias=True, exclude_none=True)
+            if response_message
+            else None,
+            status_code=status_code,
+            headers=response_headers,
+        )
+
+    def _get_session_id(self, request: Request) -> str | None:
+        """Extract the session ID from request headers."""
+        return request.headers.get(MCP_SESSION_ID_HEADER)
```
dsp-ant, line 158: It seems we often use the pattern `request.headers.get(MCP_SESSION_ID_HEADER)` should this be exposed as a public function?

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+            return
+
+        if request.method == "POST":
+            await self._handle_post_request(scope, request, receive, send)
+        elif request.method == "GET":
+            await self._handle_get_request(request, send)
+        elif request.method == "DELETE":
+            await self._handle_delete_request(request, send)
+        else:
+            await self._handle_unsupported_request(request, send)
```
dsp-ant, line 179: I am biased here, but i love `match` statements. No good reason, except it feels more functional.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+        """Application entry point that handles all HTTP requests"""
+        request = Request(scope, receive)
+        if self._terminated:
+            # If the session has been terminated, return 404 Not Found
+            response = self._create_error_response(
+                "Not Found: Session has been terminated",
+                HTTPStatus.NOT_FOUND,
+            )
+            await response(scope, receive, send)
+            return
```
dsp-ant, line 170: You can check for termination before creating the request.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+        """Check if the request accepts the required media types."""
+        accept_header = request.headers.get("accept", "")
+        accept_types = [media_type.strip() for media_type in accept_header.split(",")]
+
+        has_json = any(
+            media_type.startswith(CONTENT_TYPE_JSON) for media_type in accept_types
+        )
+        has_sse = any(
+            media_type.startswith(CONTENT_TYPE_SSE) for media_type in accept_types
+        )
```
dsp-ant, line 191: Given you have stripped accept_types as a list, you might even put them into a set and just do a check if CONTENT_TYPE_SSE is in media_type or CONTENT_TYPE_JSON is in media_type. It doesn't really matter here, but we don't need to traverse the list 3 times.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+
+        return has_json, has_sse
+
+    def _check_content_type(self, request: Request) -> bool:
+        """Check if the request has the correct Content-Type."""
+        content_type = request.headers.get("content-type", "")
```
dsp-ant, line 197: Are request.headers always lowercase?

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+        content_type_parts = [
+            part.strip() for part in content_type.split(";")[0].split(",")
+        ]
+
+        return any(part == CONTENT_TYPE_JSON for part in content_type_parts)
```
dsp-ant, line 202:
```suggestion
        content_type_parts = (
            part.strip() for part in content_type.split(";")[0].split(",")
        )

        return any(part == CONTENT_TYPE_JSON for part in content_type_parts)
```
We never use `content_type_parts` other than checking it, we either can traverse it in a for-loop or should just use a generator here to be lazy about it.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+
+    async def _handle_post_request(
+        self, scope: Scope, request: Request, receive: Receive, send: Send
+    ) -> None:
```
dsp-ant, line 206: Can we break this up? I think some of the branches, particularly the one where we handle an existing session and a non-existing session should probably be separate functions.

src/mcp/server/streamableHttp.py
```diff
@@ -0,0 +1,644 @@
...
+                if self.mcp_session_id:
+                    # Check if request has a session ID
+                    request_session_id = self._get_session_id(request)
+
+                    # If request has a session ID but doesn't match, return 404
+                    if request_session_id and request_session_id != self.mcp_session_id:
+                        response = self._create_error_response(
+                            "Not Found: Invalid or expired session ID",
+                            HTTPStatus.NOT_FOUND,
+                        )
```
dsp-ant, line 283: Are we sure this is 404 or maybe rather a 400 BAD REQUEST? What does the TS SDK do here?

dsp-ant (changes requested):

src/mcp/server/streamableHttp.py
ihrpr: @dsp-ant I renamed the files later in the stack, it's quite a few of them by now (like 9 or 10), so was very difficult to rename in the very bottom. Happy to rename in the bottom if that's needed, just let me know.

src/mcp/server/streamableHttp.py
ihrpr: We have it in the TS SDKs (both SSE and Streamable HTTP)

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
ihrpr: sure, we can make it happen, will add as a separate PR

dsp-ant (approved):
I'll accept this for now under the assumption that the fixes are done on top.

PR #554, opened 2025-04-21
StreamableHttp - stateless server support

ihrpr:
- Adding support for stateless servers using StreamableHttp transport. Each request is independent and does not require initialization.
- Adding an example of stateless server with streamable http transport.

Tested with multi-client test script in Typescript introduced in [this PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/377)
`uv run mcp-simple-streamablehttp-stateless --log-level DEBUG`
`uv run mcp-simple-streamablehttp-stateless --json-response`

## Follow ups
- Support GET SSE
- Streamable Http Client
- Resumability

Stacked on top of #553


examples/servers/simple-streamablehttp-stateless/mcp_simple_streamablehttp_stateless/server.py
```diff
@@ -0,0 +1,171 @@
...
+                    # This allows the server to run without waiting for initialization
+                    require_initialization=False,
+                )
+
+            # Start server task
+            task_group.start_soon(run_server)
+
+            # Small delay to allow the server task to start
+            # This helps prevent race conditions in stateless mode
+            await anyio.sleep(0.001)
```
jerome3o-anthropic: 🤔 that's not ideal, do you know what the race conditions are? I feel like the memory channels should account for any issues

src/mcp/server/lowlevel/server.py
```diff
@@ -479,11 +479,20 @@ async def run(
         # but also make tracing exceptions much easier during testing and when using
         # in-process servers.
         raise_exceptions: bool = False,
+        # When True, the server will wait for the client to send an initialization
+        # message before processing any other messages.
+        # False should be used for stateless servers.
+        require_initialization: bool = True,
```
jerome3o-anthropic: [nit / question] should we just spoof the initialization on these requests with capabilities that a stateless server would have (i.e. capabilities={})? that way the life cycle of the mcp server is still spec compliant - it's a bit of a pointless dance though, so I'm happy to leave it as you have it. 

The code as it stands feels like this could be interpreted as initialization being optional - that's my main concern

jerome3o-anthropic (approved):
One question around initialization, also tests for this would be good.

examples/servers/simple-streamablehttp-stateless/mcp_simple_streamablehttp_stateless/server.py
ihrpr: ugh, accidendally left ater testing...

src/mcp/server/lowlevel/server.py
ihrpr: I think I'd probaly change the name, `require_initialization` sounds like we are skipping the protocol requirement on intialization

dsp-ant:
Can we change the module name please.

jerome3o-anthropic (dismissed):
awesome!

src/mcp/server/session.py
```diff
@@ -85,11 +85,15 @@ def __init__(
...
+        standalone_mode: bool = False,
     ) -> None:
         super().__init__(
             read_stream, write_stream, types.ClientRequest, types.ClientNotification
         )
-        self._initialization_state = InitializationState.NotInitialized
+        if standalone_mode:
+            self._initialization_state = InitializationState.Initialized
+        else:
+            self._initialization_state = InitializationState.NotInitialized
```
dsp-ant:
```suggestion
        self._initialization_state = InitializationState.Initialized if standalone_mode else InitializationState.NotInitialized
```

src/mcp/server/session.py
```diff
@@ -85,11 +85,15 @@ def __init__(
         read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
         write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
         init_options: InitializationOptions,
+        standalone_mode: bool = False,
```
dsp-ant: I am not a big fan of booleans like this. If we need other modalities here, we will end up with more and more of these. I probably prefer a `mode = Mode.STATELESS` or so. Just as an idea. I am happy to go with whatever you feel is best.

If we keep the bool, let's call it `stateless`, since i think we usually don't use `standalone` as a word somewhere else to describe the stateless feature.

dsp-ant (approved):

src/mcp/server/session.py
ihrpr: updated to stateless (for enum I'm struggling to find other modes and want to make it less inviting to add more modes)

PR #555, opened 2025-04-21
Fix stdio_client kill process after timeout

cristipufu:
## Motivation and Context
Be consistent across platforms (Windows, Linux, etc). Some MCP Servers don't react to SIGTERM, so `stdio_client` will hang indefinitely when exiting the context manager.

Reference: https://github.com/modelcontextprotocol/python-sdk/pull/372

## How Has This Been Tested?
Yes

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -16,7 +16,6 @@ from mcp.shared.message import SessionMessage
 from .win32 import (
     create_windows_process,
     get_windows_executable_command,
-    terminate_windows_process,
 )
 
 # Environment variables to inherit by default
@@ -180,13 +179,15 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
         finally:
             # Clean up process to prevent any dangling orphaned processes
             try:
-                if sys.platform == "win32":
-                    await terminate_windows_process(process)
-                else:
-                    process.terminate()
+                process.terminate()
+                with anyio.fail_after(2.0):
+                    await process.wait()
             except ProcessLookupError:
                 # Process already exited, which is fine
                 pass
+            except TimeoutError:
+                # Force kill if it doesn't terminate
+                process.kill()
             await read_stream.aclose()
             await write_stream.aclose()
             await read_stream_writer.aclose()
--- a/src/mcp/client/stdio/win32.py
+++ b/src/mcp/client/stdio/win32.py
@@ -8,9 +8,7 @@ import sys
 from pathlib import Path
 from typing import BinaryIO, TextIO, cast
 
-import anyio
 from anyio import to_thread
-from anyio.abc import Process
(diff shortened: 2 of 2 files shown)


ihrpr (changes requested):
Thank you @cristipufu for working on this! 

Do you have steps to reproduce? How this changes were tested?

src/mcp/client/stdio/__init__.py
```diff
@@ -173,10 +172,13 @@ async def stdin_writer():
...
-            if sys.platform == "win32":
-                await terminate_windows_process(process)
-            else:
+            try:
                 process.terminate()
+                with anyio.fail_after(2.0):
+                    await process.wait()
+            except TimeoutError:
+                # Force kill if it doesn't terminate
+                process.kill()
```
Kludex, line 190: I think the same `TerminateProcess` is called on Windows... Did this worked locally?

https://anyio.readthedocs.io/en/stable/api.html#anyio.abc.Process.kill

surya-prakash-susarla:
Slightly orthogonal but this does not handle child process spawns which can cause the parent to leave ghost processes which are being re-parented to the root process. Addressed it here: https://github.com/modelcontextprotocol/python-sdk/pull/850

felixweinberger (changes requested):
Hi @cristipufu, thank you for working on this and for submitting this PR!

Having looked into this in some detail I think your simplification makes sense:

- There shouldn't be any change in behavior on Windows from this change, so we're good on that end
- #559 already made some changes to stream cleanup and I believe addresses SIGTERM hangs already on UNIX systems
- As you mention on #372 though, MCP servers that ignore SIGTERM may still be an issue on UNIX systems
- I was able to create a test (see below) that simulates a service that expects SIGINT and ignores SIGTERM - this PR fixes that case.

I created a draft PR #1044 with new regression tests in an attempt to actually reproduce the hanging resolved by both this change (#555), #559, and #765. 

- #559 resolves `test_stdio_client_universal_cleanup`
- #555 (this PR) resolves `test_stdio_client_sigint_only_process`
- #765 resolves `test_stdio_client_graceful_stdin_exit` and `test_stdio_client_stdin_close_ignored`

Please take a look if that reflects what you were going for here - if yes could you rebase your PR here on latest main?

cristipufu:
@felixweinberger rebased

felixweinberger:
Thank you @cristipufu for your response and again for submitting this PR! 

I spent yesterday and today trying to unify all the different approaches and fixes we have pending in this process termination space at the moment, as there are several interrelated fixes that either conflict or depend on each other - specifically #555, #729, #765, and #850.

I've added your change to #1044 as a draft with you as a co-author + added extensive regression testing. Would you be OK with consolidating this change into #1044 for the comprehensive testing & process handling introduced there?

cristipufu:
@felixweinberger yes, of course, thanks!

PR #561, opened 2025-04-22
StreamableHttp - GET request standalone SSE 

ihrpr:
As per [spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server), the client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, allowing the server to communicate to the client, without the client first sending data via HTTP POST.  This PR implements establishing SSE stream on GET request.

## Follow ups
- Streamable Http Client
- Resumability

Stacked on top of #554


jerome3o-anthropic (dismissed):
lgtm!
edit: one suggestion on adding a test for the GET stream actually being used for notifications

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/server.py
```diff
@@ -87,6 +88,9 @@ async def call_tool(
             if i < count - 1:  # Don't wait after the last notification
                 await anyio.sleep(interval)
 
+        # This will send a resource notificaiton though standalone SSE
```
jerome3o-anthropic, line 96: Actually - a test for this would be awesome

ihrpr: have a test for this in the next PR, where I have a client!

src/mcp/server/streamableHttp.py
```diff
@@ -458,13 +470,80 @@ async def _handle_get_request(self, request: Request, send: Send) -> None:
...
+            headers[MCP_SESSION_ID_HEADER] = self.mcp_session_id
+
+        # Check if we already have an active GET stream
+        if GET_STREAM_KEY in self._request_streams:
+            response = self._create_error_response(
+                "Conflict: Only one SSE stream is allowed per session",
+                HTTPStatus.CONFLICT,
+            )
+            await response(request.scope, request.receive, send)
+            return
```
dsp-ant, line 492: I am okay with this, but I don't think this is necessarily true: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#multiple-connections. A client could have multiple SSE connections open but we should just reply to one.

dsp-ant (dismissed):

Akshit97:
@ihrpr When are we planning to release this? I see that this PR has been merged but not released yet.

PR #562, opened 2025-04-22
Format files with ruff

ihrpr:
Noticed there are two unformatted files that keep popping in when I run `uv run ruff format .`

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -466,6 +466,7 @@ class FastMCP:
     async def run_sse_async(self) -> None:
         """Run the server using SSE transport."""
         import uvicorn
+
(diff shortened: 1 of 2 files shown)


jerome3o-anthropic (approved):

Kludex:
We are running this in the pipeline:

https://github.com/modelcontextprotocol/python-sdk/blob/b4c7db6a50a5c88bae1db5c1f7fba44d16eebc6e/.github/workflows/shared.yml#L20-L21

...and this locally:

https://github.com/modelcontextprotocol/python-sdk/blob/b4c7db6a50a5c88bae1db5c1f7fba44d16eebc6e/.pre-commit-config.yaml#L11-L25

We need to match both. We need to run the pre-commit job on the pipeline as well.

Issue #567, opened 2025-04-22
`Tool.from_function` does not handle asynchronous callable objects


stephanlensky:
Hi there,

I am running into an issue when attempting to use callable objects with FastMCP.

## Description / Reproducible Example

Regular, synchronous callable objects work (mostly) as expected, with the exception of having to set `self.__name__`:

```python
import asyncio
from mcp.server.fastmcp import FastMCP

class TestTool:
    def __init__(self) -> None:
        self.__name__ = "TestTool"

    def __call__(self) -> str:
        return "Hello from TestTool!"

mcp = FastMCP()
mcp.add_tool(TestTool())

async def main() -> None:
    # prints "Hello from TestTool!"
    print(await mcp.call_tool("TestTool", {}))

if __name__ == "__main__":
    asyncio.run(main())
```

However, making the `__call__` method async breaks things:

```python
import asyncio
from mcp.server.fastmcp import FastMCP

class TestTool:
    def __init__(self) -> None:
        self.__name__ = "TestTool"

    async def __call__(self) -> str:
        return "Hello from TestTool!"

mcp = FastMCP()
mcp.add_tool(TestTool())

async def main() -> None:
    # prints [TextContent(type='text', text='<coroutine object TestTool.__call__ at 0x103155300>', annotations=None)]
    print(await mcp.call_tool("TestTool", {}))

if __name__ == "__main__":
    asyncio.run(main())
```

## Simple Fix

I tracked this down to [this line](https://github.com/modelcontextprotocol/python-sdk/blob/babb477dffa33f46cdc886bc885eb1d521151430/src/mcp/server/fastmcp/tools/base.py#L51) in `Tool.from_function`, which does not work properly for async callable objects:

```python
is_async = inspect.iscoroutinefunction(fn)
```

I know this use-case is slightly out-of-the-ordinary, but it would be extremely helpful if we could improve this to work properly for callable objects as well.

There is some prior art here, Starlette checks if the provided callable is async like [this](https://github.com/encode/starlette/blob/a766a58d14007f07c0b5782fa78cdc370b892796/starlette/_utils.py#L35-L39), which works for both regular async functions and async callable objects:

```python
def is_async_callable(obj: typing.Any) -> typing.Any:
    while isinstance(obj, functools.partial):
        obj = obj.func

    return inspect.iscoroutinefunction(obj) or (callable(obj) and inspect.iscoroutinefunction(obj.__call__))
```

[...]

PR #568, opened 2025-04-22
Fix async callable object tools

stephanlensky:
Makes detection of asynchronous functions more robust in `Tool.from_function`, allowing usage of asynchronous callable objects, e.g.

```python
class MyAsyncTool:
    def __init__(self):
        self.__name__ = "MyAsyncTool"

    async def __call__(self, x: int) -> int:
        return x * 2

manager = ToolManager()
tool = manager.add_tool(MyAsyncTool())
result = await tool.run({"x": 5})  # 10
```

## Motivation and Context

Fixes https://github.com/modelcontextprotocol/python-sdk/issues/567.

## How Has This Been Tested?

Unit tests were added to verify the change.

## Breaking Changes

This is a non-breaking change.

## Types of changes

- [ ] Documentation update

## Checklist


src/mcp/server/fastmcp/tools/base.py
```diff
@@ -92,3 +93,12 @@ async def run(
             )
         except Exception as e:
             raise ToolError(f"Error executing tool {self.name}: {e}") from e
+
+
+def _is_async_callable(obj: Any) -> bool:
+    while isinstance(obj, functools.partial):
```
ihrpr, line 105: Let's limit loop iterations here

ihrpr (changes requested):

Please can you address limit of loop iterations and it's good to merge.

src/mcp/server/fastmcp/tools/base.py
stephanlensky: Thank you for the review! What would you suggest? I don't think there is any limit to how many partials you can have applied.

I do want to note that the [reference implementation](https://github.com/encode/starlette/blob/a766a58d14007f07c0b5782fa78cdc370b892796/starlette/_utils.py#L35-L39) from Starlette is used by every consumer of FastAPI and does not specify a limit.

stephanlensky: The [CPython standard library](https://github.com/python/cpython/blob/b8f55266bf873bc101c3aab7aa868b909ecbcb99/Lib/inspect.py#L308) has a very similar implementation, also without a limit.

ihrpr: scary, but consistent :)

ihrpr (approved):
Thank you!!

ihrpr:
@stephanlensky please can you merge main and I'll merge!

stephanlensky:
Should be all set now! Thanks again 🙂

src/mcp/server/fastmcp/tools/base.py
Kludex: Starlette also has the same implementation.

PR #573, opened 2025-04-23
StreamableHttp client transport

ihrpr:
Adding support for [Streamable Http ](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)client transport 

  
### Follow ups 
 - Resumability
 - Example for fast mcp
 - client example

Stacked on top of https://github.com/modelcontextprotocol/python-sdk/pull/561


src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+    Client transport for StreamableHTTP.
+
+    `sse_read_timeout` determines how long (in seconds) the client will wait for a new
+    event before disconnecting. All other HTTP operations are controlled by `timeout`.
+
+    Yields:
+        Tuple of (read_stream, write_stream, terminate_callback)
+    """
+
+    read_stream_writer, read_stream = anyio.create_memory_object_stream[
```
jerome3o-anthropic: [nit] I find these names kinda confusing - I've started using `client_to_server_read`/`client_to_server_write` etc because I find it easier to understand

I know that we use this names in a lot of places, just flagging here because I think it would be good to start changing

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+                                        by_alias=True, mode="json", exclude_none=True
+                                    ),
+                                    headers=post_headers,
+                                ) as response:
+                                    if response.status_code == 202:
+                                        logger.debug("Received 202 Accepted")
+                                        continue
+                                    # Check for 404 (session expired/invalid)
+                                    if response.status_code == 404:
+                                        if is_initialization and session_id:
```
jerome3o-anthropic: Initializing with an existing session-id seems like an odd edge case - if initializing wouldn't we expect there to be no session id? perhaps ignoring session id on initialization (or raising?) would be better / more in line with the spec?

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+                                                "POST",
+                                                url,
+                                                json=message.model_dump(
+                                                    by_alias=True,
+                                                    mode="json",
+                                                    exclude_none=True,
+                                                ),
+                                                headers=post_headers,
+                                            ) as new_response:
+                                                response = new_response
```
jerome3o-anthropic: this seems odd to me - does this work? I would have expected closing the context manager on the stream would stop if from reading?

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+CONTENT_TYPE_JSON = "application/json"
+CONTENT_TYPE_SSE = "text/event-stream"
+
+
+@asynccontextmanager
+async def streamablehttp_client(
+    url: str,
+    headers: dict[str, Any] | None = None,
+    timeout: float = 30,
+    sse_read_timeout: float = 60 * 5,
```
dsp-ant: Should this be a timedelta?

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+# Content types
+CONTENT_TYPE_JSON = "application/json"
+CONTENT_TYPE_SSE = "text/event-stream"
+
+
+@asynccontextmanager
+async def streamablehttp_client(
+    url: str,
+    headers: dict[str, Any] | None = None,
+    timeout: float = 30,
```
dsp-ant: Also here? I assume generally we have used float for timeouts, so this is likely fine, but i assume using timedeltas would be more correct here.

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
...
+
+            async with httpx.AsyncClient(
+                headers=request_headers, timeout=timeout, follow_redirects=True
+            ) as client:
+
+                async def post_writer():
+                    nonlocal session_id
+                    try:
+                        async with write_stream_reader:
+                            async for message in write_stream_reader:
```
dsp-ant: We should probably move `post_writer` out or at least find another way to not have so many indentions that the code becomes very much squeezed.

src/mcp/client/streamableHttp.py
```diff
@@ -0,0 +1,309 @@
+"""
```
jerome3o-anthropic: [non-blocking] some of the code here is getting super nested - would love to see things broken out into functions where possible

jerome3o-anthropic (approved):
Looks reasonable - some questions. Also it would be nice to refactor this to be less nested (I will re-review quick if you decide to - also, I asked claude what it [thought for a refactor](https://gist.github.com/jerome3o-anthropic/9367454238682286f6682ee6b5d00e8b) - I don't love it but it could be a good start to iterate on)

src/mcp/client/streamableHttp.py
ihrpr: I was planning to change it everywhere when I get a moment. I literally had to write it down on a piece of paper to remember which one is which, lol. I think having them in one place different will complicate things, but agree need to change it everywhere!

src/mcp/client/streamableHttp.py
ihrpr: oh, it's wrong.  Basically this is to cover the re-connect for:
> When a client receives HTTP 404 in response to a request containing an Mcp-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.

if `is_initialization` - we just need to fail
if `session_id` - try to reconnect, BUT we need to do it on the client not client transport to have the correct initializatoin handshake

src/mcp/client/streamableHttp.py
ihrpr: well, this part was working, but it was working for initialization, which is wrong... removing!

src/mcp/client/streamableHttp.py
dsp-ant, line 1: This package name is weird and not according to pep 8 recommendations: https://peps.python.org/pep-0008/#package-and-module-names. Should be all lowercase, something like mcp.client.http.streamable

dsp-ant (changes requested):

src/mcp/client/streamableHttp.py
ihrpr: oh, Typescript... thanks, will update!

ihrpr:
> Looks reasonable - some questions. Also it would be nice to refactor this to be less nested (I will re-review quick if you decide to - also, I asked claude what it [thought for a refactor](https://gist.github.com/jerome3o-anthropic/9367454238682286f6682ee6b5d00e8b) - I don't love it but it could be a good start to iterate on)

refactoring after implementation of all the features:  https://github.com/modelcontextprotocol/python-sdk/pull/595

jerome3o-anthropic (approved):

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Header names
+MCP_SESSION_ID_HEADER = "mcp-session-id"
+LAST_EVENT_ID_HEADER = "last-event-id"
+
+# Content types
+CONTENT_TYPE_JSON = "application/json"
+CONTENT_TYPE_SSE = "text/event-stream"
```
dsp-ant, line 34: Can we share these across server and client implementation somewhere in 'mcp.shared`

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+@asynccontextmanager
+async def streamablehttp_client(
+    url: str,
+    headers: dict[str, Any] | None = None,
+    timeout: timedelta = timedelta(seconds=30),
+    sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
```
dsp-ant, line 42: ❤️

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+                "Content-Type": CONTENT_TYPE_JSON,
+                **(headers or {}),
+            }
+            # Track session ID if provided by server
+            session_id: str | None = None
+
+            async with httpx.AsyncClient(
+                headers=request_headers,
+                timeout=httpx.Timeout(timeout.seconds, read=sse_read_timeout.seconds),
+                follow_redirects=True,
```
dsp-ant, line 249: We are constructing AsyncClient everywhere in the codebase but with various different options. I think it makes sense if we just have factory function that creates an async cleint with correct default values that we want everywhere, like follow_redirects.

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+CONTENT_TYPE_SSE = "text/event-stream"
+
+
+@asynccontextmanager
+async def streamablehttp_client(
+    url: str,
+    headers: dict[str, Any] | None = None,
+    timeout: timedelta = timedelta(seconds=30),
+    sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
+):
```
dsp-ant, line 43: This should have a return value.

dsp-ant: Maybe i missed this, but all the other transport implementation don't return a terminate_callback. Should we unify this so that the interface is the same?

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+            session_id: str | None = None
+
+            async with httpx.AsyncClient(
+                headers=request_headers,
+                timeout=httpx.Timeout(timeout.seconds, read=sse_read_timeout.seconds),
+                follow_redirects=True,
+            ) as client:
+                tg.start_soon(post_writer, client)
+                try:
+                    yield read_stream, write_stream, terminate_session
```
dsp-ant, line 253: I am not sure i understand why we want terminate session as a callback. We are using a context manager, shouldn't we always be able to just terminate_session if it still exists after we yielded?

src/mcp/client/streamable_http.py
```diff
@@ -0,0 +1,258 @@
...
+                                            )
+                                        except Exception as exc:
+                                            logger.exception("Error parsing message")
+                                            await read_stream_writer.send(exc)
+                                    else:
+                                        logger.warning(f"Unknown event: {sse.event}")
+
+                            except Exception as e:
+                                logger.exception("Error reading SSE stream:")
+                                await read_stream_writer.send(e)
```
dsp-ant, line 191: It would feel more intuitive for me if these branches wouldn't be inlined but separate handler functions for non-streamed vs streamed responses.

dsp-ant (changes requested):

src/mcp/client/streamable_http.py
ihrpr: streamable http is quite unique here as we have a specific DELETE request to close the session, none of the other transports have it

src/mcp/client/streamable_http.py
ihrpr: we need this so that we can have a method on a client to delete/terminate the mcp-session

src/mcp/client/streamable_http.py
ihrpr: yep, sorry, refactoring on top of the stack

src/mcp/client/streamable_http.py
ihrpr: added to follow ups

src/mcp/client/streamable_http.py
ihrpr: Adding return value in https://github.com/modelcontextprotocol/python-sdk/pull/595

ihrpr: for termination, unlike other transports, streamable http does not close session when client disconnects, it requires explicit termination by DELETE request. Session has no idea what delete request is, hence needed to have a callback to transport. 

This has it's drawbacks, like users need to know that they need to terminate the session. Alternative can be that we pass a parameter terminate_session_on_exit which defaults to  true. In this way if someone wants to have benefits of resuming a long running session later, they can, they just need to set a parameter to False, something like 

```
 async with streamablehttp_client(url, terminate_session=False) as (
```

Akshit97:
@ihrpr Thanks for this. When are we planning to release this? I see that this PR has been merged but not released yet.

HarrisonUnifyAI:
I am also curious when the maintainers think they'll make a release. If they think it may be a while, I'll gladly try working off of main

ihrpr:
> I am also curious when the maintainers think they'll make a release. If they think it may be a while, I'll gladly try working off of main


the release is planned end of this week/early next week

Issue #375, opened 2025-03-26
Passing client context to tools


samirbajaj:
(This may be related to https://github.com/modelcontextprotocol/python-sdk/issues/373)

I would like to be able to pass some "out of band" context to a tool in an MCP server. This would be separate from the arguments that are extracted by the LLM and provided to the tool function.

How do I pass this "extra" data to the tool?

Thank you for any help.

QuangTQV:
> (This may be related to [#373](https://github.com/modelcontextprotocol/python-sdk/issues/373))
> 
> I would like to be able to pass some "out of band" context to a tool in an MCP server. This would be separate from the arguments that are extracted by the LLM and provided to the tool function.
> 
> How do I pass this "extra" data to the tool?
> 
> Thank you for any help.

I ran into the same issue. Instead of struggling to figure it out, I ended up writing my own MCP from scratch — it was much faster and way more optimized.

TuanNH76:
> (This may be related to [#373](https://github.com/modelcontextprotocol/python-sdk/issues/373))
> 
> I would like to be able to pass some "out of band" context to a tool in an MCP server. This would be separate from the arguments that are extracted by the LLM and provided to the tool function.
> 
> How do I pass this "extra" data to the tool?
> 
> Thank you for any help.

I also had this problem because my app needs to add user data to the tool

martina-zxy:
I've opened a PR to hopefully solve this issue https://github.com/modelcontextprotocol/python-sdk/pull/574

mikeedjones:
@martina-zxy PR was closed but its not clear to me how to provide a tool an argument which has not been generated by the language model in the request to the tool. The answer from the bot was to use the context object - but afaict there's no way to get content from the request into the context object.

maxisbey:
Now tracked in #2098.

PR #574, opened 2025-04-23
Allow function argument to be excluded from the tool's JSON schema

martina-zxy:
Add `client_provided_arg_model` to `FuncMetadata` to allow certain arguments to be excluded from the JSON schema. This creates a cleaner interface for LLMs while still providing the necessary context at execution time.

## Motivation and Context

With the current implementation, all function arguments are included in the JSON schema when `list_tools` is called, which prompts the LLM to provide values for all arguments when calling the tool. This can be limiting in scenarios where some arguments are intended to be provided by the client at execution time rather than being part of the LLM’s input. For example, if a tool requires a customer ID or session token the LLM should not be prompted to provide these values.

This solves https://github.com/modelcontextprotocol/python-sdk/issues/375

## How Has This Been Tested?
Unit tests have been added to verify the changes are working.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/tools/base.py
+++ b/src/mcp/server/fastmcp/tools/base.py
@@ -7,7 +7,11 @@ from typing import TYPE_CHECKING, Any, get_origin
 from pydantic import BaseModel, Field
 
 from mcp.server.fastmcp.exceptions import ToolError
-from mcp.server.fastmcp.utilities.func_metadata import FuncMetadata, func_metadata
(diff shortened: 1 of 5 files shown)


mikeedjones:
The SDK's Context object can't be used to retrieve arguments from the request which were not generated by the LM. Your proposal alternative solution does not resolve the issue.

samirbajaj:
> This is a significant issue with MCP and a real solution is required.

+1

Issue #514, opened 2025-04-14
FastMCP server with SSE transport fails to shut down on a signal


growler:
A very simple server fails to shut down on a signal if it processed at least one request:

```
❯ python test-server.py --port=8085
INFO:     Started server process [216035]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8085 (Press CTRL+C to quit)
INFO:     127.0.0.1:55188 - "GET /sse HTTP/1.1" 200 OK
INFO:     127.0.0.1:55192 - "POST /messages/?session_id=e690de9733914d09aebf2b0e8c78191c HTTP/1.1" 202 Accepted
INFO:     127.0.0.1:55192 - "POST /messages/?session_id=e690de9733914d09aebf2b0e8c78191c HTTP/1.1" 202 Accepted
INFO:     127.0.0.1:55192 - "POST /messages/?session_id=e690de9733914d09aebf2b0e8c78191c HTTP/1.1" 202 Accepted
Processing request of type CallToolRequest
^CINFO:     Shutting down
INFO:     Waiting for background tasks to complete. (CTRL+C to force quit)
```

The expected behaviour would be:

```
❯ python test-server.py --port=8085
INFO:     Started server process [216006]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8085 (Press CTRL+C to quit)
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [216006]
```

Here is the code for both the server and the client.

What do I miss?

the server

```
import click
import sys
from pydantic import Field
from mcp.server.fastmcp import FastMCP

@click.command()
@click.option("--port", default=8080, help="Port to listen", type=int)
def main(port: int):
    mcp = FastMCP(
        "mcp-echo-tool",
        debug=True,
        log_level="INFO",
        port=port,
    )

    @mcp.tool(name="echo")
    async def echo_tool(string: str = Field(description="A string to echo back")) -> str:
        """Echoes back the input string"""
        return string

    mcp.run(transport='sse')
    return 0

if __name__ == "__main__":
    sys.exit(main())
```

and the client

```
import asyncio
import click
import sys
from mcp.client.session import ClientSession
from mcp.client.sse import sse_client

async def client(url, string):
    async with sse_client(url) as (read, write):
[...]

PR #586, opened 2025-04-25
Fix: memory leak and exit problems

PWZER:
## Motivation and Context

- Here `handle_sse` calls `app.run()` every time it processes a sse request, but `app.run()` keeps running and does not stop automatically, which causes `handle_sse` to never end, and finally causes a memory leak.
https://github.com/modelcontextprotocol/python-sdk/blob/b4c7db6a50a5c88bae1db5c1f7fba44d16eebc6e/examples/servers/simple-tool/mcp_simple_tool/server.py#L67-L73

- And this is also the key reason for the following exit problems [issues#541](https://github.com/modelcontextprotocol/python-sdk/issues/514#issue-2994309207), Fixed.
  ```bash
  INFO:     Waiting for background tasks to complete. (CTRL+C to force quit)
  ```

## Types of changes

## Checklist


pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "httpx-sse>=0.4",
     "pydantic>=2.7.2,<3.0.0",
     "starlette>=0.27",
-    "sse-starlette>=1.6.1",
+    "sse-starlette>=2.3.0",
```
dsp-ant, line 30: Is this needed?

PWZER: Yes, `sse-starlette` supports `client_close_handler_callable` since version 2.3.0.

nadeemcite:
This looks great, Thanks @PWZER

src/mcp/server/sse.py
```diff
@@ -120,9 +127,19 @@ async def sse_writer():
                         }
                     )
 
+        async def client_close_handler(message: Message) -> None:
+            await read_stream_writer.aclose()
+            await write_stream_reader.aclose()
+            del self._read_stream_writers[session_id]
```
ihrpr, line 133:
```suggestion
            self._read_stream_writers.pop(session_id, None)
```

ihrpr (changes requested):
Unfortunately we can't upgrade `sse-starlette` in SDK now.

We can achieve similar functionality as `client_close_handler_callable ` provides in this case by using `disconnect_listener`

Something like this should work: 

```
            response = EventSourceResponse(
                content=sse_stream_reader,
                data_sender_callable=sse_writer,
            )

            async def disconnect_listener():
                try:
                    await response.listen_for_disconnect(receive)
                    logger.error(f"Disconnect detected for session: {session_id}")
                    await client_close_handler()  # defined earlier in the PR - name should be changed probably
                except Exception as e:
                    logger.error(f"Error in disconnect listener: {e}")

            tg.start_soon(disconnect_listener)
            logger.debug("Starting SSE response task")

            ...
            yield (read_stream, write_stream)
```

src/mcp/server/sse.py
```diff
@@ -79,7 +80,13 @@ def __init__(self, endpoint: str) -> None:
         logger.debug(f"SseServerTransport initialized with endpoint: {endpoint}")
 
     @asynccontextmanager
-    async def connect_sse(self, scope: Scope, receive: Receive, send: Send):
+    async def connect_sse(
+        self,
+        scope: Scope,
+        receive: Receive,
+        send: Send,
+        callback: Callable[[], None] | None = None,
```
lcamhoa, line 88: The callback function only is called when client connection close so  the naming is not full-meaning . And I don't think it should callback here because connect_sse [yield](https://github.com/modelcontextprotocol/python-sdk/blob/v1.3.x/src/mcp/server/sse.py#L131)

ihrpr:
Closing this PR as stale. [We can't update sse-starlette for now](https://github.com/modelcontextprotocol/python-sdk/pull/586#pullrequestreview-2802781405). 

Thank you for your contribution and working on improving this project.

PR #587, opened 2025-04-25
StreamableHttp -- resumability support for servers

ihrpr:
Adding support for resumability to servers.  

As we already have Typescript client that support resumability, testing it with it and the follow up is to add support for resumability on typescript client. 
###How to test:

Start Python example server: 
```
uv run mcp-simple-streamablehttp --log-level DEBUG
```

In the new terminal start Typescript client 

```
npx tsx src/examples/client/simpleStreamableHttp.ts
```
Follow steps: 

- type `start-notifications 10 10`
- wait 20 seconds - 
- press `esc` and `Enter`
- type `connect`
- type  `start-notifications 10 10` - you'll see missing events were replayed and the tool continues to send notifications

stacked on top of https://github.com/modelcontextprotocol/python-sdk/pull/573


examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Simple in-memory implementation of the EventStore interface for resumability.
+    This is primarily intended for examples and testing, not for production use
+    where a persistent storage solution would be more appropriate.
+    """
+
+    def __init__(self):
+        self.events: dict[
+            str, tuple[str, JSONRPCMessage, float]
+        ] = {}  # event_id -> (stream_id, message, timestamp)
```
dsp-ant: Should we use a dataclass?

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
```diff
@@ -0,0 +1,75 @@
...
+This is a simple implementation intended for examples and testing,
+not for production use where a persistent storage solution would be more appropriate.
+"""
+
+import logging
+import time
+from collections.abc import Awaitable, Callable
+from uuid import uuid4
+
+from mcp.server.streamableHttp import EventId, EventStore, StreamId
```
dsp-ant: This package name is weird and not according to pep 8 recommendations: https://peps.python.org/pep-0008/#package-and-module-names. Should be all lowercase, something like `mcp.server.http.streamable`

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Find all events for this stream after the last event
+        events_to_replay = [
+            (event_id, message)
+            for event_id, (sid, message, timestamp) in self.events.items()
+            if sid == stream_id and timestamp > last_timestamp
+        ]
+
+        # Sort by timestamp to ensure chronological order
+        events_to_replay.sort(key=lambda x: self.events[x[0]][2])
```
dsp-ant:
```suggestion
        from operator import itemgetter
        events_sorted = sorted(
            (event_id, message, timestamp)
            for event_id, (sid, message, timestamp) in self.events.items()
            if sid == stream_id and timestamp > last_timestamp
        , key=itemgetter(2))
        
        events_to_replay = [(event_id, message) for event_id, message, _ in events_sorted]
```

src/mcp/server/streamableHttp.py
```diff
@@ -57,6 +58,50 @@
...
+        Returns:
+            The generated event ID for the stored event
+        """
+        pass
+
+    @abstractmethod
+    async def replay_events_after(
+        self,
+        last_event_id: EventId,
+        send_callback: Callable[[EventId, JSONRPCMessage], Awaitable[None]],
```
dsp-ant: Since we are reusing the callback signature, maybe we can move it into a type alias?

src/mcp/server/streamableHttp.py
```diff
@@ -323,7 +373,8 @@ async def _handle_post_request(
                     response_message = None
 
                     # Use similar approach to SSE writer for consistency
-                    async for received_message in request_stream_reader:
+                    async for item in request_stream_reader:
+                        received_message, _ = item  # Extract message, ignore event_id
```
dsp-ant: Why not
```suggestion
                    async for (received_message, _) in request_stream_reader:
```

src/mcp/server/streamableHttp.py
```diff
@@ -98,8 +147,9 @@ def __init__(
 
         self.mcp_session_id = mcp_session_id
         self.is_json_response_enabled = is_json_response_enabled
+        self._event_store = event_store
         self._request_streams: dict[
-            RequestId, MemoryObjectSendStream[JSONRPCMessage]
+            RequestId, MemoryObjectSendStream[tuple[JSONRPCMessage, str | None]]
```
dsp-ant: I wonder if longterm it is more manageable if we add a typed object, a databclass or typed dict here. This feels better than positional tuples.

src/mcp/server/streamableHttp.py
```diff
@@ -639,6 +711,99 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
...
+                        if stream_id and stream_id not in self._request_streams:
+                            msg_writer, msg_reader = anyio.create_memory_object_stream[
+                                tuple[JSONRPCMessage, str | None]
+                            ](0)
+                            self._request_streams[stream_id] = msg_writer
+
+                            # Forward messages to SSE
+                            async with msg_reader:
+                                async for item in msg_reader:
+                                    message, event_id = item
```
dsp-ant:
```suggestion
                                async for (message, event_id) in msg_reader:
```

dsp-ant (changes requested):

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
```diff
@@ -0,0 +1,75 @@
...
+        self.events: dict[
+            str, tuple[str, JSONRPCMessage, float]
+        ] = {}  # event_id -> (stream_id, message, timestamp)
+
+    async def store_event(
+        self, stream_id: StreamId, message: JSONRPCMessage
+    ) -> EventId:
+        """Stores an event with a generated event ID."""
+        event_id = str(uuid4())
+        self.events[event_id] = (stream_id, message, time.time())
```
dsp-ant: I think the use of `time.time()` is risky, due to potential difference in precision across multiple platforms as well as precision loss of using floats. On old windows machines apparently precision can be 15ms or so, we might get events that appear to be on the same time. In addition we might have some reliance on system clock. We might find [`time.monotonic()`](https://docs.python.org/3/library/time.html#time.monotonic`) or `time.monotonic_ns()` better suited (i tend to prefer the latter, which likely is faster anyway)

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Get the stream ID and timestamp from the last event
+        stream_id, _, last_timestamp = self.events[last_event_id]
+
+        # Find all events for this stream after the last event
+        events_to_replay = [
+            (event_id, message)
+            for event_id, (sid, message, timestamp) in self.events.items()
+            if sid == stream_id and timestamp > last_timestamp
+        ]
```
dsp-ant: We never remove events from the dictionary. Does this mean we grow unbounded in $O(events)$? If so, we should avoid that and prune appropriately

src/mcp/server/streamableHttp.py
```diff
@@ -639,6 +711,99 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
...
+                "Connection": "keep-alive",
+                "Content-Type": CONTENT_TYPE_SSE,
+            }
+
+            if self.mcp_session_id:
+                headers[MCP_SESSION_ID_HEADER] = self.mcp_session_id
+
+            # Create SSE stream for replay
+            sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[
+                dict[str, Any]
```
dsp-ant: Is it just `Any` or do we have some form or shape of the underlying data that we can type here?

src/mcp/server/streamableHttp.py
```diff
@@ -639,6 +711,99 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
...
+                        stream_id = await event_store.replay_events_after(
+                            last_event_id, send_event
+                        )
+
+                        # If stream ID not in mapping, create it
+                        if stream_id and stream_id not in self._request_streams:
+                            msg_writer, msg_reader = anyio.create_memory_object_stream[
+                                tuple[JSONRPCMessage, str | None]
+                            ](0)
+                            self._request_streams[stream_id] = msg_writer
```
dsp-ant: will this cause leaking streams and unclosed resources?

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
ihrpr: happy to change it but also don't want to overcomplicate this, it's for demo purposes only and shoud not be used in production

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
ihrpr: really want to keep this very tiny example just for demo, but it's a good idea to limit to last N saved events

src/mcp/server/streamableHttp.py
ihrpr: updated to str, as event id is string

src/mcp/server/streamableHttp.py
ihrpr: have a separate PR to review all resources and close them: https://github.com/modelcontextprotocol/python-sdk/pull/604

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
ihrpr: removed time entirely

examples/servers/simple-streamablehttp/mcp_simple_streamablehttp/event_store.py
ihrpr: updated to keep N per session

src/mcp/server/streamable_http.py
```diff
@@ -472,6 +546,12 @@ async def _handle_get_request(self, request: Request, send: Send) -> None:
 
         if not await self._validate_session(request, send):
             return
+        # Handle resumability: check for Last-Event-ID header
+        if self._event_store:
+            last_event_id = request.headers.get(LAST_EVENT_ID_HEADER)
+            if last_event_id:
+                await self._replay_events(last_event_id, request, send)
+                return
```
dsp-ant:
```suggestion
            if last_event_id := request.headers.get(LAST_EVENT_ID_HEADER):
                await self._replay_events(last_event_id, request, send)
                return
```

dsp-ant (approved):

PR #590, opened 2025-04-27
Wrap JSONRPC messages with SessionMessage for metadata support

ihrpr:
This PR introduces the `SessionMessage` type to encapsulate JSONRPC messages with metadata. This change enables clean separation between protocol messages and transport concerns.

## Problem
Previously, memory streams propagated raw JSONRPCMessage objects, which made it difficult to pass transport-specific metadata (like resumption tokens or request correlation IDs) without violating the JSONRPC spec.

## Proposed solution

Introduced a new `SessionMessage` type that wraps `JSONRPCMessage` with optional metadata:

``` @dataclass
  class SessionMessage:
      message: JSONRPCMessage
      metadata: MessageMetadata = None
```

  With specialized metadata types:
  - ClientMessageMetadata: Contains resumption_token for client-side resumability and will contain additional fields needed for resumability 
  - ServerMessageMetadata: Contains related_request_id for server-side request correlation

 ## Changes
  - All memory streams now use SessionMessage instead of raw JSONRPCMessage
  - Session classes extract the underlying JSONRPC message for protocol operations
  - Transport implementations now can  use metadata for transport-specific features:

### DIFF
--- a/src/mcp/client/__main__.py
+++ b/src/mcp/client/__main__.py
@@ -11,8 +11,8 @@ import mcp.types as types
 from mcp.client.session import ClientSession
 from mcp.client.sse import sse_client
 from mcp.client.stdio import StdioServerParameters, stdio_client
+from mcp.shared.message import SessionMessage
 from mcp.shared.session import RequestResponder
-from mcp.types import JSONRPCMessage
 
 if not sys.warnoptions:
     import warnings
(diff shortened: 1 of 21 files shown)


bhosmer-ant (approved):
LG - would it be worth adding any tests that exercise the new fields that might appear in `SessionMessage`? Could be that comes up in later PRs but just flagging

dsp-ant (dismissed):
Looks good as far as I can tell. note that eventually we also want at that high level wrapper around JSONMessage some form of original request so that servers can introspect for example HTTP header. I had an early PR for this and merged it but it got reverted. Just something to think about as we starting to introduce that high level wrapper.

Kludex:
I think it was not considered a breaking change because the assumption is that `sse_client` etc. are only used with `ClientSession`, but since is a public API, it's not necessarily true?

PR #591, opened 2025-04-27
use metadata from SessionMessage to propagate related_request_id

ihrpr:
Refactoring the way we handle related request IDs in the streamable HTTP transport. Previously, we were adding _meta fields to notification parameters to link notifications to their triggering requests. This approach was fragile and inconsistent
  with how other transports handle message metadata.

 **Changes**
  - Moves related_request_id propagation from _meta field to proper SessionMessage
   metadata
  - Updates the shared session module to pass related request ID through SessionMessage
  metadata
  - Removes the custom _meta field handling from streamable HTTP transport
  - Updates tests to reflect the new approach

Note: stacked on top of https://github.com/modelcontextprotocol/python-sdk/pull/590

Improves related request id handling in https://github.com/modelcontextprotocol/python-sdk/pull/553

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -24,7 +24,7 @@ from starlette.requests import Request
 from starlette.responses import Response
 from starlette.types import Receive, Scope, Send
 
-from mcp.shared.message import SessionMessage
+from mcp.shared.message import ServerMessageMetadata, SessionMessage
 from mcp.types import (
     INTERNAL_ERROR,
     INVALID_PARAMS,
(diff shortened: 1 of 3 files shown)


src/mcp/server/streamable_http.py
```diff
@@ -836,12 +836,17 @@ async def message_router():
                         ):
                             # Extract related_request_id from meta if it exists
                             if (
-                                (params := getattr(message.root, "params", None))
-                                and (meta := params.get("_meta"))
-                                and (related_id := meta.get("related_request_id"))
+                                session_message.metadata is not None
+                                and isinstance(
+                                    session_message.metadata,
+                                    ServerMessageMetadata,
```
bhosmer-ant, line 840: Might they be passing their own object (say with added fields) that still has a `related_request_id`? Not sure the SDK as currently implemented would be capable of handling that (without hacking anyway), but flagging in case we want to relax this to a `hasattr`

bhosmer-ant (dismissed):
LG, one question inline

dsp-ant (dismissed):

src/mcp/server/streamable_http.py
ihrpr: that's actually the reason why typed metadata was introduced, before I was using meta field in the message, which was quite bad

PR #595, opened 2025-04-28
StreamableHttp - client refactoring and resumability support 

ihrpr:
This PR implements client-side support for resumable execution in MCP, allowing
  clients to resume interrupted long-running requests, for example, after network disconnections.

NOTE: support was added only on the lowest level, to send request, as we are not ready to add it to `call_tool` API as there is ongoing discussion about long running tasks. Once we are clear on the API design for long running tools, we will add it to session API.

  ### Summary

  - Refactor client implementation, remove nesting
  - Enable clients to resume execution after connection breaks. 
  - Add resumption token support to track tool execution progress
  - Implement SSE Last-Event-ID header for automatic event  #replay

Note: stacked on top of  https://github.com/modelcontextprotocol/python-sdk/pull/591


tests/shared/test_streamable_http.py
```diff
@@ -835,3 +981,106 @@ async def test_streamablehttp_client_session_termination(
...
+        _,
+    ):
+        async with ClientSession(
+            read_stream, write_stream, message_handler=message_handler
+        ) as session:
+            # Don't initialize - just use the existing session
+
+            # Resume the tool with the resumption token
+            assert captured_resumption_token is not None
+            result = await session.resume_tool(captured_resumption_token)
```
ihrpr: not going to leave this as session method it's not applicable to the rest of the transport - just demonstrating how it can be

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
 logger = logging.getLogger(__name__)
 
-# Header names
-MCP_SESSION_ID_HEADER = "mcp-session-id"
-LAST_EVENT_ID_HEADER = "last-event-id"
 
-# Content types
-CONTENT_TYPE_JSON = "application/json"
-CONTENT_TYPE_SSE = "text/event-stream"
+MessageOrError = SessionMessage | Exception
```
bhosmer-ant: total nit: `SessionMessageOrError`?

src/mcp/client/session.py
```diff
@@ -255,9 +260,20 @@ async def unsubscribe_resource(self, uri: AnyUrl) -> types.EmptyResult:
...
 
     async def call_tool(
-        self, name: str, arguments: dict[str, Any] | None = None
+        self,
+        name: str,
+        arguments: dict[str, Any] | None = None,
+        on_resumption_token_update: ResumptionTokenUpdateCallback | None = None,
+        resumption_token: ResumptionToken | None = None,
     ) -> types.CallToolResult:
         """Send a tools/call request."""
```
bhosmer-ant, line 263: I asked claude to write a docstring for this, because it took me a minute to figure out. not sure if you think it pays its way but fwiw. (Also note the "same name and arguments" assumption Claude is making below.)

```suggestion
      """Send a tools/call request to invoke or resume a tool operation.        
      
      This method supports two modes of operation:

      Initial Invocation:
        - Tool is executed from the beginning
        - If on_resumption_token_update is provided, it will receive checkpoint tokens

      Resumption:
        - Requires a resumption_token from a previous execution
        - Continues execution from the specified checkpoint
        - Same tool name and arguments should be used as in the initial call

      Args:
          name: Tool name to call
          arguments: Optional arguments dictionary for the tool
          on_resumption_token_update: Optional callback that receives resumption tokens
              during execution, which can be used later to resume an interrupted operation
          resumption_token: Optional token to resume from a specific checkpoint
              (required for resumption mode)
          """
```

src/mcp/client/session.py
```diff
@@ -255,9 +260,20 @@ async def unsubscribe_resource(self, uri: AnyUrl) -> types.EmptyResult:
         )
 
     async def call_tool(
-        self, name: str, arguments: dict[str, Any] | None = None
```
bhosmer-ant: I'm kind of curious why you decided against a separate `resume_tool` method - at first blush it feels clearer as an API and it avoids the need to pass dummy arguments. 

edit: looking at the way `resume_tool` was implemented in the earlier commits, I can see how the refactor disentangles the logic and avoids the fake "resume_from_token" tool call, but I do wonder whether `resume_tool` is a better Session API and (if so) whether it can be implemented in a way that preserves the current improvements. Interested in your thoughts (and definitely not looking to hold up the PR over it regardless)

bhosmer-ant (commented):
@ihrpr posting some comments early for curiosity's sake, then continuing review

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
+        """Extract and store session ID from response headers."""
+        if is_initialization:
+            new_session_id = response.headers.get(MCP_SESSION_ID)
+            if new_session_id:
+                self.session_id = new_session_id
+                logger.info(f"Received session ID: {self.session_id}")
+
+    async def _handle_sse_event(
+        self,
+        sse: Any,
```
bhosmer-ant: Should this be `sse: ServerSentEvent`?

tests/shared/test_streamable_http.py
```diff
@@ -835,3 +981,110 @@ async def test_streamablehttp_client_session_termination(
...
+                resumption_token=captured_resumption_token,
+            )
+
+            # We should get a complete result
+            assert len(result.content) == 1
+            assert result.content[0].type == "text"
+            assert "Completed" in result.content[0].text
+
+            # We should have received the remaining notifications
+            assert len(captured_notifications) > 0
```
bhosmer-ant, line 1112: should we do a stricter check here? I'm assuming it would be complicated to guarantee an exact number of notifications, but what about e.g. retaining pre- and post-resumption notifications and checking that the event ids in `pre + post` are a list of incrementing ints, something like that?

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
+            if response.status_code == 404:
+                if isinstance(message.root, JSONRPCRequest):
+                    await self._send_session_terminated_error(
+                        ctx.read_stream_writer,
+                        message.root.id,
+                    )
+                return
+
+            response.raise_for_status()
+            self._extract_session_id_from_response(response, is_initialization)
```
bhosmer-ant: looks like `_extract_session_id_from_response` does nothing unless `is_initialization` is true, might be simpler to just have the `if is_initialization` guard here?

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
+        )
+
+    def _is_initialized_notification(self, message: JSONRPCMessage) -> bool:
+        """Check if the message is an initialized notification."""
+        return (
+            isinstance(message.root, JSONRPCNotification)
+            and message.root.method == "notifications/initialized"
+        )
+
+    def _extract_session_id_from_response(
```
bhosmer-ant: naming nit: I've used the `_maybe_X` naming convention for this kind of function

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
+                    message.root.id = original_request_id
+
+                session_message = SessionMessage(message)
+                await read_stream_writer.send(session_message)
+
+                # Call resumption token callback if we have an ID
+                if sse.id and resumption_callback:
+                    await resumption_callback(sse.id)
+
+                # If this is a response or error, we're done
```
bhosmer-ant: I think we're done regardless, no?

src/mcp/client/streamable_http.py
```diff
@@ -7,259 +7,466 @@
...
+
+            async for sse in event_source.aiter_sse():
+                is_complete = await self._handle_sse_event(
+                    sse,
+                    ctx.read_stream_writer,
+                    original_request_id,
+                    ctx.metadata.on_resumption_token_update if ctx.metadata else None,
+                )
+                if is_complete:
+                    break
```
bhosmer-ant, line 238: possibly a dumb question, but if we're not returning anything to indicate that we got something back before timing out, how does the client know to (e.g.) retry the resumption request?

bhosmer-ant (approved):
Continuation of earlier review - some additional comments inline but overall looks really nice, approving so that if none of the comments warrant changes you're able to go ahead and land it

src/mcp/client/session.py
ihrpr: thank you, added with slight modifications

src/mcp/client/streamable_http.py
ihrpr: oh, leftovers from refactoring.. now we are returning bool if we are done or not

src/mcp/client/streamable_http.py
ihrpr: in `_handle_sse_event` we are sending the reponses and notifications (as well as saving resumtion tokes) before we return true or false if we need to stop listening. Spec states that the last even should be response, hence we need to stop there. Client will already received the response before we got to this break

src/mcp/client/session.py
ihrpr: resume_tool will only be applicable for one transport, introducing a new API for only one transport feels confusing
But in the same way I'm not happy with calling tool to resume the tool. 

It's all back to the discussion how we support long runnng tools. Now I feel like we should just remove the resumption from session API call_tool and later introduce clear API how to do it for any transport

ihrpr: Update: removed from `call_tool` API, left only on the lowest level send_request. Once we are happy with long running tools discussion we can add it properly

dsp-ant (dismissed):

PR #596, opened 2025-04-28
Fix Windows subprocess NotImplementedError (STDIO clients)

theailanguage:
Fix subprocess creation errors on Windows for STDIO client connections.
This updates `create_windows_process` to properly handle the case where asyncio cannot create subprocess transports on Windows by falling back to using the `subprocess` module manually.

Additionally, minor import sorting and style issues have been fixed according to `ruff` linter checks.

---

## Motivation and Context

On Windows systems, MCP clients that connect via STDIO were failing with a `NotImplementedError` during `anyio.open_process()`. This is because `asyncio` on Windows does not implement full support for subprocess transports.

This change provides a clean workaround:
- Attempt async process creation first.
- If not supported, fall back to classic `subprocess.Popen` with `asyncio` wrapping.

This allows MCP STDIO clients (e.g., [streamlit_client_ui.py](https://github.com/theailanguage/mcp_client/blob/main/streamlit_client_ui.py)) to work successfully on Windows without changing user workflows.

edit: 
the default event loop comes out as (Windows 11, Python 3.13) - 
<class 'asyncio.windows_events._WindowsSelectorEventLoop'>

My understanding is that https://github.com/modelcontextprotocol/python-sdk/pull/555 might work under the ProactorEventLoop, but my patch "might" help to support environments where that’s not the default
- Frameworks like Streamlit, which may run their own event loop internally
- Python 3.13 and above (where SelectorEventLoop is often the default on Windows)
- CI setups or test runners that don’t customize the loop policy
(note - personally I have faced this with my Streamlit UI only)

## How Has This Been Tested?

- Tested manually on Windows 11, Python 3.13.
- Ran [`streamlit_client_ui.py`](https://github.com/theailanguage/mcp_client/blob/main/streamlit_client_ui.py) to launch the MCP client.
- Successfully connected to STDIO MCP servers without errors.
- Validated tool loading and interaction without subprocess creation issues.
- Verified no new `ruff` or `pyright` issues for the updated file.

---

## Breaking Changes

No breaking changes.
- The fallback mechanism is backward compatible.
- Linux/macOS users are unaffected.
- Windows users gain compatibility without needing to change their code.

---

## Types of changes

- [x] Documentation update (style improvements, minor doc updates)

## Checklist

## Additional context

This fix is especially important for making MCP tools and agents accessible on all platforms, including Windows, thereby helping broader adoption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,11 +585,14 @@ For low level server with Streamable HTTP implementations, see:
 - Stateful server: [`examples/servers/simple-streamablehttp/`](examples/servers/simple-streamablehttp/)
 - Stateless server: [`examples/servers/simple-streamablehttp-stateless/`](examples/servers/simple-streamablehttp-stateless/)
 
+
+
 The streamable HTTP transport supports:
 - Stateful and stateless operation modes
 - Resumability with event stores
-- JSON or SSE response formats
+- JSON or SSE response formats  
 - Better scalability for multi-node deployments
+
 
 ### Mounting to an Existing ASGI Server
 
--- a/README.md
+++ b/README.md
@@ -585,14 +585,11 @@ For low level server with Streamable HTTP implementations, see:
 - Stateful server: [`examples/servers/simple-streamablehttp/`](examples/servers/simple-streamablehttp/)
 - Stateless server: [`examples/servers/simple-streamablehttp-stateless/`](examples/servers/simple-streamablehttp-stateless/)
 
-
-
 The streamable HTTP transport supports:
 - Stateful and stateless operation modes
 - Resumability with event stores
-- JSON or SSE response formats  
+- JSON or SSE response formats
 - Better scalability for multi-node deployments
-
 
 ### Mounting to an Existing ASGI Server
 
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -108,7 +108,8 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
     read_stream_writer, read_stream = anyio.create_memory_object_stream(0)
     write_stream, write_stream_reader = anyio.create_memory_object_stream(0)
 
-    command = _get_executable_command(server.command)
+    try:
+        command = _get_executable_command(server.command)
 
         # Open process with stderr piped for capture
         process = await _create_platform_compatible_process(
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -182,7 +182,13 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
             if sys.platform == "win32":
                 await terminate_windows_process(process)
             else:
-                process.terminate()
+                # On Linux, terminating a process that has already exited raises ProcessLookupError.
+                # This can happen if the command failed to launch properly (as in this test case).
+                try:
+                    process.terminate()
+                except ProcessLookupError:
+                    # Process already exited — safe to ignore
+                    pass
             await read_stream.aclose()
             await write_stream.aclose()
 
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -179,18 +179,18 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
             yield read_stream, write_stream
         finally:
             # Clean up process to prevent any dangling orphaned processes
-            if sys.platform == "win32":
-                await terminate_windows_process(process)
-            else:
-                # On Linux, terminating a process that has already exited raises ProcessLookupError.
-                # This can happen if the command failed to launch properly (as in this test case).
-                try:
+            try:
+                if sys.platform == "win32":
+                    await terminate_windows_process(process)
+                else:
                     process.terminate()
-                except ProcessLookupError:
-                    # Process already exited — safe to ignore
-                    pass
+            except ProcessLookupError:
+                # Process already exited, which is fine
+                pass
             await read_stream.aclose()
             await write_stream.aclose()
+            await read_stream_writer.aclose()
+            await write_stream_reader.aclose()
 
 
 def _get_executable_command(command: str) -> str:
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -108,8 +108,7 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
     read_stream_writer, read_stream = anyio.create_memory_object_stream(0)
     write_stream, write_stream_reader = anyio.create_memory_object_stream(0)
 
-    try:
-        command = _get_executable_command(server.command)
+    command = _get_executable_command(server.command)
 
         # Open process with stderr piped for capture
         process = await _create_platform_compatible_process(
@@ -179,18 +178,12 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
             yield read_stream, write_stream
         finally:
             # Clean up process to prevent any dangling orphaned processes
-            try:
-                if sys.platform == "win32":
-                    await terminate_windows_process(process)
-                else:
-                    process.terminate()
-            except ProcessLookupError:
-                # Process already exited, which is fine
-                pass
+            if sys.platform == "win32":
+                await terminate_windows_process(process)
+            else:
+                process.terminate()
             await read_stream.aclose()
             await write_stream.aclose()
-            await read_stream_writer.aclose()
-            await write_stream_reader.aclose()
 
 
 def _get_executable_command(command: str) -> str:
--- a/src/mcp/client/stdio/win32.py
+++ b/src/mcp/client/stdio/win32.py
@@ -78,6 +78,18 @@ class DummyProcess:
         """Terminate and wait on process exit inside a thread."""
         self.popen.terminate()
         await to_thread.run_sync(self.popen.wait)
+        
+        # Close the file handles to prevent ResourceWarning
+        if self.stdin:
+            await self.stdin.aclose()
+        if self.stdout:
+            await self.stdout.aclose()
+        if self.stdin_raw:
+            self.stdin_raw.close()
+        if self.stdout_raw:
+            self.stdout_raw.close()
+        if self.stderr:
+            self.stderr.close()
 
     async def wait(self):
         """Async wait for process completion."""
--- a/src/mcp/client/stdio/win32.py
+++ b/src/mcp/client/stdio/win32.py
@@ -10,6 +10,10 @@ from typing import TextIO
 
 import anyio
 from anyio.abc import Process
+from anyio.streams.file import FileReadStream, FileWriteStream
+
+from typing import Optional, TextIO, Union
+from pathlib import Path
 
 
 def get_windows_executable_command(command: str) -> str:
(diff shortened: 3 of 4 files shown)


theailanguage:
Hi team! 

This is my first pull request. PR fixes Windows compatibility for MCP STDIO clients by introducing a fallback to subprocess.Popen wrapped with async I/O streams. It ensures that Windows users can now use streamlit and similar STDIO clients without errors.

Please let me know if you'd like any further changes. Thank you for your time and for maintaining such an awesome project! 

Regards
[Kartik / theailanguage]

theailanguage:
@dsp-ant @jerome3o-anthropic would appreciate any kind of feedback on this if possible from your end. Thanks a ton!

Regards
[Kartik / theailanguage]

BenMawnMahlauNBTC:
https://github.com/modelcontextprotocol/python-sdk/pull/555 solves this in a much cleaner way

theailanguage:
> #555 solves this in a much cleaner way

Thanks for sharing 🙏 I just tested it on my setup (Windows 11 + Python 3.13), and I’m still hitting the same error:
`raise NotImplementedError`

I think I have tested #555 well by updating the init and win32 python files in my .venv mcp path with the edits suggested here.

Baically #555 still uses await anyio.open_process(...) in both primary and fallback paths, which internally relies on asyncio.create_subprocess_exec(). Unfortunately, that call raises NotImplementedError on my machine, so the fallback logic never helps.

In my fork, I catch the failure and bypass the entire async subprocess path by:
Falling back to a sync subprocess.Popen
Then manually wrapping stdin, stdout in FileWriteStream and FileReadStream
This makes it work regardless of the event loop or OS internals

So for Windows users with newer Python versions or CI environments, my fallback is the only one that works when the async event loop doesn’t support subprocesses.

Let me know what you think. Have you tested this in event loop scenarios like the one pointed out in this PR?

BenMawnMahlauNBTC:
I am on windows 11; also using it in an event loop. are you using windows proactor for the async loop?

theailanguage:
> I am on windows 11; also using it in an event loop. are you using windows proactor for the async loop?
> 
> edit: I also have this change in the version i am using
> 
> `mcp\client\session.py`,
> 
> ```python
> import asyncio
> import os
> if os.name == 'nt':
>   asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
> ```

Aah that explains it :) 

For me it is 
`<class 'asyncio.windows_events._WindowsSelectorEventLoop'>`

This I think is the default.

I actually tried to add your session.py code but cannot get the event loop to change for some reason (I guess due to Streamlit's own eventloop). 

I understand #555 might work perfectly under the ProactorEventLoop, but this patch "might" help to support environments where that’s not the default
- Frameworks like Streamlit, which may run their own event loop internally
- Python 3.13 and above (where SelectorEventLoop is often the default on Windows)
- CI setups or test runners that don’t customize the loop policy

(note - I searched for and found the last 2 cases. Personally I have faced this with my Streamlit UI only)


That said, I really appreciate the clean structure of the other solution.

theailanguage:
One more thing, in case you'd want to look at the implementation of the streamlit code - you can have a look at 

Repo - https://github.com/theailanguage/mcp_client
UI code - https://github.com/theailanguage/mcp_client/blob/main/streamlit_client_ui.py

This is a tutorial video for the same that I built for a course
https://youtu.be/Ln-Tgz8Pmek

theailanguage:
Thanks for your reply. Yes, I'll check these.

theailanguage:
@ihrpr I checked the failing Windows checks. This doesn’t seem to be related to the code changes I made, but rather to the test setup itself. The four checks are all failing in 2 tests out of the several others that pass.

**Failed Test Details**

Both failing tests (test_stdio_context_manager_exiting and test_stdio_client) are raising:
```
PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>
ResourceWarning: unclosed file
```

Location
`tests\client\test_stdio.py:12 ` and 
`tests\client\test_stdio.py:19` on win32. 

Code - 
```
@pytest.mark.anyio
@pytest.mark.skipif(tee is None, reason="could not find tee command")
async def test_stdio_context_manager_exiting():
    ...

@pytest.mark.anyio
@pytest.mark.skipif(tee is None, reason="could not find tee command")
async def test_stdio_client():
    ...
```

**My understanding is - **
It appears the issue is with the use of the tee command. I believe on Windows, tee equivalent may not behave reliably with subprocess handling. Note sure what "tee" the CI on GitHub uses but these may not close file handles cleanly, leading to the ResourceWarning.


**Possible Next Steps**
I just want to check if that is the intention - to skip on windows? One possible solution in this case could be to explicitly skip these tests on Windows using something like:
```
@pytest.mark.skipif(
    tee is None or sys.platform.startswith("win"),
    reason="tee command not available or platform is Windows"
)

```

If not, then I believe the test files need to be updated (assuming my understanding is correct). So that might need changes to the test itself. 

Any suggestion on the next course of action to address this would be appreciated. 

Thanks
Kartik / theailanguage

theailanguage:
@ihrpr I have requested a re-review based on my last comment itself because if the test may be skipped then the checks can be passed with the existing PR code itself. No changes, kindly let me know what you think. Thanks a lot!

intval:
@ihrpr We'd appreciate if this is prioritized higher. MCP is not usable under windows.
Same issue occurs when using google agent development kit, not only streamlit. 
https://github.com/google/adk-python/issues/1321 

@theailanguage thanks for the contribution.

README.md
```diff
@@ -315,42 +315,27 @@ async def long_task(files: list[str], ctx: Context) -> str:
 Authentication can be used by servers that want to expose tools accessing protected resources.
 
 `mcp.server.auth` implements an OAuth 2.0 server interface, which servers can use by
-providing an implementation of the `OAuthAuthorizationServerProvider` protocol.
+providing an implementation of the `OAuthServerProvider` protocol.
```
felixweinberger: I assume these README changes are from merging in main and going out of sync at some point - rebase on main and re-push should likely fix.

src/mcp/client/stdio/__init__.py
```diff
@@ -108,28 +108,20 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
...
-        # Clean up streams if process creation fails
-        await read_stream.aclose()
-        await write_stream.aclose()
-        await read_stream_writer.aclose()
-        await write_stream_reader.aclose()
-        raise
+    command = _get_executable_command(server.command)
+
+    # Open process with stderr piped for capture
+    process = await _create_platform_compatible_process(
```
felixweinberger: probably unrelated changes coming from merging in main?

tests/client/test_stdio.py
```diff
@@ -56,43 +50,3 @@ async def test_stdio_client():
...
-        async with stdio_client(server_params) as (_, _):
-            pass
-
-    # The error should indicate the command was not found
-    error_message = str(exc_info.value)
-    assert (
-        "nonexistent" in error_message
-        or "not found" in error_message.lower()
-        or "cannot find the file" in error_message.lower()  # Windows error message
-    )
```
felixweinberger: these tests seem unrelated to this change (though I might be wrong), were they intentionally removed?

theailanguage:
@felixweinberger thanks for the suggestions, will work on this now!

Regards
Kartik / theailanguage

tests/client/test_stdio.py
theailanguage: Thanks for the comment. These are unrelated to the change. I have rebased to the original tests. Then I have made one change to the bad file path test 
`async def test_stdio_client_bad_path():`

The intentional related change made this time is -     
```
# Removed `-c` to simulate a real "file not found" error instead of
# executing a bad inline script.
# This ensures the subprocess fails to launch, which better matches
# the test's intent.
server_params = StdioServerParameters(command="python", args=["non-existent-file.py"])

``` 

The args initially had a -c that instructs python to execute non-existent-file.py as a literal script with just the file name rather than simulate a bad path!

Thanks
Kartik / theailanguage

src/mcp/client/stdio/__init__.py
theailanguage: Thanks for checking, yes, unrelated.

README.md
theailanguage: Yes, that's right. I have rebased. Thanks!

theailanguage:
@felixweinberger I have rebased the code, included my original changes and made two additional changes as mentioned below. Kindly review and thanks a lot for your feedback!

Regards
Kartik / theailanguage

1. src\mcp\client\stdio\\_\_init__.py

```diff
        finally:
            # Clean up process to prevent any dangling orphaned processes
            if sys.platform == "win32":
                await terminate_windows_process(process)
            else:
-              process.terminate()
+             # On Linux, terminating a process that has already exited raises ProcessLookupError.
+             # This can happen if the command failed to launch properly (as in this test case).
+             try:
+                 process.terminate()
+             except ProcessLookupError:
+                # Process already exited — safe to ignore
+                pass

```
2. tests\client\test_stdio.py
```diff
@pytest.mark.anyio
async def test_stdio_client_bad_path():
    """Check that the connection doesn't hang if process errors."""

+   # Removed `-c` to simulate a real "file not found" error instead of
+   # executing a bad inline script.
+   # This ensures the subprocess fails to launch, which better matches
+   # the test's intent.
+   server_params = StdioServerParameters(command="python", args=["non-existent-file.py"])
-    server_params = StdioServerParameters(command="python", args=["-c", "non-existent-file.py"])
    async with stdio_client(server_params) as (read_stream, write_stream):
```

src/mcp/client/stdio/__init__.py
```diff
@@ -179,18 +179,18 @@ async def stdin_writer():
...
-            except ProcessLookupError:
-                # Process already exited, which is fine
-                pass
+                except ProcessLookupError:
+                    # Process already exited — safe to ignore
+                    pass
             await read_stream.aclose()
             await write_stream.aclose()
-            await read_stream_writer.aclose()
-            await write_stream_reader.aclose()
```
felixweinberger: I don't think we should be removing these 2 `write.aclose()` lines or modifying this file at all - was this necessary to make tests pass or something? Or a potential leftover from the rebase?

README.md
```diff
@@ -585,12 +585,15 @@ For low level server with Streamable HTTP implementations, see:
...
+
+
 The streamable HTTP transport supports:
 - Stateful and stateless operation modes
 - Resumability with event stores
-- JSON or SSE response formats
+- JSON or SSE response formats  
 - Better scalability for multi-node deployments
 
+
```
felixweinberger: nit: can we make sure this PR doesn't make any unrelated changes to README?

tests/client/test_stdio.py
```diff
@@ -57,7 +57,12 @@ async def test_stdio_client():
 @pytest.mark.anyio
 async def test_stdio_client_bad_path():
     """Check that the connection doesn't hang if process errors."""
-    server_params = StdioServerParameters(command="python", args=["-c", "non-existent-file.py"])
+
+    # Removed `-c` to simulate a real "file not found" error instead of
+    # executing a bad inline script.
+    # This ensures the subprocess fails to launch, which better matches
+    # the test's intent.
+    server_params = StdioServerParameters(command="python", args=["non-existent-file.py"])
```
felixweinberger: Are the changes to this file necessary at all? The original test with `-c` works correctly and both versions test error handling adequately.

felixweinberger (changes requested):
Thank you for rebasing and making these updates so quickly!

I just wanted to understand the changes outside `win32.py` better.

Are these changes actually needed? I checked out a local copy of this PR again to run the tests and it seems to me like the only necessary changes are the ones in `src/mcp/client/stdio/win32.py`. I ran the tests on on both Windows and macOS.

I also tried my rebased version or your fix ([which only modifies `win32.py`](https://github.com/modelcontextprotocol/python-sdk/pull/1009/files)) on a Windows machine and all tests seem to pass, so curious if these other changes to `stdio` are genuinely necessary or we can keep this PR isolated to `win32.py` only.

Could you confirm if these are actually needed? If possible and it addresses the Windows issues for you I would strongly prefer to keep all fixes in this PR limited to the `win32.py` file only.

tests/client/test_stdio.py
theailanguage: No this is not strictly needed. My understanding was that this is a bad path test and the intention was for non-existent-file.py to act as a file name that does not exist. The -c flag makes it into a literal command like a print statement so I thought to include this as well. I will remove and try to push without this.

Thanks!

README.md
theailanguage: Sorry about this one, I am not sure how this slipped in. Will fix this! Was not intended.

Thanks!

src/mcp/client/stdio/__init__.py
theailanguage: No this was not an intended change. I did rebase and kept all the upstream origin code blocks while completing it. Let me just limit the changes to win32.py, do a local diff check and the commit. I think those changes should be sufficient.

Thanks!

theailanguage:
@felixweinberger I am updating the PR with just the win32.py changes - will confirm with a diff before pushing and then request to review. Thanks!

theailanguage:
Hi @felixweinberger, I’ve again pushed the code with only one file updated, the win32 client. You're right that all the CI tests (except one that fails for me) pass.

For the one test that is failing for now (test_stdio_context_manager_exiting ) - it seems unrelated to my change (which only touches win32.py). And these tests were passing earlier. 

Could I request you to please see if this test fails at your end too? Please let me know if I should do anything about this one failing test! 

Thanks
Kartik / theailanguage

felixweinberger:
> Hi @felixweinberger, I’ve again pushed the code with only one file updated, the win32 client. You're right that all the CI tests (except one that fails for me) pass.
> 
> For the one test that is failing for now (test_stdio_context_manager_exiting ) - it seems unrelated to my change (which only touches win32.py). And these tests were passing earlier.
> 
> Could I request you to please see if this test fails at your end too? Please let me know if I should do anything about this one failing test!
> 
> Thanks Kartik / theailanguage

I triggered a rebuild and the tests pass, looks like it was just a flake (happens occasionally on the windows CI builds, known problem).

felixweinberger (approved):
Huge thanks for all the iterations on this @theailanguage!

theailanguage:
@felixweinberger Thanks for reviewing and the feedback!

src/mcp/client/stdio/win32.py
```diff
@@ -44,49 +46,122 @@ def get_windows_executable_command(command: str) -> str:
         return command
 
 
+class DummyProcess:
```
ihrpr, line 49:
```suggestion
class FallbackProcess:
```

felixweinberger: I'll fix the naming in a follow up

PR #597, opened 2025-04-28
fix for SSE URL handling when a server name is specified

### LINKED ISSUE #585: Python MCP SDK: Incorrect handling of absolute paths in SSE endpoint URLs leads to incorrect URL construction.
**Describe the bug**
When the MCP server, connected via SSE, returns an endpoint event with a data value that starts with / (e.g., /mcp/message?sessionId=...), the SDK appears to be using this absolute path directly without considering the base URL of the initial SSE connection. This results in the client attempting to access an incorrect URL.

**To Reproduce**

1. Establish an SSE connection to an MCP server with a URL like: http://my.mcp-server.com/weather/sse.
2. Observe the server sending an SSE event with event: endpoint and data: /mcp/message?sessionId=your_session_id.
3. The client, using the Python MCP SDK, constructs the endpoint URL as http://my.mcp-server.com/mcp/message?sessionId=your_session_id.
4. The expected behavior, based on the MCPEx specification (clients should use the same origin as the SSE connection and append the received path), is that the URL should be http://my.mcp-server.com/weather/mcp/message?sessionId=your_session_id.

**Expected behavior**
The expected URL should be http://my.mcp-server.com/weather/mcp/message?sessionId=your_session_id.

**Possible Root Cause:**
Relevant Code Snippet (from langchain-mcp-adapters/client.py - for context):

Image

 
urllib.parse.urljoin treats paths starting with / as absolute paths, causing it to discard the path component of the base URL (http://my.mcp-server.com/weather/sse) and only use the hostname.

**Suggested Solution**
When the server returns an endpoint URL starting with /, the SDK should treat it as a path to be appended to the path of the initial SSE connection URL, rather than a completely new absolute path relative to the server root. This might involve checking if sse_data starts with / and adjusting the base URL used in the URL joining process accordingly.

Image

aarora79:
# Add custom URL join function for SSE endpoints

This PR adds a custom URL join function to properly handle SSE endpoint URLs by correctly managing path segments regardless of trailing slashes.

## Motivation and Context

When connecting to SSE endpoints, we need to reliably join base URLs with endpoint paths. For example if my SSE server lives at `https://mcp.example.com/weather/sse/` then I want to join the SSE event URL such as `/messages/?session_id=616df71373444d76bd566df4377c9629` to create `https://mcp.example.com/weather/messages/?session_id=616df71373444d76bd566df4377c9629` i.e. it includes the server name `weather`, the existing implementation does not include the server name. This custom function ensures paths are correctly joined including a server name if present in the path.

## How Has This Been Tested?

- Added comprehensive unit tests covering multiple base URL patterns:
  - `https://mcp.example.com/weather/sse/` 
  - `https://mcp.example.com/weather/clarksburg/sse/`
  - `https://mcp.example.com/sse/`
- Tested with and without trailing slashes
- Tested with session ID query parameters
- Verified behavior with actual SSE connection endpoints

## Breaking Changes

- [x] None. This is a new utility function that doesn't affect existing functionality.

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)

## Additional context

The function uses `urlparse` and `urlunparse` to properly handle all URL components, which ensures query parameters and fragments are preserved. We specifically handle the path joining logic to accommodate both trailing and non-trailing slash cases in base URLs.

Resolves #585

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -1,7 +1,7 @@
 import logging
 from contextlib import asynccontextmanager
 from typing import Any
-from urllib.parse import urljoin, urlparse
+from urllib.parse import urljoin, urlparse, urlunparse
 
 import anyio
 import httpx
@@ -16,6 +16,31 @@ logger = logging.getLogger(__name__)
 
 def remove_request_params(url: str) -> str:
     return urljoin(url, urlparse(url).path)
+
+
+def custom_url_join(base_url: str, endpoint: str) -> str:
+    """
+    Custom URL join function to handle the case where the endpoint is relative
+    to a base URL. This function ensures that the base URL and endpoint are
+    combined correctly, even if the endpoint is not a full URL.
+    """
+    # Parse the base URL
+    parsed_base = urlparse(base_url)
+
+    # Get the path prefix (e.g., '/weather')
+    path_prefix = "/".join(parsed_base.path.split("/")[:-1])
+
+    # Remove any leading slash from the endpoint
+    clean_endpoint = endpoint.lstrip("/")
+
+    # Create the new path by joining prefix and endpoint
+    new_path = f"{path_prefix}/{clean_endpoint}"
+
+    # Create a new parsed URL with the updated path
+    parsed_new = parsed_base._replace(path=new_path)
+
+    # Convert back to a string URL
+    return urlunparse(parsed_new)
 
 
 @asynccontextmanager
@@ -43,6 +68,15 @@ async def sse_client(
     async with anyio.create_task_group() as tg:
         try:
             logger.info(f"Connecting to SSE endpoint: {remove_request_params(url)}")
+
+            # extract MCP server name from URL, for example: https://mcp.example.com/weather/sse
+            # will extract 'weather'
+            path_tokens = urlparse(url).path.split("/")
+            optional_mcp_server_name = (
+                path_tokens[1:-1] if len(path_tokens) > 2 else None
+            )
+            logger.debug(f"MCP Server name (optional): {optional_mcp_server_name}")
+
             async with httpx.AsyncClient(headers=headers) as client:
                 async with aconnect_sse(
                     client,
@@ -61,7 +95,12 @@ async def sse_client(
                                 logger.debug(f"Received SSE event: {sse.event}")
                                 match sse.event:
                                     case "endpoint":
-                                        endpoint_url = urljoin(url, sse.data)
+                                        if optional_mcp_server_name:
+                                            endpoint_url = custom_url_join(
+                                                url, sse.data
+                                            )
+                                        else:
+                                            endpoint_url = urljoin(url, sse.data)
                                         logger.info(
                                             f"Received endpoint URL: {endpoint_url}"
                                         )
--- a/tests/client/test_sse.py
+++ b/tests/client/test_sse.py
@@ -1,0 +1,31 @@
+import pytest
+from urllib.parse import urlparse, urlunparse
+
+from mcp.client.sse import custom_url_join
+
+
+@pytest.mark.parametrize(
+    "base_url,endpoint,expected",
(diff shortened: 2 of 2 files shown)


dwayn:
I have been digging into this exact problem, as I am working on building a custom gateway for MCP servers, and the problem really lies on the server side that it is returning an absolute path of "/messages" regardless of the path that it is mounted on. The client will successfully connect if it is given the correct path by the server when it connects to the sse endpoint as it is currently implemented.

I found a hacky way to work around this in my dynamic router implementation by setting the fastapi request scope's root_path to "" and altering the starlette app returned by FastMCP.sse_app() to set "sse_path" and "messages_path" to include the full path to where it is mounted. This results in the starlette app acting as if you set the configuration of the FastMCP server at creation settings with sse_path="/weather/sse" and messages="/weather/messages/" and then call sse_app() and mount it at "/" on the root router (this is the reason for rewriting root_path on the incoming request before passing to the FastMCP starlette app).

Another hacky workaround that could be done with current state is to set sse_path and messages_path for the FastMCP server and then create the starlette app with sse_app(). Then rewrite the routes on it to be "/sse" and "/messages/" and mount it on the intended path.

What we need is a clean way to set/update a prefix path that it uses to announce the messages endpoint, but doesn't affect route validation. I would love if it could be late bound (after generating the starlette app) with maybe a closure, but even being able to just statically configure it as a root_path or similar property before calling sse_app and then having that prepended to the endpoint message would be helpful.

I have been fighting this problem the last few days in a dynamic gateway library I have been working on, just trying to work around it in the least hacky way. I think I may just take a shot at submitting a PR to add the ability to configure a root_path for the FastMCP server to allow us to mount on any path we like and have it work as expected.

dwayn:
> I have been digging into this exact problem, as I am working on building a custom gateway for MCP servers, and the problem really lies on the server side that it is returning an absolute path of "/messages" regardless of the path that it is mounted on. The client will successfully connect if it is given the correct path by the server when it connects to the sse endpoint as it is currently implemented.
> 
> I found a hacky way to work around this in my dynamic router implementation by setting the fastapi request scope's root_path to "" and altering the starlette app returned by FastMCP.sse_app() to set "sse_path" and "messages_path" to include the full path to where it is mounted. This results in the starlette app acting as if you set the configuration of the FastMCP server at creation settings with sse_path="/weather/sse" and messages="/weather/messages/" and then call sse_app() and mount it at "/" on the root router (this is the reason for rewriting root_path on the incoming request before passing to the FastMCP starlette app).
> 
> Another hacky workaround that could be done with current state is to set sse_path and messages_path for the FastMCP server to include full path in settings and then create the starlette app with sse_app(). Then rewrite the routes on it to be "/sse" and "/messages/" and mount it on the intended path.
> 
> What we need is a clean way to set/update a prefix path that it uses to announce the messages endpoint, but doesn't affect route validation. I would love if it could be late bound (after generating the starlette app) with maybe a closure, but even being able to just statically configure it as a root_path or similar property before calling sse_app and then having that prepended to the endpoint message would be helpful.
> 
> I have been fighting this problem the last few days in a dynamic gateway library I have been working on, just trying to work around it in the least hacky way. I think I may just take a shot at submitting a PR to add the ability to configure a root_path for the FastMCP server to allow us to mount on any path we like and have it work as expected. 

Actually, it looks like #540 is addressing exactly what I am calling out here.

dwayn:
For anyone else wanting a fix for this problem, I wrote a middleware for the server side that works around it by rewriting the endpoint url in the initial message that is sent back by the server upon connection to SSE endpoint.

Source here: https://github.com/dwayn/fastmcp-mount
It is also uploaded to PyPi: https://pypi.org/project/fastmcp-mount/

ihrpr:
Thank you, 540 should be fixing this issue

PR #604, opened 2025-04-29
Streamable Http - clean up server memory streams

ihrpr:
Saving only one read stream in `_request_streams` is not right as we need to clean up both - read and write streams.


src/mcp/server/streamable_http.py
```diff
@@ -658,19 +666,27 @@ def _terminate_session(self) -> None:
...
+                await self._read_stream_writer.aclose()
+            if self._read_stream is not None:
+                await self._read_stream.aclose()
+            if self._write_stream_reader is not None:
+                await self._write_stream_reader.aclose()
+            if self._write_stream is not None:
+                await self._write_stream.aclose()
+        except Exception as e:
+            logger.debug(f"Error closing streams: {e}")
+            pass
```
bhosmer-ant: redundant?

src/mcp/server/streamable_http.py
```diff
@@ -884,9 +911,20 @@ async def message_router():
...
+
+                # Clean up the read and write streams
+                try:
+                    await read_stream_writer.aclose()
+                    await read_stream.aclose()
+                    await write_stream_reader.aclose()
+                    await write_stream.aclose()
+                except Exception as e:
+                    logger.debug(f"Error closing streams: {e}")
+                    pass
```
bhosmer-ant: redundant?

src/mcp/server/streamable_http.py
```diff
@@ -884,9 +911,20 @@ async def message_router():
...
                 yield read_stream, write_stream
             finally:
-                for stream in list(self._request_streams.values()):
+                for stream_id in list(self._request_streams.keys()):
                     try:
-                        await stream.aclose()
-                    except Exception:
+                        await self._clean_up_memory_streams(stream_id)
+                    except Exception as e:
+                        logger.debug(f"Error closing request stream: {e}")
```
bhosmer-ant, line 915: Not part of your diff but I think the `pass` on the next line is redundant as well

src/mcp/server/streamable_http.py
```diff
@@ -239,6 +245,19 @@ def _create_event_data(self, event_message: EventMessage) -> dict[str, str]:
...
 
+    async def _clean_up_memory_streams(self, request_id: RequestId) -> None:
+        """Clean up memory streams for a given request ID."""
+        if request_id in self._request_streams:
+            try:
+                # Close the request stream
+                await self._request_streams[request_id][0].aclose()
+                await self._request_streams[request_id][1].aclose()
+            except Exception as e:
+                logger.debug(f"Error closing memory streams: {e}")
```
bhosmer-ant, line 256: What exceptions would we expect to see here? I'm slightly worried about swallowing everything, since (I think) `aclose()` itself shouldn't normally throw anything

src/mcp/server/streamable_http.py
```diff
@@ -658,19 +666,27 @@ def _terminate_session(self) -> None:
...
+        try:
+            if self._read_stream_writer is not None:
+                await self._read_stream_writer.aclose()
+            if self._read_stream is not None:
+                await self._read_stream.aclose()
+            if self._write_stream_reader is not None:
+                await self._write_stream_reader.aclose()
+            if self._write_stream is not None:
+                await self._write_stream.aclose()
+        except Exception as e:
```
bhosmer-ant, line 685: Same q about possibly narrowing down the exceptions we really want to suppress here (maybe all? but just flagging)

bhosmer-ant (dismissed):
LGTM, couple minor its inline but your call whether they're worth the time

PR #610, opened 2025-05-01
Auth SSE simple example

ihrpr:
Adding a simple example for auth which was implemented in https://github.com/modelcontextprotocol/python-sdk/pull/255/

This tiny server is useful: 
- testing auth, will need to do the same for streamable http 
- testing inspector
- testing clients 

Tested with Inspector
image


jerome3o-anthropic (dismissed):
Had a quick look over this - and had claude look at it ([here](https://gist.github.com/jerome3o-anthropic/3ac1d86326bc59930daf868533d35cde)).

I think this is good for demo purposes. I'll give it a try before approving the bigger one

dsp-ant (approved):

pcarleton (approved):
LGTM!

I tested with the inspector, and a github app I had around, and works.

I separated out the confused deputy fix in a follow up since it's going to require some SDK changes that I don't want to block this on:
https://github.com/modelcontextprotocol/python-sdk/pull/648

PR #612, opened 2025-05-02
Handle SSE Disconnects Properly

akash329d:
## Motivation and Context

The `EventSourceResponse` from https://github.com/sysid/sse-starlette returns on SSE session disconnect. However, since we start this in a sub task group, the current implementation does nothing when a disconnect happens. This causes the session `Route` to stay open forever in these cases. 

This PR modifies this behavior to close the streams in the SSE server when EventSourceResponse wraps up, and updates `ServerSession` to properly wrap up and clean up its streams as well when the upstream streams close. 

This exposes a different error in that we have this set up as a `Route` in most cases which shouldn't return `None`. By fixing the 'running forever' issue we end up seeing exceptions related to this whenever a client disconnects. I updated `mount_sse` to return an empty response to fix this. 

## How Has This Been Tested?
Tested manually and ensured that the route was closed when clients close the connection.

## Breaking Changes

It isn't breaking, but users need to change the `handle_sse` function to a `Mount` to not get `"TypeError: 'NoneType' object is not callable"` errors when client disconnects.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -19,8 +19,8 @@ from pydantic import BaseModel, Field
 from pydantic.networks import AnyUrl
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from starlette.applications import Starlette
-from starlette.requests import Request
 from starlette.routing import Mount, Route
+from starlette.types import Receive, Scope, Send
 
 from mcp.server.fastmcp.exceptions import ResourceError
 from mcp.server.fastmcp.prompts import Prompt, PromptManager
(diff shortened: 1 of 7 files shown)


src/mcp/server/fastmcp/server.py
```diff
@@ -87,7 +87,7 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     # HTTP settings
     host: str = "0.0.0.0"
     port: int = 8000
-    sse_path: str = "/sse"
+    sse_path: str = "/sse/"
```
praboud-ant: Why the trailing `/`?

akash329d: Apparently `Mount` makes it trailing by default. 

> Starlette adds a trailing slash to the mount path by default and redirects requests without a trailing slash to the path with a trailing slash.

Our test client doesn't follow redirects so didn't like it.

src/mcp/server/session.py
```diff
@@ -137,6 +137,10 @@ def check_client_capability(self, capability: types.ClientCapabilities) -> bool:
 
         return True
 
+    async def _receive_loop(self) -> None:
+        async with self._incoming_message_stream_writer:
```
praboud-ant, line 138: I think we also should remove the cleanup `_exit_stack` task for `_incoming_message_stream_writer`

src/mcp/server/fastmcp/server.py
akash329d: Discussed offline; we don't want to break current users so switched back to `Route` but return an empty `Response` to get around this.

praboud-ant (approved):

PR #615, opened 2025-05-02
When no authorization is required dont fail on missing user scope 401 Unauthorized 

yonigottesman:
While developing a mcp server that wil run locally I want to use the sse protocal and not stdio. Because im running local i dont need any authentication, but the default RequireAuthMiddleware will fail if if not isinstance(auth_user, AuthenticatedUser).

I want the solution to be
if not isinstance(auth_user, AuthenticatedUser) and self.required_scopes: so if there are no required scopes I wont have to do anything and i can just connect to the server withtout any autherization

## How Has This Been Tested?
I tested it locally to work with cursor

### DIFF
--- a/src/mcp/server/auth/middleware/bearer_auth.py
+++ b/src/mcp/server/auth/middleware/bearer_auth.py
@@ -74,7 +74,7 @@ class RequireAuthMiddleware:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         auth_user = scope.get("user")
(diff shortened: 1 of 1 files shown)


ihrpr:
@yonigottesman thank you for this PR, this actually highlights a different issue, we need to make sure that when Auth settings aren't set in FastMCP with SSE, we should not even try to go though auth middleware. We'll work on the fix and include it in the next release.

ihrpr:
should be fixed in v1.7.1

PR #625, opened 2025-05-04
Elicitation prototype

ihrpr:


src/mcp/server/fastmcp/server.py
```diff
@@ -1002,6 +1002,39 @@ async def read_resource(self, uri: str | AnyUrl) -> Iterable[ReadResourceContent
         ), "Context is not available outside of a request"
         return await self._fastmcp.read_resource(uri)
 
+    async def elicit(
+        self,
+        message: str,
+        requestedSchema: dict[str, Any],
+    ) -> dict[str, Any]:
```
Kludex: I think what you want would be something like this:

```py
async def elicit(self, schema: type[T], message: str | None = None) -> T: ...
```

And then it can be used inside the tools as:

```py
from typing_extensions import TypedDict

app = FastMCP()

class MyServerSchema(TypedDict):
    name: str

@app.tool()
async def potato_tool(ctx: Context):
    content = await ctx.elicit(MyServerSchema)
```

Kludex: I added `message` as optional because I can imagine us extracting the message from the docstring of `MyServerSchema`. For the best user experience.

ihrpr: Thank you! I,plemented `elicit`, looks nice and typesafa now!

---- 
Re message in the schema: 

The message and schema serve distinct purposes in the elicitation flow:
  - The **message** is the contextual question or prompt that explains *why* we're asking for information (e.g., "No tables available at 7 PM. Would you like to try 8 PM instead?")
  - The **schema** defines *what* information we're collecting and its structure (e.g., fields for confirmation, alternative time, etc.)

 This separation allows for:
  1. **Dynamic context** - The same schema can be reused with different messages based on the situation
  2. **Clearer intent** - The message can provide specific context that wouldn't make sense as a docstring (like "Your session will expire in 5 minutes. Save your work?")

README.md
```diff
@@ -310,6 +312,44 @@ async def long_task(files: list[str], ctx: Context) -> str:
...
+    """Book a table with confirmation"""
+    
+    # Schema must only contain primitive types (str, int, float, bool)
+    class ConfirmBooking(BaseModel):
+        confirm: bool = Field(description="Confirm booking?")
+        notes: str = Field(default="", description="Special requests")
+    
+    result = await ctx.elicit(
+        message=f"Confirm booking for {party_size} on {date}?", schema=ConfirmBooking
+    )
```
Kludex, line 404: I think it gives us a better API and user experience if `ctx.elicit(schema=SchemaT)` always return an instance of `SchemaT`, or an exception.

Kludex: My comment implies that an exception would be raised if user rejects.

ihrpr: then the problem will be how to distinguish between cancel and reject?

Kludex:
```suggestion
    try:
        result = await ctx.elicit(
            message=f"Confirm booking for {party_size} on {date}?",
            schema=ConfirmBooking
        )
    except ElicitError as exc:
        print(exc.reason)
```

README.md
```diff
@@ -310,6 +312,44 @@ async def long_task(files: list[str], ctx: Context) -> str:
...
+        return "Booking cancelled"
+    
+    # User declined or cancelled
+    return f"Booking {result.action}"
+```
+
+The `elicit()` method returns an `ElicitationResult` with:
+- `action`: "accept", "decline", or "cancel" 
+- `data`: The validated response (only when accepted)
+- `validation_error`: Any validation error message
```
Kludex, line 420: I think this should just bubble.

README.md
dsp-ant: Wouldn't this imply that we use exceptions for control flow? We should reserve exceptions for handling exceptional circumstances. I don't think decline fits into this.

I do prefer having a return value that indicates if it was accepted,declined,etc.

src/mcp/server/fastmcp/server.py
```diff
@@ -65,6 +67,21 @@
...
+
+    action: Literal["accept", "decline", "cancel"]
+    """The user's action in response to the elicitation."""
+
+    data: ElicitSchemaModelT | None = None
+    """The validated data if action is 'accept', None otherwise."""
+
+    validation_error: str | None = None
+    """Validation error message if data failed to validate."""
+
```
dsp-ant: Okay, I think I can be fine with this, but the FP person in me really doesn't like that this is not very typesafe, because `data` and `validation_error` depends on `action`.

An alternative, and something I probably prefer is using a tagged union:

```python
  from typing import Literal

  class AcceptedElicitation(BaseModel, Generic[T]):
      action: Literal["accept"]
      data: T

  class DeclinedElicitation(BaseModel):
      action: Literal["decline"]

  class CancelledElicitation(BaseModel):
      action: Literal["cancel"]

  class ValidationFailedElicitation(BaseModel):
      action: Literal["accept"]
      validation_error: str

  ElicitationResult = AcceptedElicitation[T] | DeclinedElicitation | CancelledElicitation | ValidationFailedElicitation
```

You can then do:
```python
  match result:
      case AcceptedElicitation(data=data):
          # data is guaranteed to exist and be typed
          process(data)
      case DeclinedElicitation():
          handle_decline()
      case CancelledElicitation():
          handle_cancel()
      case ValidationFailedElicitation(validation_error=error):
          handle_validation_error(error)
```

src/mcp/server/fastmcp/server.py
```diff
@@ -954,6 +1008,54 @@ async def read_resource(self, uri: str | AnyUrl) -> Iterable[ReadResourceContent
...
+        if result.action == "accept" and result.content:
+            # Validate and parse the content using the schema
+            try:
+                validated_data = schema.model_validate(result.content)
+                return ElicitationResult(action="accept", data=validated_data)
+            except ValidationError as e:
+                return ElicitationResult(action="accept", validation_error=str(e))
+        else:
+            return ElicitationResult(action=result.action)
+
```
dsp-ant: This should not be defined here. This should be in the `ServerSession` and be called from `request_context.session`.

The SDK is layered in such a way that FastMCP should always only contain convenience wrapping server/lowlevel or server/session, but never implement functionality itself.

src/mcp/server/fastmcp/server.py
ihrpr: oh, this is great!

Kludex: I don't think we want to have `validation_error` as a `str`, it should bubble up. User may want to handle it differently.

PR #632, opened 2025-05-05
Simplified client progress_callback support

davemssavage:
Add support for an optional callback to be called during tool execution to inform of progress

## Motivation and Context
Tracking progress on the client is useful for long running tasks, the low level primitives are available via the protocol however they're a little complex to setup. This patch adds an optional calllback to to the call_tool request and does the heavy lifting for the caller so they don't have to do this.

The soluiton is implemented on the BaseSession object so the approach can be used for both client->server and server->client requests. Currently the only top level method that supports this is clientsession.call_tool which seems like the most common use case. If there are other are other use cases they can be added easilly.

## How Has This Been Tested?
Unit testing only

## Breaking Changes
None expected

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
If this looks useful I can add documentation to the patch

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -250,28 +250,30 @@ class BaseSession(
         send_request = None
 
         if progress_callback is not None:
-            if request.root.params is not None:
+            if request.root.params is None:
+                progress_id = self._progress_id
+                new_params = RequestParams(
+                    _meta=RequestParams.Meta(progressToken=progress_id)
+                )
+            else:
                 if (
                     request.root.params.meta is None
                     or request.root.params.meta.progressToken is None
                 ):
                     progress_id = self._progress_id
-                    self._progress_id = progress_id + 1
                     new_params = request.root.params.model_copy(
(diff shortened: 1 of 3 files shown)


davemssavage:
I note that there is a TODO in BaseSession.send_request to support progress call backs https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/shared/session.py#L242. I had previously thought of progress as a client to server feature, I'm now thinking it's useful in both directions. If this hunch is correct I can refactor to push this behaviour down to the base session.

src/mcp/shared/session.py
```diff
@@ -214,19 +229,44 @@ async def send_request(
...
         instead.
         """
 
         request_id = self._request_id
         self._request_id = request_id + 1
 
+        progress_id = None
+        send_request = None
+
+        if progress_callback is not None:
```
davemssavage, line 252: Nasty if/else branching, ugly but functional main challenge is how to inject meta into params when the type of the request is unknown. Open to ideas on how to make this cleaner

davemssavage:
Looks like someone else already implemented this closing

PR #637, opened 2025-05-06
Fix the issue of get Authorization header fails during bearer auth

yabea:
The bearer auth retrieves the Authorization from the headers, but the HTTPConnection.headers uses lowercase authorization, resulting in authentication failure.

### DIFF
--- a/src/mcp/server/auth/middleware/bearer_auth.py
+++ b/src/mcp/server/auth/middleware/bearer_auth.py
@@ -34,7 +34,9 @@ class BearerAuthBackend(AuthenticationBackend):
         self.provider = provider
 
     async def authenticate(self, conn: HTTPConnection):
-        auth_header = conn.headers.get("Authorization") or conn.headers.get("authorization")
+        auth_header = conn.headers.get("Authorization") or conn.headers.get(
+            "authorization"
+        )
         if not auth_header or not auth_header.startswith("Bearer "):
             return None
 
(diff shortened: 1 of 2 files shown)


src/mcp/server/auth/middleware/bearer_auth.py
```diff
@@ -34,7 +34,9 @@ def __init__(
         self.provider = provider
 
     async def authenticate(self, conn: HTTPConnection):
-        auth_header = conn.headers.get("Authorization")
+        auth_header = conn.headers.get("Authorization") or conn.headers.get(
+            "authorization"
+        )
```
ihrpr:
```suggestion
        auth_header = next(
        (conn.headers.get(key) for key in conn.headers if key.lower() == "authorization"),
        None
    )
```

src/mcp/server/auth/middleware/bearer_auth.py
```diff
@@ -34,7 +34,9 @@ def __init__(
...
         if not auth_header or not auth_header.startswith("Bearer "):
```
ihrpr: would `Bearer ` have the same issue?

ihrpr (changes requested):
Thanks for addressing header case sensitivity, a few suggestions:

- improve the check for case insensitive handling 
- the "Bearer" prefix should also be checked case-insensitively
- add tests to verify the fix and prevent regression

ihrpr (approved):
thank you

Issue #9, opened 2024-10-03
`Host` abstraction for multiple client → server connections in one process


jspahrsummers:
Multiplexing has recurred for me a few times now—clients often need to connect to multiple servers simultaneously. Might be worth building something for it into the client SDK.

I think it's not too hard (send all list commands to all servers, then associate the IDs you get back with the server for follow-up gets, reads, calls), it's just annoying boilerplate.

Maybe we build an actual `Host` abstraction into the SDK, which has multiple `Client`s?

See also https://github.com/modelcontextprotocol/typescript-sdk/issues/6.

felixweinberger:
Looks like this was addressed by https://github.com/modelcontextprotocol/python-sdk/pull/639 (though not specifically with the `Host` abstraction but by having `ClientSessionGroup`)

PR #639, opened 2025-05-06
Create `ClientSessionGroup` for managing multiple session connections.

mkeid:
This abstraction concurrently manages multiple MCP session connections.

Tools, resources, and prompts are aggregated across servers. Servers may be connected to or disconnected from at any point after initialization.

This abstractions can handle naming collisions using a custom user-provided hook.

## Motivation and Context

This change alleviates the need for users to re-implement a multi-session "manager" anytime more than 1 session needs to be established.

## Testing

Code branches are covered via pytest unit tests.

## Breaking Changes

This is a no-op as it introduces a new class without modifying existing dependencies.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/session_group.py
+++ b/src/mcp/client/session_group.py
@@ -1,0 +1,207 @@
+"""
+SessionGroup concurrently manages multiple MCP session connections.
+
+Tools, resources, and prompts are aggregated across servers. Servers may
+be connected to or disconnected from at any point after initialization.
+
+This abstractions can handle naming collisions using a custom user-provided
+hook.
+"""
+
+import contextlib
+from collections.abc import Callable
+from typing import Any, TypeAlias
+
+from pydantic import BaseModel
+
+import mcp
+from mcp import types
+from mcp.client.stdio import StdioServerParameters
+from mcp.shared.exceptions import McpError
+
+
+class ClientSessionGroup:
+    """Client for managing connections to multiple MCP servers.
+
+    This class is responsible for encapsulating management of server connections.
+    It it aggregates tools, resources, and prompts from all connected servers.
+
+    For auxiliary handlers, such as resource subscription, this is delegated to
+    the client and can be accessed via the session. For example:
+      mcp_session_group.get_session("server_name").subscribe_to_resource(...)
+    """
+
+    class _ComponentNames(BaseModel):
+        """Used for reverse index to find components."""
+
+        prompts: set[str] = set()
+        resources: set[str] = set()
+        tools: set[str] = set()
+
+    # Standard MCP components.
+    _prompts: dict[str, types.Prompt]
+    _resources: dict[str, types.Resource]
+    _tools: dict[str, types.Tool]
+
+    # Client-server connection management.
+    _sessions: dict[mcp.ClientSession, _ComponentNames]
+    _tool_to_session: dict[str, mcp.ClientSession]
+    _exit_stack: contextlib.AsyncExitStack
+
+    # Optional fn consuming (component_name, serverInfo) for custom names.
+    # This is provide a means to mitigate naming conflicts across servers.
+    # Example: (tool_name, serverInfo) => "{result.serverInfo.name}.{tool_name}"
+    _ComponentNameHook: TypeAlias = Callable[[str, types.Implementation], str]
+    _component_name_hook: _ComponentNameHook | None
+
+    def __init__(
+        self,
+        exit_stack: contextlib.AsyncExitStack = contextlib.AsyncExitStack(),
+        component_name_hook: _ComponentNameHook | None = None,
+    ) -> None:
+        """Initializes the MCP client."""
+
+        self._tools = {}
+        self._resources = {}
+        self._prompts = {}
+
+        self._sessions = {}
+        self._tool_to_session = {}
+        self._exit_stack = exit_stack
+        self._component_name_hook = component_name_hook
+
+    @property
+    def prompts(self) -> dict[str, types.Prompt]:
+        """Returns the prompts as a dictionary of names to prompts."""
+        return self._prompts
+
+    @property
+    def resources(self) -> dict[str, types.Resource]:
+        """Returns the resources as a dictionary of names to resources."""
+        return self._resources
+
+    @property
+    def tools(self) -> dict[str, types.Tool]:
+        """Returns the tools as a dictionary of names to tools."""
+        return self._tools
+
+    async def call_tool(self, name: str, args: dict[str, Any]) -> types.CallToolResult:
+        """Executes a tool given its name and arguments."""
+        session = self._tool_to_session[name]
+        return await session.call_tool(name, args)
+
+    def disconnect_from_server(self, session: mcp.ClientSession) -> None:
+        """Disconnects from a single MCP server."""
+
+        if session not in self._sessions:
+            raise McpError(
+                types.ErrorData(
+                    code=types.INVALID_PARAMS,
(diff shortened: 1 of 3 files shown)


src/mcp/client/session_group.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Remove tools associated with the session.
+        for name in component_names.tools:
+            del self._tools[name]
+
+        del self._sessions[session]
+
+    async def connect_to_server(
+        self,
+        server_params: StdioServerParameters,
```
ihrpr: We need to make sure that we cover all of the transports. Right now we support three:
- stdio
- sse 
- StreamableHttp (supersedes sse)

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,207 @@
...
+        self._tools.update(tools_temp)
+        self._tool_to_session.update(tool_to_session_temp)
+
+        return session
+
+    async def _establish_session(
+        self, server_params: StdioServerParameters
+    ) -> tuple[types.Implementation, mcp.ClientSession]:
+        """Establish a client session to an MCP server."""
+        client = mcp.stdio_client(server_params)
```
ihrpr: same here, we need to support all transports

src/mcp/client/session_group.py
mkeid: Added.

One additional change was to introduce respective server parameter types to align with the existing stdio implementation (i.e., `SseServerParameters` and `StreamableHttpParameters`). These should probably not exist here long-term and instead live with their corresponding client implementations.

I avoided doing that in this change because the existing signatures for `sse_client` and `streamablehttp_client` require multiple kwargs, which is incontrast to `stdio_client` which consumes `StdioServerParameters`. It would require introducing overloaded methods if want to pursue a non-breaking refactor.

ihrpr (changes requested):
thank you so much for the quick turnaround!

there a few things we need to add:
- ClientSessionGroup to implement `__aenter__/__aexit__`
- Ability to use `async with ClientSessionGroup():`
- disconnect_from_server removes sessions from tracking but doesn't remove them from _exit_stack
- add documentation on how to use

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+ServerParameters: TypeAlias = (
+    StdioServerParameters | SseServerParameters | StreamableHttpParameters
+)
+
+
+class ClientSessionGroup:
+    """Client for managing connections to multiple MCP servers.
+
+    This class is responsible for encapsulating management of server connections.
+    It it aggregates tools, resources, and prompts from all connected servers.
```
ihrpr:
```suggestion
    It aggregates tools, resources, and prompts from all connected servers.
```

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+
+    For auxiliary handlers, such as resource subscription, this is delegated to
+    the client and can be accessed via the session. For example:
+      mcp_session_group.get_session("server_name").subscribe_to_resource(...)
```
ihrpr:
```suggestion
    the client and can be accessed via the session.
```

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Example Usage:
+        name_fn = lambda name, server_info: f"{(server_info.name)}-{name}"
+        with async ClientSessionGroup(component_name_hook=name_fn) as group:
```
ihrpr:
```suggestion
        async with  ClientSessionGroup(component_name_hook=name_fn) as group:
```

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+        self, server_params: ServerParameters
+    ) -> tuple[types.Implementation, mcp.ClientSession]:
+        """Establish a client session to an MCP server."""
+
+        session_specific_stack = contextlib.AsyncExitStack()
+        try:
+            # Create read and write streams that facilitate io with the server.
+            if isinstance(server_params, StdioServerParameters):
+                client = mcp.stdio_client(server_params)
+                read, write = await self._exit_stack.enter_async_context(client)
```
ihrpr: should this be `session_specific_stack`?

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+    ) -> None:
+        """Initializes the MCP client."""
+
+        self._tools = {}
+        self._resources = {}
+        self._prompts = {}
+
+        self._sessions = {}
+        self._tool_to_session = {}
+        self._exit_stack = exit_stack or contextlib.AsyncExitStack()
```
ihrpr:
```suggestion
         if exit_stack is None:
            self._exit_stack = contextlib.AsyncExitStack()
            self._owns_exit_stack = True
        else:
            self._exit_stack = exit_stack
            self._owns_exit_stack = False
```

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+        # If ClientSessionGroup itself is managing the lifecycle of _exit_stack
+        # (i.e., it created it), it should enter it.
+        # If _exit_stack was passed in, it's assumed the caller manages
+        # its entry/exit.
+        # For simplicity and consistency with how AsyncExitStack is often used when
+        # provided as a dependency, we might not need to enter it here if it's
+        # managed externally. However, if this class is the primary owner, entering it
+        # ensures its 'aclose' is called even if passed in. Let's assume the
+        # passed-in stack is already entered by the caller if needed.
+        # For now, we just return self as the main stack's lifecycle is tied to aclose.
```
ihrpr:
```suggestion
        # Enter the exit stack only if we created it ourselves
        if self._owns_exit_stack:
            await self._exit_stack.__aenter__()
```

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,362 @@
...
+        self,
+        _exc_type: type[BaseException] | None,
+        _exc_val: BaseException | None,
+        _exc_tb: TracebackType | None,
+    ) -> bool | None:
+        """Closes session exit stacks and main exit stack upon completion."""
+        async with anyio.create_task_group() as tg:
+            for exit_stack in self._session_exit_stacks.values():
+                tg.start_soon(exit_stack.aclose)
+        await self._exit_stack.aclose()
```
ihrpr:
```suggestion
        # Only close the main exit stack if we created it
        if self._owns_exit_stack:
            await self._exit_stack.aclose()
```

ihrpr (changes requested):
Some small changes and we are good to merge! Thank you

src/mcp/client/session_group.py
```diff
@@ -0,0 +1,372 @@
...
+    """Client for managing connections to multiple MCP servers.
+
+    This class is responsible for encapsulating management of server connections.
+    It aggregates tools, resources, and prompts from all connected servers.
+
+    For auxiliary handlers, such as resource subscription, this is delegated to
+    the client and can be accessed via the session.
+
+    Example Usage:
+        name_fn = lambda name, server_info: f"{(server_info.name)}-{name}"
```
amirh, line 80: Is it possible that both a tool name and a server_info name would conflict? would there be enough information for the hook to disambiguate?

ihrpr:
related to https://github.com/modelcontextprotocol/python-sdk/issues/9

ihrpr (approved):
Thank you!

PR #641, opened 2025-05-06
Streamable HTTP - improve usability, fast mcp and auth

ihrpr:
- introducing `StreamableHTTPSessionManager`
- adding streamable http to fastmcp
- adding auth 

[on top of auth inspector changes + some changes to handle shttp]
image

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -47,6 +47,8 @@ from mcp.server.lowlevel.server import lifespan as default_lifespan
 from mcp.server.session import ServerSession, ServerSessionT
 from mcp.server.sse import SseServerTransport
 from mcp.server.stdio import stdio_server
+from mcp.server.streamable_http import EventStore
+from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from mcp.shared.context import LifespanContextT, RequestContext
 from mcp.types import (
     AnyFunction,
@@ -89,6 +91,13 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     port: int = 8000
     sse_path: str = "/sse"
     message_path: str = "/messages/"
+    streamable_http_path: str = "/mcp"
+
+    # StreamableHTTP settings
+    json_response: bool = False
+    stateless_http: bool = (
+        False  # If True, uses true stateless mode (new transport per request)
+    )
 
     # resource settings
     warn_on_duplicate_resources: bool = True
@@ -130,6 +139,7 @@ class FastMCP:
         instructions: str | None = None,
         auth_server_provider: OAuthAuthorizationServerProvider[Any, Any, Any]
         | None = None,
+        event_store: EventStore | None = None,
         **settings: Any,
     ):
         self.settings = Settings(**settings)
@@ -161,6 +171,7 @@ class FastMCP:
                 "is specified"
             )
         self._auth_server_provider = auth_server_provider
+        self._event_store = event_store
         self._custom_starlette_routes: list[Route] = []
         self.dependencies = self.settings.dependencies
 
@@ -178,20 +189,24 @@ class FastMCP:
     def instructions(self) -> str | None:
         return self._mcp_server.instructions
 
-    def run(self, transport: Literal["stdio", "sse"] = "stdio") -> None:
+    def run(
+        self, transport: Literal["stdio", "sse", "streamable-http"] = "stdio"
+    ) -> None:
         """Run the FastMCP server. Note this is a synchronous function.
 
         Args:
-            transport: Transport protocol to use ("stdio" or "sse")
+            transport: Transport protocol to use ("stdio", "sse", or "streamable-http")
         """
-        TRANSPORTS = Literal["stdio", "sse"]
+        TRANSPORTS = Literal["stdio", "sse", "streamable-http"]
         if transport not in TRANSPORTS.__args__:  # type: ignore
             raise ValueError(f"Unknown transport: {transport}")
 
         if transport == "stdio":
             anyio.run(self.run_stdio_async)
-        else:  # transport == "sse"
+        elif transport == "sse":
             anyio.run(self.run_sse_async)
+        else:  # transport == "streamable_http"
+            anyio.run(self.run_streamable_http_async)
 
     def _setup_handlers(self) -> None:
         """Set up core MCP protocol handlers."""
@@ -567,6 +582,21 @@ class FastMCP:
         server = uvicorn.Server(config)
         await server.serve()
 
+    async def run_streamable_http_async(self) -> None:
+        """Run the server using StreamableHTTP transport."""
+        import uvicorn
+
+        starlette_app = self.streamable_http_app()
+
+        config = uvicorn.Config(
+            starlette_app,
+            host=self.settings.host,
+            port=self.settings.port,
+            log_level=self.settings.log_level.lower(),
+        )
+        server = uvicorn.Server(config)
+        await server.serve()
+
     def sse_app(self) -> Starlette:
         """Return an instance of the SSE server app."""
         from starlette.middleware import Middleware
@@ -644,9 +674,9 @@ class FastMCP:
         else:
             # Auth is disabled, no need for RequireAuthMiddleware
             # Since handle_sse is an ASGI app, we need to create a compatible endpoint
-            async def sse_endpoint(request: Request) -> None:
+            async def sse_endpoint(request: Request) -> Response:
                 # Convert the Starlette request to ASGI parameters
-                await handle_sse(request.scope, request.receive, request._send)  # type: ignore[reportPrivateUsage]
+                return await handle_sse(request.scope, request.receive, request._send)  # type: ignore[reportPrivateUsage]
 
             routes.append(
                 Route(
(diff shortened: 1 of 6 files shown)


src/mcp/server/fastmcp/server.py
```diff
@@ -669,6 +699,84 @@ async def sse_endpoint(request: Request) -> None:
...
+                    provider=self._auth_server_provider,
+                    issuer_url=self.settings.auth.issuer_url,
+                    service_documentation_url=self.settings.auth.service_documentation_url,
+                    client_registration_options=self.settings.auth.client_registration_options,
+                    revocation_options=self.settings.auth.revocation_options,
+                )
+            )
+
+        # Add the StreamableHTTP endpoint
+        if self._auth_server_provider:
```
pantanurag555: Question: Can't we club this with the condition check on line 727?

src/mcp/server/fastmcp/server.py
```diff
@@ -178,20 +189,24 @@ def name(self) -> str:
...
         """
-        TRANSPORTS = Literal["stdio", "sse"]
+        TRANSPORTS = Literal["stdio", "sse", "streamable-http"]
         if transport not in TRANSPORTS.__args__:  # type: ignore
             raise ValueError(f"Unknown transport: {transport}")
 
         if transport == "stdio":
             anyio.run(self.run_stdio_async)
-        else:  # transport == "sse"
+        elif transport == "sse":
```
bhosmer-ant: nit (and also totally a matter of preference!): could do a `match transport:` for this if/elif block

src/mcp/server/fastmcp/server.py
bhosmer-ant: fwiw (not the author) I kind of like having the streamable http endpoint addition in its own block like this for readability/separation of concerns-ish reasons, but true that it adds a redundant test

src/mcp/server/streamable_http_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        Dispatches to the appropriate handler based on stateless mode.
+
+        Args:
+            scope: ASGI scope
+            receive: ASGI receive function
+            send: ASGI send function
+        """
+        if self._task_group is None:
+            raise RuntimeError(
+                "Task group is not initialized. Make sure to use the lifespan_hook."
```
bhosmer-ant: Think maybe this message needs to be something about using `run()` as a context manager?

src/mcp/server/streamable_http_manager.py
```diff
@@ -0,0 +1,241 @@
...
+                    read_stream,
+                    write_stream,
+                    self.app.create_initialization_options(),
+                    stateless=True,
+                )
+
+                # Assert task group is not None for type checking
+                assert self._task_group is not None
+                # Start the server task
+                await self._task_group.start(run_stateless_server)
```
bhosmer-ant: ok it's late and maybe I should stop now, but... this is inside `run_stateless_server`, is that intended? (the analogous code in `_handle_stateful_request` looks more like what I was expecting)

bhosmer-ant (dismissed):
Super nice fastmcp glowup :) A couple comments inline and I wouldn't mind taking another pass in the morning if you're still working on it, but approving to unblock in the meantime

src/mcp/server/fastmcp/server.py
```diff
@@ -178,20 +189,24 @@ def name(self) -> str:
...
             anyio.run(self.run_sse_async)
+        else:  # transport == "streamable_http"
+            anyio.run(self.run_streamable_http_async)
```
000-000-000-000-000: nitpick: it might be better to have another elif here and then throw an exception in the else block. If someone introduces a new transport but misses updating this code they might believe their code is working even though it is really using a different transport.

src/mcp/server/fastmcp/server.py
```diff
@@ -669,6 +699,84 @@ async def sse_endpoint(request: Request) -> None:
             debug=self.settings.debug, routes=routes, middleware=middleware
         )
 
+    def streamable_http_app(self) -> Starlette:
+        """Return an instance of the StreamableHTTP server app."""
+        from starlette.middleware import Middleware
+        from starlette.routing import Mount
```
000-000-000-000-000, line 767: question: what's the intention behind importing these things inside this method? Is it to not force dev to have these dependencies if they are not using this method?

src/mcp/server/streamable_http_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        Use this in the lifespan context manager of your Starlette app:
+
+        @contextlib.asynccontextmanager
+        async def lifespan(app: Starlette) -> AsyncIterator[None]:
+            async with session_manager.run():
+                yield
+        """
+        async with anyio.create_task_group() as tg:
+            # Store the task group for later use
+            self._task_group = tg
```
000-000-000-000-000, line 108: nitpick: Does this mean that StreamableHTTPSessionManager is not thread safe (two threads using same manager instance will overwrite eachother's self._task_group)? Might be worth documenting.

src/mcp/server/fastmcp/server.py
ihrpr: yeah, fixed

src/mcp/server/streamable_http_manager.py
ihrpr: ah, yeah, thank you! struggled with names...

src/mcp/server/streamable_http_manager.py
ihrpr: intendation is wrong... let me add a test for stateless, must have messed up while cleaning up

ihrpr: added a test

src/mcp/server/fastmcp/server.py
```diff
@@ -669,6 +700,80 @@ async def sse_endpoint(request: Request) -> None:
             debug=self.settings.debug, routes=routes, middleware=middleware
         )
 
+    def streamable_http_app(self) -> Starlette:
```
baxen, line 764: I'm not sure if you've thought about the use case, but we're hoping to be able to use one FastAPI app with multiple routes that all have separate stateless MCP servers. 

On your branch this works:

```python
 """Example Echo MCP Server."""

 from fastapi import APIRouter

 mcp = FastMCP(name="EchoServer")


 @mcp.tool(description="A simple echo tool")
 def echo(message: str) -> str:
     return f"Echo: {message}"


 # Get the Starlette app and convert to router
 app: Starlette = mcp.streamable_http_app()
```

with a parent app like 
```python
from routers import echo

app = FastAPI(lifespan=echo.app.router.lifespan_context)
app.mount("/echo", echo.app)
```

Although that involves kind of digging into your implementation to know that the lifespan function sets up the session manager and that it ends up internally in the starlette router. It might be helpful to have this exposed on the FastMCP object to make that more clear/explicit? either way it'll take a bit of documentation

```python
from routers import echo

app = FastAPI(lifespan=lambda app: echo.mcp.session_manager.run())
app.mount("/echo", echo.mcp.streamable_http_app())
```

ihrpr: @baxen, thank you, yeah, haven't considered that use case, would definitely add documentation! 
Do you have any suggestions on how to make it more clear/explicit?

baxen: i haven't thought of anything better than having the FastMCP object having the StreamableHTTPSessionManager as a member. The error it already raises when run hasn't been called is a good hint, and as a member on FastMCP it'll be easier to find it and call it's run method (like in that last snippet)

src/mcp/server/fastmcp/server.py
```diff
@@ -181,23 +192,26 @@ def instructions(self) -> str | None:
...
-        TRANSPORTS = Literal["stdio", "sse"]
+        TRANSPORTS = Literal["stdio", "sse", "streamable-http"]
         if transport not in TRANSPORTS.__args__:  # type: ignore
             raise ValueError(f"Unknown transport: {transport}")
 
-        if transport == "stdio":
-            anyio.run(self.run_stdio_async)
-        else:  # transport == "sse"
-            anyio.run(lambda: self.run_sse_async(mount_path))
+        match transport:
```
000-000-000-000-000, line 228: suggestion: i would still implement the default match at the end since it is possible that someone updates TRANSPORTS but forgets to update this method.

ihrpr: pyright will detect that case, and the error will look something like:

``` Unhandled type: "Literal['new-thing']"
    If exhaustive handling is not intended, add "case _: pass"
```

src/mcp/server/fastmcp/server.py
ihrpr: yes, that's right, if one only uses run(transport="stdio"), they never need those dependencies

src/mcp/server/fastmcp/server.py
ihrpr: right, not super happy to add session_manager but also don't have a better option. Added to fast mcp, also adding to README this example: https://github.com/modelcontextprotocol/python-sdk/pull/664 

thank you @baxen, let me know if this works!

src/mcp/server/streamable_http_manager.py
```diff
@@ -0,0 +1,239 @@
...
+        Dispatches to the appropriate handler based on stateless mode.
+
+        Args:
+            scope: ASGI scope
+            receive: ASGI receive function
+            send: ASGI send function
+        """
+        if self._task_group is None:
+            raise RuntimeError(
+                "Task group is not initialized. Make sure to use the run()."
```
jerome3o-anthropic:
```suggestion
                "Task group is not initialized. Make sure to use run()."
```

src/mcp/server/streamable_http_manager.py
jerome3o-anthropic: +1 to this

jerome3o-anthropic (approved):
Looks good to me!

one concern around the `.run()` function being called on the same instance across multiple tasks, happy for you to do another loop on that (either document, fix, accept as is)

src/mcp/server/streamable_http_manager.py
bhosmer-ant: same, maybe not worth a fix but def worth a note in the comment header

bhosmer-ant (dismissed):
thread safety note on  is prob worth it but new changes LGTM!

jerome3o-anthropic (approved):

src/mcp/server/streamable_http_manager.py
ihrpr: updated, making only once instance being able to use run + docs

src/mcp/server/fastmcp/server.py
000-000-000-000-000: got it -- makes sense and is pretty clean since this method is only invoked during server startup as far as i understand. we should just be careful that people dont start doing this in methods that are called after server startup since then import errors will only become visible in the runtime.

src/mcp/server/streamable_http_manager.py
```diff
@@ -0,0 +1,258 @@
...
+                # Start the server task
+                await self._task_group.start(run_server)
+
+                # Handle the HTTP request and return the response
+                await http_transport.handle_request(scope, receive, send)
+        else:
+            # Invalid session ID
+            response = Response(
+                "Bad Request: No valid session ID provided",
+                status_code=HTTPStatus.BAD_REQUEST,
```
Zalewa, line 256: Shouldn't the status code be `404 NOT_FOUND`? 

See point 2.5.3 here:
https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management

> The server MAY terminate the session at any time, after which it MUST respond to requests containing that session ID with HTTP 404 Not Found.

PR #648, opened 2025-05-07
confused deputy fix in example + supporting sdk changes

pcarleton:
## Motivation and Context

See the confused deputy consideration here:
https://github.com/modelcontextprotocol/modelcontextprotocol/blob/c255349bd98d9dc73f3579c828d195bd09c0bee8/docs/specification/draft/basic/security_best_practices.mdx#21-confused-deputy-problem

In the example, we front the Github API, so we need to put a consent screen for each new client registration, otherwise the Github consent screen will be skipped for new clients, creating a path for token leakage.

## Types of changes

- [ ] Documentation update

## Checklist


src/mcp/server/auth/handlers/authorize.py
```diff
@@ -218,6 +219,26 @@ async def error_response(
             )
 
             try:
+                # Check if client has already consented
+                has_consent = await self.provider.has_client_consent(client)
```
pcarleton: todo: check `client_consent_required` before doing this

src/mcp/server/auth/provider.py
```diff
@@ -174,6 +174,27 @@ async def authorize(
         """
         ...
 
+    async def has_client_consent(self, client: OAuthClientInformationFull) -> bool:
```
pcarleton: todo: if client_consent_required, these need to be required, but otherwise they should be optional.  open to suggestions here. i think a default implementation of `return True` means we'd solve for being optional, but not for required.

felixweinberger (changes requested):
Hey @pcarleton just checking whether we're still planning to ship this? Marking it as "needs publish" for now so we can easily filter it from views.

I also assume we might still need some merge conflict resolution here.

PR #649, opened 2025-05-07
Revert "Add message queue for SSE messages POST endpoint (#459)"

ihrpr:
- In spec version 2025-03-26, Streamable HTTP transport supersedes the HTTP+SSE transport from version 2024-11-05
- We're bringing the Python SDK to the new spec version (2025-03-26), which means focusing on Streamable HTTP rather than enhancing the legacy transport, unless it's a fix
- Adding Redis integration to SSE creates maintenance burden for a transport being superseded. It would be still useful to have an MCP Server with Redis but as an example so users can reference it.


bhosmer-ant (approved):

akash329d:
> 👍
> 
> I really didn't understand why merging the redis stuff.

I saw that SSE was deprecated but had still put some time into it so wanted to get it merged 😅 . It seems like this potentially could be useful for stateful streamable HTTP still, so maybe it can get revived eventually.

PR #655, opened 2025-05-07
Introduce a function to create a standard AsyncClient with options

ihrpr:
Request as a follow up for a few PRs now, for example https://github.com/modelcontextprotocol/python-sdk/pull/284 and Streamable Http PRs

  Created `create_mcp_http_client()` that:
  - Enforces `follow_redirects=True` by default (consistent with existing usage)


src/mcp/shared/httpx_utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import Any
+
+import httpx
+
+__all__ = ["create_mcp_http_client"]
+
+
+def create_mcp_http_client(
+    headers: dict[str, Any] | None = None,
```
Kludex:
```suggestion
    headers: dict[str, str] | None = None,
```

ref.: https://github.com/encode/httpx/blob/6c7af967734bafd011164f2a1653abc87905a62b/httpx/_models.py#L139

Kludex:
```suggestion
    headers: dict[str, str] | None = None,
```

ref.: https://github.com/encode/httpx/blob/6c7af967734bafd011164f2a1653abc87905a62b/httpx/_models.py#L139

src/mcp/shared/httpx_utils.py
```diff
@@ -0,0 +1,64 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+import httpx
+
+__all__ = ["create_mcp_http_client"]
+
+
+def create_mcp_http_client(
```
Kludex: I don't think this repository follow private/public APIs properly, but maybe it should start...

```suggestion
def _create_mcp_http_client(
```

ihrpr: We need to use `create_mcp_http_client` across different client transport implementations, so making it private would cause Pyright warnings about accessing private members from other modules.

If we want to indicate that the utils module contains internal implementation details, we could consider renaming the file to something like `_internal_httpx_utils.py`. This would signal that the module itself is internal while keeping the functions public for cross-module usage.

However, given that this function is in shared/ and genuinely needs to be shared across modules, keeping it public seems more appropriate. Happy to change it if you prefer the internal module approach though or I'm missing something very obvious.

Kludex: > We need to use create_mcp_http_client across different client transport implementations, so making it private would cause Pyright warnings about accessing private members from other modules.

~Sorry, I meant making the module private `_httpx_utils.py`, and keeping `create_mcp_http_client`.~

> If we want to indicate that the utils module contains internal implementation details, we could consider renaming the file to something like _internal_httpx_utils.py. This would signal that the module itself is internal while keeping the functions public for cross-module usage.

Exactly.

Kludex: I'm not strong, since the rest here doesn't follow it.

ihrpr: cool, thank you! Applied the change. Agree, we need to start somewhere

jerome3o-anthropic (approved):

PR #659, opened 2025-05-08
Properly infer prefix for SSE messages

jlowin:
I've been following the issue in #386 and see the solution merged in #540. However, it feels very wrong. An app should never have to know its mount path in advance in order to work properly. Furthermore, this results in a cumbersome DX where *either* users specify the mount paths when creating the FastMCP server, effectively preventing it from being used in any other context *or* users double-specify mount paths when mounting the SSE app into another Starlette application, which is confusing and awkward.

With that in mind, consider this example from the docs of #540. The GitHub and browser apps will simply break if they are mounted under any other path. This is not a good user-facing outcome.

```python
# Create Starlette app with multiple mounted servers
app = Starlette(
    routes=[
        # Using settings-based configuration
        Mount("/github", app=github_mcp.sse_app()),
        Mount("/browser", app=browser_mcp.sse_app()),
        # Using direct mount path parameter
        Mount("/curl", app=curl_mcp.sse_app("/curl")),
        Mount("/search", app=search_mcp.sse_app("/search")),
    ]
)
```

Happily I don't think we need to settle, or pollute the FastMCP server with new user-facing configuration. The ASGI spec provides for pretty much this exact circumstance with a `root_path` attribute on the `Scope` object. By examining the root path when constructing the messages POST endpoint, we can ensure that it is always generated at the correct absolute path, without having to perform any complex parsing or even asking users to provide additional configuration. 

With this PR, the following works out of the box:

```python
import uvicorn

mcp = FastMCP()


@mcp.tool()
async def hello() -> str:
    return "Hello, world!"


sse_app = mcp.sse_app()
main_app = Starlette(routes=[Mount("/nested/mount/", app=sse_app)])

if __name__ == "__main__":
    uvicorn.run(main_app, host="127.0.0.1", port=8000)
```

### DIFF
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -252,3 +252,69 @@ async def test_sse_client_timeout(
         return
 
     pytest.fail("the client should have timed out and returned an error already")
+
+
+def run_mounted_server(server_port: int) -> None:
+    app = make_server_app()
+    main_app = Starlette(routes=[Mount("/mounted_app", app=app)])
+    server = uvicorn.Server(
+        config=uvicorn.Config(
+            app=main_app, host="127.0.0.1", port=server_port, log_level="error"
+        )
+    )
+    print(f"starting server on {server_port}")
+    server.run()
+
+    # Give server time to start
+    while not server.started:
+        print("waiting for server to start")
+        time.sleep(0.5)
+
+
+@pytest.fixture()
+def mounted_server(server_port: int) -> Generator[None, None, None]:
+    proc = multiprocessing.Process(
+        target=run_mounted_server, kwargs={"server_port": server_port}, daemon=True
+    )
+    print("starting process")
+    proc.start()
+
+    # Wait for server to be running
+    max_attempts = 20
+    attempt = 0
+    print("waiting for server to start")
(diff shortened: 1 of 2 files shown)


tim-watcha:
@ihrpr It seems this approach is better than my approach #540 how do you think

tests/shared/test_sse.py
```diff
@@ -252,3 +252,69 @@ async def test_sse_client_timeout(
...
+
+    # Give server time to start
+    while not server.started:
+        print("waiting for server to start")
+        time.sleep(0.5)
+
+
+@pytest.fixture()
+def mounted_server(server_port: int) -> Generator[None, None, None]:
+    proc = multiprocessing.Process(
```
ihrpr, line 276: note for the future, need to extract it into a function...

ihrpr (approved):
thank you!

dwayn:
Thank you for doing this! This is exactly the fix that was needed. 😄

mindcrime:
Does this fix also work for the Vanilla Python SDK case, where one is using SSEServerTransport directly (eg, no FastMCP)? If so, could someone be kind enough to share an example of how to configure things in that case?

Issue #631, opened 2025-05-05
How to use progress token for the tool call?


xxzero12:
See the document show below
https://gofastmcp.com/servers/context#progress-reporting

And on MCP Server need to use context.report_progress
```python
Method signature:

ctx.report_progress(progress: float, total: float | None = None)
progress: Current progress value (e.g., 24)
total: Optional total value (e.g., 100). If provided, clients may interpret this as a percentage.
Progress reporting requires the client to have sent a progressToken in the initial request. If the client doesn’t support progress reporting, these calls will have no effect.
```

But the document doesn't mention how to do it on mcp client side

See the fastmcp implement it below
```python
    async def call_tool(
        self,
        name: str,
        arguments: dict[str, Any] | None = None,
        read_timeout_seconds: timedelta | None = None,
    ) -> types.CallToolResult:
        """Send a tools/call request."""
        return await self.send_request(
            types.ClientRequest(
                types.CallToolRequest(
                    method="tools/call",
                    params=types.CallToolRequestParams(name=name, arguments=arguments),
                )
            ),
            types.CallToolResult,
            request_read_timeout_seconds=read_timeout_seconds,
        )
```

And see the mcp type.py has the below implementation
```python
class CallToolRequestParams(RequestParams):
    """Parameters for calling a tool."""

    name: str
    arguments: dict[str, Any] | None = None
    model_config = ConfigDict(extra="allow")

class RequestParams(BaseModel):
    class Meta(BaseModel):
        progressToken: ProgressToken | None = None
        """
        If specified, the caller requests out-of-band progress notifications for
        this request (as represented by notifications/progress). The value of this
        parameter is an opaque token that will be attached to any subsequent
        notifications. The receiver is not obligated to provide these notifications.
        """

        model_config = ConfigDict(extra="allow")

    meta: Meta | None = Field(alias="_meta", default=None)
```

So I think added in the meta field in arguments should pass the value into it but not working
```
        await client.call_tool("get_forecast",{ "latitude": 47.6101, "longitude": -122.2015, "meta": {"progressToken": "test"} })
```

Need help on what I miss or what to do to make progress feature work

Thank you

artorias-gh:
use alias, i.e. `"_meta": {"progressToken": "test"}`

felixweinberger:
Check out the `tool_progress` example: https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/tool_progress.py

PR #669, opened 2025-05-08
feat: add partial results as part of progress notifications

cyccolin:
This PR implements the specifications defined in [modelcontextprotocol/modelcontextprotocol#383](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/383), introducing partial result streaming capabilities within progress notifications from Server to Client.
## Motivation and Context

This enhancement addresses two key improvements to the protocol:

1.  Enables servers to stream intermediate results during execution of long-running operations, as specified in [modelcontextprotocol/modelcontextprotocol#117](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/117)
2. Adds support for clients to send metadata during call_tool operations in the FastMCP implementation, as referenced in [PR #631](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/631)

### Changes
1. Add optional "meta" parameter to the ClientSession `call_tool` api to enable feature access from [spec](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414)
2. Add "partialResults" flag that allows clients to explicitly request streaming results.
3. Add "partial_result" parameters in Server SendProgressNotification to enable notification with partial_result details.
## How Has This Been Tested?

1. Add unit tests for FastMCP reportProgress API. 
2. Manually test with custom client to receive partial results from long task server

## Breaking Changes

No breaking changes as the new attributes and parameters are all optional. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -259,6 +259,7 @@ class ClientSession(
         name: str,
         arguments: dict[str, Any] | None = None,
         read_timeout_seconds: timedelta | None = None,
+        **meta
     ) -> types.CallToolResult:
         """Send a tools/call request."""
 
@@ -266,7 +267,7 @@ class ClientSession(
             types.ClientRequest(
                 types.CallToolRequest(
                     method="tools/call",
-                    params=types.CallToolRequestParams(name=name, arguments=arguments),
+                    params=types.CallToolRequestParams(name=name, arguments=arguments, _meta=meta),
                 )
             ),
             types.CallToolResult,
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -47,7 +47,12 @@ class RequestParams(BaseModel):
         parameter is an opaque token that will be attached to any subsequent
         notifications. The receiver is not obligated to provide these notifications.
         """
-
+        partialResults: bool | None = None
+        """
+        If true, the caller is requesting that results be streamed via progress notifications.
+        When this is set to true, the final response may be empty as the complete result
+        will have been delivered through progress notifications.
+        """
         model_config = ConfigDict(extra="allow")
 
     meta: Meta | None = Field(alias="_meta", default=None)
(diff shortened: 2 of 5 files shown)


src/mcp/types.py
```diff
@@ -322,6 +327,19 @@ class PingRequest(Request[RequestParams | None, Literal["ping"]]):
     method: Literal["ping"]
     params: RequestParams | None = None
 
+class PartialResult(BaseModel): 
+    chunk: dict[str, Any]
```
000-000-000-000-000, line 331: disregard below -- these points need to go on the spec PR. i will add them there.

question: 
1. there is nothing technically wrong with modeling this as a `dict[str, Any]`, but this is not how `CallToolResult` models its content. Just curious if there is a specific reason not to keep same structure as `CallToolResult`? Intuitively i would expect them to be very similar.
2. Also a nitpick on naming but should we call this chunk? why not "content" or something (similar to `CallToolResult`)

src/mcp/client/session.py
```diff
@@ -259,14 +259,15 @@ async def call_tool(
         name: str,
         arguments: dict[str, Any] | None = None,
         read_timeout_seconds: timedelta | None = None,
+        **meta
```
pantanurag555, line 262: Can we model this as an argument with Meta type rather than treating it as **kwargs? If the idea is not to put all extra keyword arguments, that don't match other call_tool arguments, in meta then we should use an argument with type Meta instead.

src/mcp/types.py
```diff
@@ -47,7 +47,12 @@ class Meta(BaseModel):
         parameter is an opaque token that will be attached to any subsequent
         notifications. The receiver is not obligated to provide these notifications.
         """
-
+        partialResults: bool | None = None
```
pantanurag555, line 50: Might be a comment for the spec PR.
Nit: I wonder if streamPartialResults or notifyPartialResults would be more apt than adding a boolean value under partialResults?

ihrpr:
Thank you for contributing to MCP Python SDK.

Closing this PR for now as the spec changes haven't been approved yet and there were changes in progress notification in SDK to include message.

Please re-open when the spec is approved or move to draft.

Issue #652, opened 2025-05-07
Support custom OauthProviders in the client (similar to TypeScript SDK)


smurching:
**Is your feature request related to a problem? Please describe.**
I have multiple MCP servers hosted behind the same authorization server. When writing custom agents or programs that connect to the MCP servers, with the APIs currently in `main` I can manually specify headers to connect to my MCP servers:

```
    token = get_oauth_token()
    async with streamablehttp_client(url=f"{basic_server_url}/mcp/", headers={'Authorization': f'Bearer {token}'}) as (
        read_stream,
        write_stream,
        _,
    ):
        async with ClientSession(
            read_stream,
            write_stream,
        ) as session:
            await session.initialize()
            yield session
```

However, this of course isn't stable over time when I'm trying to authenticate using short-lived oauth access tokens (I'll get token expiry errors after 30 min etc).

**Describe the solution you'd like**
It seems we have a nice solution to implementing custom authentication behavior in MCP clients in the TypeScript SDK, where I can pass a custom [OauthProvider](https://github.com/modelcontextprotocol/typescript-sdk/blob/621ccea997bf318ee99c7f64ce19609e838615fa/src/client/auth.ts#L13) when instantiating my client. Would it make sense to consider the same for the Python SDK?

**Describe alternatives you've considered**

**Additional context**

omarbenhamid:
Important feature, I hope it will be merged soon

ihrpr:
linking PR: https://github.com/modelcontextprotocol/python-sdk/pull/308

felixweinberger:
Closing as https://github.com/modelcontextprotocol/python-sdk/pull/308 has landed.

PR #681, opened 2025-05-11
OAuth Client Provider for Streamable HTTP Client

aravind-segu:
## Motivation and Context
Issue: https://github.com/modelcontextprotocol/python-sdk/issues/652

Introduces a OAuthProvider for Streamable HTTP Client, which allows  to authenticate to the HTTP Client through the OAuth. This follows a similar architecture to the typescript auth provider: https://github.com/modelcontextprotocol/typescript-sdk/blob/621ccea997bf318ee99c7f64ce19609e838615fa/src/client/auth.ts#L13

## How Has This Been Tested?
Unit tests, and tested with Databricks OAuth App

## Breaking Changes
streamablehttp_client() now returns a finish_auth method which needs to be called after the OAuth redirect. If the users were unpacking this as a tuple, they now can get an extra return value in the tuple. This is not a big breaking change, but might error out for some users depending on their implementation.

## Types of changes

- [ ] Documentation update

## Checklist


ihrpr:
Thank you @aravind-segu!! Sorry I just saw this one, as discussed in previous PRs, we should have support for Auth in Client now.

PR #687, opened 2025-05-12
Move `Prompt` object instantiation from server to prompt manager

ramjibc:
The `Prompt()` decorator in `server.py` is instantiating a Prompt object before adding it to the list of prompts. This is inconsistent with how the `Tool()` decorator works; it delegates tool instantiation to the tool manager. 

## Motivation and Context
To make the decorators consistent and create clear separation of concerns: I have moved prompt object instantiation from `server.py` to `prompts/manager.py`.

## How Has This Been Tested?
I updated and ran all unit tests.

## Breaking Changes
No. Users will not be impacted as the `prompt()` decorator's signature remains unchanged.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Separation of concerns: instantiation and managements of prompt objects should be done by the prompt manager, not the server.

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -333,90 +333,107 @@ class BaseSession(
             self._read_stream,
             self._write_stream,
         ):
-            async for message in self._read_stream:
-                if isinstance(message, Exception):
-                    await self._handle_incoming(message)
-                elif isinstance(message.message.root, JSONRPCRequest):
-                    try:
-                        validated_request = self._receive_request_type.model_validate(
-                            message.message.root.model_dump(by_alias=True, mode="json", exclude_none=True)
-                        )
-                        responder = RequestResponder(
-                            request_id=message.message.root.id,
-                            request_meta=validated_request.root.params.meta if validated_request.root.params else None,
-                            request=validated_request,
-                            session=self,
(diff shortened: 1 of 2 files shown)


ihrpr (changes requested):
There are a few concerns:

1. API Change for Direct Usage: The add_prompt() method signature changes from accepting a Prompt to accepting a function. This breaks backwards compatibility.
 2. Less Flexibility: With the old design, users could create custom Prompt objects with specific configurations before
  adding them. Now they're forced to use the manager's instantiation logic.
 3. Inconsistent Manager APIs: While this makes server decorators consistent, it makes the manager APIs inconsistent:
    - ToolManager.add_tool() accepts both functions and Tool objects
    - PromptManager.add_prompt() now only accepts functions

ramjibc:
Thank you for the review @ihrpr . I have addressed your comments, but primarily from the backward compatibility perspective. Also, I don't see this in code

> Inconsistent Manager APIs: While this makes server decorators consistent, it makes the manager APIs inconsistent:
ToolManager.add_tool() accepts both functions and Tool objects
PromptManager.add_prompt() now only accepts functions

I see `ToolManager.add_tool()` taking only a function as input.

In any case, I have made changes such that the code doesn't break for existing users but moves it a bit closer to achieving clear separation of concerns. Let me know your comments.

Issue #680, opened 2025-05-10
Streamable SSE hangs forever when the server attempts any callback


jlowin:
I'm debugging the issue described in https://github.com/jlowin/fastmcp/issues/393. 

Essentially, attempting to use a callback for sampling in conjunction with a Streamable HTTP connection results in an indefinite hang. 

I've successfully traced the issue to a specific moment in the low-level interaction. The server correctly sends the sampling request to the client, and the client successfully processes it via its sampling callback and sends a response to the server. However, the server never receives that response, with the result that it blocks forever waiting. 

(Note that this also happens if the client doesn't even support sampling and the default sampling callback responds with an appropriate error -- the ultimate effect being that anytime the server requests a sample it immediately locks itself up)

Unfortunately I'm not knowledgeable enough on how the server streams work to suggest a solution, but I have created a complete unit test that I think demonstrates the issue. It is based on similar low-level code in https://github.com/modelcontextprotocol/python-sdk/blob/main/tests/shared/test_streamable_http.py and contains a single unit test that can be run with `pytest`. If you add debug prints to the session classes you can see the server and client writing messages, with the last one being the client's attempt to respond to the server.

Apologies for the length of the test but wanted to be sure everything was covered and self-contained.

```python
"""
Tests for the StreamableHTTP server and client transport.

Contains tests for both server and client sides of the StreamableHTTP transport.
"""

import multiprocessing
import socket
import time
from collections.abc import Generator

import httpx
import pytest
import uvicorn
from starlette.applications import Starlette
from starlette.routing import Mount

import mcp.types as types
from mcp.client.session import ClientSession
from mcp.client.streamable_http import streamablehttp_client
from mcp.server import Server
from mcp.server.streamable_http import (
    EventStore,
)
from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
from mcp.types import (
    TextContent,
    Tool,
)

# Test constants
SERVER_NAME = "test_streamable_http_server"
TEST_SESSION_ID = "test-session-id-12345"
INIT_REQUEST = {
    "jsonrpc": "2.0",
    "method": "initialize",
    "params": {
        "clientInfo": {"name": "test-client", "version": "1.0"},
        "protocolVersion": "2025-03-26",
        "capabilities": {},
    },
    "id": "init-1",
}


class ServerTest(Server):
    def __init__(self):
        super().__init__(SERVER_NAME)

        @self.list_tools()
[...]

PR #693, opened 2025-05-12
Fix streamable http sampling

ihrpr:
Thank you @jlowin for raising https://github.com/modelcontextprotocol/python-sdk/issues/680 🙏

The StreamableHTTP client was experiencing a deadlock when the server initiated sampling callbacks during request
   processing. The issue occurred due to synchronous request handling in the client's post_writer function.

  The Deadlock Scenario:

  1. Client sends a tool call request to the server
  2. Server processes the tool and needs to perform sampling, initiating a callback to the client
  3. Server sends the sampling request to the client as a new JSONRPC request
  4. Client's post_writer receives the sampling request but handles it synchronously
  5. Client blocks waiting for the server's response to the sampling request
  6. Server is blocked waiting for the sampling response before it can complete the original tool call
  7. Deadlock: Both client and server are waiting for each other

  The circular dependency prevented the sampling response from ever reaching the server, causing the system to hang
   indefinitely.

  **The Solution**

Concurrent request handling in the StreamableHTTP client:

  1. Added TaskGroup parameter to post_writer to enable concurrent task execution
  2. Modified request handling logic to process incoming JSONRPC requests (like sampling callbacks) in separate
  tasks using tg.start_soon(), while responses continue to be handled synchronously (notification need to be handled concurrently as we will have other issues like initialisation notifications )
  3. Fixed SSE event handling to properly detect completion and break from the event loop

  The fix ensures that request handling doesn't block the message processing pipeline, allowing the client to:
  - Process server-initiated callbacks in separate tasks
  - Continue listening for new messages in the main task
  - Send callback responses without blocking
  - Receive the server's completion of the original request

**Follow ups:**
- as we are working on integration tests - all tests for all the features and transports combinations

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -310,6 +310,8 @@ class StreamableHTTPTransport:
                         else None
                     ),
                 )
+                # If the SSE event indicates completion, like returning respose/error
+                # break the loop
                 if is_complete:
                     break
         except Exception as e:
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -15,6 +15,7 @@ from typing import Any
 
 import anyio
 import httpx
+from anyio.abc import TaskGroup
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from httpx_sse import EventSource, ServerSentEvent, aconnect_sse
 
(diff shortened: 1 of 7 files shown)


jlowin:
Thanks for the quick response @ihrpr! Confirm this fixes both my test case and the original reported bug in the FastMCP repo 🎉

src/mcp/client/streamable_http.py
```diff
@@ -309,6 +310,8 @@ async def _handle_sse_response(
                         else None
                     ),
                 )
+                if is_complete:
+                    break
```
jerome3o-anthropic, line 316: why break here?

jerome3o-anthropic (approved):
Looks allgood to me - one non-blocking question around breaking/awaiting message handling in the client

src/mcp/client/streamable_http.py
ihrpr: `_handle_sse_event` returns true if there is a response/error. this means we are done and need to stop the stream

PR #695, opened 2025-05-12
Simple Auth Example to use shttp in addition to sse

ihrpr:
We need to make sure Inspector also works with Streamable Http


lkchao-majority:
Hi @ihrpr 
First of all, a big thank you for your work! Thanks to your example, I was able to create my own MCP server using a different OAuth provider.

I have a question: does the SDK actually support refresh tokens?
When I tried to manually trigger the refresh token workflow using the Inspector tool, I received a “Bad Request” error from my server.

This leads me to believe that the server might not be expecting the current request format on the same /token endpoint.

PR #699, opened 2025-05-12
fastmcp: allow passing Tool directly to FastMCP constructor

vbarda:
This change allows passing the FastMCP `Tool` instance directly via a new `fastmcp_server.add_tool_instance` method. 

This is useful if someone needs to construct the `Tool` instance manually, without inspecting a function. For example, this can be used to convert LangChain tools to FastMCP-compatible tools and add those to an existing FastMCP server.


dsp-ant (changes requested):
I am okay with the general idea of allowing the registration of a Tool instance directly. However, I personally prefer using a separate function name over an overload, since the overload leads to `add_tool` having a very funky type signature. I personally would prefer if we add a `add_tool_instance` function instead.  Would love to hear @Kludex or @ihrpr's opinion on the matter.

vbarda:
> I am okay with the general idea of allowing the registration of a Tool instance directly. However, I personally prefer using a separate function name over an overload, since the overload leads to `add_tool` having a very funky type signature. I personally would prefer if we add a `add_tool_instance` function instead. Would love to hear @Kludex or @ihrpr's opinion on the matter.

sounds good - I will add a separate method!

vbarda:
@dsp-ant @Kludex made the changes - let me know if this looks good to you!

Kludex:
It does not. I think the `tools` can be a parameter to the `FastMCP` constructor.

vbarda:
> It does not. I think the `tools` can be a parameter to the `FastMCP` constructor.

Thanks for the feedback - I agree that it's cleaner to put it in the constructor. Updated!

src/mcp/server/fastmcp/tools/tool_manager.py
```diff
@@ -19,8 +19,16 @@
 class ToolManager:
     """Manages FastMCP tools."""
 
-    def __init__(self, warn_on_duplicate_tools: bool = True):
+    def __init__(
+        self, tools: list[Tool] | None = None, warn_on_duplicate_tools: bool = True
```
eyurtsev: This is a breaking change, let's use a `*` and put it after warn_on_duplicate_tools

src/mcp/server/fastmcp/server.py
```diff
@@ -141,6 +141,7 @@ def __init__(
         auth_server_provider: OAuthAuthorizationServerProvider[Any, Any, Any]
         | None = None,
         event_store: EventStore | None = None,
+        tools: list[Tool] | None = None,
```
eyurtsev, line 145: I think we should use a `*` here as well before any named arguments

dsp-ant (approved):

PR #308, opened 2025-03-17
Add OAuth authentication client for HTTPX

dsp-ant:
## Summary
- Adds OAuth 2.1 authentication client implementation with PKCE support
- Implements HTTP authentication for the HTTPX client
- Supports dynamic client registration, token refresh, and authorization code flow
- Provides in-memory token storage implementation

PR #700, opened 2025-05-12
Dynamic Authorization in Streamable HTTP Client

aravind-segu:
## Motivation and Context
As described in https://github.com/modelcontextprotocol/python-sdk/issues/652, the streamable HTTP client currently can be configured to include a static dictionary of headers with each request, including an Authorization header for authentication. However, to support authenticating to an MCP server using short-lived access tokens, the client needs the ability to obtain a fresh token across requests to the MCP Server. This PR introduces an AuthClientProvider base class to support custom authentication and adds the ability to pass in an instance of AuthClientProvider when instantiating a StreamableHTTPClient. If specified, the headers returned from auth_client_provider.get_headers() are passed along to each request to the Server. This resembles the setup in the TypeScript SDK.

In a follow-up PR, we'll complete parity with the TypeScript SDK & enhance the interface to support Oauth 2.0 flows, so that developers need only implement key pieces of the Oauth 2.0 authorization code flow rather than a full-fledged get_headers implementation, while still preserving the ability to override get_headers to support authenticating to an MCP server from environments where an oauth token or other auth is already available

## How Has This Been Tested?
Unit Tests
For e2e test, built an example implementation of an interface against a OAuth compliant server
```
import asyncio
from datetime import datetime, timedelta
from typing import Optional

from mcp.client.streamable_http import streamablehttp_client, AuthClientProvider

class CustomAuthClientProvider(AuthClientProvider):
    def __init__(self):
        # Initialize Auth Provider

    async def get_headers(self) -> str:
        # Implement getting dynamic headers

# Auth Provider specified here
auth_client_provider = CustomAuthClientProvider()

async def query_mcp():
        url="<URL>",
        auth_client_provider=auth_client_provider,
    ) as (read_stream, write_stream, _):
            await session.initialize()
            print(tools)


async def main():
    await query_mcp()


if __name__ == "__main__":
    asyncio.run(main())
```

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist


tests/shared/test_streamable_http.py
```diff
@@ -1223,3 +1224,61 @@ async def sampling_callback(
...
+    # Create client with dynamic token provider
+    async with streamablehttp_client(
+        f"{basic_server_url}/mcp", auth_token_provider=token_provider
+    ) as (read_stream, write_stream, _):
+        async with ClientSession(read_stream, write_stream) as session:
+            # Initialize the session
+            result = await session.initialize()
+            assert isinstance(result, InitializeResult)
+
+            # Make multiple requests to verify token updates
```
smurching, line 1272: Dumb question, where do we verify the token is actually updated?

tests/shared/test_streamable_http.py
```diff
@@ -1223,3 +1224,61 @@ async def sampling_callback(
...
+            for i in range(3):
+                tools = await session.list_tools()
+                assert len(tools.tools) == 4
+                await anyio.sleep(0.1)  # Small delay to ensure token updates
```
smurching: Is this needed? Generally it's better if we can avoid `sleep`s in tests, they can be flaky. Just calling the API multiple times should suffice right?

src/mcp/client/streamable_http.py
```diff
@@ -83,6 +95,7 @@ def __init__(
         headers: dict[str, Any] | None = None,
         timeout: timedelta = timedelta(seconds=30),
         sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
+        auth_token_provider: AuthTokenProvider | None = None,
```
smurching: Can we add some documentation for this field to the `StreamableHTTPTransport` docstring? I.e. note whether this takes precedence over headers in `headers`, etc. Also, should we align the name with the typescript SDK (IIRC it's called OauthClientProvider there ?)

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,18 @@ class RequestContext:
     sse_read_timeout: timedelta
 
 
+class AuthTokenProvider(Protocol):
+    """Protocol for providers that supply authentication tokens."""
```
smurching: Can we flesh out the docstring here a bit more? I.e. explain that this class is a base class that can be extended to provide custom behavior when authenticating from an MCP client to an MCP server?

src/mcp/client/streamable_http.py
```diff
@@ -102,6 +115,7 @@ def __init__(
             CONTENT_TYPE: JSON,
             **self.headers,
```
smurching, line 118: BTW I would expect any auth headers passed in `self.headers` to take precedence over the `auth_provider` headers - in many APIs explicitly setting a value (e.g. explicitly setting the value of the `Authorization` header via `headers`) takes precedence over implicitly inferring it. For example, I've seen that pattern often in MLflow. But will defer to the maintainers' opinion on that - in either case, it'd be good to have a test that exercises the behavior

tests/shared/test_streamable_http.py
aravind-segu: It is really hard in this testing environment to get the headers and verify the implementation. There is a mock server which hosts a list and set tools. We create a session, and add our messages to the write stream. This is then read by our transport layer and a request is sent to the server. I could not find a way to intercept or inspect this request object to verify the headers. So I just ensured the method was being called

src/mcp/client/streamable_http.py
aravind-segu: Good call, I added the behaviour to not override passed in headers, and add a test case as well

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
     sse_read_timeout: timedelta
 
 
+class AuthTokenProvider(Protocol):
+    """Protocol that can be extended to implement custom client-to-server authentication
```
smurching:
```suggestion
    """Base class that can be extended to implement custom client-to-server authentication
```

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
...
+    The get_token method is invoked before each request to the MCP Server to retrieve a
```
smurching:
```suggestion
```

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
...
+    fresh authentication token and update the request headers."""
```
smurching:
```suggestion
```

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
...
+
+    async def get_token(self) -> str:
+        """Get an authentication token.
```
smurching:
```suggestion
        """
        Get a token for authenticating to an MCP server. The token is assumed to
        be short-lived; clients may call this API multiple times per request to an MCP server.
        """
```

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
...
+
+        Returns:
+            str: The authentication token.
+        """
+        ...
+
+
 class StreamableHTTPTransport:
     """StreamableHTTP client transport implementation."""
```
smurching, line 92:
```suggestion
    """
    StreamableHTTP client transport implementation.
    
    :param auth_token_provider: Optional `AuthTokenProvider` instance for generating short-lived tokens used to authenticate to the MCP server. Note: authorization headers specified in `headers` will take precedence over tokens returned by this provider instance
    """
```

tests/shared/test_streamable_http.py
aravind-segu: We could try to build a custom app, that looks at the header, then calls the server, and returns the auth headers in the response headers. I will wait for the maintainer to chime in if they have better ideas on how to test this.

src/mcp/client/streamable_http.py
```diff
@@ -441,13 +478,21 @@ async def streamablehttp_client(
     `sse_read_timeout` determines how long (in seconds) the client will wait for a new
     event before disconnecting. All other HTTP operations are controlled by `timeout`.
 
+    `auth_client_provider` is an optional protocol that can be extended to implement
+    custom client-to-server authentication. Before each request to the MCP Server,
+    the get_token method is invoked to retrieve a fresh authentication token and
```
smurching:
```suggestion
    the auth_client_provider.get_token method is invoked to retrieve a fresh authentication token and
```

src/mcp/client/streamable_http.py
```diff
@@ -441,13 +478,21 @@ async def streamablehttp_client(
     `sse_read_timeout` determines how long (in seconds) the client will wait for a new
     event before disconnecting. All other HTTP operations are controlled by `timeout`.
 
+    `auth_client_provider` is an optional protocol that can be extended to implement
```
smurching:
```suggestion
    `auth_client_provider` is an optional instance of `AuthClientProvider` that can be passed to support
```

src/mcp/client/streamable_http.py
```diff
@@ -441,13 +478,21 @@ async def streamablehttp_client(
...
+    custom client-to-server authentication. Before each request to the MCP Server,
+    the get_token method is invoked to retrieve a fresh authentication token and
+    update the request headers. Note that if the passed in headers already
```
smurching:
```suggestion
    update the request headers. Note that if the passed in `headers` already
```

src/mcp/client/streamable_http.py
```diff
@@ -441,13 +478,21 @@ async def streamablehttp_client(
...
+    contain an authorization header, this provider will not be called.
```
smurching:
```suggestion
    contain an Authorization header, that header will take precedence over any tokens generated by this provider.
```

smurching: Just to avoid guaranteeing an implementation detail that might change later

smurching (commented):
Don't have permissions to approve, but this looks good to me, thanks @aravind-segu !

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,21 @@ class RequestContext:
     sse_read_timeout: timedelta
 
 
+class AuthClientProvider(Protocol):
+    """Base class that can be extended to implement custom client-to-server
+    authentication"""
+
+    async def get_token(self) -> str:
+        """Get a token for authenticating to an MCP server. The token is assumed to
```
smurching: I assume the linter will take care of updating the docstring formatting (?) and get rid of some of the extra spaces here

src/mcp/client/streamable_http.py
```diff
@@ -74,6 +74,20 @@ class RequestContext:
     sse_read_timeout: timedelta
 
 
+class AuthClientProvider(Protocol):
+    """Base class that can be extended to implement custom client-to-server
+    authentication"""
+
+    async def get_auth_headers(self) -> dict[str, str]:
```
smurching:
```suggestion
    async def get_headers(self) -> dict[str, str]:
```

"auth" is redundant

src/mcp/client/streamable_http.py
```diff
@@ -112,6 +128,24 @@ def _update_headers_with_session(
             headers[MCP_SESSION_ID] = self.session_id
         return headers
 
+    async def _update_headers_with_auth_headers(
+        self, base_headers: dict[str, str]
+    ) -> dict[str, str]:
+        """Update headers with auth_headers if auth client provider is specified.
+        The headers are merged giving precedence to the base_headers to
+        avoid overwriting existing Authorization headers"""
```
smurching:
```suggestion
        The headers are merged, giving precedence any headers already specified in base_headers
        """
```

src/mcp/client/streamable_http.py
```diff
@@ -112,6 +128,24 @@ def _update_headers_with_session(
...
+        if self.auth_client_provider is None:
+            return base_headers
+
+        auth_headers = await self.auth_client_provider.get_auth_headers()
```
smurching:
```suggestion
        auth_headers = await self.auth_client_provider.get_headers()
```

src/mcp/client/streamable_http.py
```diff
@@ -441,13 +474,22 @@ async def streamablehttp_client(
     `sse_read_timeout` determines how long (in seconds) the client will wait for a new
     event before disconnecting. All other HTTP operations are controlled by `timeout`.
 
+    `auth_client_provider` instance of `AuthClientProvider` that can be passed to
+    support client-to-server authentication. Before each request to the MCP Server,
+    the auth_client_provider.get_token method is invoked to retrieve a fresh
+    authentication token and update the request headers. Note that if the passed in
```
smurching:
```suggestion
    the auth_client_provider.get_headers() method is invoked to retrieve headers for authentication. Note that any headers already specified in `headers` will take precedence over headers returned by auth_client_provider.get_headers()
```

tests/shared/test_streamable_http.py
```diff
@@ -1223,3 +1224,64 @@ async def sampling_callback(
                 captured_message_params.messages[0].content.text
                 == "Server needs client sampling"
             )
+
+
+class MockAuthClientProvider:
```
smurching: I guess we don't need to actually define this class right (it's unused)? Just create a mock.Mock() inside tests? E.g. something like
```python
  client_provider = mock.Mock()
  client_provider.get_headers = AsyncMock(...
```

tests/shared/test_streamable_http.py
```diff
@@ -1223,3 +1224,64 @@ async def sampling_callback(
...
+            # Initialize the session
+            result = await session.initialize()
+            assert isinstance(result, InitializeResult)
+
+            # Make multiple requests to verify token updates
+            for i in range(3):
+                tools = await session.list_tools()
+                assert len(tools.tools) == 4
+
+    client_provider.get_auth_headers.call_count > 1
```
smurching:
```suggestion
    assert client_provider.get_auth_headers.call_count == 4
```

tests/shared/test_streamable_http.py
jerome3o-anthropic: off the cuff

I suspect you'd want to patch:
* create_mcp_http_client
* and/or aconnect_sse

In the appropriate places to catch all calls with headers, and assert your headers from the provider are there. I think the mocks could just pass through to the original function

jerome3o-anthropic (changes requested):
This looks really good - before merging I think we should definitely have tests that assert the headers are provided in the appropriate places. See my reply to @smurching's thread for an approach

ihrpr:
Thank you for working on this! Appreciate the help and contribution! 

I think we need to prioritise https://github.com/modelcontextprotocol/python-sdk/pull/308 as it's a already close to implementing the entire flow. I'll finish it on Monday, would appreciate help with the review and testing.

ihrpr:
thank you @aravind-segu for all your help! closing this PR in favour of HTTPX auth

Issue #552, opened 2025-04-21
On Windows 11 when you initialize an mcp client it hangs indefinitely


BenMawnMahlauNBTC:
**Describe the bug**
On Windows 11 when you initialize an mcp client it hangs indefinitely.

**To Reproduce**
Code to reproduce:
```python
# ruff: noqa
import asyncio

from mcp import ClientSession, StdioServerParameters
from mcp.client.sse import sse_client
from mcp.client.stdio import stdio_client


async def run():
    params = StdioServerParameters(
        command='bunx', args=['@playwright/mcp@latest']
    )
    async with stdio_client(params) as (read, write):
        print('inside client')
        async with ClientSession(read, write) as c:
            print('inside ClientSession')
            await c.initialize()
        print('exit ClientSession')
    print('exit stdio_client')


async def run_sse():
    async with sse_client('http://localhost:8931/sse') as (read, write):
        async with ClientSession(read, write) as c:
            await c.initialize()
        print('exit ClientSession')
    print('exit sse_client')


if __name__ == '__main__':
    asyncio.run(run_sse())  # works
    asyncio.run(run())  # does not work

```

**Expected behavior**
in both cases it should print both exit statements

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Python Version: Tested on both 3.13.1 and 3.12.7

**Additional context**
counterintuitively commenting out the code meant to support windows fixes this issue:
```python
import os
import sys
from contextlib import asynccontextmanager
from pathlib import Path
from typing import Literal, TextIO

import anyio
import anyio.lowlevel
from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
from anyio.streams.text import TextReceiveStream
from pydantic import BaseModel, Field

import mcp.types as types

# from .win32 import (
#     create_windows_process,
#     get_windows_executable_command,
#     terminate_windows_process,
# )

# Environment variables to inherit by default
DEFAULT_INHERITED_ENV_VARS = (
    [
        "APPDATA",
        "HOMEDRIVE",
        "HOMEPATH",
        "LOCALAPPDATA",
        "PATH",
        "PROCESSOR_ARCHITECTURE",
        "SYSTEMDRIVE",
        "SYSTEMROOT",
        "TEMP",
        "USERNAME",
        "USERPROFILE",
    ]
    if sys.platform == "win32"
    else ["HOME", "LOGNAME", "PATH", "SHELL", "TERM", "USER"]
)
[...]

PR #707, opened 2025-05-13
#552

DanielAvdar:
This pull request adds a new test to verify the fix for a Windows-specific process creation issue described in issue #552. The test ensures that the process creation function works without hanging.

### Added test for Windows process creation:

* [`tests/issues/test_552_windows_hang.py`](diffhunk://#diff-75f0c3f56e683e126a1427b31c4c5c2745e19a517433a65dbda20eff519564f4R1-R46): Introduced a new test, `test_windows_process_creation`, to validate the functionality of `_create_platform_compatible_process` on Windows. The test uses a simple command (`cmd /c echo Test successful`) and sets a timeout to prevent hanging. It verifies that the process is created successfully and produces output.


DanielAvdar:
The test `test_windows_process_creation` is currently marked as `xfail` for hanging processes, which is the expected behavior. It will pass once the issue causing the hang is resolved.
@ihrpr

tests/issues/test_552_windows_hang.py
```diff
@@ -0,0 +1,50 @@
+import asyncio
+import sys
+
+import pytest
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+
+# @pytest.mark.skipif(sys.platform != "win32", reason="Windows-specific test")
```
felixweinberger: Should this be uncommented? Tests do seem to fail on Ubuntu

tests/issues/test_552_windows_hang.py
```diff
@@ -0,0 +1,50 @@
...
+    params = StdioServerParameters(
+        command="cmd",
+        # Echo a valid JSON-RPC response message that will be parsed correctly
+        args=args,
+    )
+
+    # Directly test the fixed function that was causing the hanging issue
+    try:
+        # Set a timeout to prevent hanging
+        async with asyncio.timeout(5):
```
felixweinberger: Can we use `anyio.timeout` or `anyio.fail_after` here please to align with other tests?

tests/issues/test_552_windows_hang.py
```diff
@@ -0,0 +1,50 @@
+import asyncio
+import sys
+
+import pytest
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+
+# @pytest.mark.skipif(sys.platform != "win32", reason="Windows-specific test")
+@pytest.mark.skipif(sys.version_info < (3, 11), reason="asyncio.timeout in 3.11+")
```
felixweinberger: If you switch to `anyio` like recommended below, you shouldn't need this skip case anymore.

tests/issues/test_552_windows_hang.py
```diff
@@ -0,0 +1,50 @@
...
+                    await c.initialize()
+
+    except asyncio.TimeoutError:
+        pytest.xfail("Process creation timed out, indicating a hang issue")
+    except ProcessLookupError:
+        pytest.xfail("Process creation failed with ProcessLookupError")
+    except Exception as e:
+        assert "ExceptionGroup" in repr(e), f"Unexpected error: {e}"
+        assert "ProcessLookupError" in repr(e), f"Unexpected error: {e}"
+        pytest.xfail(f"Expected error: {e}")
```
felixweinberger: Could you simplify to use `anyio.fail_after(5)` like so?

```
      # Will fail with TimeoutError if it hangs
      with anyio.fail_after(5):
              print("inside client")
              async with ClientSession(read, write) as c:
                  print("inside ClientSession")
                  await c.initialize()
```

Then you don't need to do all the `except` statements either.

felixweinberger (dismissed):
Thank you @DanielAvdar for submitting this PR and helping us improve our test coverage!

I've left some comments on the PR - please address these and I'm happy to review quickly so we can get this landed.

felixweinberger:
Addressed comments + reformated the test to be a true _regression_ test such that it would fail _without_ the fix (previous format was causing xfail if not hanging which is counterintuitive).

PR #715, opened 2025-05-14
Adding authorization at http layer on streamable_http via httpx auth parameter

didier-durand:
…param

We add the use of param auth= to httpx_streamable.py in order to allow access to remote MCP server from local MCP clients when tcp/http level signature is in place.

## Motivation and Context

We work on use cases where MCP servers are hosted on cloud while MCP clients can be anywhere

In some environments like AWS, many applications are protected at http layer (API Gateway, Load Balancers, etc.) and require their request to be authenticated via tcp various headers.

In our case we use the tcp signatures supported by httpx_auth (see https://pypi.org/project/httpx-auth/) which works for multiple forms of authentication (incl Oauth2, AWS). See repo doc for all possible signature mechanisms.

## How Has This Been Tested?

We have run tests in AWS cloud environments with security gateways imposing signature at TCP level.

Without the TCP signature, the requests to TCP server fail with http code 403.  With the signature provided by httpx_auth, they go through the authentication gate.

## Breaking Changes

We took precautions in our code to avoid breaking changes:
- we only include auth in the httpx param when it is not None
-  we did not type it as a param in the class constructor to avoid pulling additional mandatory dependencies in the project. Otherwise, it would have pulled a dependency on AWS4Auth when httpx_auth is used (or on equivalent classes of other signing libs)

## Types of changes

- [ ] Documentation update

## Checklist

- [X ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X ] New and existing tests pass locally

## Additional context

We did not add tests here for such signatures as the validation of the signature can only happen when access to a live authorization gateway is in place (signature is time-dependent, account-dependent and validation algorithm of AWS is unknown)

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -83,6 +83,7 @@ class StreamableHTTPTransport:
         headers: dict[str, Any] | None = None,
         timeout: timedelta = timedelta(seconds=30),
         sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
+        auth= None,
     ) -> None:
         """Initialize the StreamableHTTP transport.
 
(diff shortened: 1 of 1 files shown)


didier-durand:
I'll check the various issues raised above and come back with a new PR.

PR #718, opened 2025-05-15
feat: add cursor pagination support to all client list methods

jerome3o-anthropic:
## Summary

This PR adds cursor-based pagination support to all list methods in the Python MCP client SDK. The client is now ready to handle paginated responses from servers, though server implementations don't yet support pagination.

## Changes

- Added optional `cursor` parameter to all client list methods:
  - `list_tools(cursor=None)`
  - `list_resources(cursor=None)`
  - `list_prompts(cursor=None)`
  - `list_resource_templates(cursor=None)`

- Added comprehensive tests verifying cursor parameter acceptance for all list methods

## Test Plan

Tests verify that the cursor parameter is properly accepted in various forms:
- Omitted (backward compatibility)
- `cursor=None`
- `cursor="some_value"`
- `cursor=""`

Note: The tests don't demonstrate actual pagination because no server in the SDK currently implements it. The tests only verify that the client accepts and passes the cursor parameter correctly.

## Notes

This addresses the issue reported in Slack where users couldn't access the cursor parameter in `list_tools()` and had to drop down to `session.send_request()` to use pagination.

Servers will need to be updated separately to actually implement cursor-based pagination - this PR just ensures the client is ready for it.


nbarbettini:
@ihrpr I have a server that supports pagination, and this doesn't quite work as is. It sends `request.cursor = 'xyz'` instead of `request.params.cursor = 'xyz'` which is what the [pagination spec](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/pagination#request-format) expects. I'll file an issue and propose a fix.

ihrpr:
thank you @nbarbettini, appreciate it

PR #721, opened 2025-05-15
Add progress notification callback for client

ihrpr:
It started as adding Everything server test for StreamableHttp and SSE, but then I noticed I can't test progress notifications. 

This PR adds a callback for progress notifications to the client session and sets progress token to request id where there is a callback. This approach is similar to how we are handling it in TypeScript.

Adds Everything server with tools, resources and prompts.


src/mcp/shared/session.py
```diff
@@ -231,15 +243,25 @@ async def send_request(
         ](1)
         self._response_streams[request_id] = response_stream
 
+        # Set up progress token if progress callback is provided
+        request_data = request.model_dump(by_alias=True, mode="json", exclude_none=True)
```
jerome3o-anthropic, line 247: [properly non-blocking] wondering if we can set this in the pydantic model before dumping? I'm really not fussed either way (it would just give you validation etc)

src/mcp/shared/session.py
```diff
@@ -363,6 +385,18 @@ async def _receive_loop(self) -> None:
...
                                 await self._in_flight[cancelled_id].cancel()
                         else:
+                            # Handle progress notifications callback
+                            if isinstance(notification.root, ProgressNotification):
+                                progress_token = notification.root.params.progressToken
+                                # If there is a progress callback for this token,
+                                # call it with the progress information
+                                if progress_token in self._progress_callbacks:
+                                    callback = self._progress_callbacks[progress_token]
+                                    callback(
```
jerome3o-anthropic: do we want this to support async functions?

tests/server/fastmcp/test_integration.py
```diff
@@ -138,6 +288,18 @@ def run_streamable_http_server(server_port: int) -> None:
     server.run()
 
 
+def run_everything_streamable_http_server(server_port: int) -> None:
```
jerome3o-anthropic: [non-blocking] slight preference to call this one `run_everything_server` and have the other `run_everything_legacy_sse_http_server` or something. Happy either way

tests/server/fastmcp/test_integration.py
```diff
@@ -323,3 +485,398 @@ async def test_fastmcp_stateless_streamable_http(
...
+
+    # Create notification collector
+    collector = NotificationCollector()
+
+    # Create a sampling callback that simulates an LLM
+
+    # Connect to the server with callbacks
+    async with sse_client(everything_server_url + "/sse") as streams:
+        # Set up message handler to capture notifications
+        async def message_handler(message):
```
jerome3o-anthropic, line 840: [non-blocking] I was thinking we could literally tap the memory channels that wireup the transport, but this is also good

jerome3o-anthropic (dismissed):

src/mcp/shared/session.py
ihrpr: hmm, I guess so?

tests/server/fastmcp/test_integration.py
ihrpr: I just wanted to make sure that the transports also support all the features, especially after catching an issues with sampling in shttp

src/mcp/shared/session.py
ihrpr: updated

PR #728, opened 2025-05-15
Feature/add enable disable methods tools

davila7:
Added the ability to dynamically enable and disable tools at runtime with client notifications.

## Motivation and Context
This change allows developers to temporarily restrict access to specific tools without removing them from the tool manager. This is useful for implementing feature flags, managing resource availability, or controlling access to functionality based on runtime conditions.

## How Has This Been Tested?
- Added unit tests for enabling/disabling tools
- Verified that disabled tools don't appear in tool listings and cannot be called
- Tested re-enabling previously disabled tools

## Breaking Changes
The tool list now includes the enabled state of each tool. However, all tools are initialized with enabled=True by default, maintaining backward compatibility with existing code.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/tools/tool_manager.py
+++ b/src/mcp/server/fastmcp/tools/tool_manager.py
@@ -39,8 +39,8 @@ class ToolManager:
         return self._tools.get(name)
 
     def list_tools(self) -> list[Tool]:
-        """List all registered tools."""
-        return list(self._tools.values())
+        """List all enabled registered tools."""
+        return [tool for tool in self._tools.values() if tool.enabled]
 
     def add_tool(
         self,
@@ -71,5 +71,8 @@ class ToolManager:
         tool = self.get_tool(name)
         if not tool:
             raise ToolError(f"Unknown tool: {name}")
+
+        if not tool.enabled:
+            raise ToolError(f"Tool is disabled: {name}")
 
         return await tool.run(arguments, context=context)
--- a/README.md
+++ b/README.md
@@ -247,16 +247,17 @@ async def fetch_weather(city: str) -> str:
         return response.text
 
 
-# Get a reference to the tool
 tool = mcp._tool_manager.get_tool("fetch_weather")
 
 
-# Disable the tool temporarily
-await tool.disable(mcp.get_context())
+async def disable_tool():
+    # Disable the tool temporarily
+    await tool.disable(mcp.get_context())
 
 
-# Later, re-enable the tool
-await tool.enable(mcp.get_context())
+async def enable_tool():
+    # Re-enable the tool when needed
+    await tool.enable(mcp.get_context())
 ```
 
 ### Prompts
--- a/README.md
+++ b/README.md
@@ -30,10 +30,12 @@
     - [Prompts](#prompts)
     - [Images](#images)
     - [Context](#context)
+    - [Authentication](#authentication)
   - [Running Your Server](#running-your-server)
     - [Development Mode](#development-mode)
     - [Claude Desktop Integration](#claude-desktop-integration)
     - [Direct Execution](#direct-execution)
+    - [Streamable HTTP Transport](#streamable-http-transport)
     - [Mounting to an Existing ASGI Server](#mounting-to-an-existing-asgi-server)
   - [Examples](#examples)
     - [Echo Server](#echo-server)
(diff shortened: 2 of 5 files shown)


src/mcp/server/fastmcp/tools/tool_manager.py
```diff
@@ -72,4 +72,7 @@ async def call_tool(
         if not tool:
             raise ToolError(f"Unknown tool: {name}")
 
+        if not tool.enabled:
```
bzsurbhi, line 75: Should we add similar checks in call_tool() and get_tool() as well?

bzsurbhi (commented):
I believe it needs a spec change as well. Do we already have a PR for it?

src/mcp/server/fastmcp/tools/tool_manager.py
davila7: For call_tool() we already have the necessary validation checking if the tool is enabled.

And for get_tool(), I think we don't need to add this validation. We retrieve in get_tool() and validate in call_tool() where the tool is actually used. 
This keeps the code cleaner and lets us access disabled tools when needed (for inspection, etc).

davila7:
> I believe it needs a spec change as well. Do we already have a PR for it?

I've added it to the README in this commit:
https://github.com/modelcontextprotocol/python-sdk/pull/728/commits/fd71115ed5ad04202ad3c8882a74851ee8781460#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R253

ihrpr:
Thank you for this contribution. While dynamic tool enabling/disabling could be useful for testing, we're closing this PR because:

  1. It creates API confusion (disabled vs non-existent tools)
  2. Requires MCP protocol changes to be added to SDK
  3. Adds maintenance burden without clear production benefits

  For testing, we recommend using different server configurations or mocking at the application level instead.

davila7:
@ihrpr 

I understand concerns about API complexity, but I'd like to offer a real-world perspective.

The TypeScript MCP SDK already implements this pattern via the `RegisteredTool` (https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/server/mcp.ts#L1003) type with methods like `enable()`, `disable()`, and `update()`. Their docs show a practical example for progressive auth:

```ts
const putMessageTool = server.tool("putMessage", /* ... */);
putMessageTool.disable(); // User must authenticate first

const upgradeAuthTool = server.tool("upgradeAuth", /* ... */, async ({ permission }) => {
  const { ok, previous } = await upgradeAuthAndStoreToken(permission);
  
  if (previous === "read") putMessageTool.enable();
  if (permission === "admin") upgradeAuthTool.remove();
});
```

This approach cleanly supports stateful logic, e.g., enabling tools based on auth state or feature flags, without complex conditions or restarts. It’s not just for testing; it solves real production needs where server behavior must adapt dynamically.

davila7:
> > Added the ability to dynamically enable and disable tools at runtime with client notifications.
> 
> Please can you expand on this?
> 
> The Typescript Feature mentioned is for [Dynamic Servers](https://github.com/modelcontextprotocol/typescript-sdk/pull/247), is the intention here to support it in Python SDK?

Yes, this feature implements the same Dynamic Servers capability from the TypeScript sdk.

- enable/disable tools without server restart
- tools appear/disappear based on context (auth, resources, etc.)
- automatic updates when tool lists change

Implementation:
- Added enabled: bool field to tools
- enable()/disable() methods with client notifications
- Filtered tool listing (only enabled tools shown)
- Runtime validation preventing disabled tool execution

DanilMartyniuk:
@davila7 , @bzsurbhi , I hope you will check the one failing test and approve this PR soon.

felixweinberger:
Hi @davila7 thank you for this contribution!

I believe this would constitute an enhancement of the actual MCP spec and would therefore need to go through a SEP to make sure we keep SDKs in sync and equivalent.

However, I understand this is actually trying to create parity between Typescript and Python SDKs. Is this a feature we introduced in Typescript without it being in the actual MCP Spec? I at least can't find this explicit capability in the spec itself.

@dsp-ant @ihrpr any thoughts in this case where we had a feature added to one SDK that may have been done before the SEP process was clearly established? Should we block on actually adding the feature to the spec here or prefer parity between SDKs?

felixweinberger:
Coming back to this one again, I believe we have a similar PR up that's more recent with more engagement: https://github.com/modelcontextprotocol/python-sdk/pull/1322

I'm going to close this one as a potential duplicate - feel free to chime in on that PR or submit another if you disagree.

PR #738, opened 2025-05-16
Change default binding from 0.0.0.0 to 127.0.0.1 to align with security recommendations

ciccolo-anthropic:
Change default server settings and all example code to bind to `127.0.0.1` rather than `0.0.0.0`

## Motivation and Context

The [official docs](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#security-warning) note that 
> When running locally, servers SHOULD bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0)

This change attempts to make that behavior the default and reduce the chances of successful DNS rebinding attacks.

## How Has This Been Tested?

Not tested.

## Breaking Changes

This could be breaking in the scenario where users are relying on this default behavior to make the server accessible outside of localhost. An alternative would be to remove the change to [src/mcp/server/fastmcp/server.py](https://github.com/modelcontextprotocol/python-sdk/compare/main...ciccolo-anthropic:python-sdk:main?expand=1#diff-14a2991a8f2be5b9e60dab5fae92d2230dc23e57937ad5e5641b832d9c662738) but keep the changes in example files and comments. That version would not be breaking.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/examples/servers/simple-prompt/mcp_simple_prompt/server.py
+++ b/examples/servers/simple-prompt/mcp_simple_prompt/server.py
@@ -114,7 +114,7 @@ def main(port: int, transport: str) -> int:
 
         import uvicorn
 
-        uvicorn.run(starlette_app, host="0.0.0.0", port=port)
+        uvicorn.run(starlette_app, host="127.0.0.1", port=port)
     else:
         from mcp.server.stdio import stdio_server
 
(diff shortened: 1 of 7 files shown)


ihrpr (approved):
Thank you!

PR #734, opened 2025-05-16
Allow kwargs down in AsyncClient when instantiating sse or streamable…

rumperto:
Adding support of AsynClient kwargs at the level of sse and streamable http client creation. That allows client implementations to set certain options, which are otherwise unreachable from outside the mcp code base (like max_redirects, verify, and others).

## Motivation and Context
Please have a look into the thread at https://github.com/run-llama/llama_index/issues/18752#issuecomment-2885961439

Initially, I tried to achieve a simple MCP server integration into my llama-index based agentic network and used to create a self-signed ssl certificate for local network testig. However, there is currently no option to get around the self-signed certification verification checks, so I had to monkeypatch the mcp code base in orer to set the 'verify' argument of the underlying AsyncClient. This PR extends the configuration options of the mcp clients and allows the developers to be more flexible.

## How Has This Been Tested?
Yes. I have tested ths with a patch inside the llama-index code and your code base. No productional application, just local dev.

## Breaking Changes
No

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [] [Not needed] - I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
-

PR #742, opened 2025-05-17
Configuring HTTP client for streamablehttp

fabioyamate:
## Motivation and Context
The current streamablehttp_client doesn't support customizing the new client except header and timeout. I need to customize the httpx client because my MCP server integration requires mTLS connection, thus I need to setup certification during connection.

This allow to pass a http client or default to the existing factory.

I was able to "successfully" test my MCP integration by "monkey patching" the introduced `mcp.shared._httpx_utils.create_mcp_client = my_custom_client_factory`.

Also, this supports base_url support by httpx, which is minor, but for my case, URLs are injected as they vary on deployment.

## How Has This Been Tested?
I tested my branch integrating with the deployed MCP, and it worked the same way with the monkey patch version, except that this became explicit.

## Breaking Changes
This doesn't introduce any breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This approach is similar to how other SDK configuration works, allow users to change the http client based on infrastructure requirements.

Why not kwargs? It felt that passing the client instance is more flexible than picking or passing the init constructors.

This couples with httpx.AsyncClient. Yes, in comparison to the current implementation that hides this detail. But, I didn't consider the scope of this PR, introducing a "wrapper" for the async client.

Update: I discovered a recent PR #734 after the PR opened. I didn't notice because I started to work prior it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -796,6 +796,29 @@ async def main():
             tool_result = await session.call_tool("echo", {"message": "hello"})
 ```
 
+## Configuring the HTTP client
+
+It is possible to override the httpx client to customize for your needs.
+
+* proxy
+* specific authentication needs mTLS and token
+* advanced configs supported by httpx
+
+```python
+import httpx
+from mcp.client.streamable_http import streamablehttp_client
+
+http_client = httpx.AsyncClient(
+    base_url="http://someserver",
+    proxy="http://proxy.local",
+    headers={
+        "Content": "application/json",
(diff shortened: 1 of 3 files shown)


ihrpr:
Thank you for working on this @fabioyamate! I'm closing this one as https://github.com/modelcontextprotocol/python-sdk/pull/752 will me merged.

PR #751, opened 2025-05-19
Add OAuth authentication client for HTTPX

ihrpr:
Adding support for Auth in client [using HTTPX Custom authentication schemes
](https://www.python-httpx.org/advanced/authentication/). 

NOTE: This PR adds support for Streamable HTTP, but SSE will be added as a follow up, all we need to do is to add one parameter.

This PR Implements [Auth in spec version 2025-03-26](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization)

 - Implementation supporting dynamic client registration, authorization flow, and automatic token refresh
  - Pluggable Storage System - Support for different token storage backends (in-memory and file-based providers included)
  - Streamable HTTP Integration - OAuth authentication integrated with the existing streamable HTTP transport 
  - Example Client - A complete working example demonstrating OAuth authentication flow with an interactive CLI

**Follow ups**:
 - add support for SSE

------ 

**Related Issues:**
https://github.com/modelcontextprotocol/python-sdk/issues/652
**Related PRs:**
https://github.com/modelcontextprotocol/python-sdk/pull/700
https://github.com/modelcontextprotocol/python-sdk/pull/308


examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
```diff
@@ -0,0 +1,354 @@
...
+        """Get the received state parameter."""
+        return self.callback_data["state"]
+
+
+class SimpleAuthClient:
+    """Simple MCP client with auth support."""
+
+    def __init__(self, server_url: str):
+        self.server_url = server_url
+        # Extract base URL for auth server (remove /mcp endpoint for auth endpoints)
```
pcarleton: nit: is this comment meant to be here? can't tell if it's a list of things about the server or if it used to refer to some code that's been removed for extracting the base url.

examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
```diff
@@ -0,0 +1,354 @@
...
+                client_metadata=OAuthClientMetadata.model_validate(
+                    client_metadata_dict
+                ),
+                storage=InMemoryTokenStorage(),
+                redirect_handler=_default_redirect_handler,
+                callback_handler=callback_handler,
+            )
+
+            # Initialize the auth handler and ensure we have tokens
+
```
pcarleton:
```suggestion
```

nit: think this might be lingering too?

pcarleton (changes requested):
minor nits, i also pushed a commit for scope handling.

I tested this against the typescript example, and it successfully connected after the scope fix

examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
ihrpr: yeah, sorry, refactored a bit, let me clean up all the commments

pcarleton (approved):
think it's upset about some linting errors. but otherwise lgtm!

aravind-segu:
@ihrpr Thank you very much for working on this request quickly! The OAuth implementation looks great and works well for our use case. With the Token Storage Provider in place, as long as the contract is maintained, we can also bypass the 3-legged OAuth flow and connect to MCP servers using an internally generated OAuth token - which is also perfect for us.

PR #752, opened 2025-05-19
Support custom httpx client creation

nickcoai:
## Motivation and Context
Currently the SDK defaults imply certain properties for httpx clients, two notable ones are redirect policies as well as proxies. To support various security configurations instead make the client wholly injectable. This avoids clients needing to fork the entire transport method.

## How Has This Been Tested?
Tested locally with a fake client.

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -1,6 +1,6 @@
 import logging
 from contextlib import asynccontextmanager
-from typing import Any, Callable
+from typing import Any
 from urllib.parse import urljoin, urlparse
 
 import anyio
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -1,6 +1,6 @@
 import logging
 from contextlib import asynccontextmanager
-from typing import Any
+from typing import Any, Callable
 from urllib.parse import urljoin, urlparse
 
 import anyio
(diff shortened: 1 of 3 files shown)


ihrpr (approved):
Thank you! Would be great to add instruction to README as well

Kludex:
It would have been better to pass the client itself...

rumperto:
> It would have been better to pass the client itself...

Agree with this

rumperto:
A new unit test would have been nice too.

Kludex:
And now I notice that the `timeout` parameter is not even included in the call to the callback... 😞

ihrpr:
> And now I notice that the `timeout` parameter is not even included in the call to the callback... 😞

it's not in sse, but in general is there, am I missing something?

Kludex:
> > And now I notice that the `timeout` parameter is not even included in the call to the callback... 😞
> 
> it's not in sse, but in general is there, am I missing something?

Ah, I missed that, sorry.

That said, on the sse, we use the client, so you can actually pass the timeout to the client before using the `aconnect_sse`. Can't we?

ihrpr:
> > > And now I notice that the `timeout` parameter is not even included in the call to the callback... 😞
> > 
> > 
> > it's not in sse, but in general is there, am I missing something?
> 
> Ah, I missed that, sorry.
> 
> That said, on the sse, we use the client, so you can actually pass the timeout to the client before using the `aconnect_sse`. Can't we?

yes, we should, I can just add this as there were a couple of PRs and issues about it

PR #760, opened 2025-05-20
add auth client sse

ihrpr:
Similar to Streamable Http Auth client added in #751 adding auth to SSE.

**Testing**
SSE
```
uv run mcp-simple-auth --transport sse

MCP_TRANSPORT_TYPE=sse uv run mcp-simple-auth-client
```
Check that StreamableHttp works:
```
uv run mcp-simple-auth --transport streamable-http
uv run mcp-simple-auth-client
```

### DIFF
--- a/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
+++ b/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
@@ -18,6 +18,7 @@ from urllib.parse import parse_qs, urlparse
 
 from mcp.client.auth import OAuthClientProvider, TokenStorage
 from mcp.client.session import ClientSession
+from mcp.client.sse import sse_client
 from mcp.client.streamable_http import streamablehttp_client
 from mcp.shared.auth import OAuthClientInformationFull, OAuthClientMetadata, OAuthToken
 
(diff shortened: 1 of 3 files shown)


examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
```diff
@@ -327,12 +339,14 @@ async def main():
     # Default server URL - can be overridden with environment variable
     # Most MCP streamable HTTP servers use /mcp as the endpoint
     server_url = os.getenv("MCP_SERVER_URL", "http://localhost:8000/mcp")
```
pcarleton: nit: might change this to be `/sse` if the transport is set since that's the default binding we get.  might get a little funky making sure you don't overwrite it though.

https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/server.py#L93

pcarleton (dismissed):
LGTM

Tested against the python example:

```
uv run mcp-simple-auth --transport sse

MCP_SERVER_URL=http://localhost:8000/sse MCP_TRANSPORT_TYPE=sse uv run mcp-simple-auth-client
```

examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
ihrpr: hmm, something is weird, I thought I changed that to port!

ihrpr: I didn't push it 😞

PR #523, opened 2025-04-15
llamallama-3.2-90b-vision-preview is deprecated

Arash-IND:
Updated model selection for Groq API integration

## Motivation and Context

The previous model identifier, "llama-3.2-90b-vision-preview" was causing 400 Bad Request errors with the Groq API since it was deprecated. Updated to use "meta-llama/llama-4-scout-17b-16e-instruct" properly supported by Groq's implementation.

## How Has This Been Tested?

Tested with the Groq API endpoint directly, verifying that requests now succeed without the previous 400 error.

## Breaking Changes

None. This is a transparent update to the model identifier that fixes a non-working configuration.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

The change resolves a 400 Bad Request error by providing the correct model identifier format required by Groq's API. The vision model was also replaced with a more appropriate model for text-based API interactions.

PR #777, opened 2025-05-21
Webhooks

pantanurag555:
## Motivation and Context
This is draft PR for the changes discussed in specification discussion [#523](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/523).

## How Has This Been Tested?
Tested by setting up a webhook, client and server locally. The webhook received all messages when the server was setup with streamable http. It was tested for both json_response True and False. All units tests succeeded.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist


src/mcp/server/streamable_http.py
```diff
@@ -410,9 +419,45 @@ async def _handle_post_request(
...
+                    }
+                    response = self._create_json_response(
+                        JSONRPCMessage(
+                            root=JSONRPCResponse(
+                                jsonrpc="2.0", id=message.root.id, result=result
+                            )
+                        ),
+                        HTTPStatus.OK,
+                    )
+                    asyncio.create_task(
```
000-000-000-000-000, line 444: issue: we should start the this task before responding with a 200

src/mcp/server/streamable_http.py
```diff
@@ -531,6 +576,126 @@ async def sse_writer():
...
+
+    async def _send_message_to_webhooks(
+        self,
+        webhooks: list[Webhook],
+        message: JSONRPCMessage | JSONRPCError | dict[str, str],
+    ):
+        for webhook in webhooks:
+            headers = {"Content-Type": CONTENT_TYPE_JSON}
+            # Add authorization headers
+            if webhook.authentication and webhook.authentication.credentials:
```
000-000-000-000-000, line 646: question: how much of the auth logic is duplicate from what is there for Streamable HTTP? Can auth stuff be abstracted into a helper layer if it is same?

src/mcp/server/streamable_http.py
```diff
@@ -531,6 +576,126 @@ async def sse_writer():
...
+                    )
+                elif webhook.authentication.strategy == "apiKey":
+                    headers["X-API-Key"] = webhook.authentication.credentials
+                elif webhook.authentication.strategy == "basic":
+                    try:
+                        # Try to parse as JSON
+                        creds_dict = json.loads(webhook.authentication.credentials)
+                        if "username" in creds_dict and "password" in creds_dict:
+                            # Create basic auth header from username and password
+                            import base64
```
000-000-000-000-000: suggestion: generally it is not a good idea to do imports in the runtime code like this as things that would fail at server startup may not fail until a request comes in that hits this path. Better to import this up top directly imo despite the fact that it might lead to more deps. btw is base64 part of the standard python libs?

pantanurag555: It is a python standard library. I will import it at the top.

src/mcp/server/streamable_http.py
pantanurag555: Apart from some parts that deal with bearer token, none of the other auth mechanisms are addressed in MCP. The parts that deal with bearer token are also not currently added in a way that makes them reusable for our purpose.

The auth logic here can be abstracted out into helper functions but there may be no reusability for now since it isn't used elsewhere.

src/mcp/server/streamable_http.py
pantanurag555: My understanding is that MCP should send a 200 to acknowledge the tool call prior to actually attempting to start the job. The idea is that it acts as an ack of the job and has no bearing whatsoever with the success/failure of the tool call. Ideally this should be a 202 but based on the current spec, 202 is reserved for responses when the input to the server is a response or notification.

Even if the job starts before sending the response, there won't be any functional difference since the task will be carried out asynchronously, right?

maxisbey:
Closing for now - this PR is quite stale and the SDK has changed significantly since it was opened. If webhooks get specced in the future, a fresh PR would be welcome!

PR #770, opened 2025-05-21
Remove `lstrip` in each OAuth endpoint

hongkunyoo:
`lstrip` strips left `/` which results in `ISSUER_URLauthorize`

## Motivation and Context

`/.well-known/oauth-authorization-server` output is weirld.

- expected: `"authorization_endpoint": "https://$MY_ISSUER/authorize",`
- actual: `"authorization_endpoint": "https://$MY_ISSUERauthorize",`

It seems `lstrip` is unnecessary for each OAuth endpoint

## How Has This Been Tested?

Check following result:

```
curl $MY_MCP_SERVER/.well-known/oauth-authorization-server

{
  "issuer": "$MY_ISSUER",
  "authorization_endpoint": "$MY_ISSUERauthorize",   # / is missing between $MY_ISSUER and authorize
  "token_endpoint": "$MY_ISSUERtoken",
  "registration_endpoint": "$MY_ISSUERregister",
  "scopes_supported": [
    "openid"
  ],
  "response_types_supported": [
    "code"
  ],
  "grant_types_supported": [
    "authorization_code",
    "refresh_token"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_post"
  ],
  "revocation_endpoint": "$MY_ISSUERrevoke",
  "revocation_endpoint_auth_methods_supported": [
    "client_secret_post"
  ],
  "code_challenge_methods_supported": [
    "S256"
  ]
}
```

## Breaking Changes

No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #779, opened 2025-05-22
Fix building auth metadata paths

Rodriguespn:
## Summary
Fixes OAuth endpoint URL construction to always join paths correctly, preventing missing or double slashes regardless of the issuer URL format. Removes unnecessary use of `lstrip("/")`, which caused endpoints like `$MY_ISSUERauthorize` instead of the expected `$MY_ISSUER/authorize`.

## Motivation and Context
The output of `/.well-known/oauth-authorization-server` was incorrect when the issuer URL included a path or when using URLs like `http://localhost:8000`.

Expected:

```shell
"authorization_endpoint": "https://example.com/authorize"
```

Actual:

```shell
"authorization_endpoint": "https://example.comauthorize"
```

This was due to the use of `lstrip("/")` when joining paths, which strips the leading slash and results in malformed URLs.

Note:
There is already [PR #770](url) attempting to address this, but it does not handle all cases (e.g., URLs like `http://localhost:8000` or issuer URLs with existing paths).

## How Has This Been Tested?
Tested with various issuer URLs, including:

- https://example.com
- https://example.com/
- https://example.com/auth/oidc/op/Customer
- https://example.com/auth/oidc/op/Customer
- http://localhost:8000
- http://localhost:8000/

Screenshot 2025-05-23 at 15 36 46

Verified the output of:

```shell
curl http://my-mcp-server/.well-known/oauth-authorization-server
```

Example output:

```json
{
  "issuer": "https://example.com",
  "authorization_endpoint": "https://example.com/authorize",
  "token_endpoint": "https://example.com/token",
  "registration_endpoint": "https://example.com/register",
  "revocation_endpoint": "https://example.com/revoke"
}
```

All endpoints are now correctly formed, with no missing or double slashes.

## Breaking Changes
No breaking changes

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
Related PRs
- https://github.com/modelcontextprotocol/python-sdk/pull/770 (works partially)

@hongkunyoo @pcarleton @Kludex


Kludex:
`urllib.parse.urljoin` should be enough

Rodriguespn:
> `urllib.parse.urljoin` should be enough
Thank you for the suggestion @Kludex. Refactored the `append_path` function to include `urllib.parse.urljoin` and reduce it to 3 lines to cover all use cases on the PR description

src/mcp/server/auth/routes.py
```diff
@@ -165,13 +166,24 @@ def build_metadata(
...
+
+
     authorization_url = modify_url_path(
-        issuer_url, lambda path: path.rstrip("/") + AUTHORIZATION_PATH.lstrip("/")
-    )
-    token_url = modify_url_path(
-        issuer_url, lambda path: path.rstrip("/") + TOKEN_PATH.lstrip("/")
+        issuer_url, lambda path: append_path(path, AUTHORIZATION_PATH)
     )
+    token_url = modify_url_path(issuer_url, lambda path: append_path(path, TOKEN_PATH))
```
Kludex: I think we can delete this `modify_url_path` tho. `str(issuer_url)` should work.

Rodriguespn: I was trying to modify as little code as possible, but you're right, this can be simplified by just casting the URL to a string and merging it with the endpoint (authorize, token, etc...)
I decided to convert the merged endpoint into a `AnyHttpUrl` pydantic class because the `OAuthMetadata` class is expecting `AnyHttpUrl` URLs

Let me know what you think! @Kludex

src/mcp/server/auth/routes.py
```diff
@@ -147,31 +147,19 @@ def create_auth_routes(
...
 def build_metadata(
     issuer_url: AnyHttpUrl,
     service_documentation_url: AnyHttpUrl | None,
     client_registration_options: ClientRegistrationOptions,
     revocation_options: RevocationOptions,
 ) -> OAuthMetadata:
-    authorization_url = modify_url_path(
-        issuer_url, lambda path: path.rstrip("/") + AUTHORIZATION_PATH.lstrip("/")
+    authorization_url = AnyHttpUrl(
+        str(issuer_url).rstrip("/") + AUTHORIZATION_PATH
```
Kludex, line 157: There's no need to strip in any of those.

```suggestion
        urljoin(str(issuer_url), AUTHORIZATION_PATH)
```

Rodriguespn: Thanks for the suggestion, @Kludex!

I initially thought the same — that `urljoin(str(issuer_url), "/authorize")` would be enough. But it actually breaks in some cases where the `base_url` includes a path.

For example, when the `base_url` is `https://example.com/auth/oidc/op/Customer/`, `urljoin` returns `https://example.com/authorize`, which drops the intended path entirely. That’s because `urljoin` treats the `/authorize` as an absolute path and replaces everything after the domain.

To illustrate this, I added a test suite comparing both approaches:

```python
import unittest
from urllib.parse import urljoin

class TestModifyUrlPath(unittest.TestCase):
    def test_append_authorize_to_urls_with_urljoin(self):
        """Test appending /authorize to various URL formats using urljoin"""
        test_cases = [
            ("https://example.com", "https://example.com/authorize"),
            ("https://example.com/", "https://example.com/authorize"),
            ("https://example.com/auth/oidc/op/Customer", "https://example.com/auth/oidc/op/Customer/authorize"),
            ("https://example.com/auth/oidc/op/Customer/", "https://example.com/auth/oidc/op/Customer/authorize"),
            ("http://localhost:8000", "http://localhost:8000/authorize"),
            ("http://localhost:8000/", "http://localhost:8000/authorize"),
        ]

        for base_url, expected in test_cases:
            any_http_url = AnyHttpUrl(base_url)
            with self.subTest(base_url=any_http_url):
                result = AnyHttpUrl(urljoin(str(any_http_url), "/authorize"))
                self.assertEqual(result, AnyHttpUrl(expected))

    def test_append_authorize_to_urls_with_rstrip(self):
        """Test appending /authorize to various URL formats using rstrip"""
        test_cases = [
            ("https://example.com", "https://example.com/authorize"),
            ("https://example.com/", "https://example.com/authorize"),
            ("https://example.com/auth/oidc/op/Customer", "https://example.com/auth/oidc/op/Customer/authorize"),
            ("https://example.com/auth/oidc/op/Customer/", "https://example.com/auth/oidc/op/Customer/authorize"),
            ("http://localhost:8000", "http://localhost:8000/authorize"),
            ("http://localhost:8000/", "http://localhost:8000/authorize"),
        ]

        for base_url, expected in test_cases:
            any_http_url = AnyHttpUrl(base_url)
            with self.subTest(base_url=any_http_url):
                result = AnyHttpUrl(str(any_http_url).rstrip("/") + "/authorize")
                self.assertEqual(result, AnyHttpUrl(expected))

if __name__ == "__main__":
    unittest.main()
```

So for consistency across all cases, I believe we need to keep the rstrip("/") approach.

Let me know what you think!

src/mcp/server/auth/routes.py
```diff
@@ -193,15 +181,15 @@ def build_metadata(
...
     if revocation_options.enabled:
-        metadata.revocation_endpoint = modify_url_path(
-            issuer_url, lambda path: path.rstrip("/") + REVOCATION_PATH.lstrip("/")
+        metadata.revocation_endpoint = AnyHttpUrl(
+            str(issuer_url).rstrip("/") + REVOCATION_PATH
         )
         metadata.revocation_endpoint_auth_methods_supported = ["client_secret_post"]
 
-    return metadata
+    return metadata
```
Kludex:
```suggestion
    return metadata

```

Kludex:
https://www.githubstatus.com/ 😢

Kludex:
I'll merge this once https://github.com/modelcontextprotocol/python-sdk/pull/806 lands. I need an approval from another maintainer there.

src/mcp/server/auth/routes.py
Kludex: Unfortunate, but it makes sense.

tests/client/test_auth.py
```diff
@@ -905,3 +908,76 @@ async def test_token_exchange_error_basic(self, oauth_provider, oauth_client_inf
...
+    metadata = build_metadata(
+        issuer_url=AnyHttpUrl(issuer_url),
+        service_documentation_url=AnyHttpUrl(service_documentation_url),
+        client_registration_options=ClientRegistrationOptions(
+            enabled=True, valid_scopes=["read", "write", "admin"]
+        ),
+        revocation_options=RevocationOptions(enabled=True),
+    )
+
+    assert metadata == snapshot(
```
Kludex, line 969: I've added [`inline-snapshot`](https://pypi.org/project/inline-snapshot/) as a test dependency. It's super useful.

agronholm:
The tests seem to be relying on global state, making `test_build_metadata[with-path-param]` variant fail 100% of the time locally because the snapshot from the previous invocation with different URL parameters differs from the expected metadata (`AnyHttpUrl('https://auth.example.com/') != AnyHttpUrl('https://auth.example.com/v1/mcp')`).

Kludex:
- Thanks @agronholm - opened https://github.com/modelcontextprotocol/python-sdk/pull/945

PR #786, opened 2025-05-22
Handles message type Exception in lowlevel/server.py _handle_message function. Mentioned as TODO on line 528.

AishwaryaKalloli:
Handles message type Exception in lowlevel/server.py _handle_message function.
Right now I am sending a error log if parameter raise_exception=False.
Initital commit, need feedback.

## Motivation and Context
Addressing a TODO mentioned in the code, handling one of the unhandled types of message in server.

## How Has This Been Tested?
Initial commit, testing pending, need feedback to move ahead.

## Breaking Changes
Does not break existing flow. Since the method raises exception it will break the code in case of exception, however it is expected behaviour.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -540,14 +540,13 @@ class Server(Generic[LifespanResultT]):
                     logger.error(f"Received error message: {message}")
                     if raise_exceptions:
                         raise message
-                    # Send the error as a notification since we don't have a request context
+                    # Send the error as a notification
+                    # as we don't have a request context
                     await session.send_log_message(
                         level="error",
                         data=types.ErrorData(
-                            code=types.INTERNAL_ERROR,
-                            message=str(message),
-                            data=None
-                        )
+                            code=types.INTERNAL_ERROR, message=str(message), data=None
+                        ),
                     )
 
             for warning in w:
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -536,6 +536,19 @@ class Server(Generic[LifespanResultT]):
                         )
                 case types.ClientNotification(root=notify):
                     await self._handle_notification(notify)
+                case Exception():
+                    logger.error(f"Received error message: {message}")
+                    if raise_exceptions:
+                        raise message
+                    # Send the error as a notification since we don't have a request context
+                    await session.send_log_message(
+                        level="error",
+                        data=types.ErrorData(
+                            code=types.INTERNAL_ERROR,
+                            message=str(message),
+                            data=None
+                        )
+                    )
 
             for warning in w:
                 logger.info(f"Warning: {warning.category.__name__}: {warning.message}")


AishwaryaKalloli:
@Kludex Came across [TODO](https://github.com/modelcontextprotocol/python-sdk/blob/e80c0150e1c2e45f66195d3cf7d209be31ce6e5d/src/mcp/server/lowlevel/server.py#L528) mentioned in `src/mcp/lowlevel/server.py` by you while reading the code.

Would appreciate feedback on how to handle the exception. Since there is no request attached to the message I have sent a log notification response to the client with the error message.
Please let me know if it looks good.

felixweinberger:
@AishwaryaKalloli thanks for your contribution!

Made some tweaks on top of your branch + rebased + added some test cases for exception handling. I don't think we should be using JSONRPC types for log messages, so changed this to a literal dict.

src/mcp/server/lowlevel/server.py
```diff
@@ -603,13 +603,26 @@ async def _handle_message(
...
+                    logger.error(f"Received exception from stream: {message}")
+                    if raise_exceptions:
+                        raise message
+                    await session.send_log_message(
+                        level="error",
+                        data={
+                            "message": str(message),
+                            "type": type(message).__name__,
+                            "module": type(message).__module__,
+                            "args": getattr(message, "args", None),
```
Kludex: Can we instead pass the values that would be passed to `__exit__`? 

exception_type, exception_value, exception_traceback

---

I'm assuming passing the data in this shape was your idea? Or do we do this somewhere else already?

felixweinberger: Amended.

felixweinberger:
@Kludex ready for another look

src/mcp/server/lowlevel/server.py
```diff
@@ -601,13 +601,25 @@ async def _handle_message(
...
+                case Exception():
+                    logger.error(f"Received exception from stream: {message}")
+                    if raise_exceptions:
+                        raise message
+                    await session.send_log_message(
+                        level="error",
+                        data={
+                            "exception_type": type(message).__name__,
+                            "exception_value": str(message),
+                            "exception_traceback": None,
```
Kludex: You have this information, why none?

Kludex: But actually, should we be sending the traceback? We don't want to expose this kind of info to the client.

felixweinberger: Good point - amended.

src/mcp/server/lowlevel/server.py
```diff
@@ -601,13 +601,21 @@ async def _handle_message(
...
                     await self._handle_notification(notify)
+                case Exception():
+                    logger.error(f"Received exception from stream: {message}")
+                    if raise_exceptions:
+                        raise message
+                    await session.send_log_message(
+                        level="error",
+                        data="Internal Server Error",
+                        logger="mcp.server.exception_handler",
+                    )
```
Kludex, line 657: The message is not being send it, is it?

felixweinberger: Whoops, great catch! amending...

felixweinberger:
@Kludex quick stamp?

PR #787, opened 2025-05-22
Fix `ClientSessionGroup` cleanup logic

mkeid:
The order of steps responsible for closing exist stacks is reversed. This is causing error when the `ClientSessionGroup` exits its context.

This change also includes some small doc improvements.

Fix for https://github.com/modelcontextprotocol/python-sdk/pull/639


ihrpr (approved):
Thank you

Issue #242, opened 2025-03-05
Support nginx reverse proxy or configure base_url


MiV1N:
**Is your feature request related to a problem? Please describe.**

I want to deploy multiple MCP servers behind Nginx, with Nginx using URL prefixes to reverse proxy messages to different MCP servers, but I get some errors when accessing:

```mermaid
graph LR
    subgraph "Application Host Process"

        C1[Client]

    end

    subgraph "Internet"

        N3[Nginx <br> 192.168.1.224:8889]
        S3[MCP Server <br> 192.168.1.224:18080]

        C1 --> N3
        N3 <--> S3
    end

```

**nginx.conf**：
```
server {
    listen 8889;

    location /abcdef/dddddd/ {
        rewrite ^/abcdef/dddddd/(.*)$ /$1 break;
        proxy_pass http://192.168.1.224:18080; 

        proxy_set_header Host $host;
        proxy_set_header    X-Real-Scheme $scheme;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }
}
```

**server.py**
```
import argparse
import os
from dotenv import load_dotenv
from mcp.server.fastmcp import FastMCP
from starlette.applications import Starlette
from mcp.server.sse import SseServerTransport
from starlette.requests import Request
from starlette.routing import Mount, Route
from mcp.server import Server
from datetime import datetime
import logging
import warnings
import uvicorn

MCP_SERVER_NAME = "elasticsearch-mcp-sse"

logging.basicConfig(
    level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
)
logger = logging.getLogger(MCP_SERVER_NAME)

mcp = FastMCP(MCP_SERVER_NAME)


@mcp.tool()
def get_time() -> str:   
    return datetime.now().isoformat()


def create_starlette_app(mcp_server: Server, *, debug: bool = False) -> Starlette:
    """Create a Starlette application that can server the provied mcp server with SSE."""
    sse = SseServerTransport("/messages/")

    async def handle_sse(request: Request) -> None:
        async with sse.connect_sse(
                request.scope,
                request.receive,
                request._send,
        ) as (read_stream, write_stream):
            await mcp_server.run(
                read_stream,
                write_stream,
                mcp_server.create_initialization_options(),
            )

    return Starlette(
        debug=debug,
        routes=[
[...]

PR #797, opened 2025-05-24
Add root_path support to properly route subpaths for SSE and streamable-http endpoints when behind Nginx or other HTTP reverse proxies

LiangYang666:
This PR adds support for `root_path` in subpath routing of SSE andstreamable-http endpoints when using Nginx or other HTTP middleware as a reverse proxy.

## Motivation and Context

When deploying applications behind a reverse proxy (e.g., Nginx), it's common to route requests via subpaths like `/server1/` or  `/server2/` . Without proper handling of these subpaths, SSE and streamable-http endpoints may fail to receive correctly routed requests.

This issue was originally reported in https://github.com/modelcontextprotocol/python-sdk/issues/242

Similar functionality exists in other web applications:
- [base_url in jupyter](https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#command)
- [base_url in filebrowser](https://filebrowser.org/cli/filebrowser-config-set#options) 
- [root_url in grafana](https://grafana.com/tutorials/run-grafana-behind-a-proxy/#alternative-for-serving-grafana-under-a-sub-path)

## How Has This Been Tested?

Yes, this has been tested and running stably in production environments for several days. Both sse and streamable-http endpoints have been verified to work correctly under subpaths behind Nginx.

## Breaking Changes

No breaking changes are introduced. Users do not need to update their code or configuration files. This is an optional enhancement that activates only when `root_path` is provided.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This change ensures that the application respects the `root_path` setting for streaming endpoints by utilizing Uvicorn’s built-in support for it. It does not introduce any custom routing logic but instead makes sure that streaming responses behave consistently with regular HTTP responses under subpath deployments.

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -89,6 +89,7 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     # HTTP settings
     host: str = "0.0.0.0"
     port: int = 8000
+    root_path: str = "/" # root path, same with root_path in uvicorn config, for sse or streamable http
     mount_path: str = "/"  # Mount path (e.g. "/github", defaults to root path)
     sse_path: str = "/sse"
     message_path: str = "/messages/"
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -605,6 +605,7 @@ class FastMCP:
             starlette_app,
             host=self.settings.host,
             port=self.settings.port,
+            root_path=self.settings.root_path,
             log_level=self.settings.log_level.lower(),
         )
         server = uvicorn.Server(config)
@@ -620,6 +621,7 @@ class FastMCP:
             starlette_app,
             host=self.settings.host,
             port=self.settings.port,
+            root_path=self.settings.root_path,
             log_level=self.settings.log_level.lower(),
         )
         server = uvicorn.Server(config)
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -89,7 +89,7 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     # HTTP settings
     host: str = "0.0.0.0"
     port: int = 8000
-    root_path: str = "/" # root path, same with root_path in uvicorn config, for sse or streamable http
+    root_path: str = "" # root path, same with root_path in uvicorn config, for sse or streamable http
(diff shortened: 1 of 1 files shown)


LiangYang666:
cc @ihrpr

lukehsiao (commented):
We're running into this issue as well. This seems an elegant fix!

lukehsiao:
@felixweinberger, I confirmed that mount_path does NOT address the use case. In fact, I don't see any difference in behavior.

However, I did find a different workaround. In my particular case, I'm running FastMCP mounted to a FastAPI server, that is served [behind a proxy, which strips prefixes](https://fastapi.tiangolo.com/advanced/behind-a-proxy/).

It turns out FastAPI has a bug, which, if I workaround that, also seems to resolve the issue here.

In this setting, FastAPI docs instruct us to provide `root_path`:

> Passing the root_path to FastAPI would be the equivalent of passing
> the `--root-path` command line option to Uvicorn or Hypercorn.

This is wrong. It is NOT equivalent.

There is a very old PR to address this: https://github.com/fastapi/fastapi/pull/11160

Long story short, switching to providing `--root-path` via command line option, rather than as a parameter to FastAPI, also resolved our issue here.

felixweinberger:
> @felixweinberger, I confirmed that mount_path does NOT address the use case. In fact, I don't see any difference in behavior.
> 
> However, I did find a different workaround. In my particular case, I'm running FastMCP mounted to a FastAPI server, that is served [behind a proxy, which strips prefixes](https://fastapi.tiangolo.com/advanced/behind-a-proxy/).
> 
> It turns out FastAPI has a bug, which, if I workaround that, also seems to resolve the issue here.
> 
> In this setting, FastAPI docs instruct us to provide `root_path`:
> 
> > Passing the root_path to FastAPI would be the equivalent of passing
> > the `--root-path` command line option to Uvicorn or Hypercorn.
> 
> This is wrong. It is NOT equivalent.
> 
> There is a very old PR to address this: [fastapi/fastapi#11160](https://github.com/fastapi/fastapi/pull/11160)
> 
> Long story short, switching to providing `--root-path` via command line option, rather than as a parameter to FastAPI, also resolved our issue here.

Thank you for reporting back and the additional details on your workaround! 

Given this sounds like the workaround needed to address a FastAPI specific issue, I'm going to close this PR for now. If you find yourself needing this feature in future for some other purpose, happy to look at a resubmission.

PR #798, opened 2025-05-24
Session validation error

moldhouse:
This PR adds exception handling to validation of incoming grpc requests in the session.

## Motivation and Context

While the public interface of the session (e.g. `session.send_request`) makes assumptions about the data types send to the session, the `StreamableHttpServerTransport` does not use this, and rather sends data directly to the [channel](https://github.com/moldhouse/python-sdk/blob/ab9c9ed0afb3f18935b61acf8e33b5c9c3cdbc82/src/mcp/server/streamable_http.py#L515). Therefore, [the validation](https://github.com/moldhouse/python-sdk/blob/ab9c9ed0afb3f18935b61acf8e33b5c9c3cdbc82/src/mcp/shared/session.py#L355) of the incoming data in the receive loop of the session may fail. This can be integration tested by sending an invalid method to the server, which crashes the task of the particular session.

## How Has This Been Tested?

I added a unit test on the session level that verifies that the session is unresponsive after a bad input. Construction of the session in the test feels a bit awkward, and there might be better ways to test this particular piece of code. One refactor idea might be to move the logic in the session loop to a separate method to better unit test it.

## Breaking Changes

This change does not break any public apis. It does however change behaviour, in that the server does not crash on invalid messages, but swallows them, similar how it is already happening for notifications which can not be validated.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -351,26 +351,34 @@ class BaseSession(
                 if isinstance(message, Exception):
                     await self._handle_incoming(message)
                 elif isinstance(message.message.root, JSONRPCRequest):
-                    validated_request = self._receive_request_type.model_validate(
-                        message.message.root.model_dump(
-                            by_alias=True, mode="json", exclude_none=True
+                    try:
+                        validated_request = self._receive_request_type.model_validate(
+                            message.message.root.model_dump(
+                                by_alias=True, mode="json", exclude_none=True
+                            )
                         )
-                    )
-                    responder = RequestResponder(
-                        request_id=message.message.root.id,
-                        request_meta=validated_request.root.params.meta
-                        if validated_request.root.params
-                        else None,
-                        request=validated_request,
-                        session=self,
-                        on_complete=lambda r: self._in_flight.pop(r.request_id, None),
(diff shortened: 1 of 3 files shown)


src/mcp/shared/session.py
```diff
@@ -351,26 +351,34 @@ async def _receive_loop(self) -> None:
...
-                    await self._received_request(responder)
+                        self._in_flight[responder.request_id] = responder
+                        await self._received_request(responder)
 
-                    if not responder._completed:  # type: ignore[reportPrivateUsage]
-                        await self._handle_incoming(responder)
+                        if not responder._completed:  # type: ignore[reportPrivateUsage]
+                            await self._handle_incoming(responder)
+                    except Exception as e:
+                        logging.warning(
```
felixweinberger, line 378: we should send back some feedback about invalid request to the client here (and update the tests to check for that)

felixweinberger:
Looks like this should already be addressed by https://github.com/modelcontextprotocol/python-sdk/pull/967, hence closing this one to clean up our backlog.

Thank you again for this contribution!

PR #806, opened 2025-05-26
Pin uv version

Kludex:
uv changed slightly how the lockfile looks like on the most recent versions. To avoid having conflicts between environments, I'm pinning uv.


pyproject.toml
```diff
@@ -55,6 +56,7 @@ dev = [
     "pytest-xdist>=3.6.1",
     "pytest-examples>=0.0.14",
     "pytest-pretty>=1.2.0",
+    "inline-snapshot>=0.23.0",
```
Kludex, line 59: I've included this because I'll be using in another PR. I'm preemptively including it here because it doesn't matter much since it's a dev dependency.

evalstate (approved):
lgtm

samuelcolvin (approved):
LGTM

PR #816, opened 2025-05-27
Support for http request injection propagation to tools

ihrpr:
This PR implements request context propagation in the MCP Python SDK for SSE transport, build on top of PR #380 and addressing concerns in the comments. The implementation captures HTTP request metadata (headers, method, URL, etc.) for each individual request and makes it available to request handlers for authentication and observability purposes.

  Commends addressed based on PR #380

  1. Per-Request Context:  implementation captures context for each individual POST request
  2. Non-Breaking: No changes to public API signatures - adding default to type definition
  3. Added tests verifying header propagation and context isolation
-----
  ## Implementation Details

  - Modified ServerMessageMetadata to carry request context through the message pipeline
  - Updated SSE transport to extract request metadata for each POST message
  - Server class extracts context from message metadata and makes it available via server.request_context
  - Request handlers can access HTTP headers and other metadata for authentication/authorization decisions

## Follow ups
- Implement request context propagation for Streamable Http
- Refactor SSE tests + Streamable Http

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -559,10 +559,8 @@ class Server(Generic[LifespanResultT]):
             try:
                 # Extract request context from message metadata
                 request_data = None
-                if (
-                    hasattr(message, "message_metadata")
-                    and message.message_metadata
-                    and isinstance(message.message_metadata, ServerMessageMetadata)
+                if message.message_metadata is not None and isinstance(
+                    message.message_metadata, ServerMessageMetadata
                 ):
                     request_data = message.message_metadata.request_context
 
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -72,11 +72,12 @@ import logging
 import warnings
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterable
 from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager
-from typing import Any, Generic, TypeVar
+from typing import Any, Generic
 
 import anyio
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from pydantic import AnyUrl
+from typing_extensions import TypeVar
 
 import mcp.types as types
 from mcp.server.lowlevel.helper_types import ReadResourceContents
@@ -91,9 +92,10 @@ from mcp.shared.session import RequestResponder
 logger = logging.getLogger(__name__)
 
 LifespanResultT = TypeVar("LifespanResultT")
+RequestT = TypeVar("RequestT", default=Any)
 
 # This will be properly typed in each Server instance's context
-request_ctx: contextvars.ContextVar[RequestContext[ServerSession, Any]] = (
+request_ctx: contextvars.ContextVar[RequestContext[ServerSession, Any, Any]] = (
     contextvars.ContextVar("request_ctx")
 )
 
@@ -111,7 +113,7 @@ class NotificationOptions:
 
 
 @asynccontextmanager
-async def lifespan(server: Server[LifespanResultT]) -> AsyncIterator[object]:
+async def lifespan(server: Server[LifespanResultT, RequestT]) -> AsyncIterator[object]:
     """Default lifespan context manager that does nothing.
 
     Args:
@@ -123,14 +125,15 @@ async def lifespan(server: Server[LifespanResultT]) -> AsyncIterator[object]:
     yield {}
 
 
-class Server(Generic[LifespanResultT]):
+class Server(Generic[LifespanResultT, RequestT]):
     def __init__(
         self,
         name: str,
         version: str | None = None,
         instructions: str | None = None,
         lifespan: Callable[
-            [Server[LifespanResultT]], AbstractAsyncContextManager[LifespanResultT]
+            [Server[LifespanResultT, RequestT]],
+            AbstractAsyncContextManager[LifespanResultT],
         ] = lifespan,
     ):
         self.name = name
@@ -215,7 +218,9 @@ class Server(Generic[LifespanResultT]):
         )
 
     @property
-    def request_context(self) -> RequestContext[ServerSession, LifespanResultT]:
+    def request_context(
+        self,
+    ) -> RequestContext[ServerSession, LifespanResultT, RequestT]:
         """If called outside of a request context, this will raise a LookupError."""
         return request_ctx.get()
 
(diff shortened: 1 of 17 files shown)


src/mcp/server/sse.py
```diff
@@ -203,7 +204,19 @@ async def handle_post_message(
...
-        session_message = SessionMessage(message)
+        # Extract request headers and other context
+        request_context: RequestData = {
+            "headers": dict(request.headers),
+            "method": request.method,
+            "url": str(request.url),
+            "client": request.client,
+            "path_params": request.path_params,
+            "query_params": dict(request.query_params),
+        }
```
dsp-ant: Can we instead of passing an open dict, just pass the request object? This gives the full power to the consumer. They can still `isinstance` check to see what specific object they are getting. This way we are a bit more generic than specifiying an open dict.

ihrpr: My only concern with that is that we are putting starlette.requests into the core and types. Not sure I'm super happy about that, but equally we use starlette everywhere in the SDK

src/mcp/server/lowlevel/server.py
```diff
@@ -215,7 +215,9 @@ def get_capabilities(
         )
 
     @property
-    def request_context(self) -> RequestContext[ServerSession, LifespanResultT]:
+    def request_context(
+        self,
+    ) -> RequestContext[ServerSession, LifespanResultT]:
```
Kludex: Unrelated, but I think we should change the `line-length` to 120 characters, as I mentioned before.

```suggestion
    def request_context(self) -> RequestContext[ServerSession, LifespanResultT]:
```

src/mcp/server/sse.py
Kludex: You should pass the `request.scope`, which is a dictionary, and contains all the connection information the user will ever need.

tests/server/fastmcp/test_integration.py
```diff
@@ -435,6 +430,181 @@ async def test_fastmcp_without_auth(server: None, server_url: str) -> None:
...
+            if ctx.request_context.request:
+                context_data["headers"] = ctx.request_context.request.get("headers", {})
+                context_data["method"] = ctx.request_context.request.get("method")
+                context_data["url"] = ctx.request_context.request.get("url")
+        except Exception:
+            pass
+        return json.dumps(context_data)
+
+    # Create the SSE app
+    app: Starlette = mcp.sse_app()
```
Kludex: Do you need to do this because you use PyCharm?

```suggestion
    app = mcp.sse_app()
```

tests/server/fastmcp/test_integration.py
```diff
@@ -435,6 +430,181 @@ async def test_fastmcp_without_auth(server: None, server_url: str) -> None:
...
+            "method": None,
+            "url": None,
+        }
+        try:
+            if ctx.request_context.request:
+                context_data["headers"] = ctx.request_context.request.get("headers", {})
+                context_data["method"] = ctx.request_context.request.get("method")
+                context_data["url"] = ctx.request_context.request.get("url")
+        except Exception:
+            pass
```
Kludex: What are you handling here here?

tests/shared/test_sse.py
```diff
@@ -318,3 +319,192 @@ async def test_sse_client_basic_connection_mounted_app(
...
+                    inputSchema={
+                        "type": "object",
+                        "properties": {"request_id": {"type": "string"}},
+                        "required": ["request_id"],
+                    },
+                ),
+            ]
+
+
+def run_context_server(server_port: int) -> None:
```
Kludex, line 367: Can we avoid running uvicorn in the test suite and just use the `httpx.AsyncClient` with the `ASGITransport`?

tests/server/fastmcp/test_integration.py
ihrpr: removed in all cases

tests/shared/test_sse.py
ihrpr: will do as a follow up, as sse tests are using it everywhere

tests/server/fastmcp/test_integration.py
ihrpr: removed

tests/shared/test_sse.py
Kludex: That's probably why I found some flaky issues the other day. We shouldn't have those _sleeps_ around.

src/mcp/server/sse.py
Kludex: The latest changes use `request`, not the `request.scope` I mentioned tho 👀

ihrpr: I saw `request.scope` only after pushed the update... will update in a few hours

dsp-ant: @Kludex would that work for any transport? the problem with the general RequestData is that it's:
1. Has no clear structure.
2. Might de-facto depend on HTTP specifics

I was wondering if we can make this a generic and actually pass the origianl request objects. This could be in gRPC context something different than Websocket or HTTP.

ihrpr: Honestly none of the options are great here. I'd love to use generics to make this properly typed, but that would break backwards compatibility for everyone using `RequestContext[SessionT, 
  LifespanContextT]` - they'd get errors about missing type arguments.

 So for now, i'd suggest to go with request.scope as a dict. It's not perfect but:
  - No breaking changes
  - No framework dependencies in core types
  - Gets us all the request data we need

For WebSocket, it'll work the same way - we can pass the WebSocket's ASGI scope which contains initial handshake headers etc

The scope structure is consistent across HTTP and WebSocket per the ASGI spec, so tools can handle both transports uniformly.

Not ideal, but it gets the job done without breaking existing code. I'd love to revisit this in sdk v2 though

Kludex: > Honestly none of the options are great here. I'd love to use generics to make this properly typed, but that would break backwards compatibility for everyone using RequestContext[SessionT,  LifespanContextT] - they'd get errors about missing type arguments.

The way to do for this is to set a default value on the `TypeVar` - so people don't get errors.

```py
from typing_extensions import TypeVar

T = TypeVar("T", default=...)`
```

I don't consider type changes as breaking changes btw.

ihrpr: ah, cool, thanks, didn't realise typing_extensions had TypeVar with default

bahattincinic:
When will you plan to release this PR @ihrpr ?

RRRajput:
Is there a guide on how this is supposed to be used?

RRRajput:
Found out how to use it:
```python

def get_authentication_headers(context: Context) -> dict[str, str]:
    """A function that gives the headers for the request.
    """
    request_object = context.request_context.request
    assert request_object is not None
    assert hasattr(request_object, "headers")
    headers: dict[str, str] = request_object.headers
    return headers
```

PR #821, opened 2025-05-27
RFC: Lower log levels used for connections

nickcoai:
It's not ideal, but currently some servers encode authorization in their URI. To reduce risk of such servers in other situations lower the log level of messages including the URI so we don't log potentially auth-containing data by default.

## Motivation and Context
As above really, it could be done by altering log levels but it's unclear this granularity is needed so instead proposing changing it.

## How Has This Been Tested?
Ran locally

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -53,7 +53,7 @@ async def sse_client(
 
     async with anyio.create_task_group() as tg:
         try:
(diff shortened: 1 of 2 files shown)


ihrpr (approved):
Thank you!

ugh, yeah, we need to implement proper URI sanitization, adding to the list. For now it's Okay to lower the level.

Issue #820, opened 2025-05-27
`tools/call` with string `arguments` breaks mcp server until restart


Sillocan:
**Describe the bug**

When making a `tools/call` JSON-RPC request to the mcp server, passing the `arguments` parameter as a string instead of an object causes the server to stop responding to all HTTP requests. The server does not return an informative error, and requires a restart to recover.

**This makes it possible for a single malformed client request to break the server for everyone.**

See the MRE at https://github.com/Sillocan/fastmcp-mre-breaking-task-group-with-invalid-arguments/blob/main/python_sdk/simple_echo_mre.py

**To Reproduce**
1. Run the following command in your shell:
   ```sh
   uv run --with=git+https://github.com/Sillocan/fastmcp-mre-breaking-task-group-with-invalid-arguments#subdirectory=python_sdk mre
   ```
2. Observe that a single malformed request (`arguments` is a string) returns 200, but further requests return 500 and the server must be restarted.
3. See error output below.

**Expected behavior**
- The server should return a clear error response (such as HTTP 400) when a request has an invalid format—specifically, when `arguments` is a string instead of a dict/object.
- The server should remain operational and handle all further requests correctly, even after processing an invalid or malformed request.

**Logs**

```sh
$ uv run --with=git+https://github.com/Sillocan/fastmcp-mre-breaking-task-group-with-invalid-arguments#subdirectory=python_sdk mre
      Built fastmcp-mre-breaking-task-group-with-invalid-arguments @ git+https://github.com/Sillocan/fastmcp-mre-breaking-task-group-with-invalid-arguments@3eeb959518a5eaa685c7619ebc7b20f5330f6235#subdirectory=python_sdk
Installed 32 packages in 5ms
[05/27/25 12:40:16] DEBUG    Using selector: EpollSelector                                                                                                                                                              selector_events.py:64
INFO:     Started server process [1783696]
INFO:     Waiting for application startup.
                    INFO     StreamableHTTP session manager started                                                                                                                                            streamable_http_manager.py:109
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
[05/27/25 12:40:17] DEBUG    connect_tcp.started host='127.0.0.1' port=8000 local_address=None timeout=5.0 socket_options=None                                                                                                   _trace.py:87
[...]

PR #822, opened 2025-05-27
Fix uncaught exception in MCP server

ddworken:
## Motivation and Context
This fixes an uncaught exception in the MCP server

## How Has This Been Tested?
Via unit tests and locally. 

## Breaking Changes
None. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -377,9 +377,9 @@ class BaseSession(
                     except Exception as e:
                         # For request validation errors, send a proper JSON-RPC error
                         # response instead of crashing the server
-                        logging.warning(
-                            f"Failed to validate request: {e}. "
-                            f"Message was: {message.message.root}"
+                        logging.warning(f"Failed to validate request: {e}")
+                        logging.debug(
+                            f"Message that failed validation: {message.message.root}"
                         )
                         error_response = JSONRPCError(
                             jsonrpc="2.0",
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -14,6 +14,7 @@ from typing_extensions import Self
 from mcp.shared.exceptions import McpError
 from mcp.shared.message import MessageMetadata, ServerMessageMetadata, SessionMessage
 from mcp.types import (
+    INVALID_PARAMS,
     CancelledNotification,
     ClientNotification,
     ClientRequest,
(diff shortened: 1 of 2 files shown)


src/mcp/shared/session.py
```diff
@@ -351,26 +352,47 @@ async def _receive_loop(self) -> None:
...
+                        await self._received_request(responder)
 
-                    if not responder._completed:  # type: ignore[reportPrivateUsage]
-                        await self._handle_incoming(responder)
+                        if not responder._completed:  # type: ignore[reportPrivateUsage]
+                            await self._handle_incoming(responder)
+                    except Exception as e:
+                        # For request validation errors, send a proper JSON-RPC error
+                        # response instead of crashing the server
+                        logging.warning(
```
johnw188: Should this be a warning? We should also probably put the actual message text in a logging.debug to avoid logging potential PII in orgs where that's a concern.

ddworken: My personal guess is that this should be a warning since this is a pretty malformed message. But if you'd prefer to move it to something lower (info?) that also SGTM. 

And done, moved the invalid message to logging.debug.

bruno-oliveira:
So to clarify, currently this is blocking from using the `streamable-http` transport inside clients such as Windsurf or Claude Desktop correct?

Sillocan:
@bruno-oliveira This shouldn't be blocking anyone, but it does fix a huge security issue with remote servers. Currently anyone can lock up a server just by sending a bad payload https://github.com/modelcontextprotocol/python-sdk/issues/820

NAVNAV221:
The [`except` part](https://github.com/modelcontextprotocol/python-sdk/pull/822/files#diff-aaa1596dd4db91f5cd7f743c8afbe58596daf01a2ecec4a0024768f7409fb1dcR398):
```python
self._write_stream.send(session_message)
```
write session_message back to the client.

Maybe there is a need to standarize the message back so mcp clients such as [mcp-remote](https://github.com/geelen/mcp-remote) can support that.

@ddworken

src/mcp/shared/session.py
```diff
@@ -354,27 +355,47 @@ async def _receive_loop(self) -> None:
...
+                            await self._handle_incoming(responder)
+                    except Exception as e:
+                        # For request validation errors, send a proper JSON-RPC error
+                        # response instead of crashing the server
+                        logging.warning(f"Failed to validate request: {e}")
+                        logging.debug(
+                            f"Message that failed validation: {message.message.root}"
+                        )
+                        error_response = JSONRPCError(
+                            jsonrpc="2.0",
```
NAVNAV221, line 388: Is it alwayes: `2.0`?

src/mcp/shared/session.py
```diff
@@ -354,27 +355,47 @@ async def _receive_loop(self) -> None:
...
+                            message_metadata=message.metadata,
+                        )
+                        self._in_flight[responder.request_id] = responder
+                        await self._received_request(responder)
 
-                    if not responder._completed:  # type: ignore[reportPrivateUsage]
-                        await self._handle_incoming(responder)
+                        if not responder._completed:  # type: ignore[reportPrivateUsage]
+                            await self._handle_incoming(responder)
+                    except Exception as e:
```
NAVNAV221, line 380: I think is better to handle a specific exceptions before the general one, such as `RuntimeError` (Which mentioned in [this issue](https://github.com/modelcontextprotocol/python-sdk/issues/820))

ddworken: Given that the risk is the server becoming unresponsive, I believe catching all exceptions to isolate errors to a single request is correct.

src/mcp/shared/session.py
ddworken: Yes, this matches the rest of the code in this file which is always `2.0`.

Sillocan:
I'm a bit baffled that a security issue like this is still open

johnw188 (approved):
LGTM

src/mcp/shared/session.py
```diff
@@ -354,27 +355,47 @@ async def _receive_loop(self) -> None:
...
+                            jsonrpc="2.0",
+                            id=message.message.root.id,
+                            error=ErrorData(
+                                code=INVALID_PARAMS,
+                                message="Invalid request parameters",
+                                data="",
+                            ),
+                        )
+                        session_message = SessionMessage(
+                            message=JSONRPCMessage(error_response))
```
Kludex, line 397: How this got merged?

src/mcp/shared/session.py
Kludex: I don't think this is a valid answer from the package's point of view.

It makes it considerably hard to maintain the package if everything is `except Exception`.

PR #833, opened 2025-05-28
Support for http request injection propagation in StreamableHttp

ihrpr:
Follow up from https://github.com/modelcontextprotocol/python-sdk/pull/816

Adding support for propagating Request Context in Streamable Http


bhosmer-ant (approved):
LGTM!

PR #841, opened 2025-05-29
Fix: Prevent session manager shutdown on individual session crash

soby:
Previously, an unhandled exception within a single MCP session's `MCPServer.run()` task could propagate to the `StreamableHTTPSessionManager`'s main task group. This would cause the entire task group to cancel, effectively shutting down the session manager and terminating all active sessions.

This commit addresses the issue by:
1. Wrapping the `self.app.run(...)` call within the `run_server` (for stateful requests) and `run_stateless_server` (for stateless requests) inner functions in `StreamableHTTPSessionManager` with a `try...except Exception` block.
2. Logging any caught exceptions along with the session ID (for stateful requests) to aid in debugging the crashed session.

This change ensures that if a single session encounters an unexpected error and crashes, it only affects that specific session. The `StreamableHTTPSessionManager` will continue to run, and other active sessions will remain operational. This significantly improves the robustness and availability of the server.

## Motivation and Context

Unhandled exceptions such as a network error would render the server unusable until restart

## How Has This Been Tested?

Yes, specifically by generating client disconnects and observing the server log the unhandled error but remain running and stable

## Breaking Changes

No breaking changes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -51,7 +51,6 @@ class StreamableHTTPSessionManager:
         json_response: Whether to use JSON responses instead of SSE streams
         stateless: If True, creates a completely fresh transport for each request
                    with no session tracking or state persistence between requests.
-
     """
 
     def __init__(
@@ -171,12 +170,15 @@ class StreamableHTTPSessionManager:
             async with http_transport.connect() as streams:
                 read_stream, write_stream = streams
                 task_status.started()
-                await self.app.run(
-                    read_stream,
-                    write_stream,
-                    self.app.create_initialization_options(),
-                    stateless=True,
-                )
+                try:
+                    await self.app.run(
+                        read_stream,
+                        write_stream,
+                        self.app.create_initialization_options(),
+                        stateless=True,
+                    )
+                except Exception as e:
+                    logger.error(f"Stateless session crashed: {e}", exc_info=True)
 
         # Assert task group is not None for type checking
         assert self._task_group is not None
@@ -235,12 +237,33 @@ class StreamableHTTPSessionManager:
                     async with http_transport.connect() as streams:
                         read_stream, write_stream = streams
                         task_status.started()
-                        await self.app.run(
-                            read_stream,
-                            write_stream,
-                            self.app.create_initialization_options(),
-                            stateless=False,  # Stateful mode
-                        )
+                        try:
+                            await self.app.run(
+                                read_stream,
+                                write_stream,
+                                self.app.create_initialization_options(),
+                                stateless=False,  # Stateful mode
+                            )
+                        except Exception as e:
+                            logger.error(
+                                f"Session {http_transport.mcp_session_id} crashed: {e}",
+                                exc_info=True,
+                            )
+                        finally:
+                            # Cleanup logic
+                            if (
+                                http_transport.mcp_session_id
+                                and http_transport.mcp_session_id
+                                in self._server_instances
+                            ):
+                                logger.info(
+                                    "Cleaning up crashed/terminated session "
+                                    f"{http_transport.mcp_session_id} from "
+                                    "active instances."
+                                )
+                                del self._server_instances[
+                                    http_transport.mcp_session_id
+                                ]
 
                 # Assert task group is not None for type checking
                 assert self._task_group is not None
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -1,9 +1,12 @@
 """Tests for StreamableHTTPSessionManager."""
 
+from unittest.mock import AsyncMock
+
 import anyio
 import pytest
 
 from mcp.server.lowlevel import Server
+from mcp.server.streamable_http import MCP_SESSION_ID_HEADER
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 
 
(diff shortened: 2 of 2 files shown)


soby:
To address a test failure, I updated the cleanup logic to only remove error sessions and not those that have been explicitly terminated. I see that the terminated sessions are retained and used for the 404 vs 400 return code logic. However, I do not see any place where those are ever removed so I imagine they just accumulate until server restart. It's not within the scope of this PR to address but that's likely not desirable behavior.

UYousafzai:
any update on this @soby? maybe they implemented this in some other issue perhaps?

soby:
@UYousafzai it's ready to merge IMO. I've been running it in a fork and it works as intended to keep the server up and stable. Hopefully it can make it into tomorrow's release. @ihrpr ?

src/mcp/server/streamable_http_manager.py
```diff
@@ -171,12 +170,15 @@ async def run_stateless_server(
             async with http_transport.connect() as streams:
                 read_stream, write_stream = streams
                 task_status.started()
-                await self.app.run(
-                    read_stream,
-                    write_stream,
-                    self.app.create_initialization_options(),
-                    stateless=True,
-                )
+                try:
```
NAVNAV221, line 175: Did we look into `.run` function to see how to handle the error there?

soby: @NAVNAV221 That should also happen, but there should be no scenarios in which a per-session error is allowed to destabilize the entire server until reboot. This is a catch-all to make sure that the server as a whole survives any errors where proper error handling was missed.

Sillocan:
Fairly certain this is a duplicate of https://github.com/modelcontextprotocol/python-sdk/issues/820 and https://github.com/modelcontextprotocol/python-sdk/pull/822 which is associated with a hackerone issue

soby:
@Sillocan Those are for handling of specific exception sources within a connection (what @NAVNAV221 was referencing). This PR implements catch-all exception handling for all stream/sessions such that a future unhandled exceptions within the stream/session are logged but do not corrupt the whole server until restarted

soby:
@NAVNAV221 Is this good to merge?

src/mcp/server/streamable_http_manager.py
```diff
@@ -233,12 +235,33 @@ async def run_server(*, task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORE
...
-                        )
+                        try:
+                            await self.app.run(
+                                read_stream,
+                                write_stream,
+                                self.app.create_initialization_options(),
+                                stateless=False,  # Stateful mode
+                            )
+                        except Exception as e:
+                            logger.warning(
```
bendavis78: Why are you using `log.warning` for the crash exception? Seems like it should be `log.error`?

soby: I originally logged it at error but it was lighting up my bug reporting system with uncaught exceptions related to trivial things like the client closing the tcp connection unexpectedly. "Be usable in the real world" has to take priority.

tests/server/test_streamable_http_manager.py
```diff
@@ -71,3 +74,124 @@ async def send(message):
...
+    mock_mcp_run = AsyncMock(return_value=None)
+    # This will be called by StreamableHTTPSessionManager's run_server -> self.app.run
+    app.run = mock_mcp_run
+
+    sent_messages = []
+
+    async def mock_send(message):
+        sent_messages.append(message)
+
+    scope = {"type": "http", "method": "POST", "path": "/mcp", "headers": []}
```
felixweinberger: I think you need the headers here to make CI pass:

```
    scope = {
        "type": "http",
        "method": "POST",
        "path": "/mcp",
        "headers": [(b"content-type", b"application/json")],
    }
```

tests/server/test_streamable_http_manager.py
```diff
@@ -71,3 +74,124 @@ async def send(message):
...
+
+    async def mock_send(message):
+        sent_messages.append(message)
+        # If an exception occurs, the transport might try to send an error response
+        # For this test, we mostly care that the session is established enough
+        # to get an ID
+        if message["type"] == "http.response.start" and message["status"] >= 500:
+            pass  # Expected if TestException propagates that far up the transport
+
+    scope = {"type": "http", "method": "POST", "path": "/mcp", "headers": []}
```
felixweinberger: I think you need the headers here to make CI pass:

```
    scope = {
        "type": "http",
        "method": "POST",
        "path": "/mcp",
        "headers": [(b"content-type", b"application/json")],
    }
```

src/mcp/server/streamable_http_manager.py
```diff
@@ -233,12 +235,33 @@ async def run_server(*, task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORE
...
+                                f"Session {http_transport.mcp_session_id} crashed: {e}",
+                                exc_info=True,
+                            )
+                        finally:
+                            # Only remove from instances if not terminated
+                            if (
+                                http_transport.mcp_session_id
+                                and http_transport.mcp_session_id in self._server_instances
+                                and not (
+                                    hasattr(http_transport, "_terminated") and http_transport._terminated  # pyright: ignore
```
felixweinberger: To avoid having to override the linter, should we have `http_transport` expose a public property for checking termination state rather than directly access a private property?

```
  # streamable_http.py
  @property
  def is_terminated(self) -> bool:
      """Check if this transport has been explicitly terminated."""
      return getattr(self, '_terminated', False)
      
  # here
  if (
    http_transport.mcp_session_id
    and http_transport.mcp_session_id in self._server_instances
    and not http_transport.is_terminated
  ):
```

tests/server/test_streamable_http_manager.py
```diff
@@ -71,3 +74,124 @@ async def send(message):
...
+    # It's possible handle_request itself might raise an error if the TestException
+    # isn't caught by the transport layer before propagating.
+    # The key is that the session manager's internal task for MCPServer.run
+    # encounters the exception.
+    try:
+        await manager.handle_request(scope, mock_receive, mock_send)
+    except TestException:
+        # This might be caught here if not handled by StreamableHTTPServerTransport's
+        # error handling
+        pass
```
felixweinberger: Do we actually need this try catch and explanation? I ran this test locally 50x without the Try/Catch and the test worked fine without it. Can we remove the try catch and comment?

When would the exception not be caught by the app.run try-catch you added in the manager?

felixweinberger (changes requested):
Thank you for working on this and adding comprehensive tests! `try-catching` the `app.runs()` makes sense to me to stop errors propagating to other concurrent requests.

For the changes to logging severity - did you see any specific errors that were flooding your terminal that we might be able to isolate?

I think we may miss important errors this way if we hide them. Could we instead do something like this:

```
      # Expected when client disconnects - this is normal
      logger.warning("Client disconnected from SSE stream")
  except asyncio.CancelledError:
      # Expected during shutdown
      logger.warning("SSE writer cancelled")
      raise  # Must re-raise cancellation!
  except Exception as e:
      # Unexpected error - this is a real problem!
      logger.exception(f"Unexpected error in SSE writer: {e}")
```

That would help us clearly see the difference between expected errors and actual errors that we should look into.

src/mcp/server/streamable_http_manager.py
```diff
@@ -173,12 +172,15 @@ async def run_stateless_server(*, task_status: TaskStatus[None] = anyio.TASK_STA
...
-                )
+                try:
+                    await self.app.run(
+                        read_stream,
+                        write_stream,
+                        self.app.create_initialization_options(),
+                        stateless=True,
+                    )
+                except Exception as e:
+                    logger.error(f"Stateless session crashed: {e}", exc_info=True)
```
ihrpr:
```suggestion
                    logger.exception("Stateless session crashed")
```

soby:
@ihrpr gtg

ihrpr (approved):
LGTM

Issue #547, opened 2025-04-21
MCP Server Hangs Indefinitely After KqueueSelector Log on macOS (Python 3.12, Both STDIO & SSE)


jlcases:
**Describe the bug**

When running an MCP server using `mcp-sdk` (both `FastMCP` and the low-level `Server` with `mcp.server.stdio.stdio_server`) on macOS with Python 3.12, the server process hangs indefinitely immediately after the `DEBUG - Using selector: KqueueSelector` log message appears. This happens during the `server.run()` or `mcp.run()` call.

The hang occurs regardless of whether the STDIO transport (`transport="stdio"`) or the default web/SSE transport is used. It also occurs even with a minimal server configuration with no custom tools, resources, or plugins registered.

**Environment:**

*   **OS:** macOS 15 Sequoia (Darwin 24.5.0)
*   **Python:** 3.12.9 (via Homebrew - *Please update if incorrect*)
*   **mcp-sdk Version:** 1.6.1.dev14+babb477 (Installed via `pip install -e .` from Git main branch using `mcp @ git+https://github.com/modelcontextprotocol/python-sdk.git` in `pyproject.toml` with `hatchling` and `allow-direct-references = true`)
*   **Key Dependencies:**
    *   uvicorn: 0.30.3
    *   starlette: 0.37.2
    *   anyio: 4.9.0

**To Reproduce**

1.  Ensure Python 3.12.9 and `mcp-sdk` (tested with version `1.6.1.dev14+babb477` installed from Git main branch) are installed in your environment on macOS.
2.  Save the following minimal STDIO server example (based on the SDK documentation) as `mcp_stdio_test.py`:
    ```python
    # From https://github.com/modelcontextprotocol/python-sdk README (Low-Level Server example)
    import mcp.server.stdio
    import mcp.types as types
    from mcp.server.lowlevel import NotificationOptions, Server
    from mcp.server.models import InitializationOptions
    import logging # Add logging
    import asyncio # Add asyncio

    # Configure logging
    logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')
    logger = logging.getLogger(__name__)


    # Create a server instance
    logger.debug("Creating Server instance...")
    server = Server("example-server")
    logger.debug("Server instance created.")


    @server.list_prompts()
    async def handle_list_prompts() -> list[types.Prompt]:
        logger.debug("Handling list_prompts request...")
        return [
            types.Prompt(
[...]

PR #765, opened 2025-05-21
Align stdio shutdown sequence with spec

davenpi:
Align the stdio shutdown sequence with the [MCP spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle#stdio): close the server’s input, wait for exit, send SIGTERM if needed, and SIGKILL as a last resort.

## Motivation and Context
The MCP spec recommends a careful shutdown sequence for stdio transport to avoid leaving orphaned or stuck server processes. This change follows those steps, making shutdowns more reliable and predictable.

## How Has This Been Tested?
All tests pass (except one unrelated skipped test). I reviewed the new shutdown procedure to match the spec and tested it in the standard test suite.

## Breaking Changes
None

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
None

PR #729, opened 2025-05-16
Subprocess kill win32

jingx8885:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #850, opened 2025-05-30
Fix child process kill error with npx based servers

surya-prakash-susarla:
## Motivation and Context
When running npx based servers in stdio mode, the npx process spawns a sub-process. During exit, the main process does not propagate the SIGTERM to the child process which leads to a failed kill signal and the stdin & stdout of the root process getting mangled.

The current change creates a new process group when launching the fork for the server and ensures that a kill signal is sent for the process group is the process does not shutdown cleanly. This would ensure that the process group (which includes all the child process) would be killed successfully to avoid any potential issues.

## How Has This Been Tested?
Yes, a simple script like this:
```
from fastmcp import Client
import asyncio

async def test():
    config = {
        "mcpServers": {
            "everything": {
                "command": "npx",
                "args": ["-y", "@modelcontextprotocol/server-everything"]
            }
        }
    }

    async with Client(config) as client:
      print("TOOlS: {tools}".format(tools=tools))

asyncio.run(test())
```
The above script will currently block the stdio of the parent server when run using as: `python test.py`
After the change the npx root & the node child process die succesfully allowing the control to flow back to the parent process ensuring a clean shutdown.

## Breaking Changes
No, this should not be a breaking change since the process group flag is ignored in non-POSIX systems, and the cleanup via process group functionality is only applied for POSIX style launches.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [ x ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ x ] My code follows the repository's style guidelines
- [ x ] New and existing tests pass locally
- [ x ] I have added appropriate error handling
- [ x ] I have added or updated documentation as needed

## Additional context
Please refer to the issue here (slightly different context): 
https://github.com/modelcontextprotocol/python-sdk/issues/547

### DIFF
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -1,4 +1,5 @@
 import os
+import signal
 import sys
 from contextlib import asynccontextmanager
 from pathlib import Path
@@ -190,6 +191,13 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                     await terminate_windows_process(process)
                 else:
                     process.terminate()
+
+                    try:
+                        with anyio.fail_after(5):
+                            await process.wait()
+                    except TimeoutError:
+                        pgid = os.getpgid(process.pid)
+                        os.killpg(pgid, signal.SIGKILL)
             except ProcessLookupError:
                 # Process already exited, which is fine
                 pass
@@ -230,7 +238,7 @@ async def _create_platform_compatible_process(
         process = await create_windows_process(command, args, env, errlog, cwd)
     else:
         process = await anyio.open_process(
-            [command, *args], env=env, stderr=errlog, cwd=cwd
+            [command, *args], env=env, stderr=errlog, cwd=cwd, start_new_session=True
         )
 
     return process


surya-prakash-susarla:
Failing test seems to be a streamable http test which was not modified in this change. Please let me know in case it's a side-effect of this change.

felixweinberger:
Thank you @surya-prakash-susarla for submitting this PR and the detailed information on reproduction of the issue!

I spent yesterday and today trying to unify all the different approaches and fixes we have pending in this process termination space at the moment, as there are several interrelated fixes that either conflict or depend on each other - specifically #555, #729, #765, and #850.

I've added your change to #1044 as a draft with you as a co-author + added extensive regression testing. Would you be OK with consolidating this change into #1044 for the comprehensive testing & process handling introduced there? Would love your feedback on the PR as well if you have time!

felixweinberger (changes requested):
Requesting changes for now to put this PR in your queue - please TAL at my comment above regarding #1044.

felixweinberger:
Closing this PR in favor of https://github.com/modelcontextprotocol/python-sdk/pull/1078

Please let me know if that does not resolve your issue here!

PR #851, opened 2025-05-30
normalize string ID to int in server messages for compatibility

wangxm345566462:
Some non-standard SSE servers may return numeric IDs as strings, even when the original request used an integer. This change ensures that message.root.id is always an integer to maintain consistency and avoid type issues in downstream processing.

Normalize `message.root.id` to integer if it is a numeric string returned by non-standard SSE servers.

## Motivation and Context

Some SSE servers return numeric IDs as strings, causing mismatches between request and response IDs. This leads to callbacks not being triggered and the client waiting indefinitely. The change improves compatibility and stability by normalizing the ID type.

## How Has This Been Tested?

Tested with mock SSE server responses containing string IDs; verified that messages with string numeric IDs are correctly converted to int and processed without errors.

## Breaking Changes

No breaking changes. This is a backward-compatible bug fix.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This fix addresses compatibility issues with SSE servers that do not strictly follow numeric ID typing conventions. It ensures the client logic for callback matching remains reliable.

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -98,6 +98,21 @@ async def sse_client(
                                             message = types.JSONRPCMessage.model_validate_json(  # noqa: E501
                                                 sse.data
                                             )
+                                            
+                                            # Normalize ID to int if it's a numeric string.
+                                            # Some non-standard SSE servers return numeric IDs as strings,
+                                            # even if the original request used an integer ID.
+                                            if isinstance(message.root, types.JSONRPCResponse):
+                                                msg_id = message.root.id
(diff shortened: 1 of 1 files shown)


src/mcp/types.py
```diff
@@ -34,7 +34,7 @@
 ProgressToken = str | int
 Cursor = str
 Role = Literal["user", "assistant"]
-RequestId = str | int
+RequestId = Annotated[int | str, Field(union_mode="left_to_right")]
```
Kludex, line 37: This is really the only change that is necessary.

It will try to validate `int` first, then `str`.

PR #856, opened 2025-05-30
Use 120 characters instead of 88

Kludex:
I brought this on our Discord server before, and we said to wait a bit to some PRs.

I think this will improve the readability given the many layers of context managers we have in the source code.


wangxm345566462:
Formatting affects too much code, brother，Although I am also puzzled about why it has to be so short, maybe the screen font is too large, hahaha

PR #869, opened 2025-06-02
Add `streamable_http_client` and deprecate old usage

Kludex:
The module is named `streamable_http_client`, it doesn't make sense for the function to be called `streamablehttp_client`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,12 @@
     - [Prompts](#prompts)
     - [Images](#images)
     - [Context](#context)
+    - [Authentication](#authentication)
   - [Running Your Server](#running-your-server)
     - [Development Mode](#development-mode)
     - [Claude Desktop Integration](#claude-desktop-integration)
     - [Direct Execution](#direct-execution)
+    - [Streamable HTTP Transport](#streamable-http-transport)
     - [Mounting to an Existing ASGI Server](#mounting-to-an-existing-asgi-server)
   - [Examples](#examples)
     - [Echo Server](#echo-server)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -30,17 +30,20 @@
     - [Prompts](#prompts)
     - [Images](#images)
     - [Context](#context)
+    - [Authentication](#authentication)
```
Kludex, line 33: My IDE added those... Which seems correct.

src/mcp/client/streamable_http.py
```diff
@@ -81,8 +82,8 @@ def __init__(
         self,
         url: str,
         headers: dict[str, Any] | None = None,
-        timeout: timedelta = timedelta(seconds=30),
-        sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
+        timeout: float | timedelta = 30,
+        sse_read_timeout: float | timedelta = 60 * 5,
```
Kludex, line 86: This is backwards compatible.

src/mcp/client/streamable_http.py
```diff
@@ -446,6 +460,52 @@ async def streamablehttp_client(
...
+    warnings.warn(
+        "`streamablehttp_client` is deprecated. Use `streamable_http_client` instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    async with streamable_http_client(
+        url,
+        headers,
+        timeout.total_seconds(),
+        sse_read_timeout.total_seconds(),
```
Kludex, line 478: We should use `total_seconds()`, not `seconds`. Only `seconds` was wrong.

tests/client/test_session_group.py
```diff
@@ -279,45 +280,42 @@ async def test_disconnect_non_existent_server(self):
...
                 StreamableHttpParameters(
                     url="http://test.com/stream", terminate_on_close=False
                 ),
-                "streamablehttp",
-                "mcp.client.session_group.streamablehttp_client",
+                "mcp.client.session_group.streamable_http_client",
             ),  # url, headers, timeout, sse_read_timeout, terminate_on_close
         ],
     )
     async def test_establish_session_parameterized(
```
Kludex, line 301: I've added type hints, and simplified this test.

This test shouldn't exist as is, it's very confusing, and full of mocking, but I just adapted it for the sake of my changes.

PR #878, opened 2025-06-03
Unit Test Fix - Update concurrency test duration threshold and logging for clarity

pandasanjay:
## Motivation and Context

I made two important changes to fix failing tests in the MCP Python SDK project:

1. **The Concurrency Test Fix**  
   **Original Problem:**  
   The concurrency test (`test_messages_are_executed_concurrently`) was failing because it had an overly optimistic timing expectation. The test was asserting that 20 concurrent tasks (10 tool calls and 10 resource reads), each with a 0.01-second sleep, would complete in under 0.06 seconds (`6 * _sleep_time_seconds`). However, the actual execution time was around 0.11 seconds.

   **Why the fix is needed:**  
   This test was meant to verify that operations happen concurrently, but the threshold was unrealistically strict. The assertion was failing because:
   - There's always some overhead in task scheduling and management in async systems
   - CI environments can have variable performance characteristics
   - Even on local development machines, exact timing can fluctuate based on system load

   **Solution:**  
   I increased the time threshold to a more realistic value (`12 * _sleep_time_seconds`) that still validates the concurrency behavior. The test still confirms that tasks run in parallel (since serial execution would take `20 * _sleep_time_seconds`), but it's now more tolerant of real-world execution environments.

2. **The Snapshot Test Fix**  
   **Original Problem:**  
   The `test_build_metadata` test was failing with a `UsageError: snapshot value should not change` error. The test was using inline-snapshot to compare OAuth metadata objects. The error specifically mentioned that the issuer URL value had changed from `https://auth.example.com/` to `https://auth.example.com/v1/mcp` for the `with-path-param` test case.

   **Why the fix is needed:**  
   Snapshot tests are useful for catching unintended changes, but in this case:
   - The test was parametrized to run with different URL examples
   - The snapshot appeared to be capturing values from one test run but failing on another
   - The URLs might be normalized differently in different scenarios
   Since the test is parametrized, it needs to handle different expected values

   **Solution:**  
   I replaced the rigid snapshot comparison with direct attribute-by-attribute assertions. This approach:
   - Still thoroughly validates all fields in the OAuth metadata object
   - Works correctly with all parametrized test cases
   - Is more resilient to minor formatting differences in URLs
   - Makes the test more maintainable as it's more explicit about what's being verified

Both of these changes maintain the original intent of the tests while making them more reliable, especially in CI environments where timing and exact object representation can vary.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -10,7 +10,6 @@ from urllib.parse import parse_qs, urlparse
 
 import httpx
 import pytest
-from inline_snapshot import snapshot
 from pydantic import AnyHttpUrl
 
 from mcp.client.auth import OAuthClientProvider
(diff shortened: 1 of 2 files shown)


ihrpr:
Thank you for your contribution! Sorry for the delayed PR reviews. test_188_concurrency was rewritten recently and test_auth is part of this PR https://github.com/modelcontextprotocol/python-sdk/pull/945 - we can continue the discussion there.

PR #879, opened 2025-06-03
relax validation

### LINKED ISSUE #814: OAuthMetadata as define is too restrivtive, hence produces a ValidationError
**Describe the bug**
The current implementation of OauthMetadata is too restrictive 

**To Reproduce**
1. Initialize an OAuth config as described in the [README](https://github.com/modelcontextprotocol/python-sdk/blob/6e418e62f9446e3204f4bf889efcd9e6df76aef8/README.md#oauth-authentication-for-clients)
2. Use the paypal MCP server (https://mcp.paypal.com/sse) as an SSE endpoint
3. Initialize the MCP server
4. Observe the logs producing pydantic incompatibility errors.

**Expected behavior**
I would expect the pedantic schema for OAuthMetadata to be less restrictive and hence it would not produce errors if the schema is a little bit off.

**Additional context**
```
ERROR:mcp.client.auth:Failed to discover OAuth metadata
  File ".../.venv/lib/python3.13/site-packages/mcp/client/auth.py", line 154, in _discover_oauth_metadata
    return OAuthMetadata.model_validate(metadata_json)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.13/site-packages/pydantic/main.py", line 705, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        obj, strict=strict, from_attributes=from_attributes, context=context, by_alias=by_alias, by_name=by_name
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
pydantic_core._pydantic_core.ValidationError: 2 validation errors for OAuthMetadata
token_endpoint_auth_methods_supported.0
  Input should be 'none' or 'client_secret_post' [type=literal_error, input_value='client_secret_basic', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/literal_error
code_challenge_methods_supported.0
  Input should be 'S256' [type=literal_error, input_value='plain', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/literal_error
```

dr3s:
- fixed broken tests on `main` due to async handling
- Relax auth validation to permit additional responses in list fields
- Correct token_type and normalize to align with spec

Fixes #814

## Motivation and Context
The validation prevented usage with 3rd party Oauth AS.  The pydantic models are both used for validating hosted Oauth endpoints and validating responses from 3rd party AS.

## How Has This Been Tested?
tested the simple auth client example against a remote MCP server with a 3rd party AS

## Breaking Changes
n/a

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import AnyHttpUrl, BaseModel, Field, field_validator
+from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_validator
 
 
 class OAuthToken(BaseModel):
@@ -41,7 +41,7 @@ class OAuthClientMetadata(BaseModel):
     for the full specification.
     """
 
-    redirect_uris: list[AnyHttpUrl] = Field(..., min_length=1)
+    redirect_uris: list[AnyUrl] = Field(..., min_length=1)
     # token_endpoint_auth_method: this implementation only supports none &
     # client_secret_post;
     # ie: we do not support client_secret_basic
@@ -80,7 +80,7 @@ class OAuthClientMetadata(BaseModel):
                 raise InvalidScopeError(f"Client was not registered with scope {scope}")
         return requested_scopes
 
-    def validate_redirect_uri(self, redirect_uri: AnyHttpUrl | None) -> AnyHttpUrl:
+    def validate_redirect_uri(self, redirect_uri: AnyUrl | None) -> AnyUrl:
         if redirect_uri is not None:
             # Validate redirect_uri against client's registered redirect URIs
             if redirect_uri not in self.redirect_uris:
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -11,7 +11,7 @@ from urllib.parse import parse_qs, urlparse
 import httpx
 import pytest
 from inline_snapshot import snapshot
-from pydantic import AnyHttpUrl
+from pydantic import AnyHttpUrl, AnyUrl
 
 from mcp.client.auth import OAuthClientProvider
 from mcp.server.auth.routes import build_metadata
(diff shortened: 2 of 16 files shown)


src/mcp/shared/auth.py
```diff
@@ -111,13 +120,13 @@ class OAuthMetadata(BaseModel):
     token_endpoint: AnyHttpUrl
     registration_endpoint: AnyHttpUrl | None = None
     scopes_supported: list[str] | None = None
-    response_types_supported: list[Literal["code"]] = ["code"]
+    response_types_supported: list[Literal["code"] | str] = ["code"]
     response_modes_supported: list[Literal["query", "fragment"]] | None = None
     grant_types_supported: (
-        list[Literal["authorization_code", "refresh_token"]] | None
+        list[Literal["authorization_code", "refresh_token"] | str] | None
```
dr3s: keeping the literals is redundant.  We can also make them the default instead.

pcarleton: Can you make these all `str`? An alternative would be to enumerate all the valid values from RFC 8414, but we want the client to parse without failure even if the server has something odd with its metadata.

dr3s: Sure I had done that but wanted to check because it's usage on the server. Do you want defaults?

I think most of them do not have enums and can accept any string in the spec.

dr3s: updated to be strings.  I started to put all acceptable values in the rfc but most of them have registries that can change

pcarleton (approved):

sakalys:
Thank you, sir

tests/client/test_auth.py
```diff
@@ -348,11 +350,13 @@ async def test_register_oauth_client_failure(self, oauth_provider):
                         None,
                     )
 
-    def test_has_valid_token_no_token(self, oauth_provider):
+    @pytest.mark.anyio
+    async def test_has_valid_token_no_token(self, oauth_provider):
```
Kludex, line 354: Why did you change those?

PR #896, opened 2025-06-06
Use `total_seconds` instead of `seconds` on timeout

Kludex:
Since we accept `timedelta`, we are ignoring the other _non-seconds_ fields, which shouldn't be the case.


samuelcolvin (approved):
Lgtm

Issue #668, opened 2025-05-08
Will not work if multiple MCP inspector windows are open


Tylersuard:
**Describe the bug**
Not really a bug, it's just been driving me crazy for the past hour and I figured out what the problem was.  I had multiple inspector windows open.
**To Reproduce**
1. uv run mcp dev simple_echo.py
2. Open inspector in a browser
3. Shut down the server using ctrl+c
4. uv run mcp dev simple_echo.py again
5. Observe the errors
**Expected behavior**
It would be nice if it did not do this.
**Screenshots**

![Image](https://github.com/user-attachments/assets/6f91a247-7fff-4edf-aa8f-78cfd52979a7)

**Desktop (please complete the following information):**
Windows 11, Firefox

Kludex:
Closing it as stale. If it's still an issue, please let me know so I can reopen it.

Issue #548, opened 2025-04-21
Use fastapi mount fastmcp and configure custome path but show 404


MarshallEriksen-shaomingyang:
self.fast_mcp= FastMCP(
                self.server_config.get("name", "MCP Server"), message_path="/mcp/messages"
            )
Fastapi.mount(path="/mcp", app=self.main_server.sse_app())

![Image](https://github.com/user-attachments/assets/ff2a831b-4a9a-4bda-999a-8bdbf74aa1e1)

zhangzp409:
I have encountered the same problem, please tell me how to solve it.

PR #898, opened 2025-06-07
feat: implement MCP-Protocol-Version header requirement for HTTP transport

felixweinberger:
## Summary

This PR implements the MCP-Protocol-Version header requirement as specified in modelcontextprotocol/modelcontextprotocol#548. The implementation ensures that:

- Clients extract and store the negotiated protocol version from the initialization response
- Clients include the `MCP-Protocol-Version` header in all subsequent HTTP requests after initialization
- Servers validate the presence and correctness of this header for non-initialization requests
- Servers return 400 Bad Request for missing or invalid protocol version headers

## High-Level Approach

The implementation follows a straightforward approach:

1. **Client-side (StreamableHTTP)**:
   - Extract the `protocolVersion` from the initialization response's result field
   - Store it in the transport instance
   - Include it as `MCP-Protocol-Version` header in all subsequent requests via `_prepare_request_headers()`

2. **Server-side (StreamableHTTP)**:
   - Add validation in `_validate_protocol_version()` that checks:
     - Header presence (returns 400 if missing)
     - Protocol version validity (returns 400 if unsupported)
   - Apply validation to all non-initialization requests in POST, GET, and DELETE handlers
   - Consolidate header validation logic in `_validate_request_headers()` for extensibility

## Spec Changes Addressed

This PR addresses the breaking changes introduced in modelcontextprotocol/modelcontextprotocol#548 (see DIFF.md):

- **Requirement**: "If using HTTP, the client **MUST** include the `MCP-Protocol-Version: <protocol-version>` HTTP header during any subsequent requests to the MCP server"
- **Error Handling**: "If the server receives a request with a missing, invalid, or unsupported MCP-Protocol-VERSION, it **MUST** respond with `400 Bad Request`"

Note: This implementation follows the stricter requirements from PR #548, not the backwards-compatible version proposed in PR #668.

## Testing

- Updated all existing tests to include the MCP-Protocol-Version header
- Added helper function to extract protocol version from SSE responses
- Added specific tests for protocol version validation
- All tests pass successfully

## Breaking Changes

This is a breaking change that requires both clients and servers to be updated. Clients that don't send the MCP-Protocol-Version header will receive 400 Bad Request errors from updated servers.


src/mcp/client/streamable_http.py
```diff
@@ -139,19 +143,36 @@ def _maybe_extract_session_id_from_response(
...
+        message: JSONRPCMessage,
+    ) -> None:
+        """Extract protocol version from initialization response message."""
+        if isinstance(message.root, JSONRPCResponse) and message.root.result:
+            # Check if result has protocolVersion field
+            result = message.root.result
+            if "protocolVersion" in result:
+                self.protocol_version = result["protocolVersion"]
+                logger.info(f"Negotiated protocol version: {self.protocol_version}")
+
```
felixweinberger: inspecting the response in this way feels a bit wrong on the transport, but it looks like we're "peeking" into the response in various parts of the transport already

it felt a lot cleaner than doing something like creating a callback, passing that back to the session and have the session jam the version back into the transport, but keen for thoughts here

tests/shared/test_streamable_http.py
```diff
@@ -914,9 +914,9 @@ async def message_handler(
                     assert str(notif.root.params.uri) == "http://test_resource/"
                     resource_update_found = True
 
-            assert (
-                resource_update_found
-            ), "ResourceUpdatedNotification not received via GET stream"
+            assert resource_update_found, (
+                "ResourceUpdatedNotification not received via GET stream"
+            )
```
felixweinberger: ruff

src/mcp/client/streamable_http.py
```diff
@@ -40,6 +40,7 @@
 GetSessionIdCallback = Callable[[], str | None]
 
 MCP_SESSION_ID = "mcp-session-id"
+MCP_PROTOCOL_VERSION = "MCP-Protocol-Version"
```
felixweinberger: capitalization on these headers is inconsistent across the codebase, should we align on `kebab-case-everywhere` for headers?

Kludex: Yes, we should.

Kludex:
```suggestion
MCP_PROTOCOL_VERSION = "mcp-protocol-version"
```

felixweinberger: #911

tests/shared/test_streamable_http.py
```diff
@@ -1425,3 +1476,125 @@ async def test_streamablehttp_request_context_isolation(
...
+            assert isinstance(tool_result.content[0], TextContent)
+            headers_data = json.loads(tool_result.content[0].text)
+
+            # Verify protocol version header is present
+            assert "mcp-protocol-version" in headers_data
+            assert headers_data[MCP_PROTOCOL_VERSION_HEADER] == negotiated_version
+
+
+def test_server_validates_protocol_version_header(basic_server, basic_server_url):
+    """Test that server returns 400 Bad Request version if header missing or invalid."""
```
ochafik: Nit: update comment here and in PR desc to remove mention of missing since default handled / tested below

tests/shared/test_streamable_http.py
```diff
@@ -1425,3 +1476,125 @@ async def test_streamablehttp_request_context_isolation(
...
+        headers={
+            "Accept": "application/json, text/event-stream",
+            "Content-Type": "application/json",
+        },
+        json=INIT_REQUEST,
+    )
+    assert init_response.status_code == 200
+    session_id = init_response.headers.get(MCP_SESSION_ID_HEADER)
+
+    # Test request without mcp-protocol-version header (backwards compatibility)
```
ochafik, line 1511: Wondering if we can/should test the "default to 2025-03-26" behaviour (now appreciate your earlier chat about returning the version as header, would make this easier). An indirect hacky way could be to send a tool call w/ [structuredContent](https://modelcontextprotocol.io/specification/draft/server/tools#structured-content) which was only added after that version?

src/mcp/client/streamable_http.py
```diff
@@ -139,19 +141,36 @@ def _maybe_extract_session_id_from_response(
...
 
+    def _maybe_extract_protocol_version_from_message(
+        self,
+        message: JSONRPCMessage,
+    ) -> None:
+        """Extract protocol version from initialization response message."""
+        if isinstance(message.root, JSONRPCResponse) and message.root.result:
+            # Check if result has protocolVersion field
+            result = message.root.result
+            if "protocolVersion" in result:
```
ochafik: Since this method is only called if is_initialization, we may want to log an error if protocolVersion isn't set.

src/mcp/client/streamable_http.py
```diff
@@ -139,19 +141,36 @@ def _maybe_extract_session_id_from_response(
             self.session_id = new_session_id
             logger.info(f"Received session ID: {self.session_id}")
 
+    def _maybe_extract_protocol_version_from_message(
```
ochafik, line 136: nit: _handle_initialization_response

src/mcp/server/streamable_http.py
```diff
@@ -732,6 +741,26 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
...
         return True
 
+    async def _validate_protocol_version(self, request: Request, send: Send) -> bool:
+        """Validate the protocol version header in the request."""
+        # Get the protocol version from the request headers
+        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
+
+        # If no protocol version provided, assume version 2025-03-26
+        if not protocol_version:
+            protocol_version = "2025-03-26"
```
ochafik: maybe define some IMPLICIT_PROTOCOL_VERSION?

src/mcp/server/streamable_http.py
```diff
@@ -732,6 +741,26 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
 
         return True
 
+    async def _validate_protocol_version(self, request: Request, send: Send) -> bool:
+        """Validate the protocol version header in the request."""
+        # Get the protocol version from the request headers
+        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
+
+        # If no protocol version provided, assume version 2025-03-26
+        if not protocol_version:
```
ochafik: nit: `if protocol_version is None` (or else empty protocol string will not be treated as invalid)

src/mcp/server/streamable_http.py
```diff
@@ -559,8 +560,9 @@ async def _handle_get_request(self, request: Request, send: Send) -> None:
             await response(request.scope, request.receive, send)
             return
 
-        if not await self._validate_session(request, send):
+        if not await self._validate_request_headers(request, send):
```
ochafik, line 518: Do we need to test this too? (and propagate the version in the headers sent by sse_client)

src/mcp/client/streamable_http.py
felixweinberger: ack, but keeping it consistent wtih `_maybe_extract_session_id_from_response()` above

src/mcp/server/streamable_http.py
felixweinberger: added DEFAULT_PROTOCOL_VERSION

tests/shared/test_streamable_http.py
felixweinberger: responding on the comment further down on `_validate_request_headers`

src/mcp/server/streamable_http.py
felixweinberger: This is actually an interesting point 🤔 while we have the headers for version in a few places, we don't actually seem to _do_ anything with it on the server as far as I can tell (at least in the Python SDK).

I don't think we modify function behavior based on protocol version anywhere, so I'm not actually sure where (if anywhere) we would propagate this information to?

If we had the server respond with a version header it would make sense to jam it into the response object, but because we don't I don't see where we should propagate it to 🤔 

Maybe @ihrpr or @bhosmer-ant have a thought here?

src/mcp/types.py
```diff
@@ -23,6 +23,7 @@
 """
 
 LATEST_PROTOCOL_VERSION = "2025-03-26"
+DEFAULT_PROTOCOL_VERSION = "2025-03-26"
```
ihrpr: I'd call it default negotiated version and add comment linking to the spec explaining why we need it

src/mcp/server/streamable_http.py
ihrpr: right now the SDK is supporting all features from all version in backwards compatible way, hence we do not use the negotiated version in features and just checking that serve and client negotiated the version successfully. 
Note, this will change in the future.

src/mcp/client/streamable_http.py
```diff
@@ -131,19 +135,38 @@ def _maybe_extract_session_id_from_response(
...
+        self,
+        message: JSONRPCMessage,
+    ) -> None:
+        """Extract protocol version from initialization response message."""
+        if isinstance(message.root, JSONRPCResponse) and message.root.result:
+            # Check if result has protocolVersion field
+            result = message.root.result
+            if "protocolVersion" in result:
```
ihrpr: there is a type for it `InitializeResult`

ihrpr (dismissed):
Thank you for working on this 🙏 

Please can you address a few small things and good to go

src/mcp/server/streamable_http.py
```diff
@@ -682,6 +692,26 @@ async def _validate_session(self, request: Request, send: Send) -> bool:
...
+        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
+
+        # If no protocol version provided, assume default version
+        if protocol_version is None:
+            protocol_version = DEFAULT_PROTOCOL_VERSION
+
+        # Check if the protocol version is supported
+        if protocol_version not in SUPPORTED_PROTOCOL_VERSIONS:
+            response = self._create_error_response(
+                f"Bad Request: Unsupported protocol version: {protocol_version}",
```
ihrpr: would be good to include which versions are supported in the message

src/mcp/client/streamable_http.py
felixweinberger: amended

src/mcp/server/streamable_http.py
felixweinberger: added

src/mcp/client/streamable_http.py
```diff
@@ -128,19 +133,39 @@ def _maybe_extract_session_id_from_response(
...
+    def _maybe_extract_protocol_version_from_message(
+        self,
+        message: JSONRPCMessage,
+    ) -> None:
+        """Extract protocol version from initialization response message."""
+        if isinstance(message.root, JSONRPCResponse) and message.root.result:
+            try:
+                # Parse the result as InitializeResult for type safety
+                init_result = InitializeResult.model_validate(message.root.result)
+                self.protocol_version = str(init_result.protocolVersion)
```
ihrpr, line 145: something unsettling about this 😅 technically it will work, but can we handle None?

ihrpr: wait, actually it will not work, as now we'll have "None" and will not set it to the default

ihrpr (changes requested):
please can you double check handling None?

src/mcp/client/streamable_http.py
felixweinberger: `protocolVersion` can only be `str | int`:

```py
class InitializeResult(Result):
    """After receiving an initialize request from the client, the server sends this."""

    protocolVersion: str | int
    """The version of the Model Context Protocol that the server wants to use."""
    capabilities: ServerCapabilities
    serverInfo: Implementation
    instructions: str | None = None
    """Instructions describing how to use the server and its features."""
```

Do you mean handling when `message.root.result === None` implying a broken server? if that happens the try-catch ensures we don't crash out, `self.protocol_version` just stays unset (this is on the client).

felixweinberger: We could also do something like this, where we set the `self.protocol_version = DEFAULT_NEGOTIATED_VERSION` but I would argue against this because we're on the _client_ here and that might mask server issues:

```py
try:
    # Parse the result as InitializeResult for type safety
    init_result = InitializeResult.model_validate(message.root.result)
    self.protocol_version = str(init_result.protocolVersion)
    logger.info(f"Negotiated protocol version: {self.protocol_version}")
except Exception as exc:
    # Assume the default version if parsing fails for any reason
    self.protocol_version = DEFAULT_NEGOTIATED_VERSION
    logger.warning(f"Failed to parse initialization response as InitializeResult: {exc}")
    logger.warning(f"Raw result: {message.root.result}")
```

If the server is returning incorrect responses I feel like client shouldn't be "hiding" that?

felixweinberger:
> please can you double check handling None?

Left some additional comments in the thread - not quite sure what the concern about `None` is, whether it's handling a broken server response or incorrect parsing?

The protocolVersion field on `InitializeResult` can't be `None` according to the types.

src/mcp/client/streamable_http.py
ihrpr: sorry, I mean just remove `str(init_result.protocolVersion)` and instead do 
```
self.protocol_version = str(init_result.protocolVersion) if init_result.protocolVersion is not None else None
```

Issue #827, opened 2025-05-28
SseServerTransport should better handle the quoting the post message endpoint.


pradeepvishwakarma:
Just pointing out the issue where it malformed the endpoint url. Can we send full endpoint instead of passing the route as some sse integration requires full endpoints. 

**_>> sse.py -> SseServerTransport -> connect_sse_**

from urllib.parse import quote
from uuid import uuid4

session_id = uuid4()
root_path = "https://example.com/api_prefix"
full_message_path_for_client = root_path.rstrip("/") + "/messages"
client_post_uri_data = (
    f"{quote(full_message_path_for_client)}?session_id={session_id.hex}"
)
print(client_post_uri_data)

Output :  **https%3A//example.com/api_prefix/messages?session_id=0bb8b48cc4d1447c992e6872f2668b05**
Expected :  **https://example.com/api_prefix/messages?session_id=0bb8b48cc4d1447c992e6872f2668b05**

AishwaryaKalloli:
Hi, I am trying to understand the problem, I understood how to fix the issue related to % symbol.
However, I was curious about how this scenario would be triggered where we would have the full_url including "https://example.com".

As per my understanding so far, `client_post_uri_data` would resolve to `"/api_prefix/messages?session_id=0bb8b48cc4d1447c992e6872f2668b05"` because `scope("root_path")` would be `"/api_prefix"`.


If you could give an example of how this scenario is triggered it would be helpful.

AishwaryaKalloli:
To add details:
[The client appends the original url to the server's endpoint event.](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/sse.py#L74) This allows for more secure connection.

pradeepvishwakarma:
Thanks Aishwarya !

For Context, we are working with MS Copilot and currently it has some limitation on how the mcp server needs to be configured. Per blog post : https://learn.microsoft.com/en-us/microsoft-copilot-studio/agent-extend-action-mcp

![Image](https://github.com/user-attachments/assets/2e0422df-73fb-4f79-b79c-329cac70cd48)

There are few known issues (full url and content type) and wanted to do a fix or patch to resolve that.

pradeepvishwakarma:
Closing as copilot not supports Streamable http. Thanks for support and updates.

PR #900, opened 2025-06-07
Doc update + validation in SseServerTransport + existing test fixes: addresses Issue: #827

AishwaryaKalloli:
1. Doc update to correctly reflect that `SseServerTransport` only supports partial url for `endpoint` + validating the same.
2. `test_messages_are_executed_concurrently`: Removed flakiness from the test.
3. `test_build_metadata`: Removed usage of snapshot in assert condition in the test. Using snapshot is failing the test even if the function itself is working. This was because of parameterisation. After some research online found that not using snapshot after creation would be a better way to handle this scenario.

## Motivation and Context
This relates to the issue reported here: https://github.com/modelcontextprotocol/python-sdk/issues/827
The doc string is slightly misleading as it mentioned that we can pass the absolute URL as well, which is not true in code implementation or best practices.

Fixed tests that were breaking when I run `uv run pytest`.

## How Has This Been Tested?
Existing test cases pass.

## Breaking Changes
No.

## Types of changes

- [x] Documentation update

## Checklist


tests/client/test_auth.py
```diff
@@ -968,8 +967,7 @@ def test_build_metadata(
         revocation_options=RevocationOptions(enabled=True),
     )
 
-    assert metadata == snapshot(
-        OAuthMetadata(
+    assert metadata == OAuthMetadata(
```
Kludex: What is the error you were having?

AishwaryaKalloli: when I run `uv run pytest` I get the following error:
image

As a fix I tried running:
`uv run pytest --inline-snapshot=fix tests/client/test_auth.py` but got the same error again, alternatively I tried using Is for parameterised variables, but that did not seem like fully correct either.

AishwaryaKalloli: Few observations
1. This test is passing in GitHub workflow
2. In local, when I delete __pycache__ and run `uv run pytest tests/client/test_auth.py` it passes
3. In local, when I run `uv run pytest` it fails again

Since there does not seem to be an issue in the workflow I have reverted this change, or alternatively we can create explicit snapshot for each parameter similar to mentioned [here](https://github.com/15r10nk/inline-snapshot/issues/65).

AishwaryaKalloli:
updated the PR
1. removed unrelated changes
2. added some more checks
3. added tests for the change
let me know if it looks good!

PR #908, opened 2025-06-09
Add .envrc to gitignore with example file

felixweinberger:
## Summary
- Added `.envrc` to `.gitignore`
- Created `.envrc.example` as a template

## Motivation
Providing an example file to help setup .venv automatically on `cd` with e.g. direnv.

## Changes
1. Updated `.gitignore` to exclude `.envrc` files
2. Added `.envrc.example` with placeholder environment variables

## Test plan
N/A

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,0 +1,3 @@
+source .venv/bin/activate
+export UV_INDEX=https://pypi.org/simple
+unset PYTEST_DISABLE_PLUGIN_AUTOLOAD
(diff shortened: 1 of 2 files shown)


Kludex:
This PR seems to come from an Anthropic member, so I'll leave this to you folks. 

I personally don't accept this kind of PR on the repositories I maintain because I believe the `.gitignore` should only contain what the repository itself builds/uses/generates that should be ignored. Any user specific files should be in their global `.gitignore`.

PR #924, opened 2025-06-10
Add error log for client stdio

pengwa:
When use stdio mode, the server module is provided by developers (like me). But sometimes, the module not only call FastMCP().run(), but also other code.  If the other code print something, then standard output will contains those text. 

When stdio client read the standard output of the server module, it will fail to parse the content in expected content. 

For example: 
```
[50] Add mon5_log_handler to original_logger fsspec.mapping{"jsonrpc":"2.0","id":0,"result":{"protocolVersion":"2025-03-26","capabilities":{"experimental":{},"prompts":{"listChanged":false},"resources":{"subscribe":false,"listChanged":false},"tools":{"listChanged":false}},"serverInfo":{"name":"draft_2025-06-10 13:37:18","version":"0.1.0"}}} """

```

`[50] Add mon5_log_handler to original_logger ` is the log I printed by mistake before calling FastMCP().run(). 

## Motivation and Context

It is really hard to finally locate the root cause, so we need a better logging showing such error clearly. 

## How Has This Been Tested?

Tested in a stdio mode. Let me know if you have any suggestions on what else test to be covered.

## Breaking Changes

NO!

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist


src/mcp/client/stdio/__init__.py
```diff
@@ -149,6 +152,9 @@ async def stdout_reader():
                         try:
                             message = types.JSONRPCMessage.model_validate_json(line)
                         except Exception as exc:
+                            logger.error(
+                                f"Failed to parse JSONRPC message from server: {line}, exception: {exc}",
+                            )
```
Kludex:
```suggestion
                            logger.exception("Failed to parse JSONRPC message from server")
```

Kludex (dismissed):

src/mcp/client/stdio/__init__.py
pengwa: Thanks @Kludex for the review. May I know your concerns on exposing the message content in the log? Without the content, it is still not easy to know things are broken

Kludex: The `exception` method already provides that information.

```py
import logging
import sys

# Create stdout handler
stdout_handler = logging.StreamHandler(sys.stdout)
stdout_handler.setLevel(logging.DEBUG)
logging.basicConfig(level=logging.DEBUG, handlers=[stdout_handler])

try:
    raise Exception('test')
except Exception:
    logger.exception('I can see both the exception and the line here.')
```

pengwa: Got it! Thanks @Kludex

pengwa:
Resolved format issue. Please take a further look. @Kludex. THanks!

src/mcp/client/stdio/__init__.py
```diff
@@ -24,6 +24,8 @@
 
 logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
+
 # Environment variables to inherit by default
```
felixweinberger, line 29: Looks like duplication - possibly because of merging in latest updates.

src/mcp/client/stdio/__init__.py
```diff
@@ -24,6 +24,8 @@
 
 logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
```
felixweinberger: Duplicate lines, can we remove this? Might be because of syncing changes from main into the branch, if so apologies!

felixweinberger (changes requested):
Hi @pengwa thanks for this contribution! And apologies for the time it took to get back to you on this.

Change looks good to me, let's just fix the duplicate logger instantiation - happy to get back to this quickly if you have bandwidth to make the update?

src/mcp/client/stdio/__init__.py
pengwa: Thanks! Just sync main and fix the duplicated logger issue. pls review.

pengwa:
> Hi @pengwa thanks for this contribution! And apologies for the time it took to get back to you on this.
> 
> Change looks good to me, let's just fix the duplicate logger instantiation - happy to get back to this quickly if you have bandwidth to make the update?

Thanks @felixweinberger , thanks for reminding me there is such a PR here. Just updated the code, feel free to review again.

PR #928, opened 2025-06-10
Add stateless mode support for STDIO transport

### LINKED ISSUE #912: Support the `stateless` flag for STDIO transport

For many other MCP implementations, especially those that predate the `initialize` and `notifications/initialized` handshake, it's convenient to be able to test or execute new tools with a shell, like this:

```
echo '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}' | uv run main.py
```

And similar commands. This isn't possible anymore - one has to include the initialization messages as well.

This also greatly raises the bar for writing toy clients for educational purposes, which is my main goal. When learning about the MCP protocol, the initialization parts aren't as interesting as `tools/list` and `tools/call` to folks that are learning about how LLMs consume tools. 

Streamable HTTP Transport supports a settings flag `stateless` which initializes the Session state to be Initialized:

https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/session.py#L95

```
        self._initialization_state = (
            InitializationState.Initialized
            if stateless
            else InitializationState.NotInitialized
        )
```

A similar flag for STDIO would be awesome. It's easily patched like this:

```
from mcp.server.stdio import stdio_server

class MyFastMCP(FastMCP):
    async def run_stdio_async(self) -> None:
        """Run the server using stdio transport."""
        async with stdio_server() as (read_stream, write_stream):
            await self._mcp_server.run(
                read_stream,
                write_stream,
                stateless=True
            )

# Initialize FastMCP server                                                                                                              
mcp = MyFastMCP("mymcp")
```

which is sufficient for testing the flag itself. It would be nice if I could pass this flag somehow. For HTTP it's done with a Setting: https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/server.py#L100

And that would be fine here as well, but it's not plumbed through.

None.

**Additional context**

I think the diff here is pretty straightforward. I'm filing this because `CONTRIBUTING.md` says:

```
Choose the correct branch for your changes:
    For bug fixes to a released version: use the latest release branch (e.g. v1.1.x for 1.1.3)
    For new features: use the main branch (which will become the next minor/major version)
    If unsure, ask in an issue first
```

I'm unsure, so I'm asking here first.

efj-amzn:
## Motivation and Context

This change adds support for the `stateless` flag to STDIO transport, addressing issue #912. Currently, STDIO transport requires a full initialization handshake, making simple command line interactions like `echo '{"jsonrpc":"2.0",...}' | python script.py` impossible.

The stateless flag already exists for HTTP transport but was missing for STDIO. This feature is particularly helpful for:
- Simple CLI debugging and testing
- Support for older MCP clients that don't do the handshake
- Educational scenarios where users are learning about MCP without needing to understand the initialization protocol
- Quick command-line tool interactions

## How Has This Been Tested?

Aside from the added tests, I created a basic server and ran some commands. 

```

# Create FastMCP server
mcp = FastMCP("StatelessTest")

# Register a simple echo tool
@mcp.tool()
def echo(message: str) -> str:
    """Echo a message back to the client."""
    return f"Echo: {message}"

if __name__ == "__main__":
    # Run in STDIO mode
    mcp.run()
```

```
% echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"echo","arguments":{"message":"Hello world"}}}' | FASTMCP_STATELESS_STDIO=true uv run test_stateless.py
[06/10/25 14:52:17] INFO     Processing request of type CallToolRequest                                                                                                                                                                                                                                                                          server.py:561
{"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"Echo: Hello world"}],"isError":false}}

% echo '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"echo","arguments":{"message":"Hello world"}}}' | FASTMCP_STATELESS_STDIO=false uv run test_stateless.py
<runtime exception>
```

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

I forked from gitish 1a9ead07f5319abf941b54668b911e3f3773a1cb and the tests occasionally hang on that commit.

Fixes #912

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -87,6 +87,9 @@ class Settings(BaseSettings, Generic[LifespanResultT]):
     debug: bool = False
     log_level: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "INFO"
 
+    # STDIO settings
+    stateless_stdio: bool = False
+
     # HTTP settings
     host: str = "127.0.0.1"
     port: int = 8000
@@ -597,6 +600,7 @@ class FastMCP:
                 read_stream,
                 write_stream,
                 self._mcp_server.create_initialization_options(),
+                stateless=self.settings.stateless_stdio
             )
 
     async def run_sse_async(self, mount_path: str | None = None) -> None:
--- a/tests/server/test_session.py
+++ b/tests/server/test_session.py
@@ -200,3 +200,105 @@ async def test_server_session_initialize_with_older_protocol_version():
 
     assert received_initialized
     assert received_protocol_version == "2024-11-05"
+
+
+@pytest.mark.anyio
+async def test_server_session_requires_initialization():
+    """Test that ServerSession requires initialization before accepting requests."""
+    server_to_client_send, server_to_client_receive = anyio.create_memory_object_stream[
+        SessionMessage
+    ](1)
+    client_to_server_send, client_to_server_receive = anyio.create_memory_object_stream[
+        SessionMessage | Exception
+    ](1)
+
+    try:
+        init_options = InitializationOptions(
+            server_name="TestServer",
+            server_version="1.0",
+            capabilities=ServerCapabilities(),
+        )
+
+        async with ServerSession(
+            client_to_server_receive,
+            server_to_client_send,
+            init_options,
+            stateless=False,
+        ) as server_session:
+            request = types.ClientRequest(
+                root=types.CallToolRequest(
+                    method="tools/call",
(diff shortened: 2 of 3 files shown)


ihrpr:
Thank you for contributing to the MCP Python SDK!

Statelessness only applies to the StreamableHttp transport, as the protocol itself is stateful and requires initialization. With StreamableHttp, the client must still initialize with one node before sending requests to any other nodes.
Given the protocol's emphasis on proper initialization handshakes, I don't believe introducing this complexity to STDIO is justified for testing purposes.

Issue #936, opened 2025-06-11
Inconsistent timeout and sse_read_timeout Types in sse_client and streamablehttp_client


yinglj:
# Inconsistent `timeout` and `sse_read_timeout` Types in `sse_client` and `streamablehttp_client`

## Description

In the `modelcontextprotocol/python-sdk` repository (version 1.9.3), the `timeout` and `sse_read_timeout` parameters in `mcp/client/sse.py` and `mcp/client/streamable_http.py` have inconsistent types, causing confusion and runtime errors in downstream applications:

- `mcp/client/sse.py` (`sse_client`):
  ```python
  @asynccontextmanager
  async def sse_client(
      url: str,
      headers: dict[str, Any] | None = None,
      timeout: float = 5,
      sse_read_timeout: float = 60 * 5,
      ...
  )
  ```
  - `timeout` and `sse_read_timeout` are typed as `float` (seconds).

- `mcp/client/streamable_http.py` (`streamablehttp_client`):
  ```python
  @asynccontextmanager
  async def streamablehttp_client(
      url: str,
      headers: dict[str, Any] | None = None,
      timeout: timedelta = timedelta(seconds=30),
      sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
      ...
  )
  ```
  - `timeout` and `sse_read_timeout` are typed as `timedelta`.

This inconsistency forces developers to handle different types for similar parameters across MCP transport implementations, leading to errors. For example, in the `openai-agents-python` project, passing a `float` timeout to `MCPServerStreamableHttp` (which uses `streamablehttp_client`) causes a `TaskGroup` exception:

```
2025-06-11 19:17:20,478 server.py:132 - mcp_client - [MainThread:5432] - ERROR - Error initializing MCP server: unhandled errors in a TaskGroup (1 sub-exception)
```

The lack of clear error messages from `streamablehttp_client` (e.g., type mismatch) further complicates debugging.

## Steps to Reproduce

1. Configure an `MCPServerStreamableHttp` (from `openai-agents-python`) with a `float` timeout (e.g., `timeout=30`).
2. Call `server.connect()`, which invokes `streamablehttp_client`.
3. Observe a `TaskGroup` exception due to `timedelta` type mismatch.
4. Compare with `MCPServerSse`, which accepts `float` without issues.

## Expected Behavior

[...]

PR #941, opened 2025-06-12
Allow to pass timeout as float

Kludex:
I personally think that `timedelta` adds an unnecessary complexity, and is inconsistent with the SSE implementation. Even if it's considered the "right type", I don't think those timeouts should be used with any other unit than seconds/milliseconds.

I'm implementing it like this because it is the least breaking change, and @dsp-ant said he'd prefer it.

---

- Closes https://github.com/modelcontextprotocol/python-sdk/issues/936

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -11,7 +11,6 @@ from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Any
 
 import anyio
 import httpx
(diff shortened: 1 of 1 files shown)


src/mcp/client/streamable_http.py
```diff
@@ -80,9 +75,9 @@ class StreamableHTTPTransport:
     def __init__(
         self,
         url: str,
-        headers: dict[str, Any] | None = None,
-        timeout: timedelta = timedelta(seconds=30),
-        sse_read_timeout: timedelta = timedelta(seconds=60 * 5),
+        headers: dict[str, str] | None = None,
```
Kludex, line 78: The values are not `Any`, they are `str`.

PR #947, opened 2025-06-12
Rename ResourceReference to ResourceTemplateReference

### LINKED ISSUE #916: Rename ResourceReference to ResourceTemplateReference

ihrpr:
Closes #916 

Details in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/639


Kludex:
Those are breaking changes - fyi

ihrpr:
> Those are breaking changes - fyi

@Kludex , have type alias for this in case anyone is using it directly, any other use case is missing?
```
# Deprecated: Use ResourceTemplateReference instead
ResourceReference = ResourceTemplateReference
```

Kludex:
> > Those are breaking changes - fyi
> 
> @Kludex , have type alias for this in case anyone is using it directly, any other use case is missing?
> 
> ```
> # Deprecated: Use ResourceTemplateReference instead
> ResourceReference = ResourceTemplateReference
> ```

ah, then sorry, all good.

ihrpr:
> > > Those are breaking changes - fyi
> > 
> > 
> > @Kludex , have type alias for this in case anyone is using it directly, any other use case is missing?
> > ```
> > # Deprecated: Use ResourceTemplateReference instead
> > ResourceReference = ResourceTemplateReference
> > ```
> 
> ah, then sorry, all good.

thanks for checking!!

src/mcp/types.py
```diff
@@ -962,6 +962,10 @@ class ResourceReference(BaseModel):
     model_config = ConfigDict(extra="allow")
 
 
+# Deprecated: Use ResourceTemplateReference instead
+ResourceReference = ResourceTemplateReference
```
Kludex: You want something like this:

```suggestion
@deprecated("`ResourceReference` is deprecated, you should use `ResourceTemplateReference`.")
class ResourceReference(ResourceTemplateReference):
    """A reference to a resource or resource template definition."""
```

ihrpr: ah, thank you!

Kludex (approved):

PR #948, opened 2025-06-12
Fix simple-auth server example AuthorizationCode.redirect_uri type

basoko:
The AuthorizationCode.redirect_uri field type is AnyUrl, but the example used AnyHttpUrl type.

This mismatch causes the redirect URI check in TokenHandler to fail, since it compares against AuthorizationCodeRequest.redirect_uri, which also expects AnyUrl.

## Motivation and Context
It makes the simple-auth server example to work with simple-auth client example. Without the change when client requests to get the token it fails with a "Bad Request" error.

## How Has This Been Tested?
Manually by running server and client examples locally.

## Breaking Changes
No.

## Types of changes

- [ ] Documentation update

## Checklist

- [] I have added appropriate error handling
- [] I have added or updated documentation as needed

### DIFF
--- a/examples/servers/simple-auth/mcp_simple_auth/server.py
+++ b/examples/servers/simple-auth/mcp_simple_auth/server.py
@@ -6,7 +6,7 @@ import time
 from typing import Any, Literal
 
 import click
-from pydantic import AnyHttpUrl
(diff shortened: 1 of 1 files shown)


pcarleton:
this part of the example has been simplified now

PR #966, opened 2025-06-16
Include context into completions

ihrpr:
Details: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/598
Closes: https://github.com/modelcontextprotocol/python-sdk/issues/917

--- 

**Implementation**:
  - Added `CompletionContext` type with optional arguments field to store previously resolved parameters
  - Updated `CompleteRequestParams` to include optional context parameter
  - Modified client's `complete()` method to accept context_arguments parameter
  - Updated server completion handler signature to receive the context

  **API Improvements:**
  - Added `completion()` decorator method to FastMCP for cleaner API (lowlevel already had it)

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -364,6 +364,24 @@ class FastMCP:
 
         return decorator
 
+    def completion(self):
+        """Decorator to register a completion handler.
+
+        The completion handler receives:
+        - ref: PromptReference or ResourceTemplateReference
+        - argument: CompletionArgument with name and partial value
+        - context: Optional CompletionContext with previously resolved arguments
+
+        Example:
+            @mcp.completion()
+            async def handle_completion(ref, argument, context):
+                if isinstance(ref, ResourceTemplateReference):
+                    # Return completions based on ref, argument, and context
+                    return Completion(values=["option1", "option2"])
+                return None
+        """
+        return self._mcp_server.completion()
+
     def add_resource(self, resource: Resource) -> None:
         """Add a resource to the server.
 
--- a/tests/server/test_completion_with_context.py
+++ b/tests/server/test_completion_with_context.py
@@ -1,0 +1,133 @@
+"""
+Tests for completion handler with context functionality.
+"""
+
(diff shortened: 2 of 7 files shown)


tests/server/test_completion_with_context.py
```diff
@@ -0,0 +1,133 @@
...
+    ) -> Completion | None:
+        # Simulate database/table completion scenario
+        if isinstance(ref, ResourceTemplateReference):
+            if ref.uri == "db://{database}/{table}":
+                if argument.name == "database":
+                    # Complete database names
+                    return Completion(values=["users_db", "products_db", "analytics_db"], total=3, hasMore=False)
+                elif argument.name == "table":
+                    # Complete table names based on selected database
+                    if context and context.arguments:
```
bhosmer-ant, line 101: maybe overpolishing, but do we want a test where the server errors on not enough context (e.g. "need a database to complete tables" and the client receives it?

src/mcp/server/fastmcp/server.py
```diff
@@ -364,6 +364,24 @@ def decorator(fn: AnyFunction) -> AnyFunction:
 
         return decorator
 
+    def completion(self):
```
bhosmer-ant, line 367: Is it weird to route all completions to the same function in the high level API? Alternative would be to defer adding the fastmcp decorator for now and do something fancier that routes on URI, but agree that it's nice to bring fastmcp decorator coverage to parity with the lowlevel server. 

Also a) maybe it's natural for "real" servers to route all completion requests through the same handler, and b) we could always add new decorators (`resource_completion`, `prompt_completion`) later.

ihrpr: yeah, it is quite weird even on the lowlevel, TS has completions handling on a field that needs to be completed, which I found quite nice. 

I'd love to make it nicer, but for now to have at least a way to handle completions on FastMCP level and not do to things like `_server`.

tests/server/test_completion_with_context.py
ihrpr: thank you, added!

PR #974, opened 2025-06-17
Add resource Link

ihrpr:
Details: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/603
Closes https://github.com/modelcontextprotocol/python-sdk/issues/944

------- 
- Added resource link to the tool output types
- Added a test to integrations tests

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -733,14 +733,27 @@ class EmbeddedResource(BaseModel):
     model_config = ConfigDict(extra="allow")
 
 
-Content = TextContent | ImageContent | AudioContent | EmbeddedResource
+class ResourceLink(Resource):
+    """
+    A resource that the server is capable of reading, included in a prompt or tool call result.
+
+    Note: resource links returned by tools are not guaranteed to appear in the results of `resources/list` requests.
+    """
+
+    type: Literal["resource_link"]
+
+
+ContentBlock = TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource
+"""A content block that can be used in prompts and tool results."""
+
+Content = ContentBlock
 
 
 class PromptMessage(BaseModel):
     """Describes a message returned as part of a prompt."""
 
     role: Role
-    content: Content
+    content: ContentBlock
     model_config = ConfigDict(extra="allow")
 
 
@@ -859,7 +872,7 @@ class CallToolRequest(Request[CallToolRequestParams, Literal["tools/call"]]):
 class CallToolResult(Result):
     """The server's response to a tool call."""
 
-    content: list[Content]
+    content: list[ContentBlock]
     isError: bool = False
 
 
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -746,7 +746,8 @@ class ResourceLink(Resource):
 ContentBlock = TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource
 """A content block that can be used in prompts and tool results."""
 
-Content = ContentBlock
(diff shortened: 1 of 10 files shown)


src/mcp/types.py
```diff
@@ -733,14 +733,27 @@ class EmbeddedResource(BaseModel):
...
+    Note: resource links returned by tools are not guaranteed to appear in the results of `resources/list` requests.
+    """
+
+    type: Literal["resource_link"]
+
+
+ContentBlock = TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource
+"""A content block that can be used in prompts and tool results."""
+
+Content = ContentBlock
```
bhosmer-ant: Assuming this is to reduce blast radius for the PR, but if so, should we maybe leave it named `Content` and then do the `Content` -> `ContentBlock` change in a followup, rather than having both names floating around?

src/mcp/types.py
```diff
@@ -733,14 +733,27 @@ class EmbeddedResource(BaseModel):
...
 
 
 class PromptMessage(BaseModel):
     """Describes a message returned as part of a prompt."""
 
     role: Role
-    content: Content
+    content: ContentBlock
```
bhosmer-ant, line 757: per above comment

src/mcp/types.py
```diff
@@ -859,7 +872,7 @@ class CallToolRequest(Request[CallToolRequestParams, Literal["tools/call"]]):
 class CallToolResult(Result):
     """The server's response to a tool call."""
 
-    content: list[Content]
+    content: list[ContentBlock]
```
bhosmer-ant, line 876: per above comment

bhosmer-ant (approved):
Inline comment re `Content`/`ContentBlock`, but your call

src/mcp/types.py
ihrpr: need to add deprecated!

src/mcp/types.py
```diff
@@ -733,14 +733,28 @@ class EmbeddedResource(BaseModel):
...
+    """
+
+    type: Literal["resource_link"]
+
+
+ContentBlock = TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource
+"""A content block that can be used in prompts and tool results."""
+
+Content: TypeAlias = ContentBlock
+# """DEPRECATED: Content is deprecated, you should use ContentBlock directly."""
```
ihrpr, line 750: leaving content just in case someone uses it from SDK and not to break them

bhosmer-ant (approved):
LGTM!

bhosmer-ant:
test failure looks like a linux flake?
```
[gw3] linux -- Python 3.11.12 /home/runner/work/python-sdk/python-sdk/.venv/bin/python

    @pytest.mark.anyio
    async def test_messages_are_executed_concurrently():
        server = FastMCP("test")
    
        @server.tool("sleep")
        async def sleep_tool():
            await anyio.sleep(_sleep_time_seconds)
            return "done"
    
        @server.resource(_resource_name)
        async def slow_resource():
            await anyio.sleep(_sleep_time_seconds)
            return "slow"
    
        async with create_session(server._mcp_server) as client_session:
            start_time = anyio.current_time()
                for _ in range(10):
                    tg.start_soon(client_session.call_tool, "sleep")
                    tg.start_soon(client_session.read_resource, AnyUrl(_resource_name))
    
            end_time = anyio.current_time()
    
            duration = end_time - start_time
>           assert duration < 10 * _sleep_time_seconds
E           assert 0.10288924399999644 < (10 * 0.01)
```

PR #979, opened 2025-06-18
Handle '&scope=' in authorize request to return empty list

### LINKED ISSUE #977: Empty but defined scope query param ("&scope=") validates as [""] and not [] during authorization.
### Initial Checks

### Description

Hello, we're trying to get MCP server working with auth in ChatGPT, and I believe I encountered a small bug. OpenAI sends an auth request with `&scope=`. This seems to imply that it is requesting no scopes.

However, in `mcp/shared/auth.py`, the code fails with the error message `Client+was+not+registered+with+scope+` because `requested_scope.split(" ")` parses as `[""]`, and so it checks that `""` is an allowed scope.

The code causing the error is as follows:

```python
    def validate_scope(self, requested_scope: str | None) -> list[str] | None:
        if requested_scope is None:
            return None
            if scope not in allowed_scopes:
                raise InvalidScopeError(f"Client was not registered with scope {scope}")
        return requested_scopes
```

It seems that the code should probably be changed to look something like this:

```python
    def validate_scope(self, requested_scope: str | None) -> list[str] | None:
        if requested_scope is None:
            return None
        if requested_scope == "":
            return []
            if scope not in allowed_scopes:
                raise InvalidScopeError(f"Client was not registered with scope {scope}")
        return requested_scopes
```

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Python version 3.12.7

`mcp==1.9.4`
```

dwreeves:
Resolves #977

## Motivation and Context

I am attempting to authorize to the ChatGPT web client using Okta OAuth. I got it working, but with significant rewrites and patching to the internals of the MCP Python SDK.

I'm not comfortable proposing support for everything I needed to do as many of those are client side issues with OpenAI's implementation, including lack of PKCE support. However, one of the things that tripped me up seems to be a minor issue on the MCP Python SDK's side, which is that the ChatGPT client was requesting `&scope=`. I'm not 100% sure what's going on as I give the client a default scope when it registers that it in turn is not requesting.

In any case, if `None` is a valid value for the scopes, then it seems that `[]` should also be, and furthermore that `&scope=` in the query params should correspond to either `None` or `[]`. However, right now the MCP SDK only sets the scopes to `None` if the query parameter is not defined at all, whereas the ChatGPT web client sends `&scope=`, which parses as `""`, which in turn fails validation unless `""` is defined as a valid scope in the `ClientRegistrationOptions`.

## How Has This Been Tested?
I have a patched version of this running and connecting to ChatGPT's web client.

## Breaking Changes
Should be no breaking changes.

## Types of changes
- [ ] Documentation update

## Checklist
- [x] My code follows the repository's style guidelines **(pre-commit run -a)**
- [x] I have added or updated documentation as needed **(n/a)**


dwreeves:
@felixweinberger I'm not actually if this is still an issue with OpenAI's implementation of OAuth-secured MCP! They've changed it quite a bit since I wrote this PR, including adding PKCE support.

We got it working exactly two times under very specific situations, but I gave up on it after too many issues, all of which seemed to be on their end (including random 500s) since we had no issues on Anthropic or running locally.

I still think it is a sensible behavior to not interpret `scope=` as `""` but it is unclear to me today if it fixes any extant issues with OpenAI.

I think I can get this working on Friday and see why the PR is failing; I'm busy today and tomorrow.

felixweinberger:
> @felixweinberger I'm not actually if this is still an issue with OpenAI's implementation of OAuth-secured MCP! They've changed it quite a bit since I wrote this PR, including adding PKCE support.
> 
> We got it working exactly two times under very specific situations, but I gave up on it after too many issues, all of which seemed to be on their end (including random 500s) since we had no issues on Anthropic or running locally.
> 
> I still think it is a sensible behavior to not interpret `scope=` as `""` but it is unclear to me today if it fixes any extant issues with OpenAI.
> 
> I think I can get this working on Friday and see why the PR is failing; I'm busy today and tomorrow.

Gotcha that would be great thank you - agreed that it still makes sense to handle this case explicitly.

dwreeves:
Actually, my months of intermittent testing is all coming back to me. OpenAI _does_ now support user scopes in their OAuth implementation, so they will return e.g. `&scope=user` and not just an empty param. So, this is definitely far less necessary.

felixweinberger:
> Actually, my months of intermittent testing is all coming back to me. OpenAI _does_ now support user scopes in their OAuth implementation, so they will return e.g. `&scope=user` and not just an empty param. So, this is definitely far less necessary.

Coming back to this, sounds like this is no longer a problem? To manage our inbound load I'm closing this one for now, but feel free to ping here if it's something you'd want to come back to.

Thanks again for the contribution!

etiennecaldichouryatnapta:
Hi team,
Seems like we have the exact same issue when adding MCP tool via Copilot: scope in URL is empty and thus we get "Client was not registered with scope" error

PR #981, opened 2025-06-18
[feat] Add `user` to `Context`

dwreeves:
## Motivation and Context

Accessing the `ctx.request_context.request.user` is a simple way to retrieve the scopes available to a user (as well as access token and client ID, albeit client ID is already available as an attr to the Context class), which is useful for some request handling. E.g. in some MCP clients (OpenAI's DeepResearch MCP implementations), scopes cannot necessarily be handled at the middleware level because everything needs to hit a specific `fetch` tool, so scopes need to be handled within the request callback. This means scopes need to be available within that callback. Adding a `user` property to the ctx makes it easier for a developer to discern where they might get access to the user's scopes.

## How Has This Been Tested?
I have been able to access the `user` attribute via `ctx.request_context.request.user` in my own code.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -27,6 +27,7 @@ from starlette.types import Receive, Scope, Send
 
 from mcp.server.auth.middleware.auth_context import AuthContextMiddleware
 from mcp.server.auth.middleware.bearer_auth import (
+    AuthenticatedUser,
     BearerAuthBackend,
     RequireAuthMiddleware,
 )
@@ -1046,6 +1047,11 @@ class Context(BaseModel, Generic[ServerSessionT, LifespanContextT, RequestT]):
             logger=logger_name,
             related_request_id=self.request_id,
         )
+
+    @property
+    def user(self) -> AuthenticatedUser | None:
+        """Get the authenticated user if available."""
+        return self.request_context.request.user if isinstance(self.request_context.request, Request) else None
 
     @property
     def client_id(self) -> str | None:


src/mcp/server/fastmcp/server.py
```diff
@@ -1047,6 +1048,11 @@ async def log(
             related_request_id=self.request_id,
         )
 
+    @property
+    def user(self) -> AuthenticatedUser | None:
+        """Get the authenticated user if available."""
+        return self.request_context.request.user if isinstance(self.request_context.request, Request) else None
```
felixweinberger, line 1054: Not a huge fan of having a helper that depends on introspecting the type of the Request as we can't guarantee `.user` is available.

Maybe we can use the same pattern as in `client_id` instead with `getattr`?

```
    @property
    def user(self) -> AuthenticatedUser | None:
        """Get the authenticated user if available."""
        return getattr(self.request_context.request, "user", None)
```

Also would be good to update the README at https://github.com/modelcontextprotocol/python-sdk/blob/346e794d599e5662479db8ea3d9fba67553286d2/README.md#L569

dwreeves:
Hi @felixweinberger,

I'm open to closing this one. I don't know that it is really necessary.

felixweinberger:
> Hi @felixweinberger,
> 
> I'm open to closing this one. I don't know that it is really necessary.

Cool thank you for the response! In the interest of keeping changes contained I'll close this one in that case.

Issue #921, opened 2025-06-10
auth spec implementation


ihrpr:

maia-iyer:
Hello! Super interested in this, assuming this relates to the current draft version of MCP that treats MCP Servers as resource serverse. May I ask what is the status and ETA for this?

Issue #338, opened 2025-03-20
Expose session_id back to server when SSE connections are established


nimeshmc:
**Is your feature request related to a problem? Please describe.**
When a new SSE connection is established, I'm creating state specific to that session within my server. Requests coming into the /messages endpoint need to be able to access and modify the state created upon connection establishment. When a SSE connection is established, the SDK currently sends the session ID to the client, attached to the /messages URI. However, the server can't access this session ID right now without intercepting the streamed data.

**Describe the solution you'd like**
Expose the session ID for the SSE connection from the sse.connect_sse async context manager. Also expose the session ID to the request handlers.

**Describe alternatives you've considered**
Intercept the data being written to the outgoing stream to get the session ID from the endpoint event.
Doing something more contrived that involves introducing another primitive to duplicate what the session ID actually does, but is accessible to the server.

nimeshmc:
Update: so I've found a way to do this using the server lifespan to share state across all the request handlers. However, the problem is that I have to initialize the state to something that is specific to that session, and that happens when the SSE connection is created. So to make this possible, I'm actually creating new Server instances for every new SSE session and initializing its lifespan to that initial state. I haven't found any performance issues with this yet, but I'll report back if I do.

felixweinberger:
We've moved on from SSE to StreamableHTTP as of the March Spec: https://modelcontextprotocol.io/specification/2025-03-26/changelog#major-changes

Given SSE is deprecated we won't look into this anymore.

PR #982, opened 2025-06-18
MCP server separation into Authorization Server (AS) and Resource Server (RS) roles per spec PR #338

ihrpr:
Background: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/338
Closes: https://github.com/modelcontextprotocol/python-sdk/issues/921

---

  - MCP servers now act as Resource Servers (RS) only, not Authorization Servers
  - Implements RFC 9728 Protected Resource Metadata for AS discovery

  Implementation:
  - Added `TokenVerifier` protocol to decouple token validation from OAuth provider logic
  - Resource Servers use token introspection to validate tokens with Authorization Servers
  - Added WWW-Authenticate header with resource_metadata parameter pointing to /.well-known/oauth-protected-resource

  Example:
  - Split example into separate AS (auth_server.py) and RS (server.py) implementations
  - AS handles OAuth flows and token issuance
  - RS validates tokens via introspection and serves protected resources

Follow up:

- [ ] remove github from Auth server examples, it brings too munch complexity
- [x] https://www.rfc-editor.org/rfc/rfc8707.html
- [ ] refactor AnyHttpUrl

### DIFF
--- a/src/mcp/server/auth/settings.py
+++ b/src/mcp/server/auth/settings.py
@@ -25,5 +25,6 @@ class AuthSettings(BaseModel):
     # Resource Server settings (when operating as RS only)
     resource_server_url: AnyHttpUrl | None = Field(
         ...,
-        description="OAuth authorization server URL that issues tokens for this resource server.",
+        description="The URL of the MCP server to be used as the resource identifier "
+        "and base route to look up OAuth Protected Resource Metadata.",
     )


src/mcp/client/auth.py
```diff
@@ -120,6 +121,42 @@ def _get_authorization_base_url(self, server_url: str) -> str:
...
+                    response.raise_for_status()
+                    metadata_json = response.json()
+                    logger.debug(f"Protected resource metadata discovered (no MCP header): {metadata_json}")
+                    return ProtectedResourceMetadata.model_validate(metadata_json)
+                except Exception:
+                    logger.exception("Failed to discover protected resource metadata")
+                    return None
+            except Exception:
+                logger.exception("Failed to discover protected resource metadata")
+                return None
```
Kludex: I suggest we stop developing so defensive in this code source.

src/mcp/client/auth.py
```diff
@@ -120,6 +121,42 @@ def _get_authorization_base_url(self, server_url: str) -> str:
...
+        """
+        Discover protected resource metadata from server's well-known endpoint.
+        RFC 9728 Protected Resource Metadata.
+        """
+        # Extract base URL per MCP spec
+        auth_base_url = self._get_authorization_base_url(server_url)
+        url = urljoin(auth_base_url, "/.well-known/oauth-protected-resource")
+        headers = {MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION}
+
+        async with httpx.AsyncClient() as client:
```
Kludex: We aren't supposed to be creating many clients inside an `httpx.Auth`.

The idea is to use the client that is already created.
Check: https://www.python-httpx.org/advanced/authentication/#custom-authentication-schemes

Specifically this part:

```py
    def auth_flow(self, request):
        request.headers["X-Authentication"] = self.access_token
        response = yield request

        if response.status_code == 401:
            # If the server issues a 401 response, then issue a request to
            # refresh tokens, and resend the request.
            refresh_response = yield self.build_refresh_request()
            self.update_tokens(refresh_response)

            request.headers["X-Authentication"] = self.access_token
            yield request
```

src/mcp/client/auth.py
```diff
@@ -120,6 +121,42 @@ def _get_authorization_base_url(self, server_url: str) -> str:
...
+            try:
+                response = await client.get(url, headers=headers)
+                if response.status_code == 404:
+                    return None
+                response.raise_for_status()
+                metadata_json = response.json()
+                logger.debug(f"Protected resource metadata discovered: {metadata_json}")
+                return ProtectedResourceMetadata.model_validate(metadata_json)
```
Kludex: We don't need to convert to JSON first.
```suggestion
                metadata = ProtectedResourceMetadata.model_validate_json(response.content)
                logger.debug(f"Protected resource metadata discovered: {metadata}")
                return metadata
```

Kludex (commented):
Sorry for reviewing a draft, but I'm not sure if I'll have time to help further here.

src/mcp/client/auth.py
ihrpr: I might be missing something, this is specifically for `_discover_protected_resource_metadata`, discovery happens outside the context of an actual API request. 

I see how refresh token can be refactored, it's a good point

Kludex: There are many other clients being created in this file, but everything is under the auth flow, which could have been using the original client that is using the Auth. I didn't notice the others before.

ihrpr: refactoring all to use request

ihrpr: removed all uses of `httpx.AsyncClient()`

src/mcp/server/auth/middleware/bearer_auth.py
```diff
@@ -59,27 +58,70 @@ class RequireAuthMiddleware:
...
+
+        www_authenticate = f"Bearer {', '.join(www_auth_parts)}"
+
+        # Send response
+        await send(
+            {
+                "type": "http.response.start",
+                "status": status_code,
+                "headers": [
+                    (b"content-type", b"application/json"),
```
Kludex, line 116: I don't think it makes sense for MCP to be using the low level ASGI API. I know it's not introduced in this PR, but it makes me sad. 😞 

---

This is missing `content-length` btw.

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+            raise
+        except Exception as e:
+            logger.error("Unexpected error", exc_info=e)
+            return JSONResponse(
+                status_code=500,
+                content={
+                    "error": "server_error",
+                    "error_description": "Unexpected error",
+                },
+            )
```
Kludex: This is not necessary. It will already respond with a 500, and you'll be able to see in the logs.

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+        state = request.query_params.get("state")
+
+        if not code or not state:
+            raise HTTPException(400, "Missing code or state parameter")
+
+        try:
+            redirect_uri = await oauth_provider.handle_github_callback(code, state)
+            return RedirectResponse(url=redirect_uri, status_code=302)
+        except HTTPException:
+            raise
```
Kludex: This is redundant.
```suggestion
```

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+                    "scope": " ".join(access_token.scopes),
+                    "exp": access_token.expires_at,
+                    "iat": int(time.time()),
+                    "token_type": "Bearer",
+                }
+            )
+
+        except Exception as e:
+            logger.exception("Token introspection error")
+            return JSONResponse({"active": False, "error": str(e)}, status_code=500)
```
Kludex: Sending back `e` can expose private user's information here!

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+            except ValueError as e:
+                if "No GitHub token found" in str(e):
+                    return JSONResponse({"error": "no_github_token"}, status_code=404)
+                elif "GitHub API error" in str(e):
+                    return JSONResponse({"error": "github_api_error"}, status_code=502)
+                raise
+
+        except Exception as e:
+            logger.exception("GitHub user info error")
+            return JSONResponse({"error": str(e)}, status_code=500)
```
Kludex: Please let's stop using `except Exception`.

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+
+    routes.append(
+        Route(
+            "/github/user",
+            endpoint=cors_middleware(github_user_handler, ["GET", "OPTIONS"]),
+            methods=["GET", "OPTIONS"],
+        )
+    )
+
+    return Starlette(debug=True, routes=routes)
```
Kludex:
```suggestion
    return Starlette(routes=routes)
```

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
```diff
@@ -0,0 +1,259 @@
...
+    logger.info("Configure GitHub App callback URL: " + settings.github_callback_path)
+    logger.info("=" * 80)
+
+    await server.serve()
+
+
+@click.command()
+@click.option("--port", default=9000, help="Port to listen on")
+@click.option("--host", default="localhost", help="Host to bind to")
+def main(port: int, host: str) -> int:
```
Kludex: This will give a lot of maintenance burden. People will ask a lot of uvicorn's parameters to be added here.

examples/servers/simple-auth/mcp_simple_auth/github_oauth_provider.py
```diff
@@ -0,0 +1,251 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class GitHubOAuthSettings(BaseSettings):
+    """Common GitHub OAuth settings."""
+
+    # GitHub OAuth settings - MUST be provided via environment variables
+    github_client_id: str  # MCP_GITHUB_CLIENT_ID env var
+    github_client_secret: str  # MCP_GITHUB_CLIENT_SECRET env var
```
Kludex: What is the prefix `MCP_` in the comment about?

examples/servers/simple-auth/mcp_simple_auth/legacy_as_server.py
```diff
@@ -0,0 +1,173 @@
...
+
+class ServerSettings(GitHubOAuthSettings):
+    """Settings for the simple GitHub MCP server."""
+
+    model_config = SettingsConfigDict(env_prefix="MCP_")
+
+    # Server settings
+    host: str = "localhost"
+    port: int = 8000
+    server_url: AnyHttpUrl = AnyHttpUrl("http://localhost:8000")
```
Kludex, line 39: Pydantic is appending a trailing slash on `AnyHttpUrl` - I think we should stop using `AnyHttpUrl` and use plain `str` in the whole code source.

Even if you are setting up `http://localhost:8000`, the real value will be `http://localhost:8000/`.

examples/servers/simple-auth/mcp_simple_auth/legacy_as_server.py
```diff
@@ -0,0 +1,173 @@
...
+    github_callback_path: str = "http://localhost:8000/github/callback"
+
+    def __init__(self, **data):
```
Kludex: What is this for? Why is there no type hints?

src/mcp/server/auth/token_verifier.py
```diff
@@ -0,0 +1,14 @@
...
+from mcp.server.auth.provider import AccessToken
+
+
+@runtime_checkable
+class TokenVerifier(Protocol):
+    """Protocol for verifying bearer tokens."""
+
+    async def verify_token(self, token: str) -> AccessToken | None:
+        """Verify a bearer token and return access info if valid."""
+        ...
```
Kludex: No need when using docstring.
```suggestion
```

examples/servers/simple-auth/mcp_simple_auth/github_oauth_provider.py
ihrpr: leftover from refactoring, removed

examples/servers/simple-auth/mcp_simple_auth/legacy_as_server.py
ihrpr: Without the `__init__`, you get the error "Arguments missing for parameters 'github_client_id', 'github_client_secret'" because Pydantic expects these to be explicitly passed, not understanding they should come from environment variables. One option would be to just make then optional

Kludex: That's the nature of `BaseSettings`. Is the idea that they can either be set programmatically or via env var?

examples/servers/simple-auth/mcp_simple_auth/legacy_as_server.py
ihrpr: noted, this should be done outside of this PR, added to follow ups

examples/servers/simple-auth/mcp_simple_auth/legacy_as_server.py
ihrpr: yes, moved to optional. Using github in the example servers brings too much complexity, like having secrets etc, will remove it in the follow up PR.

examples/servers/simple-auth/mcp_simple_auth/auth_server.py
ihrpr: The purpose of this file is demonstrate mcp oauth concepts, not to be a production server. Removed the host as it's not needed, but port is useful

Kludex: Ah, my bad. All good here.

andreasoie:
@ihrpr Hi! Thanks so much for driving this change, do you have any sense of when this PR might be merged? 🙏 (ballpark estimate)

src/mcp/server/auth/settings.py
```diff
@@ -15,9 +15,15 @@ class RevocationOptions(BaseModel):
 class AuthSettings(BaseModel):
     issuer_url: AnyHttpUrl = Field(
         ...,
-        description="URL advertised as OAuth issuer; this should be the URL the server " "is reachable at",
+        description="Base URL where this server is reachable. For AS: OAuth issuer URL. For RS: Resource server URL.",
```
pcarleton: `issuer` means something specific in OAuth parlance, so I'd prefer we not overload it here if we can avoid it.  (mentioned [here](https://datatracker.ietf.org/doc/html/rfc8414#section-2) for AS metadata, and [here](https://www.rfc-editor.org/rfc/rfc9728.html#name-protected-resource-metadata) for PRM metadata) 

I put a commit here that changes this around: https://github.com/modelcontextprotocol/python-sdk/commit/be4ddd3b0e8b3f97202bbaad65888b14156dca08

I didn't push directly (still need to test it), but curious for your thoughts on that.

That means we'll always do the PRM endpoint, even if the server is an "all-in-one".  (Which to clarify, all-in-one is still supported! but in that case we do want both the PRM and the AS metadata endpoints, the PRM will just point to itself.)

so with that change `issuer_url` will always be the URL of the authorization server.

The most unfortunate thing in that change is crafting the "server" URL, in order to put it in the `www-authenticate` and as the `resource` parameter.

I added a function to do it from `host` + `port`, but it feels a little jank.  The other alternative would be to pass in a new parameter that's the server URL. lmkwyt

ihrpr: >  The other alternative would be to pass in a new parameter that's the server URL. lmkwyt

I think it's cleaner to have a parameter

rkondra-eightfold:
@ihrpr  @Kludex  Is it possible to cleanup AuthSettings and the authorization server provider code ... MCP sdk doesn't need to have utils to implement auth server. This is too much unnecessary complexity, and looks very unclean. Can we please just point to existing libraries which implement oauth very well already https://github.com/authlib/authlib for Authorization server

ihrpr:
> @ihrpr @Kludex Is it possible to cleanup AuthSettings and the authorization server provider code ... MCP sdk doesn't need to have utils to implement auth server. This is too much unnecessary complexity, and looks very unclean. Can we please just point to existing libraries which implement oauth very well already https://github.com/authlib/authlib for Authorization server

With the [new spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization), MCP server acts as RS, so there is no need to support AS in the future, for now leaving it for backwards compatibility, and yes, ideally it should be cleaned up.

src/mcp/server/auth/settings.py
```diff
@@ -15,9 +15,15 @@ class RevocationOptions(BaseModel):
...
     )
     service_documentation_url: AnyHttpUrl | None = None
     client_registration_options: ClientRegistrationOptions | None = None
     revocation_options: RevocationOptions | None = None
     required_scopes: list[str] | None = None
+
+    # Resource Server settings (when operating as RS only)
+    resource_server_url: AnyHttpUrl | None = Field(
+        ...,
+        description="OAuth authorization server URL that issues tokens for this resource server.",
```
pcarleton:
```suggestion
        description="The URL of the MCP server to be used as the resource identifier and base route to look up OAuth Protected Resource Metadata.",
```

pcarleton (dismissed):
pending doc string, otherwise LGTM

PR #895, opened 2025-06-06
Support Cursor OAuth client registration

sam-tombury:
Relaxes the restriction on `redirect_uri` during OAuth client registration (and token flows) from `AnyHttpUrl` to `AnyUrl`.

## Motivation and Context

Cursor uses [deeplinks](https://docs.cursor.com/deeplinks#markdown) with a custom scheme for `redirect_uri` when registering OAuth clients for MCP, which currently fails validation (as it is not a valid `AnyHttpUrl`).

## How Has This Been Tested?

Tested an MCP server using this branch and successfully registered Cursor as an OAuth client.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #989, opened 2025-06-19
Compare redirect URIs as strings

sp94:
## Motivation and Context

When testing [the simple auth example server](https://github.com/modelcontextprotocol/python-sdk/tree/main/examples/servers/simple-auth), I encountered 400 status codes from `POST /token` because `TokenHandler.handle` was evaluating `token_request.redirect_uri != authorize_request_redirect_uri` as `True`.

This occurred because `token_request.redirect_uri` was of type `AnyUrl`, while `authorize_request_redirect_uri` was `AnyHttpUrl`, and comparing these two different Pydantic types always returns `False` - even if the underlying URLs are the same.

(Example: `AnyUrl("http://google.com") == AnyHttpUrl("http://google.com")` is `False`.)

I believe the example auth server worked before https://github.com/modelcontextprotocol/python-sdk/pull/895 changed some `AnyHttpUrl`s to `AnyUrl`s in the validation code, but the example server was not updated.

## How Has This Been Tested?

Tested using Claude Code as the MCP client to connect to the example server:

`claude mcp add --transport sse github http://localhost:8000/sse`

The auth flow succeeded only after applying this patch.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update  

## Checklist

## Additional context

We could also update the example server to declare the redirect uri as `AnyHttpUrl`, but I think comparing the URIs as strings would make the validation less brittle.

### DIFF
--- a/src/mcp/server/auth/handlers/token.py
+++ b/src/mcp/server/auth/handlers/token.py
@@ -151,7 +151,7 @@ class TokenHandler:
                     authorize_request_redirect_uri = auth_code.redirect_uri
                 else:
                     authorize_request_redirect_uri = None
-                if token_request.redirect_uri != authorize_request_redirect_uri:
+                if str(token_request.redirect_uri) != str(authorize_request_redirect_uri):
                     return self.response(
                         TokenErrorResponse(
                             error="invalid_request",


sp94:
I just saw a draft PR that also addresses this: https://github.com/modelcontextprotocol/python-sdk/pull/982/files#diff-0bfdf5a468abdd078a0287db095a4ad5ebfcb04ea3e437026cdc8038c137ecab

I'll leave it to a maintainer if they want to close this or merge it in as a bug fix in the meantime

PR #991, opened 2025-06-20
RFC 8707 Resource Indicators Implementation

ihrpr:
# RFC 8707 Resource Indicators Implementation

**Background**: 

[Spec Changes PR](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/734)
[rfc8707](https://www.rfc-editor.org/rfc/rfc8707.html)
[TS Implementation](https://github.com/modelcontextprotocol/typescript-sdk/pull/638)
[Additional changes for not requiring strict matching ](https://github.com/modelcontextprotocol/typescript-sdk/pull/664 )

Closes: https://github.com/modelcontextprotocol/python-sdk/issues/962

------

Implements Resource Indicators validation for OAuth 2.0 (RFC 8707; [spec change](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/734))
.

- Set the `resource` parameter in OAuth authorization and token exchange flows to bind tokens to the MCP server
- Updated `token_verifier.py` to show how to perform resource verification (disabled by default, run with `--oauth-strict` to enable)
- Validation of resource uses hierarchical matching

## Motivation and Context

Facilitate prevention of token theft/confusion attacks where a malicious MCP server steals tokens meant for other services by explicitly binding tokens to their intended resources and showing how a server can check this binding.

This security vulnerability was identified in https://github.com/modelcontextprotocol/modelcontextprotocol/issues/544.

## Breaking Changes

While the change is breaking at a protocol level, it should not require code changes from SDK users (just SDK version bumping).

- **Client developers**: No code changes required. The SDK automatically extracts and includes the resource parameter from the server URL
- **Server developers**: 
  - Resource validation is left to implementors (see `token_verifier.py`)
  - The Authorization Server (if external) needs to support / relay the resource param, *and* enough of the clients need to update their SDK to start setting the resource param in their auth requests. This is why `server.py` only validates the `resource` if `--oauth-strict` is set.
  - Each server should ideally treat each MCP server as a distinct resource in their PRM. If they host many servers under the same domain, the resource should be specific to a single server, not be domain-wide (**otherwise any "bad apple" server might be able to benefit from resource confusion w/ its siblings**).
- **Auth providers**: should be updated to accept and relay resource parameter.


examples/servers/simple-auth/mcp_simple_auth/token_verifier.py
```diff
@@ -54,12 +63,48 @@ async def verify_token(self, token: str) -> AccessToken | None:
...
+                    resource=data.get("aud") or data.get("resource"),  # Include resource in token
                 )
             except Exception as e:
                 logger.warning(f"Token introspection failed: {e}")
                 return None
+
+    def _validate_resource(self, token_data: dict) -> bool:
+        """Validate token was issued for this resource server."""
+        if not self.server_url or not self.resource_url:
+            return True  # No validation if server URL not configured
```
ochafik: nit: I'd probably return False here. If the user asks for strict validation, should fail if they voluntarily emptied out the server_url setting

examples/servers/simple-auth/mcp_simple_auth/token_verifier.py
```diff
@@ -54,12 +63,48 @@ async def verify_token(self, token: str) -> AccessToken | None:
...
+        aud = token_data.get("aud")
+        if isinstance(aud, list):
+            for audience in aud:
+                if self._is_valid_resource(audience):
+                    return True
+            return False
+        elif aud:
+            return self._is_valid_resource(aud)
+
+        # Check custom 'resource' claim if no 'aud'
```
ochafik: Is this standard? We [only check for aud in the TS SDK](https://github.com/modelcontextprotocol/typescript-sdk/pull/638/files#diff-3866da951a54ae8122efa966efff28239a164b4282565b2dab5ddd4d2fdf56ad) and assuming [the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) applies it doesn't seem to be a registered claim.

https://datatracker.ietf.org/doc/html/rfc7519#section-4.1

src/mcp/shared/auth_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        Canonical resource URL string
+    """
+    # Convert to string if needed
+    url_str = str(url)
+
+    # Parse the URL
+    parsed = urlparse(url_str)
+
+    # Create canonical form: lowercase scheme and host, no fragment
+    canonical = urlunparse(
```
ochafik: nit: consider `urlunsplit(urlsplit(url_str)._replace(fragment=''))` instead

src/mcp/client/auth.py
ochafik, line 1: Would be good to add expectations on resource param to tests/client/test_auth.py similar to https://github.com/modelcontextprotocol/typescript-sdk/pull/638/files#diff-68dc66b14905d1fc01dc142e244a7e13fbdf88062d3a0b426983c333ccdfc9ed

(doesn't look like there's tests for authorize.py or for provider.py yet, not sure how hard they'd be to add)

ochafik (changes requested):
Looks good, couple of nits + could benefit from more tests

examples/servers/simple-auth/mcp_simple_auth/token_verifier.py
ihrpr: ugh, this is actually completely wrong, thank you! fixed

examples/servers/simple-auth/mcp_simple_auth/token_verifier.py
ihrpr: removed, using only aud

pcarleton (approved):
LGTM!

pcarleton (approved):

dr3s:
Heads up that I'm going to submit a PR imminently to omit the `resource` parameter when the PRM is not available and the server version is 3-26-2025 or earlier.  Just caught up with @pcarleton on the need to retain backwards compatibility for earlier servers.  

Whether servers should be able to support more recent protocol versions without the resource parameter is an open question.

PR #685, opened 2025-05-11
WIP: add tool outputSchema and DataContent type to support structured content 

davemssavage:
Enable tool to return data content and schema.

## Motivation and Context
As per discussion in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/356 there are use cases where being able to know the schema of a tool useful prior to calling it. This patch is an experiment to enable me to test this behaviour, it is far from finished and depends on the above RFC being accepted before it is merged, however opening a draft pull request for comment

## Breaking Changes
At the moment yes, this patch results in DataContent items being returned whenever a complex object is returned from a tool, this may result in breaking changes in clients. Some thought needs to go into how best to tackle this.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context
Updates to documentation will be required if this patch is approved

PR #993, opened 2025-06-20
feat: Add structured output support for tool functions

bhosmer-ant:
Adds support for structured output in FastMCP and Client, allowing tool functions to return well-typed, validated data that clients can easily process and validate.

 ## Key Changes

**FastMCP:**
Adds structured tool support to FastMCP's `@mcp.tool()` decorator. By default, functions with structured-output-compatible return types become structured tools, other functions become unstructured tools. This behavior can be overridden by passing `structured_output=False` to the decorator. (`structured_output=True` will force FastMCP to attempt to create a structured tool, which may be useful in certain edge cases.)

To convert a function into a structured tool,
  - the function return type annotation is converted into Pydantic model (see "Supported return types" below)
  - the tool's `outputSchema` is generated from this model
  - `structuredContent` results are validated on the server before sending, via the lowlevel server's structured output support

**Client:** adds validation of `structuredContent` results against tool `outputSchema`
**Tests:** lots
**Docs:** added to README.md
**Examples:** added `examples/fastmcp/weather_structured.py` showing various return type options

### Supported return types

`@mcp.tool()` supports the following function return types for structured output:
- Pydantic models: Used directly to validate function results and generate `outputSchema`
- dict\[str, \*\], TypedDict, NamedTuple, classes with typed fields: Converted to Pydantic models
- Primitives, lists, generics: Wrapped in `{\"result\": value}`

### Note on backward compatibility

In keeping with the [MCP specification regarding backward compatibility](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content), structured tools will provide both structured results and backwards-compatible unstructured results. In order to provide backward compatibility with previous versions of FastMCP in particular, we **do not use** the generic JSON serialization provided by the lowlevel server for this purpose. Instead we use the existing FastMCP unstructured results pipeline, which performs several ad hoc conversions of function results. **Future versions of FastMCP will instead use the generic lowlevel server pipeline.**

## Motivation and Context
Support structured output as defined in [spec rev 2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content)

## How Has This Been Tested?
New tests added

## Breaking Changes
* Protocol defines structured output in a backwards-compatible way
* SDK API changes should be backwards compatible

## Types of changes
- [X] Documentation update

## Checklist

## Additional context
Many thanks to @davemssavage, who [prototyped an initial version](https://github.com/modelcontextprotocol/python-sdk/pull/685) of this functionality.


Kludex:
I don't think the `structured_output` is needed on the `FastMCP`. Is it?

ihrpr:
> I don't think the `structured_output` is needed on the `FastMCP`. Is it?

@Kludex, Structured output is part of [the spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content). Since FastMCP in this repository serves as a high-level API and is the primary entry point for users to define tools, it needs to include structured_output. This gives users the ability to define tools with structured output support, which is a standard feature of the MCP protocol.

or you mean there is no much need of of the param itself as it's already part of the spec?

src/mcp/server/fastmcp/server.py
```diff
@@ -315,6 +311,7 @@ def add_tool(
         title: str | None = None,
         description: str | None = None,
         annotations: ToolAnnotations | None = None,
+        structured_output: bool = False,
```
ihrpr: Is there a reason we want to have a default set to `True`, given the backwards compatibility, where we also send an unstructured output. 

 As per other comments, we might just remove this entirely, it will be a nice way to migrate most of the tool to return structure output. 

```suggestion
        structured_output: bool = True,
```

Kludex:
Sorry, I meant "does the flag need to exist? Can the behavior be always true and based on the type hints?"

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -68,6 +84,28 @@ async def call_fn_with_arg_validation(
             return fn(**arguments_parsed_dict)
         raise TypeError("fn must be either Callable or Awaitable")
 
+    def to_validated_dict(self, result: Any) -> dict[str, Any]:
+        """Validate and convert the result to a dict after validation."""
+        if self.output_model is None:
```
ihrpr: this was already checked before calling `to_validated_dict`

tests/server/fastmcp/test_server.py
```diff
@@ -350,6 +353,338 @@ def mixed_list_fn() -> list:
...
+            assert tool.outputSchema is not None
+            assert tool.outputSchema["type"] == "object"
+            assert tool.outputSchema["additionalProperties"]["type"] == "string"
+
+            # Call tool
+            result = await client.call_tool("get_settings", {})
+            assert result.isError is False
+            assert result.structuredContent == {"theme": "dark", "language": "en", "timezone": "UTC"}
+
+    @pytest.mark.anyio
```
ihrpr: here tests started using lowlevel server

ihrpr (changes requested):
- I think we might just use HighLevel API (FastMCP) to always return structured and unstructured output. Given that it always has a return type we can return structured output and unstructured will be non-wrapped output behaviour.
- I think it would make sense to have validation on the lowlevel, what is the reason of not having the validation there?

src/mcp/server/lowlevel/server.py
```diff
@@ -381,18 +382,42 @@ async def handler(_: Any):
...
             async def handler(req: types.CallToolRequest):
                 try:
                     results = await func(req.params.name, (req.params.arguments or {}))
-                    return types.ServerResult(types.CallToolResult(content=list(results), isError=False))
+
+                    if isinstance(results, dict):
+                        json_content = pydantic_core.to_json(results, fallback=str, indent=2).decode()
+                        return types.ServerResult(
+                            types.CallToolResult(
+                                content=[types.TextContent(type="text", text=json_content)],
```
ihrpr: this will not be fully backwards compatible, would leaving this as  `content=list(results)` work?

bhosmer-ant:
@ihrpr per offline convo, recapping here
> I think we might just use HighLevel API (FastMCP) to always return structured and unstructured output. Given that it always has a return type we can return structured output and unstructured will be non-wrapped output behaviour.

Problems with this are:
- not all return types yield a structured output schema (e.g. classes without annotated properties) - if we wanted to define FastMCP as supporting only tools that do, we could do that, but it would be a pretty big (BC breaking) change
- BC in general - tools might be returning results that aren't easily represented as structured output, irrespective of return type annotations 
- new tools might _want_ to return results that aren't easily represented as structured output. (this is more of a "what tools do you want FastMCP to support" design question)

> I think it would make sense to have validation on the lowlevel, what is the reason of not having the validation there?
- lowlevel doesn't do input validation, so I went with what seemed to be the intent there. But per discussion, it sounds like lowlevel is more heavily used than I realized - will add (output) schema validation there

bhosmer-ant:
> Sorry, I meant "does the flag need to exist? Can the behavior be always true and based on the type hints?"

@Kludex see my reply to @ihrpr - making it always true would make some functions ineligible as tools, and would silently change the results returned by others (the latter also an issue with making it default-true)

src/mcp/server/fastmcp/server.py
bhosmer-ant: BC problem with this is that existing tools don't (or may not) map cleanly to structured output - and even when they do, we would need to decide for a given tool whether to employ 
* the "new structured tool BC conversion" (serializing the json to text and stuffing it into a single TextContent block) or 
* the "old unstructured tool BC conversion" (running the return value through the old conversion, which leaves content blocks in place and does a few other fairly ad hoc things)

...so defaulting structured_output=True would produce silent breaking changes for a certain class of tools, which seems bad. if there's a way to tighten the valid-structured-return-type rule so that *only* tools for which the conversion produces the same result make it through without error (and others produce a "please add `structured_output=False` to your decorator error" message), then we could dodge the silent-BC-break issue, will look into that

src/mcp/server/fastmcp/utilities/func_metadata.py
bhosmer-ant: true in the current codepath, but it's (currently) a public method

tests/server/fastmcp/test_server.py
bhosmer-ant: [edit] ah misunderstood your comment - will move lowlevel server tests outta here :)

src/mcp/server/lowlevel/server.py
bhosmer-ant: the `dict` return type from a tool is new, so this codepath should be new (for any code that typechecked). Don't think `list(results)` would work since that would give you `[<some python dict>]` rather than `[TextContent(<serialized python dict>)]`.

Kludex:
On the FastMCP repository, they have an analogous PR that seems to not include any parameter: https://github.com/jlowin/fastmcp/pull/901 cc @jlowin

jlowin:
Hey! Just to clarify, https://github.com/jlowin/fastmcp/pull/901 implements output schemas but *not* structured content (yet) because I'm actually waiting for this low-level PR to be merged so I can fully test those outputs.

That said, the approach we are taking toward structured content in FastMCP 2 is much more opinionated - if a function has a serializable return annotation, it automatically gets structured output, no flags needed. This feels more aligned with what users expect from a high-level SDK (automatic behavior based on type hints) vs low-level SDK (explicit configuration).

That said, I understand the backwards compatibility concerns that motivate the structured_output=True flag approach. Different design philosophies for different use cases!

Happy to share more thoughts on our implementation if it would be helpful. The automatic dict→structured content conversion is definitely a tricky area where user expectations can and will vary.

bhosmer-ant:
Updates: 
- made `structured_output` decorator param optional. If not passed, uses type sniffing to decide whether the tool should be structured or unstructured. (Can override by passing the flag explicitly to force either structured or unstructured.)
- rebased onto lowlevel server schema validation, although it turns out the multiple ad hoc things FastMCP does to both incoming arguments and results mean we have to jump through some hoops to preserve compatibility. Later versions of FastMCP should break compatibility and simplify.

README.md
```diff
@@ -249,6 +250,93 @@ async def fetch_weather(city: str) -> str:
...
+- Generic types (list, tuple, set) - wrapped in `{"result": value}`
+- Union types and Optional - wrapped in `{"result": value}`
+
+Structured results are automatically validated against the output schema 
+generated from the annotation. This ensures the tool returns well-typed, 
+validated data that clients can easily process:
+
+**Note:** For backward compatibility, unstructured results are also
+returned. Unstructured results are provided strictly for compatibility 
+with previous versions of FastMCP.
```
ihrpr:
```suggestion
with previous MCP specification revisions.
```

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -162,8 +230,215 @@ def func_metadata(func: Callable[..., Any], skip_names: Sequence[str] = ()) -> F
...
+                output_conversion = "class" if output_model is not None else "none"
+        else:
+            if structured_output:
+                raise InvalidSignature(
+                    f"Function {func.__name__}: return type {annotation} is not supported for structured output. "
+                )
+
+    import sys
+
+    print(f"HEY Function {func.__name__} metadata: {output_model=} {output_conversion=}", file=sys.stderr)
```
ihrpr: debugging leftovers? 
```suggestion
```

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -61,12 +68,51 @@ async def call_fn_with_arg_validation(
...
+        Note: we return unstructured content here **even though the lowlevel server
+        tool call handler provides generic backwards compatibility serialization of
+        structured content**. This is for FastMCP backwards compatibility: we need to
+        retain FastMCP's ad hoc conversion logic for constructing unstructured output
+        from function return values (see _convert_to_content in mcp.server.fastmcp.server),
+        whereas the lowlevel server simply serializes the structured output.
+
+        This backwards compatibility provision will be removed in a future version of FastMCP.
+        """
+        from mcp.server.fastmcp.server import _convert_to_content  # type: ignore
```
ihrpr: why here is typeignore? 

should we move private method to this utils?

Kludex: So, the right way to do this in python is to have a private module, and then you can import anywhere in the package.

In this case, `from mcp.server.fastmcp._server import convert_to_content` would be fine, because `_server` is a private module, but since `func_metadata` is in the package, then you can import objects from the private module. But you can't import a private object from a module within the package.

src/mcp/server/fastmcp/utilities/func_metadata.py
bhosmer-ant: 😬

README.md
bhosmer-ant: This is specifically about past versions of FastMCP - i.e. the particular quirks of FastMCP's conversion have been preserved for the sake of backward compatibility, rather than using lowlevel's generic conversion that would produce different unstructured outputs for the same tools.

src/mcp/server/fastmcp/utilities/func_metadata.py
bhosmer-ant: [edit: went ahead and moved it to the right place/added comments, no private module needed]

Yeah this odd move was to leave the existing function completely undisturbed, for blame continuity - it's the quirky converter FastMCP uses to produce unstructured tool output., which the current setup opts to leave in place for the sake of compatibility (i.e., clients depending on the particular structure the old FastMCP would return for particular return types). 

In the light of day I need to think more about whether this extreme compatibility choice is the right one but if so I'll at least change the method name and add clarification to the comment right about this line.

ihrpr (approved):

ihrpr (approved):
Thank you!

PR #1003, opened 2025-06-22
feat(client): add _meta support to call_tool for tool calls\n\n- Adds…

pyrotank41:
Adds optional `_meta` parameter support to `ClientSession.call_tool` and `ClientSessionGroup.call_tool` methods, enabling users to pass request metadata (such as progress tokens or user context) in a clean, explicit, and backward-compatible way.

## Motivation and Context

Currently, users who need to pass metadata (like progress tokens or user context) to tool calls have to work around the SDK's limitations by either:
1. Creating custom session classes that override `call_tool`
2. Passing metadata through the `arguments` parameter (which is not semantically correct)

This change provides a proper, first-class way to pass request metadata while maintaining full backward compatibility.

## How Has This Been Tested?

- Added comprehensive unit tests for both `ClientSession.call_tool` and `ClientSessionGroup.call_tool` with `_meta` parameter
- Verified that existing functionality remains unchanged (backward compatibility)
- Tested that `_meta` data is properly passed through to the underlying request structure

## Breaking Changes

No breaking changes. The `_meta` parameter is optional and defaults to `None`, so existing code will continue to work without modification.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- The implementation leverages the existing `RequestParams.Meta` structure that was already defined in the types
- The `_meta` parameter is passed as the `_meta` field in `CallToolRequestParams`, which inherits from `RequestParams`
- This change is minimal and focused, affecting only the client-side tool calling interface
- The feature enables advanced use cases like progress tracking, user context passing, and other metadata requirements without requiring users to create custom session classes

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -282,8 +282,12 @@ class ClientSession(
         arguments: dict[str, Any] | None = None,
         read_timeout_seconds: timedelta | None = None,
         progress_callback: ProgressFnT | None = None,
+        _meta: dict[str, Any] | None = None,
     ) -> types.CallToolResult:
-        """Send a tools/call request with optional progress callback support."""
+        """Send a tools/call request with optional progress callback."""
+
+        # Create the Meta object if _meta is provided
+        meta_obj = types.RequestParams.Meta(**_meta) if _meta else None
 
         return await self.send_request(
             types.ClientRequest(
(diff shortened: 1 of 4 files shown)


ihrpr:
Thank you for your contribution! 

One of the examples mentioned in PR is passing progress token. SDK handless it automatically, when the `progress_callback` is set on the client tool call. This change will break the functionality to automatically assign the progress token in SDK.

User context use case sounds more like an extension and should be handled outside of the SDK.

Please re-open the PR in case there are concrete examples for passing _meta and I missed something.

Issue #938, opened 2025-06-11
Working example- httpstreamable


tampiss:
### Question

Hello, has anyone a functioning py example running /mcp via streamable? No matter what I do, the py does not work. Tried similar approach with ts, and there it works well. I'd rather use py so asking. Even if I create as simple server with fastmcp as possible, it does not work. Thank you in an advance 🙏🏼

mjunaidca:
Here's a one using streamable http stateless transport in python with a simple httpx client:

https://github.com/panaversity/learn-agentic-ai/tree/main/03_ai_protocols/02_model_context_protocol/01_server_development/01_hello_mcp_server

tampiss:
> Here's a one using streamable http stateless transport in python with a simple httpx client:
> 
> https://github.com/panaversity/learn-agentic-ai/tree/main/03_ai_protocols/02_model_context_protocol/01_server_development/01_hello_mcp_server

Thank you, but unfortunately, it is the same as with other approaches. It does not work with the clients I use due to the redirect issue, which still occurs in the Py SDK.

INFO:     10.78.80.197:53202 - "GET /mcp HTTP/1.1" 307 Temporary Redirect
INFO:     10.78.80.197:53202 - "GET /mcp/ HTTP/1.1" 200 OK

xuzexin-hz:
Yes, whether printing mcp logs or checking other tools calling mcp, the /mcp/ address is ok. Although I don't know why the underlying layer does this

pamelafox:
This one works for me:

```
"""MCP Streamable HTTP Client"""
import argparse
import asyncio
import logging

from mcp import ClientSession
from mcp.client.streamable_http import streamablehttp_client

async def main():
    """Main function to run the MCP client"""
    parser = argparse.ArgumentParser()
    # If the server returns a redirect for /mcp to /mcp/, then just specify /mcp/ in the URL
    parser.add_argument(
        "--server", type=str, default="http://localhost:8000/mcp", help="Server hostname"
    )
    args = parser.parse_args()

    async with streamablehttp_client(args.server) as (
        read_stream,
        write_stream,
        _,
    ):
        # Create a session using the client streams
        async with ClientSession(read_stream, write_stream) as session:
            # Initialize the connection
            await session.initialize()
            # List available tools
            tools_list = await session.list_tools()
            print("Available tools:", ", ".join(tool.name for tool in tools_list.tools))
            # Call a tool
[...]

PR #1007, opened 2025-06-23
feat: default streamable_http_path to "/" for consistency

tewatia:
feat: default streamable_http_path to "/" for consistency

Previously, this default was "/mcp", which led to URLs like /mcp/mcp when the base was already mounted at /mcp. This nested path confused the developers. Switching to "/" aligns with the TypeScript SDK behavior and prevents double mounting.

Summary of changes:
- Updated default of `streamable_http_path` to `/`.
- Added inline comment in `MCPServerConfig`.

## Motivation and Context

Avoids double-mount confusion and brings parity with the TypeScript SDK.

## How Has This Been Tested?

- Ran `uv run pytest`.
- Verified types with `uv run pyright`.
- Ensured linting/formatting via `uv run ruff check .` and `uv run ruff format .`.
- Tested local server routing on both `/` and `/mcp` mounts.

## Types of changes

- [ ] Documentation update

## Checklist

-  [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
-  [x] My code follows the repository's style guidelines
-  [x] New and existing tests pass locally
-  [x] I have added appropriate error handling
-  [x] I have added or updated documentation as needed

**Breaking changes:**
Users relying on the old default must now explicitly set:
```python
settings={"streamable_http_path": "/mcp"}

### CODE REVIEW

**Comment by tewatia** [comment]
@dsp-ant Hi David, could you take a look at this? I discovered that when the app is mounted at /mcp, the server endpoint ends up at /mcp/mcp, which isn’t documented and prevented my client from connecting—this cost me two days of debugging. I’ve seen others report the same issue in the tracker, and the TypeScript SDK doesn’t exhibit this behavior. I’m confident that updating the default path to / will resolve it. Please let me know if there’s anything else I can do to help get this merged, or if there’s a specific reason for the current default.

**Comment by akolotov** [comment]
I would vote for keeping it `/mcp` by default for backward compatibility but provide ability to reconfigure it.

**Comment by tewatia** [comment]
@akolotov @dsp-ant I just went through the first 2 pages of issues list and found atlesst 2 issues related to this - #938 #1053 
I think it's time either this is changed or it is made abundantly clear in the documentation that the server is reachable at base_url/mcp/mcp.
I think this behaviour must has caused frustration for even those who eventually got the server working.

**Comment by akolotov** [comment]
Let's discover a common ground

> it is made abundantly clear in the documentation that the server is reachable at base_url/mcp/mcp.
I think this behaviour must has caused frustration for even those who eventually got the server working.

In my case the server work ` base_url/mcp/` without additional `mcp`:

```python
import uvicorn

mcp = FastMCP(name=SERVER_NAME, instructions=composed_instructions)

mcp.settings.stateless_http = True
mcp.settings.json_response = True

asgi_app = mcp.streamable_http_app()

uvicorn.run(asgi_app, host=http_host, port=http_port)
```

What will be changed after your PR merged in my setup if I don't add `settings={"streamable_http_path": "/mcp"}`?

felixweinberger (changes requested):
Thank you for the contribution and bringing attention to the discrepancy between typescript and python behavior here!

I looked at the issues https://github.com/modelcontextprotocol/python-sdk/issues/1053 and https://github.com/modelcontextprotocol/python-sdk/issues/938 you linked but I'm not sure how they relate to this particular setup?

Given the significant potential for breakage here I'd also err on the side of improving docs for now like @akolotov instead of introducing a breaking change - my understanding is that the doubled up `/mcp/mcp` mount only actually applies if in your own routing setup you set `/mcp` as the root and mount a sub-app - it doesn't always just mount at `/mcp/mcp`, it could be `/xyz/mcp` or whatever you set as your root in the parent app?

Nonetheless I think the discrepancy between typescript and python is not ideal so we should probably make this change in a future version that permits breaking changes - I'm keeping this PR open and adding it to the milestone for SDK v2.0.0

I've added some proposed improved docs here for now - would this have helped you avoid this rabbit hole? https://github.com/modelcontextprotocol/python-sdk/pull/1172

felixweinberger:
Closing this one as it's been a while and for the above given reasons. We're going through older PRs right now to ensure we're prioritizing maintainer time appropriately.

Feel free to reopen or ping here if this is still an issue you would want to come back to.

PR #1008, opened 2025-06-23
fix: Add exception handling for POST request failures in StreamableHTTP

Kulaxyz:
Added exception handling in the `_handle_post_request` method to gracefully handle transport errors and return proper JSON-RPC error responses.

## Motivation and Context

When an MCP server returns HTTP 5xx errors (e.g., when the server is down) after initial connection, these errors were not being propagated to the caller. This made it impossible for client applications to detect and handle server failures appropriately.
## How Has This Been Tested?

- Verified that HTTP exceptions are properly forwarded to read streams
- Confirmed error responses are delivered immediately with correct JSON-RPC format
- All existing transport tests continue to pass

## Breaking Changes

no
## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

[Strands SDK](https://github.com/strands-agents/sdk-python) is an example of an implementation that is hanging when MCP Server is down.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -255,40 +255,50 @@ class StreamableHTTPTransport:
         headers = self._prepare_request_headers(ctx.headers)
         message = ctx.session_message.message
         is_initialization = self._is_initialization_request(message)
+        try:
+            async with ctx.client.stream(
+                "POST",
+                self.url,
+                json=message.model_dump(by_alias=True, mode="json", exclude_none=True),
+                headers=headers,
+            ) as response:
+                if response.status_code == 202:
+                    logger.debug("Received 202 Accepted")
+                    return
 
-        async with ctx.client.stream(
-            "POST",
-            self.url,
-            json=message.model_dump(by_alias=True, mode="json", exclude_none=True),
-            headers=headers,
-        ) as response:
-            if response.status_code == 202:
-                logger.debug("Received 202 Accepted")
-                return
+                if response.status_code == 404:
+                    if isinstance(message.root, JSONRPCRequest):
+                        await self._send_session_terminated_error(
+                            ctx.read_stream_writer,
+                            message.root.id,
+                        )
+                    return
 
-            if response.status_code == 404:
-                if isinstance(message.root, JSONRPCRequest):
-                    await self._send_session_terminated_error(
+                response.raise_for_status()
+                if is_initialization:
+                    self._maybe_extract_session_id_from_response(response)
+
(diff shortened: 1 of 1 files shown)


yangmianmian:
When will this commit merge to the master?

src/mcp/client/streamable_http.py
```diff
@@ -255,40 +255,50 @@ async def _handle_post_request(self, ctx: RequestContext) -> None:
...
+                if content_type.startswith(JSON):
+                    await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
+                elif content_type.startswith(SSE):
+                    await self._handle_sse_response(response, ctx, is_initialization)
+                else:
+                    await self._handle_unexpected_content_type(
+                        content_type,
+                        ctx.read_stream_writer,
+                    )
+        except Exception as exc:
```
felixweinberger, line 292: This is too broad for the specific case you mention - we shouldn't be doing this broad exception handling specifically for 5xx= failures as we might mask other issues this way.

src/mcp/client/streamable_http.py
```diff
@@ -255,40 +255,50 @@ async def _handle_post_request(self, ctx: RequestContext) -> None:
...
-                return
-
-            if response.status_code == 404:
-                if isinstance(message.root, JSONRPCRequest):
-                    await self._send_session_terminated_error(
-                        ctx.read_stream_writer,
-                        message.root.id,
-                    )
-                return
+                if response.status_code == 404:
```
felixweinberger, line 269: We should instead expand this or add a case like

```
if response.status_code in (500, 502, 503, ...):
```

That does specific handling of only those specific errors you're encountering and that are expected that we want to handle. That would prevent us from masking other errors that might occur.

Other unexpected HTTP errors should be handled by raise_for_status() lower down.

felixweinberger (changes requested):
Hi @Kulaxyz thank you for this contribution! And apologies for the time it took to get back to you on this.

The problem raised here looks valid, but we should be less broad with Exception handling.

It would also be great if we could add tests to demonstrate the functionality and prevent regressions in future.

If you have the bandwidth to work on this I'd be happy to come back to this quickly!

felixweinberger:
Similar PR: https://github.com/modelcontextprotocol/python-sdk/pull/1194

Kulaxyz:
@felixweinberger accidentally requested changes. I'm working on this, will ping you once I'm done.

maxisbey:
Closing due to inactivity. There's a similar PR at #1194 also addressing error handling in this area.

PR #1014, opened 2025-06-24
Fix `/.well-known/oauth-authorization-server` dropping path

ihrpr:
As identified in https://github.com/modelcontextprotocol/typescript-sdk/pull/687 we need fix it in Python SDK as well

Fixed OAuth discovery URL construction to comply with RFC 8414 by preserving path components. Previously, /.well-known/oauth-authorization-server discovery was dropping the path portion of authorization server URLs (e.g.,
  https://example.com/path/mcp became https://example.com/.well-known/... instead of https://example.com/.well-known/.../path/mcp).

  This fix enables operation with auth servers hosted at subpaths, matching the same issue recently fixed in the TypeScript SDK. 

 *Fallback for backwards compatibility*:
  - If path-aware discovery returns 404, automatically retries at root: https://example.com/.well-known/oauth-authorization-server
  - No fallback for root paths (/) to avoid unnecessary requests
  - No fallback for non-404 errors to preserve error semantics

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -106,6 +106,10 @@ class OAuthContext:
     # State
     lock: anyio.Lock = field(default_factory=anyio.Lock)
 
+    # Discovery state for fallback support
+    discovery_base_url: str | None = None
+    discovery_pathname: str | None = None
+
     def get_authorization_base_url(self, server_url: str) -> str:
         """Extract base URL by removing path component."""
         parsed = urlparse(server_url)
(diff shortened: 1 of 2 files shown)


src/mcp/client/auth.py
```diff
@@ -197,18 +201,53 @@ async def _handle_protected_resource_response(self, response: httpx.Response) ->
             except ValidationError:
                 pass
 
+    def _build_well_known_path(self, pathname: str) -> str:
+        """Construct well-known path for OAuth metadata discovery."""
+        well_known_path = f"/.well-known/oauth-authorization-server{pathname}"
+        if pathname.endswith("/"):
```
pcarleton, line 207: This tripped me up for a minute, b/c i thought it was referring to `//` at the end, but I see you have a test for it, and it follow this from the RFC:

   terminating "/" MUST be removed before inserting "/.well-known/" and
   the well-known URI suffix between the host component and the path
   component.  The client would make the following request when the
   issuer identifier is "https://example.com/issuer1" and the well-known
   URI suffix is "oauth-authorization-server" to obtain the metadata,
   since the issuer identifier contains a path component:

 >   `GET /.well-known/oauth-authorization-server/issuer1 HTTP/1.1
     Host: example.com`

pcarleton (approved):
LGTM!

PR #1017, opened 2025-06-24
Make "resource" optional on earlier protocols

dr3s:
Make "resource" optional on earlier protocols in order to remain backwards compatible with servers on earlier versions.

## Motivation and Context
https://github.com/modelcontextprotocol/python-sdk/pull/991 introduced breaking changes for servers on earlier protocol versions that don't support the resource param.  This change requires the param only if either is true: PRM is available or the protocol is 2025-06-18 or later.

## How Has This Been Tested?
unit tests

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -95,6 +95,7 @@ class OAuthContext:
     protected_resource_metadata: ProtectedResourceMetadata | None = None
     oauth_metadata: OAuthMetadata | None = None
     auth_server_url: str | None = None
+    protocol_version: str | None = None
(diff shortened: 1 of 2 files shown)


dr3s:
@pcarleton it seems like my implementation is more narrow than [yours](https://github.com/modelcontextprotocol/typescript-sdk/pull/693) as I only exclude "resource" if it's also an older protocol

pcarleton (approved):
lgtm, thanks!

PR #1041, opened 2025-06-26
Update MCP to use Pydantic>=2.11

medaminezghal:
Update tests to be compatible with `pydantic>=2.11`

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -15,7 +15,6 @@ from mcp.shared.version import SUPPORTED_PROTOCOL_VERSIONS
 
 DEFAULT_CLIENT_INFO = types.Implementation(name="mcp", version="0.1.0")
 
-logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger("client")
 
 


medaminezghal:
https://github.com/pydantic/pydantic/blob/42298a08b4cb999242b2ce231adda03539858567/HISTORY.md?plain=1#L293

medaminezghal:
@bhosmer-ant Would you like to review this PR?

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "anyio>=4.5",
     "httpx>=0.27",
     "httpx-sse>=0.4",
-    "pydantic>=2.7.2,<3.0.0",
+    "pydantic>=2.11.0,<3.0.0",
```
Kludex: I don't think we can do this. It will bother a lot of package maintainers that depend on versions in between.

Kludex (changes requested):
We can update the lockfile and the tests, but we can't update the `pyproject.toml`.

src/mcp/server/auth/routes.py
```diff
@@ -32,7 +32,8 @@ def validate_issuer_url(url: AnyHttpUrl):
     """
 
     # RFC 8414 requires HTTPS, but we allow localhost HTTP for testing
-    if url.scheme != "https" and url.host != "localhost" and not url.host.startswith("127.0.0.1"):
+    host = cast(str, url.host)
```
Kludex: what's the issue here?

medaminezghal: I've solved the problem from pyright

Kludex: but what was the problem?

medaminezghal: error: "startswith" is not a known attribute of "None" (reportOptionalMemberAccess)

medaminezghal: @Kludex Is there anythink I can fix? 

I didn't understand why the tests fails.

medaminezghal: It's fixed after merging new updates

Kludex: We need to check if it's None instead of casting here.

medaminezghal: @Kludex something like that?
```
if (
    url.scheme != "https" 
    and url.host != "localhost" 
    and (url.host is None or not url.host.startswith("127.0.0.1"))
):
```

Kludex: `is not None` I guess?

medaminezghal:
```
url.host is not None and not url.host.startswith("127.0.0.1")
```
what about this?

Kludex: yes

medaminezghal:
@Kludex Is everything good now?

medaminezghal:
@Kludex I think it can be merged now

bhosmer-ant:
@Kludex @medaminezghal unfortunately the changes here are incompatible with the older versions of pydantic we still want to support (due to pydantic-core changes from 2.27 to 2.33). I'll try to forward fix but may need to revert

Kludex:
> @Kludex @medaminezghal unfortunately the changes here are incompatible with the older versions of pydantic we still want to support (due to pydantic-core changes from 2.27 to 2.33). I'll try to forward fix but may need to revert

You need to check the comments in this PR.

This PR doesn't bump the minimum version of pydantic. It bumps on this project pipeline only.

bhosmer-ant:
> > @Kludex @medaminezghal unfortunately the changes here are incompatible with the older versions of pydantic we still want to support (due to pydantic-core changes from 2.27 to 2.33). I'll try to forward fix but may need to revert
> 
> You need to check the comments in this PR.
> 
> This PR doesn't bump the minimum version of pydantic. It bumps on this project pipeline only.

Yup I'd read the comments. The problem is that we're _not_ bumping the minimum version - it means that we're allowing older versions of pydantic with which the fixes here are incompatible. (The issue isn't with pydantic itself but with pydantic-core: the new version of pydantic pins a version of pydantic-core which has the changes that necessitated these changes.)

> It will bother a lot of package maintainers that depend on versions in between.

Can you elaborate? Are there concrete instances you're aware of, or was it just on general principles?

medaminezghal:
@bhosmer-ant This change will only affect how tests work with `Pydantic>=2.11`

I don't think it will cause any problem with other packages.

bhosmer-ant:
> @bhosmer-ant This change will only affect how tests work with `Pydantic>=2.11`
> 
> I don't think it will cause any problem with other packages.

@medaminezghal the problem is when Pydantic < 2.11 is installed (as allowed by pyproject.toml). Earlier versions pin an earlier version of `pydantic-core`, under which the `is not None` check is redundant (typechecker warning) and the generated schema in the test is the earlier one (so it fails with this change).

The root of the problem is `pyproject.toml` - we'll need to narrow the versions of pydantic so that we're using only one version of pydantic-core. (And then sync `uv.lock` with it.)

Kludex:
I don't understand the problem here.

If this repository wants to the right thing, it needs to test the lowest and the latest version allowed. The pydantic version on the lockfile was already far above the minimum before this PR.

Python package repositories usually make sure the latest release are tested without breaking the previous versions - or test lowest and latest.

---

About my comment on the "other maintainers" I meant that people may be pinning pydantic for some reason, and it forces them to work and update the project, which is annoying.

medaminezghal:
@bhosmer-ant the `is not None` is releated with new version of `Pyright` not `Pydantic`.

PR #1043, opened 2025-06-26
Adding roots changed and initialized notification handlers

gspencergoog:
I've added two server-side notification handlers, one for `InitializedNotification` and one for `RootsListChangedNotification`.  

## Motivation and Context
This allows an MCP server to fully implement roots support, since once the initialization is complete, it needs to be able to request roots from clients which support them, and it needs to be able to react to changes in the client's list of roots if they change.

## How Has This Been Tested?
I wrote new tests for the notifications, and merged it with the existing progress notification tests.

## Breaking Changes
No breaking changes were introduced.

## Types of changes

## Checklist


gspencergoog:
@felixweinberger I'm new to this repo: do I need to do anything that I haven't done to proceed to a review?

src/mcp/server/lowlevel/server.py
```diff
@@ -677,7 +715,11 @@ async def _handle_notification(self, notify: Any):
             logger.debug("Dispatching notification of type %s", type(notify).__name__)
 
             try:
-                await handler(notify)
+                sig = inspect.signature(handler)
+                if "session" in sig.parameters:
+                    await handler(notify, current_session_ctx.get())
+                else:
+                    await handler(notify)
```
felixweinberger: I don't think we should do introspection based switches in our `lowlevel` server code - we should be as explicit as possible. If we're going to have different behavior based on the type of request, we should explicitly have match/case statements that handle those different cases.

Can we instead consistently pass the session into the `handler` like we do for `requests` if we need it? Some handlers just won't use it and that's fine.

For backwards compatibility we could have the `progress_notification` handler have an optional session argument.

src/mcp/server/lowlevel/server.py
```diff
@@ -104,6 +105,9 @@ async def main():
 # This will be properly typed in each Server instance's context
 request_ctx: contextvars.ContextVar[RequestContext[ServerSession, Any, Any]] = contextvars.ContextVar("request_ctx")
 
+# Context variable to hold the current ServerSession, accessible by notification handlers
+current_session_ctx: contextvars.ContextVar[ServerSession] = contextvars.ContextVar("current_server_session")
+
```
felixweinberger: What's the motivation for creating this session variable to pass through to the handlers?

Is it only to be able to call `ServerSession.list_changed()`? The tests don't really demonstrate how it will be used.

src/mcp/server/lowlevel/server.py
```diff
@@ -596,22 +630,26 @@ async def run(
...
     ):
-        with warnings.catch_warnings(record=True) as w:
-            # TODO(Marcelo): We should be checking if message is Exception here.
-            match message:  # type: ignore[reportMatchNotExhaustive]
-                case RequestResponder(request=types.ClientRequest(root=req)) as responder:
-                    with responder:
-                        await self._handle_request(message, req, session, lifespan_context, raise_exceptions)
-                case types.ClientNotification(root=notify):
-                    await self._handle_notification(notify)
+        session_token = current_session_ctx.set(session)
```
felixweinberger: Would be great not to have to manage this global state by just passing the `session` to `_handle_notification`

src/mcp/server/lowlevel/server.py
gspencergoog: Yes, that makes good sense.  Okay, I switched to just handling the type error if it doesn't take a session.

src/mcp/server/lowlevel/server.py
gspencergoog: Yes, it was just to allow calling `list_changed`, since the roots list can change.

src/mcp/server/lowlevel/server.py
gspencergoog: Done.

gspencergoog:
> Hi @gspencergoog thank you for this contribution! And apologies for the time it took to get back to you on this, currently getting a handle on managing all the inbound contributions like yours.

No worries, I know how it is. :-)

Hopefully this is closer to what you were expecting now.

gspencergoog:
@felixweinberger Do you have some time to finish this review?

gspencergoog:
Closing this since it's been months since it has had any attention.  Please feel free to reopen if you still would like this change.

ronytesler:
any update on supporting https://github.com/modelcontextprotocol/python-sdk/issues/1126 ("Specifying capabilities w/ FastMCP Server")?

PR #559, opened 2025-04-22
Fix hanging on streams when stdio_client exiting

LionNatsu:
Triggers a TaskGroup cancel after reaching the finally code block.

## Motivation and Context
`stdin_writer` in the task group hangs when pressing Ctrl+c.
Now we explicitly close the streams blocking.

## How Has This Been Tested?
Tested in a new simple no-op test and a real Uvicorn Starlette application.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1044, opened 2025-06-26
Add regression tests for #559 and #555

felixweinberger:
- **Simplify stdio cleanup: unified stream cleanup across platforms**
- **Add regression tests for stdio cleanup hanging**

## Types of changes

- [ ] Documentation update

## Checklist


jingx8885:
**Verification of Windows Process Tree Termination Fix**  
*(Originally encountered in #729  context)*  

**Scenario:**  
In a Python MCP server (`uv-pyproject` structure) using stdio mode launched via:  
`uv run --env-file=xxx mcp_server.py`  

We observed the following process hierarchy:  
```
Python parent process  
  → Python subprocess (middle layer)  
    → uv.exe  
      → MCP stdio service  
```  

**Failure:**  
Terminating the **middle Python subprocess** failed to terminate its descendants, leaving:  
1. `uv.exe` process  
2. MCP stdio service process  

...as orphaned/zombie processes on Windows.  

**Validation:**  
➡️ **This specific failure mode** was resolved by the process termination logic **originally implemented in #729 **.  
➡️ The **equivalent solution** now exists in #1044's unified approach.  
➡️ This logic has been **production-validated** on Windows 11 across:  
   - Live production systems  
   - Daily development/testing cycles  
   - High-frequency restart scenarios  

**Conclusion:**  
#1044 successfully addresses this Windows-specific process tree termination flaw. The child process cleanup behavior required for this scenario is correctly implemented.

felixweinberger:
After receiving offline feedback from @dsp-ant I'm splitting this PR into 2:

- This PR will keep the changes from #555 #559 #765
- #1078 will house the changes #729 #850 - there's some rework needed to avoid race conditions

FYI for @jingx8885 @surya-prakash-susarla, it's currently in draft and still needs work, I'll tag you once its ready

Kludex:
@agronholm Should we have to create OSS specific logic for termination when using `anyio.open_process`?

src/mcp/client/stdio/__init__.py
```diff
@@ -178,15 +177,38 @@ async def stdin_writer():
...
                     process.terminate()
+                    with anyio.fail_after(2.0):
+                        await process.wait()
+                except TimeoutError:
+                    # Process didn't respond to SIGTERM, force kill it
+                    process.kill()
+                    await process.wait()
+                except ProcessLookupError:
+                    # Process already exited, which is fine
+                    pass
```
agronholm: This seems redundant as the outer `except ProcessLookupError` would catch this.

agronholm:
> @agronholm Should we have to create OSS specific logic for termination when using `anyio.open_process`?

I'm not sure I follow. Can you give me more context? I'm very new to this project.

src/mcp/client/stdio/__init__.py
felixweinberger: Good point! Removed the redundant inner check, thank you.

felixweinberger:
Hi @Kludex, @agronholm thanks for taking a look!

For context, we actually don't spawn windows processes with `anyio.open_process` on Windows anymore since https://github.com/modelcontextprotocol/python-sdk/pull/596. We now use `subprocess.Popen` - the reason was that trying to spawn processes with `anyio` way was causing `NotImplementedError` on Windows when using the `SelectorEventLoop` (apparently the default for Python 3.13+ and e.g. in Streamlit, Google AgentKit https://github.com/google/adk-python/issues/1321), which was blocking Windows development of MCP servers in those environments.

However, your comment made me realize that we should probably explicitly guard for that case and still use `anyio` by default if possible as a first try, so I added a commit on top of this PR to attempt with `anyio` first and only in the `NotImplementedError` case use the `FallbackProcess` wrapper.

If this could somehow be fixed upstream in `anyio` that would of course be fantastic - though we did get feedback from Microsoft that Python's `asyncio` is relatively underdeveloped on Windows, which might be a root cause why we ran into this issue in the first place and might make it difficult to address there?

agronholm:
Any idea why `SelectorEventLoop` would be the default there? It stopped being the Python default on Windows on Python 3.8.

felixweinberger:
> Any idea why `SelectorEventLoop` would be the default there? It stopped being the Python default on Windows on Python 3.8.

@agronholm Doing some research based on what @theailanguage raised in #596:

1. Streamlit:
  - Discussion in #596 concerns Streamlit on Windows (cc @theailanguage)

2. MCP Adapters:
  - https://github.com/langchain-ai/langchain-mcp-adapters/issues/25 open, showing the same NotImplementedError with SelectorEventLoop on Windows

3. Jupyter Notebooks (outdated, but as examples):
  - https://github.com/jupyter/notebook/issues/5916 opened January 2021 for Jupyter Notebook 6.1.6
  - https://github.com/jupyterlab/jupyterlab/issues/12545 opened May 2022 for JupyterLab with Python 3.10
  - Both resolved now but show that subprocesses (in this case ipykernel?) were using unexpected event loops on windows.
  
It seems that while the `ProactorEventLoop` is the default on Python 3.8+ on Windows, that doesn't necessarily carry over to all applications that might set the event loop internally to `SelectorEventLoop` for some reason? In such cases this leads to a `NotImplementedError`

tests/client/test_stdio.py
```diff
@@ -90,3 +94,121 @@ async def test_stdio_client_nonexistent_command():
...
+
+        if cancel_scope.cancelled_caught:
+            raise TimeoutError("Test timed out")
+
+        end_time = time.time()
+        elapsed = end_time - start_time
+
+        # Should complete quickly even with SIGTERM-ignoring process
+        # This will fail if cleanup only uses process.terminate() without fallback
+        assert elapsed < 5.0, (
```
ihrpr: nit, would prefer some constants here and the reason documented

src/mcp/client/stdio/win32.py
```diff
@@ -159,24 +199,3 @@ async def create_windows_process(
             bufsize=0,
         )
         return FallbackProcess(popen_obj)
-
-
-async def terminate_windows_process(process: Process | FallbackProcess):
```
ihrpr, line 164: this is a public method - hope it's not used but worth keeping it with deprecation method, or at least we can include this change into minor version release and document in release notes

src/mcp/client/stdio/__init__.py
```diff
@@ -180,10 +179,12 @@ async def stdin_writer():
         finally:
             # Clean up process to prevent any dangling orphaned processes
             try:
-                if sys.platform == "win32":
-                    await terminate_windows_process(process)
-                else:
-                    process.terminate()
+                process.terminate()
+                with anyio.fail_after(2.0):
```
ihrpr: nit: constant

ihrpr (approved):
LGTM

left a few comments

tests/client/test_stdio.py
felixweinberger: done

src/mcp/client/stdio/win32.py
felixweinberger: marked as deprecated instead

src/mcp/client/stdio/__init__.py
felixweinberger: added

PR #1055, opened 2025-06-29
Embed code snippets for README from executable examples

ihrpr:
We already had an integration test that verifies all current functionality works as expected, so we can use it as both a server example **and** as the source for README snippets. These snippets will be automatically updated and verified through CI checks. In the future, we'll also add an everything client example and include it in both the README and integration tests.

1. **Automated README Snippets**
- Added `scripts/update_readme_snippets.py` that extracts code from example files into README
- Uses `<!-- snippet-source path/to/file.py -->` markers to identify updatable sections
- Automatically adds links to full example files on GitHub with line highlighting
- Supports both full-file and line-range extraction (e.g., `#L10-L20`)

2. **New Example Structure**
- Created `examples/snippets/` with single-feature examples:
  - `basic_tool.py` - Simple tool example
  - `basic_resource.py` - Resource handling
  - `basic_prompt.py` - Prompt creation
  - `tool_progress.py` - Progress reporting
  - `sampling.py` - LLM sampling
  - `elicitation.py` - User interaction
  - `completion.py` - Argument completion
  - `notifications.py` - Logging and notifications
- Each example is self-contained 

3. **Everything Server Extraction**
- Moved the comprehensive `make_everything_fastmcp()` function from tests into a standalone example at `examples/servers/everything/`
- Everything server remains available for testing all MCP features in one place

4. **CI Integration**
- Added `readme-snippets` job to `.github/workflows/shared.yml` to verify snippets are current
- Added pre-commit hook that checks snippet freshness on relevant file changes
- Both use `--check` mode to fail if snippets are outdated
image


README.md
```diff
@@ -419,21 +419,27 @@ def create_thumbnail(image_path: str) -> Image:
 
 The Context object gives your tools and resources access to MCP capabilities:
 
+<!-- snippet-source examples/servers/everything/src/everything/server.py#L43-L58 -->
```
ihrpr: this is only one example for this PR, when we agree on the process, will replace the rest of the Server examples (and later client) in README as separate PRs

examples/fastmcp/weather_structured.py
```diff
@@ -90,7 +90,7 @@ def get_weather_alerts(region: str) -> list[WeatherAlert]:
             WeatherAlert(
                 severity="high",
                 title="Heat Wave Warning",
-                description="Temperatures expected to exceed 40°C",
+                description="Temperatures expected to exceed 40 degrees",
```
ochafik, line 93: nit: revert to celsius to avoid frowns from US users?

scripts/update_readme_snippets.py
```diff
@@ -0,0 +1,196 @@
...
+{indent}```python
+{code}
+{indent}```
+{indent}_Full example: [{file_path}]({github_url})_
+{indent}<!-- /snippet-source -->"""
+
+        # In check mode, only check if code has changed
+        if check_mode:
+            # Extract existing code from the match
+            existing_content = match.group(5)
```
ochafik: move up w/ the others?

scripts/update_readme_snippets.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            # No line range - use whole file
+            code = file.read_text().rstrip()
+            github_url = f"https://github.com/modelcontextprotocol/python-sdk/blob/main/{file_path}"
+            line_ref = ""
+
+        # Build the replacement block
+        replacement = f"""{indent}<!-- snippet-source {file_path}{line_ref} -->
+{indent}```python
+{code}
```
ochafik: Might want to indent the code here too `{indent}{code.replace('\n', f'\n{indent}')}` (markdown should remove it)

examples/snippets/servers/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+This package contains simple examples of MCP server features.
+Each server demonstrates a single feature and can be run as a standalone server.
+"""
+
+import importlib
+import sys
+from typing import Literal
+
+# Available snippet modules
+SNIPPET_MODULES = [
```
ochafik: Given you're not using the #line syntax, i wonder if you could just get the list of snippets on the fly from readme parser, avoid checking in these generated files / evaluate them on the fly (maybe create a temp directory, add it to the PYTHONPATH programmatically and load adhoc $tmp/snippet$i.py modules?)

(would also remove the need for a file name)

Kludex (changes requested):
As mentioned on Discord, I don't think this will provide a good experience for readers.

We should have a setup in place where every example is an MRE in the documentation.

---

For inspiration, FastAPI and PydanticAI (popular python projects) have already this kind of setup in their repositories - although their setup is for the MkDocs, not for the README.

ihrpr:
> We should have a setup in place where every example is an MRE in the documentation.

@Kludex, the examples are now all standalone files, what am I missing?

Kludex (approved):
LGTM

PR #1057, opened 2025-06-30
fix json response

notsobad:
Fix JSON array response handling in StreamableHTTP client transport

## Motivation and Context

The StreamableHTTP client transport's `_handle_json_response` method was not properly handling JSON array responses from servers. When a server returns multiple JSONRPCMessage objects as an array (e.g., batch responses), the current implementation would fail to parse them correctly. This issue was discovered when testing with servers that return array responses for certain operations like `tools/list`.

## How Has This Been Tested?

- Tested with servers returning single JSONRPCMessage responses (existing functionality)
- Tested with servers returning JSON array responses containing multiple JSONRPCMessage objects
- Verified that both response types are correctly parsed and processed
- Confirmed that protocol version extraction still works correctly for initialization responses

## Breaking Changes

None. This is a backward-compatible fix that maintains existing functionality while adding support for array responses.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix refactors the JSON parsing logic to:
1. First parse the raw JSON content to determine its structure
2. Handle both single object and array responses appropriately
3. Maintain proper error handling for malformed responses
4. Preserve existing functionality for protocol version extraction and message processing

This change ensures compatibility with MCP servers that may return batch responses or multiple messages in a single HTTP response, which is a valid pattern in JSON-RPC 2.0 specifications.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -11,6 +11,7 @@ from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import timedelta
+import json
 
 import anyio
 import httpx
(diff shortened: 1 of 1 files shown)


notsobad:
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/streamableHttp.ts#L473

The TypeScript SDK can handle this situation correctly.

ihrpr:
Thank you for your contribution. The latest spec 2025-06-18 does not support batching. Python SDK is not supporting batching either, adding it only to the client will be confusing (Typescript is not fully supporting batching either). If there is a need for the spec to support batching, please create a PR/issue in https://github.com/modelcontextprotocol/modelcontextprotocol

PR #1058, opened 2025-06-30
feat: Allow any value type in get_prompt arguments

### LINKED ISSUE #749: Support anything other than strings in GetPromptRequestParams
My MCP server looks like that and the main focus should be on the fact that I expect `keywords` as a list of strings.
```python
from fastmcp import FastMCP
from fastmcp.prompts.prompt import PromptMessage, TextContent

mcp = FastMCP(
    name="My MCP Server",
    instructions="Dummy instruction.",
)


@mcp.prompt()
def create_joke(keywords: list[str]) -> PromptMessage:
    """Create joke using words: {', '.join(keywords)}"""
    content = f"Using the following keywords, write a short joke. Keywords: {", ".join(keywords)}"
    return PromptMessage(role="user", content=TextContent(type="text", text=content))


if __name__ == "__main__":
    mcp.run(transport="streamable-http", host="127.0.0.1", port=8000, path="/mcp")
```

The issue comes from the client side of MCP protocol, which expects only strings [here](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L617) and thus Pydantic is not happy with me providing list of strings. Is there any strong reason, why only strings are expected or we could relax the requirements to anything JSON serializable?

Simply that would do:
```python
class GetPromptRequestParams(RequestParams):
    """Parameters for getting a prompt."""

    name: str
    """The name of the prompt or prompt template."""
    arguments: dict[str, Any] | None = None
    """Arguments to use for templating the prompt."""
```

I'd be happy to submit a PR.

Currently, there seems to be no other solution than to provide strings.

arjunaskykok:
The type hints for the `arguments` parameter in `ClientSession.get_prompt` and the corresponding `GetPromptRequestParams` were previously restricted to `dict[str, str]`.

This was overly restrictive and prevented passing arguments with non-string values, such as numbers or booleans. This commit changes the type to `dict[str, Any]`, allowing for more flexible prompt argument structures.

A corresponding test case has been added to verify that calling `get_prompt` with an integer argument now works as expected.

Fixes #749 

## Motivation and Context

Imagine this MCP server code:

```python

mcp = FastMCP("My MCP Server", host="127.0.0.1", port=9000)

@mcp.prompt()
def get_employee_profile(employee_id: int) -> str:
    return f"Using SQL, get the employee profile from the id: {employee_id}"

if __name__ == "__main__":
```

And this client code:

```python
import os
import json
import asyncio
from mcp import ClientSession, StdioServerParameters, types

MCP_SERVER_URL = "http://127.0.0.1:9000/mcp/"

async def run():
    async with streamablehttp_client(MCP_SERVER_URL) as (read, write, _):
        async with ClientSession(
            read, write
        ) as session:
            await session.initialize()

            # Read prompt
            output = await session.get_prompt("get_employee_profile", {"employee_id": "77"}) # this works
            output = await session.get_prompt("get_employee_profile", {"employee_id": 77}) # this fails but we have declared the argument type to be int in the server code. BUG

            print(output.messages[0].content.text)


if __name__ == "__main__":
    asyncio.run(run())
```

It's very strange that we have to send a string argument although the argument type in the MCP server is int.

This PR also gives us flexibility to send an argument with a different type, like a list of strings, as referenced in the bug report.

The argument in calling the tool is not restricted to string. Why do we need to restrict it in the prompt case?

## How Has This Been Tested?

I have added a unit test.

## Breaking Changes

Nope.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

I was not sure where to put the unit test. The `test_session.py` is not ideal. But I couldn't find a better place.

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -344,7 +344,7 @@ class ClientSession(
             types.ListPromptsResult,
         )
 
-    async def get_prompt(self, name: str, arguments: dict[str, str] | None = None) -> types.GetPromptResult:
+    async def get_prompt(self, name: str, arguments: dict[str, Any] | None = None) -> types.GetPromptResult:
         """Send a prompts/get request."""
         return await self.send_request(
             types.ClientRequest(
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -639,7 +639,7 @@ class GetPromptRequestParams(RequestParams):
 
     name: str
     """The name of the prompt or prompt template."""
-    arguments: dict[str, str] | None = None
+    arguments: dict[str, Any] | None = None
     """Arguments to use for templating the prompt."""
     model_config = ConfigDict(extra="allow")
 
--- a/tests/client/test_session.py
+++ b/tests/client/test_session.py
@@ -495,3 +495,72 @@ async def test_client_capabilities_with_custom_callbacks():
     assert received_capabilities.roots is not None  # Custom list_roots callback provided
     assert isinstance(received_capabilities.roots, types.RootsCapability)
     assert received_capabilities.roots.listChanged is True  # Should be True for custom callback
+
+
+@pytest.mark.anyio
+async def test_get_prompt_with_non_string_argument():
+    client_to_server_send, client_to_server_receive = anyio.create_memory_object_stream[SessionMessage](1)
+    server_to_client_send, server_to_client_receive = anyio.create_memory_object_stream[SessionMessage](1)
+
+    async def mock_server():
+        await client_to_server_receive.receive()
+
+        await server_to_client_send.send(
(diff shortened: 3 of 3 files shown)


ihrpr:
Thank you for contributing! While this PR addresses a real usability issue, it diverges from the MCP protocol specification which explicitly defines prompt arguments as string-only (`{ [key: string]: string }`). The Python SDK should remain compliant with the official protocol. If non-string prompt arguments are needed, this should be addressed at the protocol specification level first, not in individual SDK implementations. Consider opening an issue in the MCP specification repository to discuss expanding prompt argument types to match tool argument flexibility. Closing this PR for now, we can re-open if the spec changes are merged.

Issue #976, opened 2025-06-18
enhance auth server discovery with OAuth2 and OIDC metadata support


ihrpr:
On hold until https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797 is merged

LucaButBoring:
@ihrpr Are you already working on implementing this, or planning to prior to the next spec release? Hoping to pick it up if not, as it'd enable using [Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/federation-endpoints.html) as an AS instead of needing a custom proxy. Looks like a small code change, as you already fixed the path routing for RFC8414 metadata earlier today (which this depends on).

manthram:
@LucaButBoring For Cognito, we have another issue with the metadata published on the endpoint. Though it supports S256 and PKCE, Cognito doesnt publish it. As the client specification expects it and checks for it, we end up with an error. I had to write a proxy to get the metadata from Cognito and add the S256 code_challenge_methods_supported option before responding to the client.

LucaButBoring:
Raised a PR for this (#1061).

> As the client specification expects it and checks for it

@manthram is that the case on the latest SDK version? From what I can see, the client code [does assume](https://github.com/modelcontextprotocol/python-sdk/blob/6f43d1f17022d9d45b51955304e39ad43ca48014/src/mcp/client/auth.py#L332-L343) S256 is supported, but doesn't appear to actually enforce that it's declared. `code_challenge_methods_supported` isn't defined for OIDC to begin with. The metadata type (`OAuthMetadata`) doesn't appear to enforce S256 during validation, either.

PR #1061, opened 2025-06-30
Support falling back to OIDC metadata for auth

LucaButBoring:
Implements basic OIDC support for working with authorization servers that don't expose OAuth metadata, following the semantics described in [RFC 8414 Section 5](https://datatracker.ietf.org/doc/html/rfc8414#autoid-11).

As part of this change, I modified the internal fallback behavior from a simple boolean flag to instead represent discovery as a stack of methods, which are consumed sequentially until one fallback method succeeds. This also helps clean up the duplicated code from working with `AsyncGenerator`.

## Motivation and Context
Enables falling back to OIDC 1.0 metadata when a server does not support OAuth metadata according to RFC 8414.

## How Has This Been Tested?
Added/updated unit tests.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
#976


ihrpr:
Thank you! Sorry, we had back and forth with this spec changes. For now it's on hold until https://github.com/modelcontextprotocol/modelcontextprotocol/pull/797 is merged

src/mcp/client/auth.py
```diff
@@ -251,32 +254,60 @@ async def _handle_protected_resource_response(self, response: httpx.Response) ->
...
+        if not base_url:
+            raise OAuthFlowError("No base URL available for fallback discovery")
+
+        # Fallback to root discovery for legacy servers
+        return urljoin(base_url, f"/.well-known/{well_known_endpoint}")
+
+    def _build_oidc_fallback_path(self, pathname: str, well_known_endpoint: str) -> str:
+        """Construct fallback well-known path for OIDC metadata discovery in legacy servers."""
+        # Strip trailing slash from pathname to avoid double slashes
+        clean_pathname = pathname[:-1] if pathname.endswith("/") else pathname
```
dsp-ant:
```suggestion
        clean_pathname = pathname.rstrip('/')
```

src/mcp/client/auth.py
```diff
@@ -511,6 +559,26 @@ def _add_auth_header(self, request: httpx.Request) -> None:
...
+            # If path-aware discovery fails with 404, try fallback to root
+            self._discover_oauth_metadata_fallback,
+            # If root discovery fails with 404, fall back to OIDC 1.0 following
+            # RFC 8414 path-aware semantics (see RFC 8414 section 5)
+            self._discover_oidc_metadata,
+            # If path-aware OIDC discovery failed with 404, fall back to OIDC 1.0
+            # following OIDC 1.0 semantics (see RFC 8414 section 5)
+            self._discover_oidc_metadata_fallback,
+        ]
+
```
dsp-ant: Can we just return URIs here and just call them, since they all eventually just pass through `_try_metadata_discovery`

src/mcp/client/auth.py
```diff
@@ -544,15 +612,12 @@ async def async_auth_flow(self, request: httpx.Request) -> AsyncGenerator[httpx.
...
-                    if not handled:
-                        fallback_request = await self._discover_oauth_metadata_fallback()
-                        fallback_response = yield fallback_request
-                        await self._handle_oauth_metadata_response(fallback_response, is_fallback=True)
+                    oauth_discovery_stack = self._create_oauth_discovery_stack()
+                    while len(oauth_discovery_stack) > 0:
+                        oauth_discovery = oauth_discovery_stack.pop()
+                        oauth_request = await oauth_discovery()
+                        oauth_response = yield oauth_request
+                        await self._handle_oauth_metadata_response(oauth_response, oauth_discovery_stack)
```
dsp-ant: I think a for loop with a break is more readable and pythonic. I am not fully sure but I think we should be explicit about breaking out of the loop once we handle the oauth_metadata response.

src/mcp/client/auth.py
```diff
@@ -286,17 +317,37 @@ async def _discover_oauth_metadata(self) -> httpx.Request:
...
+        See https://www.rfc-editor.org/rfc/rfc8414.html#section-5
+        """
+        return await self._discover_well_known_metadata("openid-configuration")
+
+    async def _discover_oidc_metadata_fallback(self) -> httpx.Request:
+        """
+        Build fallback OIDC metadata discovery request for legacy servers.
+        See https://www.rfc-editor.org/rfc/rfc8414.html#section-5
+        """
+        url = self._build_oidc_fallback_url("openid-configuration")
```
dsp-ant: We should simplify all url building into one method. It shouldn't be complicated to generate 4 urls in one method. This avoid a lot of duplication. We should build the URLs upfront and don't requre handlers.

dsp-ant (changes requested):
I'll add a claude generated review in addition, since I reviewed it and found it quite useful.

dsp-ant:
Here is Claude's take, which in this instance, I found helpful:


Thank you for implementing RFC 9728 support with OIDC fallback functionality. While the implementation works, I have concerns about the complexity of the approach, particularly around the callback-based discovery stack pattern.

## Main Concerns

### 1. Over-engineered Discovery Stack Pattern

The current implementation uses a callback-based stack with function references:
```python
OAuthDiscoveryStack = list[Callable[[], Awaitable[httpx.Request]]]
```

This pattern, combined with stack reversal and while-loop popping, adds unnecessary complexity:
- The callbacks don't encapsulate varying behavior - they just generate URLs
- The stack reversal (`discovery_attempts.reverse()`) followed by `pop()` is non-idiomatic
- `while len(oauth_discovery_stack) > 0:` should be `while oauth_discovery_stack:`

### 2. Scattered URL Building Logic

The implementation has four separate methods for URL construction:
- `_build_well_known_path()`
- `_build_well_known_fallback_url()`
- `_build_oidc_fallback_path()`
- `_build_oidc_fallback_url()`

This fragmentation makes it harder to understand the discovery order and URL patterns.

### 3. Unnecessary State Management

The code stores intermediate state as instance variables:
```python
self.context.discovery_base_url = base_url
self.context.discovery_pathname = parsed.path
```

These are only used for fallback logic and could be handled more cleanly.

## Suggested Simplification

Replace the callback stack with a straightforward URL list approach:

```python
def _get_discovery_urls(self) -> list[tuple[str, str]]:
    """Generate ordered list of (url, type) tuples for discovery attempts."""
    urls = []
    auth_server_url = self.context.auth_server_url or self.context.server_url
    parsed = urlparse(auth_server_url)
    
    # RFC 8414: Path-aware OAuth discovery
    if parsed.path and parsed.path != "/":
        oauth_path = f"/.well-known/oauth-authorization-server{parsed.path.rstrip('/')}"
        urls.append((urljoin(base_url, oauth_path), "oauth"))
    
    # OAuth root fallback
    urls.append((urljoin(base_url, "/.well-known/oauth-authorization-server"), "oauth"))
    
    # RFC 8414 section 5: Path-aware OIDC discovery  
    if parsed.path and parsed.path != "/":
        oidc_path = f"/.well-known/openid-configuration{parsed.path.rstrip('/')}"
        urls.append((urljoin(base_url, oidc_path), "oidc"))
    
    # OIDC 1.0 fallback (appends to full URL per spec)
    oidc_fallback = urljoin(auth_server_url.rstrip("/"), "/.well-known/openid-configuration")
    urls.append((oidc_fallback, "oidc"))
    
    return urls
```

Then in `async_auth_flow()`:

```python
# Step 2: Discover OAuth metadata (with fallback for legacy servers)
discovery_urls = self._get_discovery_urls()
for url, discovery_type in discovery_urls:
    request = httpx.Request("GET", url, headers={MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION})
    response = yield request
    
    if response.status_code == 200:
        try:
            metadata = OAuthMetadata.model_validate_json(content)
            self.context.oauth_metadata = metadata
            # Apply default scope if needed
            if self.context.client_metadata.scope is None and metadata.scopes_supported is not None:
                self.context.client_metadata.scope = " ".join(metadata.scopes_supported)
            break
        except ValidationError:
            continue
    elif response.status_code != 404:
        break  # Non-404 error, stop trying
```

## Benefits of This Approach

1. **Explicit discovery order**: All URLs are visible in one place
2. **No callbacks**: Direct, simple URL generation
3. **Pythonic iteration**: Standard for-loop instead of while/pop
4. **Less state**: No instance variables for intermediate values
5. **Easier testing**: URL generation can be unit tested separately
6. **Clearer intent**: The fallback chain is immediately apparent

## Minor Issues

- Line 295: Use `return response_status == 404 and discovery_stack` instead of `len(discovery_stack) > 0`
- The comment on line 577-578 about declaring the stack backwards is a code smell

## Overall

The implementation is functionally correct and follows the RFCs properly. My suggestions are about reducing complexity and making the code more Pythonic. The callback pattern might make sense in other contexts, but here it adds unnecessary abstraction for what is essentially "try these URLs in order until one works."

Would you consider simplifying the approach? Happy to discuss further or help with the refactoring.

dsp-ant:
@ihrpr how do you want to sequence this? My suggestion is waiting for this PR to be in an acceptable state, then merge the spec change and the python sdk pr at the same time.

LucaButBoring:
Will make the changes by EOD tomorrow (juggling other things as well) - I agree with what you called out in general.

---

For context, the reason I did it like this originally was because the entire OAuth flow was duplicated in the `async_auth_flow` function due to generator syntax limitations, so I was attempting to avoid explicitly writing the entire fallback sequence twice. #1071 was merged earlier today which was able to remove the duplicated OAuth flow, so the discovery stack thing doesn't make sense anymore.

ihrpr:
> @ihrpr how do you want to sequence this? My suggestion is waiting for this PR to be in an acceptable state, then merge the spec change and the python sdk pr at the same time.

yes, that works, adding tracking for this as well

dsp-ant (approved):
LGTM

PR #1067, opened 2025-07-01
tests: add lowest dependency tests

Kludex:


.github/workflows/shared.yml
```diff
@@ -45,4 +45,7 @@ jobs:
 
       - name: Run pytest
         run: uv run --frozen --no-sync pytest
-    continue-on-error: true
```
Kludex, line 51: This shouldn't be set, should it be?

.github/workflows/shared.yml
```diff
@@ -26,6 +26,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
+    continue-on-error: true
```
Kludex, line 32: I put it up because it makes more sense since it's a job configuration, not a step one.

.github/workflows/shared.yml
Kludex: I don't think it's needed, but I figured out why it's here: so if other tests on the matrix fail, it will continue to run.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
     "anyio>=4.5",
     "httpx>=0.27",
     "httpx-sse>=0.4",
-    "pydantic>=2.7.2,<3.0.0",
+    "pydantic>=2.8.0,<3.0.0",
```
Kludex, line 28: 2.7.2 resolves one of the tests wrong, and I found that pushing up 1 minor is not that damaging - pushing it to 2.11+ would be too much right now. 

2.8.0 is from July 2024.

pyproject.toml
```diff
@@ -58,6 +58,7 @@ dev = [
     "pytest-examples>=0.0.14",
     "pytest-pretty>=1.2.0",
     "inline-snapshot>=0.23.0",
+    "dirty-equals>=0.9.0",
```
Kludex, line 60: This goes with the same line as `inline-snapshot`s. It's a utility package that helps a lot.

See: https://dirty-equals.helpmanual.io/latest/

Kludex: This is not necessary - but it did make the test cleaner. I can remove it if requested.

tests/issues/test_188_concurrency.py
```diff
@@ -46,15 +47,3 @@ async def slow_resource():
...
-def main():
-    anyio.run(test_messages_are_executed_concurrently)
-
-
-if __name__ == "__main__":
-    import logging
-
-    logging.basicConfig(level=logging.DEBUG)
-
-    main()
```
Kludex, line 60: This shouldn't have been merged - it seems it was here for debugging purposes.

tests/issues/test_188_concurrency.py
```diff
@@ -3,14 +3,15 @@
...
-)
+from mcp.shared.memory import create_connected_server_and_client_session as create_session
 
 _sleep_time_seconds = 0.01
 _resource_name = "slow://slow_resource"
 
 
+@pytest.mark.filterwarnings(
+    "ignore:coroutine 'test_messages_are_executed_concurrently.<locals>.slow_resource' was never awaited:RuntimeWarning"
+)
```
Kludex: I couldn't find the reason or package that triggered this - I tried bumping anyio, but that didn't seem to be the reason - but this seems an issue only with the test suite, so it doesn't have any impact.

tests/server/fastmcp/test_func_metadata.py
```diff
@@ -202,11 +203,8 @@ def func_dict_any() -> dict[str, Any]:
         return {"a": 1, "b": "hello", "c": [1, 2, 3]}
 
     meta = func_metadata(func_dict_any)
-    assert meta.output_schema == {
-        "additionalProperties": True,
-        "type": "object",
-        "title": "func_dict_anyDictOutput",
-    }
+
+    assert meta.output_schema == IsPartialDict(type="object", title="func_dict_anyDictOutput")
```
Kludex, line 207: This means that the `output_schema` has AT LEAST the keys I've added inside the `IsPartialDict`.

pyproject.toml
```diff
@@ -43,7 +43,6 @@ ws = ["websockets>=15.0.1"]
 mcp = "mcp.cli:app [cli]"
 
 [tool.uv]
-resolution = "lowest-direct"
```
Kludex, line 46: Now, we actually want the resolution to be the "highest" - which is the default. Because we have a test job in the pipeline that sets this resolution.

This way we can have the highest resolution on the lockfile, and actually test both highest and lowest-direct.

pyproject.toml
```diff
@@ -36,14 +36,13 @@ dependencies = [
 
 [project.optional-dependencies]
 rich = ["rich>=13.9.4"]
-cli = ["typer>=0.12.4", "python-dotenv>=1.0.0"]
+cli = ["typer>=0.16.0", "python-dotenv>=1.0.0"]
```
Kludex, line 39: I had to bump typer because click broke it recently.

I don't think bumping this package is an issue given that people can usually easily bump this without coding changes on their applications.

Kludex: pydantic or starlette would be more concerning.

ihrpr (approved):
lgtm

Issue #1052, opened 2025-06-28
The resource URL path is ignored when building the protected resource metadata URL


yurikunash:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

The current MCP Server implementation appears to use a fixed URL pattern `[domain]/.well-known/oauth-protected-resource` for the protected resource URL. After reviewing RFC9728, I believe this doesn't fully align with the specification's requirements.

**Expected behavior:**
If the resource URL is `https://resource.example.com/resource1`, the protected resource metadata URL to be `https://resource.example.com/.well-known/oauth-protected-resource/resource1`

**Current behavior:**
If the resource URL is `https://resource.example.com/resource1`, the protected resource metadata URL to be `https://resource.example.com/.well-known/oauth-protected-resource`

Link to the source code:
https://github.com/modelcontextprotocol/python-sdk/blob/794218433656554deff37477c0bef8cb7deb40f6/src/mcp/server/auth/routes.py#L219

```
return [
        Route(
            "/.well-known/oauth-protected-resource",
            endpoint=cors_middleware(handler.handle, ["GET", "OPTIONS"]),
            methods=["GET", "OPTIONS"],
        )
    ]
```

**RFC 9728**
> Protected resources supporting metadata MUST make a JSON document containing metadata as specified in [Section 2](https://datatracker.ietf.org/doc/html/rfc9728#PRMetadata) available at a **URL formed by inserting a well-known URI string into the protected resource's resource identifier between the host component and the path and/or query components, if any.**
> ...
>  The consumer of the metadata would make the following request when the resource identifier is https://resource.example.com/resource1 and the well-known URI path suffix is oauth-protected-resource to obtain the metadata, since the resource identifier contains a path component:
> 
>   GET /.well-known/oauth-protected-resource/resource1 HTTP/1.1
>   Host: resource.example.com

As the RFC correctly states: 
> Using path components enables supporting multiple resources per host. This is required in some multi-tenant hosting configurations.
which is difficult to achieve with the current implementation.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Latest version
```

pcarleton:
Hi @yurikunash thanks for opening this, I agree with you that we don't implement the more involved path discovery flow currently, and think we should add it like you've described.

[...]

PR #1071, opened 2025-07-02
Implement RFC9728 - Support WWW-Authenticate header by MCP client

### LINKED ISSUE #1054: WWW-Authentication header is not respected by Client SDK
### Initial Checks

### Description

As per MCP specification:

> MCP clients MUST be able to parse WWW-Authenticate headers and respond appropriately to HTTP 401 Unauthorized responses from the MCP server.

Link: https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization

At the same time, the Client SDK calculates the protected resource metadata URL and ignores the header:

```
    async def _discover_protected_resource(self) -> httpx.Request:
        """Build discovery request for protected resource metadata."""
        auth_base_url = self.context.get_authorization_base_url(self.context.server_url)
        url = urljoin(auth_base_url, "/.well-known/oauth-protected-resource")
        return httpx.Request("GET", url, headers={MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION})
```
Link to the source code: https://github.com/modelcontextprotocol/python-sdk/blob/794218433656554deff37477c0bef8cb7deb40f6/src/mcp/client/auth.py#L206C5-L211C1

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Letest
```

yurikunash:
Resolves #1054 
This PR implements proper WWW-Authenticate header parsing for protected resource metadata URL discovery as required by RFC9728 and the MCP specification. This change ensures full compliance with MCP authentication requirements and includes several related improvements to the authentication flow.

### Primary Changes
- ✅ **WWW-Authenticate Header Parsing**: Implemented parsing of WWW-Authenticate headers for protected resource metadata URL discovery per RFC9728
- ✅ **MCP Specification Compliance**: Now properly handles HTTP 401 Unauthorized responses as required by the MCP specification

### Indirect Improvements
- **Optimized Authorization Flow**: Stop making unnecessary Authorize requests when the initial request doesn't respond with 401 status code
- **Code Cleanup**: Removed code duplication in authentication flow logic

## Motivation and Context

As per MCP specification:
> MCP clients MUST be able to parse WWW-Authenticate headers and respond appropriately to HTTP 401 Unauthorized responses from the MCP server.

This implementation was missing from the current MCP client, creating a gap in specification compliance that has real-world implications. The absence of proper WWW-Authenticate header handling prevents hosting multiple MCP servers that share the same domain but function as separate OAuth resources. 

## How Has This Been Tested?

The changes were verified by the comprehensive unit-test coverage for different scenarios, as well as simulating the Authorization scenario by executing the client locally.

## Breaking Changes

Authorization Flow Change: The client now follows proper OAuth flow by sending an initial request to the resource and only performing authorization upon receiving a 401 status code. This aligns with OAuth specifications and security best practices.
Impact: Clients that relied on the previous behavior (where authorization was attempted regardless of initial response) may need to be updated. However, this change improves security and compliance with OAuth standards.
Migration: Most properly configured MCP servers should work without changes. Servers that expect immediate authorization without sending a 401 response should be updated to follow OAuth specifications.

## Types of changes

- [] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -520,59 +520,56 @@ class OAuthClientProvider(httpx.Auth):
             # Capture protocol version from request headers
             self.context.protocol_version = request.headers.get(MCP_PROTOCOL_VERSION)
 
+            if not self.context.is_token_valid() and self.context.can_refresh_token():
+                # Try to refresh token
+                refresh_request = await self._refresh_token()
+                refresh_response = yield refresh_request
+
+                if not await self._handle_refresh_response(refresh_response):
+                    # Refresh failed, need full re-authentication
+                    self._initialized = False
+
             if self.context.is_token_valid():
                 self._add_auth_header(request)
 
             response = yield request
 
             if response.status_code == 401:
-                if self.context.can_refresh_token():
-                    # Try to refresh token
-                    refresh_request = await self._refresh_token()
-                    refresh_response = yield refresh_request
+                # Perform full OAuth flow
+                try:
+                    # OAuth flow must be inline due to generator constraints
+                    # Step 1: Discover protected resource metadata (RFC9728 with WWW-Authenticate support)
+                    discovery_request = await self._discover_protected_resource(response)
+                    discovery_response = yield discovery_request
+                    await self._handle_protected_resource_response(discovery_response)
 
-                    if not await self._handle_refresh_response(refresh_response):
-                        # Refresh failed, need full re-authentication
-                        self._initialized = False
-                else:
-                    self.context.clear_tokens()
+                    # Step 2: Discover OAuth metadata (with fallback for legacy servers)
+                    oauth_request = await self._discover_oauth_metadata()
+                    oauth_response = yield oauth_request
+                    handled = await self._handle_oauth_metadata_response(oauth_response, is_fallback=False)
 
-                # If we don't have valid tokens after refresh, perform OAuth flow
-                if not self.context.is_token_valid():
-                    try:
-                        # OAuth flow must be inline due to generator constraints
-                        # Step 1: Discover protected resource metadata (RFC9728 with WWW-Authenticate support)
-                        discovery_request = await self._discover_protected_resource(response)
-                        discovery_response = yield discovery_request
-                        await self._handle_protected_resource_response(discovery_response)
+                    # If path-aware discovery failed with 404, try fallback to root
+                    if not handled:
+                        fallback_request = await self._discover_oauth_metadata_fallback()
+                        fallback_response = yield fallback_request
+                        await self._handle_oauth_metadata_response(fallback_response, is_fallback=True)
 
-                        # Step 2: Discover OAuth metadata (with fallback for legacy servers)
-                        oauth_request = await self._discover_oauth_metadata()
-                        oauth_response = yield oauth_request
-                        handled = await self._handle_oauth_metadata_response(oauth_response, is_fallback=False)
+                    # Step 3: Register client if needed
+                    registration_request = await self._register_client()
+                    if registration_request:
+                        registration_response = yield registration_request
+                        await self._handle_registration_response(registration_response)
 
-                        # If path-aware discovery failed with 404, try fallback to root
-                        if not handled:
-                            fallback_request = await self._discover_oauth_metadata_fallback()
-                            fallback_response = yield fallback_request
-                            await self._handle_oauth_metadata_response(fallback_response, is_fallback=True)
+                    # Step 4: Perform authorization
+                    auth_code, code_verifier = await self._perform_authorization()
 
-                        # Step 3: Register client if needed
-                        registration_request = await self._register_client()
-                        if registration_request:
-                            registration_response = yield registration_request
-                            await self._handle_registration_response(registration_response)
-
-                        # Step 4: Perform authorization
-                        auth_code, code_verifier = await self._perform_authorization()
-
-                        # Step 5: Exchange authorization code for tokens
-                        token_request = await self._exchange_token(auth_code, code_verifier)
-                        token_response = yield token_request
-                        await self._handle_token_response(token_response)
-                    except Exception:
-                        logger.exception("OAuth flow error")
-                        raise
+                    # Step 5: Exchange authorization code for tokens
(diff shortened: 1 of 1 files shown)


dead8309:
Hi, Any updates on it ? @ihrpr 

we rely on correct parsing of Auth header

lucheng-hugh:
@ihrpr Can we raise attention to this bug? It blocks the MCP client from correctly finding the resource metadata, leading to incorrect AS. Thank you

src/mcp/client/auth.py
```diff
@@ -490,92 +520,60 @@ async def async_auth_flow(self, request: httpx.Request) -> AsyncGenerator[httpx.
...
+
             response = yield request
 
-            # Handle 401 responses
-            if response.status_code == 401 and self.context.can_refresh_token():
-                # Try to refresh token
-                refresh_request = await self._refresh_token()
-                refresh_response = yield refresh_request
+            if response.status_code == 401:
+                if self.context.can_refresh_token():
```
pcarleton: I'm not sure I agree with this part... if we know our current tokens are invalid, and we know we can refresh them, shouldn't we do that before sending a request we know is going to 401? 

I think that would mean moving this `if/else` block ahead of line 526 where we yield request.

pcarleton (changes requested):

src/mcp/client/auth.py
yurikunash: Thank you for the suggestion @pcarleton 
I was hesitant about it, so I implemented the "try first" approach, as it seemed more reliable.
As this [article](https://www.oauth.com/oauth2-servers/making-authenticated-requests/refreshing-an-access-token/) states, it's also perfectly fine to have that kind of optimization and refresh the token preemptively, so I updated the code as you proposed.

pcarleton (approved):
LGTM

Thanks for the thorough PR with tests, and being open to feedback!

Note to self: we'll still need to follow up to fix the path-based URL issue https://github.com/modelcontextprotocol/python-sdk/issues/1052

dsp-ant:
# Review of PR #1071: OIDC Fallback Implementation

Thank you for implementing RFC 9728 support with OIDC fallback functionality. While the implementation works, I have concerns about the complexity of the approach, particularly around the callback-based discovery stack pattern.

## Main Concerns

### 1. Over-engineered Discovery Stack Pattern

The current implementation uses a callback-based stack with function references:
```python
OAuthDiscoveryStack = list[Callable[[], Awaitable[httpx.Request]]]
```

This pattern, combined with stack reversal and while-loop popping, adds unnecessary complexity:
- The callbacks don't encapsulate varying behavior - they just generate URLs
- The stack reversal (`discovery_attempts.reverse()`) followed by `pop()` is non-idiomatic
- `while len(oauth_discovery_stack) > 0:` should be `while oauth_discovery_stack:`

### 2. Scattered URL Building Logic

The implementation has four separate methods for URL construction:
- `_build_well_known_path()`
- `_build_well_known_fallback_url()`
- `_build_oidc_fallback_path()`
- `_build_oidc_fallback_url()`

This fragmentation makes it harder to understand the discovery order and URL patterns.

### 3. Unnecessary State Management

The code stores intermediate state as instance variables:
```python
self.context.discovery_base_url = base_url
self.context.discovery_pathname = parsed.path
```

These are only used for fallback logic and could be handled more cleanly.

## Suggested Simplification

Replace the callback stack with a straightforward URL list approach:

```python
def _get_discovery_urls(self) -> list[tuple[str, str]]:
    """Generate ordered list of (url, type) tuples for discovery attempts."""
    urls = []
    auth_server_url = self.context.auth_server_url or self.context.server_url
    parsed = urlparse(auth_server_url)
    
    # RFC 8414: Path-aware OAuth discovery
    if parsed.path and parsed.path != "/":
        oauth_path = f"/.well-known/oauth-authorization-server{parsed.path.rstrip('/')}"
        urls.append((urljoin(base_url, oauth_path), "oauth"))
    
    # OAuth root fallback
    urls.append((urljoin(base_url, "/.well-known/oauth-authorization-server"), "oauth"))
    
    # RFC 8414 section 5: Path-aware OIDC discovery  
    if parsed.path and parsed.path != "/":
        oidc_path = f"/.well-known/openid-configuration{parsed.path.rstrip('/')}"
        urls.append((urljoin(base_url, oidc_path), "oidc"))
    
    # OIDC 1.0 fallback (appends to full URL per spec)
    oidc_fallback = urljoin(auth_server_url.rstrip("/"), "/.well-known/openid-configuration")
    urls.append((oidc_fallback, "oidc"))
    
    return urls
```

Then in `async_auth_flow()`:

```python
# Step 2: Discover OAuth metadata (with fallback for legacy servers)
discovery_urls = self._get_discovery_urls()
for url, discovery_type in discovery_urls:
    request = httpx.Request("GET", url, headers={MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION})
    response = yield request
    
    if response.status_code == 200:
        try:
            metadata = OAuthMetadata.model_validate_json(content)
            self.context.oauth_metadata = metadata
            # Apply default scope if needed
            if self.context.client_metadata.scope is None and metadata.scopes_supported is not None:
                self.context.client_metadata.scope = " ".join(metadata.scopes_supported)
            break
        except ValidationError:
            continue
    elif response.status_code != 404:
        break  # Non-404 error, stop trying
```

## Benefits of This Approach

1. **Explicit discovery order**: All URLs are visible in one place
2. **No callbacks**: Direct, simple URL generation
3. **Pythonic iteration**: Standard for-loop instead of while/pop
4. **Less state**: No instance variables for intermediate values
5. **Easier testing**: URL generation can be unit tested separately
6. **Clearer intent**: The fallback chain is immediately apparent

## Minor Issues

- Line 295: Use `return response_status == 404 and discovery_stack` instead of `len(discovery_stack) > 0`
- The comment on line 577-578 about declaring the stack backwards is a code smell

## Overall

The implementation is functionally correct and follows the RFCs properly. My suggestions are about reducing complexity and making the code more Pythonic. The callback pattern might make sense in other contexts, but here it adds unnecessary abstraction for what is essentially "try these URLs in order until one works."

Would you consider simplifying the approach? Happy to discuss further or help with the refactoring.

PR #1078, opened 2025-07-03
Improve child process termination on POSIX & Windows

felixweinberger:
## Motivation and Context

#729, and #850 address important failures to shut down childprocesses when MCP servers spawn subprocesses.

This is not a trivial thing to fix, as there's no great cross-platform library for terminating process-trees on all platforms so we have to fork between Unix (using `os.killpg`) and Windows (using `JobObjects` via `pywin32` which is the best supported external library for this: https://github.com/mhammond/pywin32)

This PR adds regressions tests to clearly demonstrate the edge cases addressed + unifies the approaches from the PRs listed above.

## How Has This Been Tested?

New regression tests added (each fails without the relevant fix). These were developed and tested on _both_ POSIX and Windows systems.

### For #729 and #850 (properly terminate processes and their children across platforms):
- Original authors: @jingx8885, @surya-prakash-susarla

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import signal
 import sys
 from contextlib import asynccontextmanager
 from pathlib import Path
@@ -14,16 +13,16 @@ from anyio.streams.text import TextReceiveStream
 from pydantic import BaseModel, Field
 
 import mcp.types as types
-from mcp.shared.message import SessionMessage
-
-from .win32 import (
+from mcp.os.posix.utilities import terminate_posix_process_tree
+from mcp.os.win32.utilities import (
     FallbackProcess,
     create_windows_process,
     get_windows_executable_command,
     terminate_windows_process_tree,
 )
+from mcp.shared.message import SessionMessage
 
-logger = logging.getLogger("client.stdio")
+logger = logging.getLogger(__name__)
 
 # Environment variables to inherit by default
 DEFAULT_INHERITED_ENV_VARS = (
@@ -247,47 +246,20 @@ async def _create_platform_compatible_process(
     return process
 
 
-async def _terminate_process_tree(process: Process | FallbackProcess, timeout: float = 2.0) -> None:
+async def _terminate_process_tree(process: Process | FallbackProcess, timeout_seconds: float = 2.0) -> None:
     """
     Terminate a process and all its children using platform-specific methods.
 
     Unix: Uses os.killpg() for atomic process group termination
     Windows: Uses Job Objects via pywin32 for reliable child process cleanup
+
+    Args:
+        process: The process to terminate
+        timeout_seconds: Timeout in seconds before force killing (default: 2.0)
     """
     if sys.platform == "win32":
-        await terminate_windows_process_tree(process, timeout)
+        await terminate_windows_process_tree(process, timeout_seconds)
     else:
-        pid = getattr(process, "pid", None) or getattr(getattr(process, "popen", None), "pid", None)
-        if not pid:
-            return
-
-        try:
-            pgid = os.getpgid(pid)
-            os.killpg(pgid, signal.SIGTERM)
-
-            with anyio.move_on_after(timeout):
-                while True:
-                    try:
-                        # Check if process group still exists (signal 0 = check only)
-                        os.killpg(pgid, 0)
-                        await anyio.sleep(0.1)
-                    except ProcessLookupError:
-                        return
-
-            try:
-                os.killpg(pgid, signal.SIGKILL)
-            except ProcessLookupError:
-                pass
-
-        except (ProcessLookupError, PermissionError, OSError) as e:
-            logger.warning(f"Process group termination failed for PID {pid}: {e}, falling back to simple terminate")
-            try:
-                process.terminate()
-                with anyio.fail_after(timeout):
-                    await process.wait()
-            except Exception as term_error:
-                logger.warning(f"Process termination failed for PID {pid}: {term_error}, attempting force kill")
-                try:
-                    process.kill()
-                except Exception as kill_error:
-                    logger.error(f"Failed to kill process {pid}: {kill_error}")
+        # FallbackProcess should only be used for Windows compatibility
+        assert isinstance(process, Process)
+        await terminate_posix_process_tree(process, timeout_seconds)
--- a/src/mcp/os/__init__.py
+++ b/src/mcp/os/__init__.py
@@ -1,0 +1,1 @@
+"""Platform-specific utilities for MCP."""
--- a/src/mcp/os/posix/__init__.py
+++ b/src/mcp/os/posix/__init__.py
@@ -1,0 +1,1 @@
+"""POSIX-specific utilities for MCP."""
--- a/src/mcp/os/posix/utilities.py
+++ b/src/mcp/os/posix/utilities.py
@@ -1,0 +1,60 @@
+"""
+POSIX-specific functionality for stdio client operations.
+"""
+
+import logging
+import os
+import signal
+
+import anyio
+from anyio.abc import Process
+
+logger = logging.getLogger(__name__)
+
+
+async def terminate_posix_process_tree(process: Process, timeout_seconds: float = 2.0) -> None:
+    """
+    Terminate a process and all its children on POSIX systems.
+
+    Uses os.killpg() for atomic process group termination.
+
+    Args:
+        process: The process to terminate
+        timeout_seconds: Timeout in seconds before force killing (default: 2.0)
+    """
+    pid = getattr(process, "pid", None) or getattr(getattr(process, "popen", None), "pid", None)
+    if not pid:
+        # No PID means there's no process to terminate - it either never started,
+        # already exited, or we have an invalid process object
+        return
+
+    try:
+        pgid = os.getpgid(pid)
+        os.killpg(pgid, signal.SIGTERM)
+
+        with anyio.move_on_after(timeout_seconds):
+            while True:
+                try:
+                    # Check if process group still exists (signal 0 = check only)
+                    os.killpg(pgid, 0)
+                    await anyio.sleep(0.1)
+                except ProcessLookupError:
+                    return
+
+        try:
+            os.killpg(pgid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+
+    except (ProcessLookupError, PermissionError, OSError) as e:
+        logger.warning(f"Process group termination failed for PID {pid}: {e}, falling back to simple terminate")
+        try:
+            process.terminate()
+            with anyio.fail_after(timeout_seconds):
+                await process.wait()
+        except Exception as term_error:
+            logger.warning(f"Process termination failed for PID {pid}: {term_error}, attempting force kill")
+            try:
+                process.kill()
+            except Exception as kill_error:
+                logger.error(f"Failed to kill process {pid}: {kill_error}")
--- a/src/mcp/os/win32/__init__.py
+++ b/src/mcp/os/win32/__init__.py
@@ -1,0 +1,1 @@
+"""Windows-specific utilities for MCP."""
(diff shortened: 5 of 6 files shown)


felixweinberger:
@dsp-ant as discussed platform native ways to terminate processes for Unix & Windows without race conditions between child enumeration and parent shutdown.

@jingx8885 I had to break this out of #1044 to isolate this particular problem. If you'd be willing to test whether this branch still resolves your issue like you did on #1044 that would be greatly appreciated! The tests still pass, but just to make sure there's nothing we missed.

jingx8885:
@felixweinberger got it, I'll deal with it in a few days.

felixweinberger:
@Kludex @agronholm given I've seen you both on Python process management related PRs, tagging y'all here in case you have any thoughts on whether this could or should be simpler with `anyio` 🙂

This is an edge case where terminating a process doesn't automatically terminate children via `anyio`'s `terminate()`, but I'm also not sure whether it should in the first place as Python's `asyncio` doesn't do so either AFAIK.

For context, we need the subprocesses to be killed in the `stdio` case of MCP, because the MCP "server" is started as a subprocess of the "client". When spinning down the client, it should also spin down that server to free up resources.

tests/client/test_stdio.py
```diff
@@ -212,3 +215,306 @@ def sigint_handler(signum, frame):
             )
         else:
             raise
+
+
+class TestChildProcessCleanup:
```
felixweinberger, line 227: Aware that using "timeout" anywhere in a test is usually not great practice - however, cross platform challenges and the subject matter being tested (processes) make this a little trickier to resolve than usual, as these tests are fundamentally there to detect whether processes are being closed (which in these tests should take a negligible, but non-zero amount of time).

I ran a deep research on alternative approaches - here's @claude 's POV:

<details>

> ⏺ Deep Analysis: Is the Time-Based Testing Approach Justified?
> 
>   After researching testing best practices and ultrathinking about this specific case, yes, the approach is justified. Here's why:
> 
>   The Tests Are Actually Condition-Based, Not Time-Based
> 
>   The tests check a condition: "Is the file size stable?" This indicates process termination. The sleep() calls create measurement windows, not synchronization points. This is fundamentally
>   different from hoping timing works out.
> 
>   Why Alternative Approaches Don't Work Here
> 
>   1. Process polling (poll()/wait()): Only works for known process handles. This PR tests termination of entire process trees including unknown grandchildren.
>   2. Mock-based testing: Process groups and Windows Job Objects require real OS interaction. The bugs being fixed are OS-level issues that mocks can't reproduce.
>   3. Event/signal-based coordination: Would require modifying child processes, but the goal is to test termination of arbitrary processes (like npx).
>   4. OS monitoring APIs: Platform-specific and still require polling. The file-based approach is simpler and more portable.
> 
>   The Approach Is Well-Designed
> 
>   - Black-box testing: Verifies actual behavior (processes stop working) rather than just process state
>   - Cross-platform: Works identically on Windows and Unix without platform-specific test code
>   - Real-world scenario: Tests continuous I/O patterns common in development servers
>   - Minimal timing: Uses 0.1s intervals and reasonable timeouts
>   - Comprehensive: Tests single children, nested trees, and race conditions
> 
>   Critical Context
> 
>   These tests provide regression coverage for production bugs where child processes weren't properly terminated. The ability to reliably test this behavior outweighs the theoretical concerns
>   about time-based tests.
> 
>   Recommendations
> 
>   1. Consider adding a comment explaining why the file-based approach is used:
>   ```
>   # We use file writing as a cross-platform way to verify process liveness
>   # without needing to know child PIDs or use platform-specific APIs
>   ```
>   2. The ResourceWarning filter is appropriately documented and justified.
>   3. Future enhancement: Consider adding a process count assertion on Unix using ps to supplement the file-based checks.
> 
>   Verdict
> 
>   This PR provides a robust, well-tested solution to critical process management bugs. The testing approach, while using timing windows, is actually condition-based and represents the most
>   pragmatic way to verify process tree termination across platforms. The implementation quality is high with proper platform abstractions and error handling.
> 

</details>

src/mcp/client/stdio/__init__.py
```diff
@@ -219,11 +226,65 @@ async def _create_platform_compatible_process(
...
+        pid = getattr(process, "pid", None) or getattr(getattr(process, "popen", None), "pid", None)
+        if not pid:
+            return
+
+        try:
+            pgid = os.getpgid(pid)
+            os.killpg(pgid, signal.SIGTERM)
+
+            deadline = anyio.current_time() + timeout
+            while anyio.current_time() < deadline:
```
agronholm: Have you considered `anyio.move_on_after()`?

felixweinberger: Great point! I've actually used it elsewhere but forgot about it here - will amend.

src/mcp/client/stdio/__init__.py
```diff
@@ -219,11 +226,65 @@ async def _create_platform_compatible_process(
...
+
+    Unix: Uses os.killpg() for atomic process group termination
+    Windows: Uses Job Objects via pywin32 for reliable child process cleanup
+    """
+    if sys.platform == "win32":
+        await terminate_windows_process_tree(process, timeout)
+    else:
+        pid = getattr(process, "pid", None) or getattr(getattr(process, "popen", None), "pid", None)
+        if not pid:
+            return
```
dsp-ant: Is not having a PID mean that the process is dead?

src/mcp/client/stdio/__init__.py
ihrpr: this wasn't applied, right?

src/mcp/client/stdio/win32.py
```diff
@@ -13,6 +14,23 @@
...
 from anyio.streams.file import FileReadStream, FileWriteStream
 
+logger = logging.getLogger("client.stdio.win32")
+
+# Windows-specific imports for Job Objects
+if sys.platform == "win32":
+    import pywintypes
+    import win32api
+    import win32con
+    import win32job
```
jingx8885, line 25: @felixweinberger  I saw you used win32api instead of psutil,  and used win32job.TerminateJobObject in terminate_windows_process_tree function to kill all processes, which is more simple and elegance, I tested in local and it‘s worked, great job!

felixweinberger: Thank you for testing it, really appreciate you taking the time!

src/mcp/client/stdio/__init__.py
felixweinberger: amended now!

ihrpr (dismissed):
Thank you!

src/mcp/client/stdio/__init__.py
```diff
@@ -14,10 +17,14 @@
 from mcp.shared.message import SessionMessage
 
 from .win32 import (
+    FallbackProcess,
     create_windows_process,
     get_windows_executable_command,
+    terminate_windows_process_tree,
 )
 
+logger = logging.getLogger("client.stdio")
```
dsp-ant: any particular reason we are providing a name instead of allowing Python to pick a name?

src/mcp/client/stdio/__init__.py
```diff
@@ -187,7 +194,7 @@ async def stdin_writer():
                     await process.wait()
             except TimeoutError:
                 # If process doesn't terminate in time, force kill it
-                process.kill()
+                await _terminate_process_tree(process)
```
dsp-ant, line 196: why the underscore?

src/mcp/client/stdio/__init__.py
```diff
@@ -222,11 +229,65 @@ async def _create_platform_compatible_process(
...
+            env=env,
+            stderr=errlog,
+            cwd=cwd,
+            start_new_session=True,
+        )
 
     return process
+
+
+async def _terminate_process_tree(process: Process | FallbackProcess, timeout: float = 2.0) -> None:
```
dsp-ant: why not 
```suggestion
async def terminate_process_tree(process: Process | FallbackProcess, timeout: float = 2.0) -> None:
```

dsp-ant: Also, I know i am that person, but I really like having proper types for variables when possible. A timeout of a float can be anything, milliseconds, seconds, etc. In fact I just fixed a bug in a separate codebase where a float timeout was millisecond but considered seconds. I personally would use `time.Duration` and then use `timeout.total_seconds()` to pass to `move_on_after`.

src/mcp/client/stdio/__init__.py
```diff
@@ -222,11 +229,65 @@ async def _create_platform_compatible_process(
...
+            try:
+                process.terminate()
+                with anyio.fail_after(timeout):
+                    await process.wait()
+            except Exception as term_error:
+                logger.warning(f"Process termination failed for PID {pid}: {term_error}, attempting force kill")
+                try:
+                    process.kill()
+                except Exception as kill_error:
+                    logger.error(f"Failed to kill process {pid}: {kill_error}")
```
dsp-ant: makle this separate methods, one win32.terminate_process_tree() and one `posix.terminate_process_tree`, allowing us to separate platform specific things into a platforms directory

We likely want something like 
```
src/
  - os/
    - win32/utilites.py
    - posix/utilities.py
```

src/mcp/client/stdio/__init__.py
agronholm: I guess you meant `timedelta`? And this is for the internal API, isn't it? I'm guessing that's why it had the leading underscore. And timeouts are normally expressed in seconds (given that both the stdlib and AnyIO use it for timeouts/sleep calls); milliseconds as unit are much less often used. Documenting the unit is probably good enough, as importing `timedelta` may feel bothersome.

src/mcp/client/stdio/__init__.py
felixweinberger: `_prefix` is usually to mark something as a private method / function in Python (though there's no specific enforcement of it).

This is a client internal implementation detail I felt we wanted to keep internal and not have external programs using.

src/mcp/client/stdio/__init__.py
sungigant: > Also, I know i am that person, but I really like having proper types for variables when possible. A timeout of a float can be anything, milliseconds, seconds, etc. In fact I just fixed a bug in a separate codebase where a float timeout was millisecond but considered seconds. I personally would use `time.Duration` and then use `timeout.total_seconds()` to pass to `move_on_after`.

src/mcp/client/stdio/__init__.py
sungigant: > makle this separate methods, one win32.terminate_process_tree() and one `posix.terminate_process_tree`, allowing us to separate platform specific things into a platforms directory
> 
> We likely want something like
> 
> ```
> src/
>   - os/
>     - win32/utilites.py
>     - posix/utilities.py
> ```

src/mcp/client/stdio/__init__.py
felixweinberger: added a comment

src/mcp/client/stdio/__init__.py
felixweinberger: Renamed to `timeout_seconds` for clarity and kept it as `float` to match existing python stdlib conventions

src/mcp/client/stdio/__init__.py
felixweinberger: fair point, I just spotted `getLogger("client")` and `getLogger("server")` elsewhere in the code so I just copied that pattern, but `__name__` is perfectly adequate, amended.

src/mcp/client/stdio/__init__.py
felixweinberger: added a comment

src/mcp/client/stdio/__init__.py
felixweinberger: Amended! thanks for the suggestion, much cleaner

dsp-ant (approved):

surya-prakash-susarla:
Thank you for pushing the fix!

sungigant:
[Uploading hello-vault-a8a65c85bbae57c879fa3c42e0320353623159ea.zip…]()

PR #1089, opened 2025-07-04
Implemented runtime tools

NikBellini:
This feature implements runtime tools. Basically generates the tools at runtime giving for example the possibility of showing a list of tools instead of others based on the user or other factors.

## Motivation and Context

We have a very long list of tools available, but this tools must be available based on the preferences of the user. This way the LLM won't see tools that the user doesn't want or doesn't have access to, avoiding possible wrong calls.

## How Has This Been Tested?

For now this has been tested in a small demo. The scenario is showing the available tools based on the user.

## Breaking Changes

No, no need to update the code. This will only add a functionality

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,36 @@ def get_temperature(city: str) -> float:
     # Returns: {"result": 22.5}
 ```
 
+#### Runtime tools
+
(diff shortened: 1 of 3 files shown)


dsp-ant:
I don't think we want to support this directly in the SDK. Implementors who are interested in this functionality should either use the lowlevel server directly, or inherit from FastMCP and change the appropriate functions.

NikBellini:
Absolutely, no problem!

Issue #1073, opened 2025-07-03
Inconsistent Field Naming Between `outputSchema` and `model_dump`


akolotov:
If the model to return by a tool is defined like:

```python
class Foo(BaseModel):
    field_first: str | None = Field(
        default=None, alias="first", description="The first field."
    )
    field_second: str | None = Field(
        default=None, alias="second", description="The second field."
    )
```

In `outputSchema` of `tools/list` they will be displayed as "first" and "second".

At the same time `model_dump` in
https://github.com/modelcontextprotocol/python-sdk/blob/6f43d1f17022d9d45b51955304e39ad43ca48014/src/mcp/server/fastmcp/utilities/func_metadata.py#L114
will produce the dump with "field_first" and "field_second".

To keep the schema and the tool response consistent, consider adding `by_alias=True` to the call of `model_dump`.

PR #1099, opened 2025-07-07
Fix Pydantic field alias consistency in structured output

sjwhole:
# Fix Pydantic field alias consistency in structured output

## Summary
Fixes issue #1073. Fix inconsistency between schema and structured output when using Pydantic models with field aliases in tool return types. The schema would display aliased field names but the actual output would use original field names.
## Motivation and Context

When defining Pydantic models with field aliases for tool return types, there was an inconsistency where:
- The output schema would correctly show aliased field names (e.g., "first", "second") 
- The actual structured tool output would use the original field names (e.g., "field_first", "field_second")

This inconsistency made it difficult for clients to reliably consume structured tool outputs, as the field names in the schema didn't match the actual response data.

## How Has This Been Tested?

- Added comprehensive test case `test_structured_output_aliases()` that verifies both schema generation and output use consistent alias names
- All existing pytest tests continue to pass

## Breaking Changes

No breaking changes. This is a bug fix that makes the behavior consistent with what the schema advertises. Users who were working around this inconsistency may need to update their code to use the aliased field names consistently.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix was a simple one-line change adding `by_alias=True` to the `model_dump()` call in `func_metadata.py:114`. This ensures that when Pydantic models are serialized for structured output, they use the same field names that appear in the generated JSON schema.

The test case covers both the schema generation and the actual output conversion to ensure they remain consistent. This addresses the core issue described in the original bug report where the schema and output used different field naming conventions.

### Before the fix:
```python
class Foo(BaseModel):
    field_first: str | None = Field(default=None, alias="first", description="The first field.")
    field_second: str | None = Field(default=None, alias="second", description="The second field.")
```

- Schema would show: `"first"` and `"second"`
- Output would contain: `"field_first"` and `"field_second"`

### After the fix:
- Both schema and output consistently use: `"first"` and `"second"`

### Code change:
```python
# Before
structured_content = validated.model_dump(mode="json")

# After  
structured_content = validated.model_dump(mode="json", by_alias=True)
```

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -111,7 +111,7 @@ class FuncMetadata(BaseModel):
 
             assert self.output_model is not None, "Output model must be set if output schema is defined"
             validated = self.output_model.model_validate(result)
-            structured_content = validated.model_dump(mode="json")
+            structured_content = validated.model_dump(mode="json", by_alias=True)
 
             return (unstructured_content, structured_content)
 
--- a/tests/server/fastmcp/test_func_metadata.py
+++ b/tests/server/fastmcp/test_func_metadata.py
@@ -839,3 +839,48 @@ def test_structured_output_unserializable_type_error():
         func_metadata(func_returning_namedtuple, structured_output=True)
     assert "is not serializable for structured output" in str(exc_info.value)
     assert "Point" in str(exc_info.value)
+
+
+def test_structured_output_aliases():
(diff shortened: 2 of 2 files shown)


sjwhole:
I’m wondering why I can’t get some tests to pass on Windows.

> Thank you for working on this!
> 
> QQ:
> 
> > Added comprehensive test case test_structured_output_aliases() that verifies both schema generation and output use consistent alias names
> 
> I don't see any tests added, please can this be checked?

PR #1102, opened 2025-07-07
Extract BaseSession request ID increment logic

ronmrdechai:
See title

## Motivation and Context
This allows the request ID increment logic to be overwritten by
subclasses. I have a unique need to share sessions' underlying input
and output streams, but not the sessions themselves. This would require
keeping track of request IDs globally to avoid overlap.

Having this logic overridable lets me do that.

## How Has This Been Tested?
CI

## Breaking Changes
N/A

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A


ronmrdechai:
For reference, the way I've worked around this currently is that I added these two methods to a subclass of `BaseSession`:

```
class MySession(ClientSession):
    _global_request_id: int = 1

    @override
    @property
    def _request_id(self) -> int:
        """
        This setter/getter pair ensures request IDs are globally unique for the
        lifetime of the process. Client session channels for the same MCP server
        are shared, so tracking the request ID globally is required.
        """
        return MySession._global_request_id
​​​
    @_request_id.setter
    def _request_id(self, value: int) -> None:
        diff = value - MySession._global_request_id
        MySession._global_request_id += diff
```

This is obviously brittle and I'd like a proper interface for it.

felixweinberger:
Hi @ronmrdechai thank you for this contribution! Apologies for the time it took to get back to this PR.

The solution you have seems like the right way to handle very application specific needs - I don't think we want to take on this flexibility inside the SDK without additional motivation that this is a common request / use case.

PR #865, opened 2025-05-31
fix: Expose completions capability

fredericbarthelet:
Advertising support for completions in servers where a completion handler has been defined.
Similar to https://github.com/modelcontextprotocol/typescript-sdk/pull/546

## Motivation and Context
While implementing https://github.com/modelcontextprotocol/inspector/pull/441, I found out that the only way to advertise completions support was defining it manually in constructor argument.

## How Has This Been Tested?
Added corresponding unit tests.

## Breaking Changes
None

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1107, opened 2025-07-09
Fix completions

ihrpr:
We merged https://github.com/modelcontextprotocol/python-sdk/pull/865 and the PR had an old interface for the completions (pre spec 2025-08-16)

Something is wrong with CI - checking as a separate PR - this should have been caught, but main seems green

(CI fix is here https://github.com/modelcontextprotocol/python-sdk/pull/1108)


felixweinberger (approved):
LGTM

PR #1113, opened 2025-07-09
Avoiding to parse string arguments containing valid JSON if the argument annotation is str

cjg:
Every time a parameter of a tool is passed as a string containing a valid JSON it is parsed and replaced. From the comment of pre_parse_json() "This is to handle cases like `["a", "b", "c"]` being passed in as JSON inside a string rather than an actual list"

Nevertheless this method doesn't check if the annotation of parameters indicates that it is supposed to be a str rather than a dict or a list

## Types of changes
- [ ] Documentation update

## Checklist


bhosmer-ant (approved):
@cjg thank you for this!

Issue #1076, opened 2025-07-03
**Memory Leak**  in MCP Server – will be a memory leak until OOM (Out of Memory) occurs


MadanMaram:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

As part of our usage of FastMCP and Python MCP SDK, we observed a memory leak in the MCP server. When the client continuously calls a simple tool (e.g., a greet function), the server's memory usage steadily increases over time and does not get released, eventually leading to an Out of Memory (OOM) condition.

This issue is reproducible both in the fastmcp library and the  MCP SDK.

### Issue
Here is the Logs:Memory will keep increasing
[memory_log_pythonmcp.txt](https://github.com/user-attachments/files/21033314/memory_log_pythonmcp.txt)

### Example Code

```Python
Server.py
----------------------------------------------------
from mcp.server.fastmcp import FastMCP

# Create an MCP server
mcp = FastMCP("Demo Server")

# Add a simple tool
@mcp.tool()
def greet(name: str) -> str:
    """Return a greeting message."""
    return f"Hello, {name}!"

if __name__ == "__main__":
    # Run the server with streamable HTTP transport
    mcp.run(transport="streamable-http")

--------------------------------------------------
client.py(（with a while loop）
---------------------------------------------------
import asyncio
from mcp.client.streamable_http import streamablehttp_client
from mcp.client.session import ClientSession

async def call_tool(name: str):
    # Connect to the MCP server using streamable HTTP
    async with streamablehttp_client("http://127.0.0.1:8000/mcp") as (read, write, _):
        async with ClientSession(read, write) as session:
            # Initialize the connection
            await session.initialize()
            # Call the 'greet' tool
            result = await session.call_tool("greet", {"name": name})
            print(result)

async def main():
    while True:
        await call_tool("Bob")

if __name__ == "__main__":
    asyncio.run(main())
--------------------------------------------
```

### Python & MCP Python SDK

```Text
Environment:

Python: 3.12.1

MCP Python SDK: 1.10.1
```

yaronday:
Hi @MadanMaram,

If I understood the observed issue correctly, it looks like your current client implementation simply creates a new connection per call, which might result in such a monotonic memory usage growth on the MCP server. 

[...]

PR #1116, opened 2025-07-10
Close server session after handle stateless request

hopeful0:
## Motivation and Context

The `ServerSession` created by stateless Streamable HTTP requests is not being properly cleaned up, leading to memory leaks, which may be related to #1067.
This PR allows the transport to immediately close the session after handling the request, which I believe is reasonable, as the session will no longer be needed once the request has been processed.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

At the same time, for stateful requests, when the client does not send a DELETE request due to abnormal termination or network issues, the service session will not be cleaned up either, which similarly leads to memory leaks.
I think a heartbeat mechanism can be used to determine whether the client is alive. allowing the server to timely clean up expired sessions. I would be happy to open a new PR for this or continue addressing this issue in the current PR.

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -190,6 +190,9 @@ class StreamableHTTPSessionManager:
         # Handle the HTTP request and return the response
         await http_transport.handle_request(scope, receive, send)
 
+        # Terminate the session after the request is handled
+        await http_transport._terminate_session()  # type: ignore[reportPrivateUsage]
+
     async def _handle_stateful_request(
         self,
         scope: Scope,


src/mcp/server/streamable_http_manager.py
```diff
@@ -190,6 +190,9 @@ async def run_stateless_server(*, task_status: TaskStatus[None] = anyio.TASK_STA
         # Handle the HTTP request and return the response
         await http_transport.handle_request(scope, receive, send)
 
+        # Terminate the session after the request is handled
+        await http_transport._terminate_session()  # type: ignore[reportPrivateUsage]
```
felixweinberger: Can we create a public `cleanup()` method in `StreamableHTTPSeverTransport` and use that internally inside `_terminate_session()` instead of using a private method? 

There's no "Session" in a stateless request, so I think we shouldn't be terminating a non-existent session and extract the cleanup logic instead.

Something like this:

```python
# streamable_http.py

    async def cleanup(self) -> None:
        """Clean up all transport resources (streams, connections).

        This method closes all streams and cleans up resources but does NOT
        mark the session as terminated. Use this for stateless cleanup.
        """
        # We need a copy of the keys to avoid modification during iteration
        request_stream_keys = list(self._request_streams.keys())

        # Close all request streams asynchronously
        for key in request_stream_keys:
            await self._clean_up_memory_streams(key)

        # Clear the request streams dictionary immediately
        self._request_streams.clear()
        try:
            if self._read_stream_writer is not None:
                await self._read_stream_writer.aclose()
            if self._read_stream is not None:
                await self._read_stream.aclose()
            if self._write_stream_reader is not None:
                await self._write_stream_reader.aclose()
            if self._write_stream is not None:
                await self._write_stream.aclose()
        except Exception as e:
            # During cleanup, we catch all exceptions since streams might be in various states
            logger.debug(f"Error closing streams: {e}")

    async def _terminate_session(self) -> None:
        """Terminate the current session, closing all streams.

        Once terminated, all requests with this session ID will receive 404 Not Found.
        """
        self._terminated = True
        logger.info(f"Terminating session: {self.mcp_session_id}")
        await self.cleanup()
```

felixweinberger (changes requested):
Thank you for this contribution!

Do you have a way of reproducing this issue? An regression test as part of this PR would be great.

src/mcp/server/streamable_http_manager.py
hopeful0: Thank you for your review, I strongly agree with avoiding calls to internal methods.

> There's no "Session" in a stateless request

I agree with your point, but I think we can also view it as a temporary session that only exists during this stateless request

> Can we create a public `cleanup()` method in `StreamableHTTPSeverTransport` and use that internally inside `_terminate_session()` instead of using a private method?

I prefer to modify the `_terminal_session` method directly to be public rather than extracting another method from it, because the transport itself should also be terminated after handling the stateless request, which means the `_terminated` should also be set to `True`. I think `terminate` might be a good name, because we actually directly terminate the transport itself, and the session is automatically terminated due to the stream being closed.

And this terminated method needs to be called not only in stateless requests but also in stateful requests. If the server-side manager cannot actively clean up these transports or sessions, it means that as long as the client does not send delete requests, they will exist in memory **FOREVER**, and the same problem exists even in stateful requests.

I think a *"session"* can exist forever, but a *session instance* should not.

hopeful0:
> Do you have a way of reproducing this issue? 

You can use the *Example Code* in #1076, just change the FastMCP server as stateless

```
mcp = FastMCP("Demo Server", stateless_http=True)
```

Or for stateful servers, just add `os._exit` after `print(result)` on the client side to simulate a program crash without an opportunity to exit the context manager.

> An regression test as part of this PR would be great.

I tried many methods, but there was no way to implement a reasonable test, because in a `StreamableHTTPSessionManager`, I couldn't get a `StreamableHTTPServerTransport` instance (which only exists in `_handle_stateless_request`), nor could I get a `ServerSession` instance (which only exists in the `Server.run` function). Test cases similar to `test_stateless_session_cleanup_on_graceful_exit` could only get a `StreamableHTTPSessionManager` and a `ServerSession`.

src/mcp/server/streamable_http_manager.py
felixweinberger: Sure, works for me! Do you want to make that change to this PR? In that case rename `_terminate_session` to just `terminate()`

hopeful0: Thank you, I have submitted this change.

I am handling the termination of a stateful server session. Do you think I should continue in this PR or open a new one?

felixweinberger (approved):
Thanks again for the contribution! I'll figure out how we can have at least some kind of basic test for this :)

felixweinberger:
I made an (imperfect) attempt to have _some_ kind of test coverage for this, feel free to take a look if you agree with the approach: https://github.com/modelcontextprotocol/python-sdk/pull/1140

src/mcp/server/streamable_http_manager.py
```diff
@@ -190,6 +190,9 @@ async def run_stateless_server(*, task_status: TaskStatus[None] = anyio.TASK_STA
         # Handle the HTTP request and return the response
         await http_transport.handle_request(scope, receive, send)
 
+        # Terminate the transport after the request is handled
+        await http_transport.terminate()
```
Kludex, line 194: Can `handle_request` fail? I didn't check, but I think ideally this transport should run in a context manager, which would terminate on exit.

hopeful0: I agree that `handle_request` might fail, perhaps we could use `try...finally...` to handle this situation. Context managers may not be suitable for stateful requests.

PR #1127, opened 2025-07-11
[fix] add type field to client session config BaseModels

bfineran:
Adds a pre-set `type` field to `SseServerParameters`, `StreamableHttpParameters`, and `StdioServerParameters` so serialized parameters will easily be loaded into the correct class.

## Motivation and Context
All fields of `SseServerParameters` overlap with `StreamableHttpParameters`, so a third party application that would want to connect to multiple session with a config such as:

```python
class ClientConfig(BaseModel):
    mcp_configs: List[Union[StreamableHttpParameters, SseServerParameters]]
```
would not have a way to be instantiated from a serialized config (like a yaml file) since the following is ambiguous.
```yaml
mcp_configs:
  - url: ""
```

Additionally, if someone were to serialize either the sse or http config classes they would not be able to reload the second class listed in the union.

Adding the preset `type` field should fix this in a backwards compatible way.

## How Has This Been Tested?
Existing tests cover backwards compatibility, confirmed this PR addresses the described use case

## Breaking Changes
No, added fields include the only valid Literal default.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/session_group.py
+++ b/src/mcp/client/session_group.py
@@ -13,7 +13,7 @@ import logging
 from collections.abc import Callable
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, TypeAlias
+from typing import Any, Literal, TypeAlias
 
 import anyio
 from pydantic import BaseModel
@@ -30,6 +30,9 @@ from mcp.shared.exceptions import McpError
 class SseServerParameters(BaseModel):
     """Parameters for intializing a sse_client."""
 
+    # The type of server connection.
+    type: Literal["sse"] = "sse"
+
     # The endpoint URL.
     url: str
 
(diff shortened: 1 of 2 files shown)


felixweinberger:
Hi @bfineran thank you for your contribution! And apologies for the time it took to get back tot his.

> I don't think this is necessary. You can handle the serialization logic in your application.

Agree with this, this seems like a concern that should be handled at the application layer and not in the SDK - within the SDK we can leverage `isinstance` to compare the Pydantic Types of already instantiated objects.

If you're storing these configurations somehow and need to discriminate between different types, you could have a wrapper class that does this for you without having to modify the SDK types for it.

Something like this:

```python
  class ServerConfig(BaseModel):
      type: Literal["stdio", "sse", "streamable_http"]
      config: dict[str, Any]

      def to_server_params(self) -> Union[StdioServerParameters, ...]:
          if self.type == "stdio":
              # do stuff
          elif self.type == "sse":
              # do stuff
          elif self.type == "streamable_http":
              # do stuff
```

Issue #952, opened 2025-06-13
Add `Audio` to `fastmcp/types.py`


Kludex:
### Description

There's a `Image` in FastMCP that can be used to send image content back, but we don't have `Audio` there.

dragonier23:
Hi I'll like to work on this. I've just taken a look at the code, and this issue seems to have been resolved. Could I verify if this is indeed the case?

Kludex:
No, I meant in FastMCP: https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/utilities/types.py

dragonier23:
thanks for the clarification! ill work on it this week

PR #1130, opened 2025-07-12
Added Audio to FastMCP

dragonier23:
## Motivation and Context

Adding Audio in FastMCP that can be used to send audio content back. Related to Issue Closes: #952 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/utilities/types.py
+++ b/src/mcp/server/fastmcp/utilities/types.py
@@ -3,7 +3,7 @@
 import base64
 from pathlib import Path
 
-from mcp.types import ImageContent
+from mcp.types import AudioContent, ImageContent
 
 
 class Image:
@@ -52,3 +52,52 @@ class Image:
             raise ValueError("No image data available")
 
         return ImageContent(type="image", data=data, mimeType=self._mime_type)
+
+
+class Audio:
+    """Helper class for returning audio from tools."""
+
+    def __init__(
+        self,
+        path: str | Path | None = None,
+        data: bytes | None = None,
+        format: str | None = None,
+    ):
+        if path is None and data is None:
+            raise ValueError("Either path or data must be provided")
+        if path is not None and data is not None:
+            raise ValueError("Only one of path or data can be provided")
+
+        self.path = Path(path) if path else None
+        self.data = data
+        self._format = format
+        self._mime_type = self._get_mime_type()
+
+    def _get_mime_type(self) -> str:
+        """Get MIME type from format or guess from file extension."""
+        if self._format:
+            return f"audio/{self._format.lower()}"
+
+        if self.path:
+            suffix = self.path.suffix.lower()
+            return {
+                ".wav": "audio/wav",
+                ".mp3": "audio/mpeg",
+                ".ogg": "audio/ogg",
+                ".flac": "audio/flac",
+                ".aac": "audio/aac",
+                ".m4a": "audio/mp4",
+            }.get(suffix, "application/octet-stream")
+        return "audio/wav"  # default for raw binary data
+
+    def to_audio_content(self) -> AudioContent:
+        """Convert to MCP AudioContent."""
+        if self.path:
+            with open(self.path, "rb") as f:
+                data = base64.b64encode(f.read()).decode()
+        elif self.data is not None:
+            data = base64.b64encode(self.data).decode()
+        else:
+            raise ValueError("No audio data available")
+
+        return AudioContent(type="audio", data=data, mimeType=self._mime_type)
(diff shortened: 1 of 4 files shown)


src/mcp/server/fastmcp/utilities/types.py
```diff
@@ -52,3 +52,52 @@ def to_image_content(self) -> ImageContent:
...
+        """Convert to MCP AudioContent."""
+        if self.path:
+            with open(self.path, "rb") as f:
+                data = base64.b64encode(f.read()).decode()
+        elif self.data is not None:
+            data = base64.b64encode(self.data).decode()
+        else:
+            raise ValueError("No audio data available")
+
+        return AudioContent(type="audio", data=data, mimeType=self._mime_type)
```
RavindraGDP, line 103: I think there are a lot of similarity between Audio class and Image class, maybe we can extract the similarity and make it into a base class that both Audio and Image will inherit?, this is so that the code is not duplicated

tests/server/fastmcp/test_server.py
```diff
@@ -299,6 +303,27 @@ async def test_tool_image_helper(self, tmp_path: Path):
...
+
+        mcp = FastMCP()
+        mcp.add_tool(audio_tool_fn)
+        async with client_session(mcp._mcp_server) as client:
+            result = await client.call_tool("audio_tool_fn", {"path": str(audio_path)})
+            assert len(result.content) == 1
+            content = result.content[0]
+            assert isinstance(content, AudioContent)
+            assert content.type == "audio"
+            assert content.mimeType == "audio/wav"
```
ochafik, line 321: Could you also add something to exercice the suffix-based mime type detection?

dragonier23: I just added a test to test the mimetypes, do let me know if this is what you are looking for!

tests/server/fastmcp/test_server.py
```diff
@@ -371,6 +429,47 @@ def mixed_list_fn() -> list:
...
+            assert isinstance(content1, TextContent)
+            assert content1.text == "text message"
+            # Check audio conversion
+            content2 = result.content[1]
+            assert isinstance(content2, AudioContent)
+            assert content2.mimeType == "audio/wav"
+            assert base64.b64decode(content2.data) == b"test audio data"
+            # Check dict conversion
+            content3 = result.content[2]
+            assert isinstance(content3, TextContent)
```
Kludex: Remove this test and add Audio on the above please.

tests/server/fastmcp/test_server.py
```diff
@@ -299,6 +303,60 @@ async def test_tool_image_helper(self, tmp_path: Path):
...
+        # Test different audio file extensions
+        test_cases = [
+            ("test.wav", "audio/wav"),
+            ("test.mp3", "audio/mpeg"),
+            ("test.ogg", "audio/ogg"),
+            ("test.flac", "audio/flac"),
+            ("test.aac", "audio/aac"),
+            ("test.m4a", "audio/mp4"),
+            ("test.unknown", "application/octet-stream"),  # Unknown extension fallback
+        ]
```
Kludex: Use `pytest.mark.parametrize`, please.

src/mcp/server/fastmcp/utilities/types.py
```diff
@@ -52,3 +52,52 @@ def to_image_content(self) -> ImageContent:
...
+
+class Audio:
+    """Helper class for returning audio from tools."""
+
+    def __init__(
+        self,
+        path: str | Path | None = None,
+        data: bytes | None = None,
+        format: str | None = None,
+    ):
```
Kludex, line 65: Let's use `@typing.override`.

src/mcp/server/fastmcp/utilities/types.py
```diff
@@ -52,3 +52,52 @@ def to_image_content(self) -> ImageContent:
...
+        if path is None and data is None:
+            raise ValueError("Either path or data must be provided")
+        if path is not None and data is not None:
+            raise ValueError("Only one of path or data can be provided")
```
Kludex: I think you can do `bool(path) ^ bool(data)`, and have a meaningful message: "Either path or data can be provided".

src/mcp/server/fastmcp/utilities/types.py
dragonier23: I would like to get some clarity as to what we are tying to override, given the audio class dosent inherit from any class at the moment

src/mcp/server/fastmcp/utilities/types.py
dragonier23: thanks for this suggestion! should i also make these changes to the Image class?

ochafik: If you're up for it, please feel free to send a tiny follow up PR w/ both changes :-)

ochafik (approved):
Thanks again!

PR #1133, opened 2025-07-12
feat: ClientConnectionManager to manage multiple sessions without async with

### LINKED ISSUE #922: Create and Close multiple client session result "RuntimeError: Attempted to exit a cancel scope that isn't the current tasks's current cancel scope"
This is originally found in [issue#788](https://github.com/modelcontextprotocol/python-sdk/issues/788)， but after some debugging, I believe this related to how ClientSession is implemented.

    import contextlib
    async def sessions():

        context1 = contextlib.AsyncExitStack()
        client1 = stdio_client(server_params_list['filesystem'])
        read1, write1 = await context1.enter_async_context(client1)
        session1 = await context1.enter_async_context(ClientSession(read1, write1))
        await session1.initialize()
        tools1 = await session1.list_tools()
        print(f"Session1 has {len(tools1.tools)} tools")
    
        context2 = contextlib.AsyncExitStack()
        client2 = sse_client(url = server_params_list['excel'].url)
        read2, write2 = await context2.enter_async_context(client2)
        session2 = await context2.enter_async_context(ClientSession(read2, write2))
        await session2.initialize()
        tools2 = await session2.list_tools()
        print(f"Session2 has {len(tools2.tools)} tools")
    
        await context1.aclose()
        await context2.aclose()
  
This is actually the simplified way how ClientSessionGroup() did it. This code will except in "await context1.aclose()". Please note if change the order:

        await context2.aclose()
        await context1.aclose()
    
This will go through without exception.
 
After some further looking, the ClientSession is use BaseSession, iwhere n its __aenter__(), BaseSession created a Task Group (to run receive_loop), and try to __aexit__() the task group during BaseSession's __aexit__()

The Task Group, among creation, will bind a "cancel scope" with "current_task". So when second ClientSession created, a new "cancel scope 2" is binding the "current_task", replace the first one. When you try to tear down the session1 now, evantually it try to call task_group.__aexit__() which lead to CancelScope().__exit__(), there it find the current_task() is binding to a different cancel_scope, thus the runtime.

The code below, demostrate the task group issue (without any mcp releated code):

    async def taskgroup():
        import anyio
        from contextlib import AsyncExitStack
    
        async def task1(id):
            print(f"Task {id} started")
            await anyio.sleep(1)
            print(f"Task {id} running...")
    
        tg1 = anyio.create_task_group()
        await tg1.__aenter__()
        tg1.start_soon(task1, 1)
    
        tg2 = anyio.create_task_group()
        await tg2.__aenter__()
        tg2.start_soon(task1, 2)
    
        await anyio.sleep(3) 
        await tg1.__aexit__(None, None, None)
        await tg2.__aexit__(None, None, None)
    
Maybe BaseSession could consider other way instead of use anyio.create_task_group()

hanhan1123:
This PR introduces a `ClientConnectionManager` that allows users to manage multiple sessions without needing to explicitly use `async with`. This provides more flexibility in long-running or dynamic session use cases.
This PR also resolves #922 by allowing users to manage multiple sessions without async with

## Motivation and Context
In developing AI agents, the MCP client is often just one part of a larger system. However, the requirement to use `async with` when initiating sessions makes it difficult to integrate MCP into more complex workflows, especially when other asynchronous components are involved.

To address this, I implemented a `ClientConnectionManager` that completely removes the need for `async with` during usage. This provides much greater flexibility in designing and structuring other components, particularly in long-running or modular systems.

## How Has This Been Tested?
Code branches are covered via pytest unit tests.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist


ihrpr:
Thank you for your contribution. The issue mentioned is related to ClientSessionGroup. If there is a specific concern around it, we need to address it in ClientSessionGroup. Adding an additional, very similar abstraction creates API fragmentation and maintenance burden.

hanhan1123:
@ihrpr 
Thank you for the feedback.

Regarding ClientSessionGroup, I’ve evaluated it before proposing this PR. However, it doesn’t fundamentally resolve the session management problem.

ClientSessionGroup still relies on the async with lifecycle, which makes it tricky when you need sessions to persist across a wider application scope. More importantly, it can still hit this error:

And with the current architecture, there’s really no clean way to solve that, because of how async context managers and cancellation scopes work across tasks. That’s why I explored a different approach in this PR.

I’m happy to further discuss how this could evolve to better align with the project’s direction while effectively addressing these session lifecycle constraints.

qrzeng:
nice, this pr is what i need.

PR #1136, opened 2025-07-14
README - replace code snippets with examples 

ihrpr:
Part 2, based on https://github.com/modelcontextprotocol/python-sdk/pull/1055

- Replaced hardcoded Quickstart example with `fastmcp_quickstart.py`
- Replaced lifespan example with `lifespan_example.py` (fixed type safety)
- Updated "Writing MCP Clients" to use `stdio_client.py`
- Replaced completion client fragment with full `completion_client.py`
- Removed redundant "## Examples" section

Follow up:

- [ ] replace all of the examples
- [ ] re-write README (only after we re-do the docs section in modelcontextprotocol repo)


examples/snippets/clients/completion_client.py
```diff
@@ -0,0 +1,80 @@
...
+
+from mcp import ClientSession, StdioServerParameters
+from mcp.client.stdio import stdio_client
+from mcp.types import PromptReference, ResourceTemplateReference
+
+# Create server parameters for stdio connection
+server_params = StdioServerParameters(
+    command="uv",  # Using uv to run the server
+    args=["run", "server", "completion", "stdio"],  # Server with completion support
+    env=os.environ.copy(),  # Pass environment variables
```
Kludex: This is dangerous, I think we should be more conscious about it.

Can you pass a single environment variable instead to show the point?

examples/snippets/pyproject.toml
```diff
@@ -12,7 +12,9 @@ requires = ["setuptools", "wheel"]
...
 [tool.setuptools]
-packages = ["servers"]
+packages = ["servers", "clients"]
 
 [project.scripts]
-server = "servers:run_server"
\ No newline at end of file
+server = "servers:run_server"
+client = "clients.stdio_client:main"
+completion-client = "clients.completion_client:main"
```
Kludex:
```suggestion
completion-client = "clients.completion_client:main"

```

Kludex (approved):

PR #1139, opened 2025-07-14
Set LifespanResultT default=Any to prevent reportMissingTypeArgument

strawgate:
Set default type for LifespanResultT to Any to avoid the default behavior of, when not providing an explicit type argument, instantiating generic class without a type leading to `unknown` type with type checkers.

## Motivation and Context

Improve type-checking behavior when Lifespan is not used.

## Types of changes

- [ ] Documentation update

## Checklist


Kludex:
I've applied this on https://github.com/modelcontextprotocol/python-sdk/pull/1198/.

PR #1140, opened 2025-07-14
Add regression test for stateless request memory cleanup

felixweinberger:
## Summary

This PR adds a regression test for the memory leak fix from PR #1116. The test ensures that transport resources are properly cleaned up after handling stateless requests in the StreamableHTTPSessionManager.

## Problem

PR #1116 identified a memory leak where transport instances were not being terminated after handling stateless requests. This could lead to resource accumulation over time as streams and other resources were not properly closed.

## Solution

The test verifies that:
- The transport's `terminate()` method is called after each stateless request
- The transport's internal state (`_terminated` flag and `_request_streams`) is properly cleaned up
- The test fails if the cleanup is not performed, providing regression protection

## Test Implementation

The test creates real StreamableHTTPServerTransport instances and verifies the cleanup behavior rather than relying on mocks. This provides a more robust test that validates the actual behavior of the system.

Related to #1116

### DIFF
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -1,10 +1,11 @@
 """Tests for StreamableHTTPSessionManager."""
 
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, patch
 
 import anyio
 import pytest
 
+from mcp.server import streamable_http_manager
 from mcp.server.lowlevel import Server
 from mcp.server.streamable_http import MCP_SESSION_ID_HEADER
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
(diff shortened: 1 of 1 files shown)


felixweinberger:
@hopeful0 FYI

hopeful0:
Thanks, I learned that the patch can helped us track down the transport, forgive my unfamiliarity with unittest.

tests/server/test_streamable_http_manager.py
```diff
@@ -197,3 +198,68 @@ async def mock_receive():
...
+                    "type": "http.request",
+                    "body": b"",
+                    "more_body": False,
+                }
+
+            # Send a request
+            await manager.handle_request(scope, mock_receive, mock_send)
+
+            # Give async tasks time to complete
+            await anyio.sleep(0.1)
```
ihrpr: why do we need sleep here?

felixweinberger: good catch, not necessary - removed.

PR #1142, opened 2025-07-15
fix: attempt to query resource-specific protected resource metadata before root PRM

LucaButBoring:
Currently, the resource-specific protected resource metadata endpoint is not queried to discover RFC 9728 metadata, which is incorrect per the [specification](https://datatracker.ietf.org/doc/html/rfc9728#name-protected-resource-metadata-).

Given an MCP server at `https://example.com/mcp`, the PRM must be located at `https://example.com/.well-known/oauth-protected-resource/mcp`. However, no matter what path the MCP server is hosted on, the SDK currently attempts to discover PRM at the root, e.g. `https://example.com/.well-known/oauth-protected-resource`. This breaks compatibility with any MCP server with authorization hosted outside of the root, unless the MCP server is double-hosting metadata to account for this.

This PR attempts to handle both cases (giving consideration to the existing behavior potentially being expected now) by querying the most specific resource path first, and then falling back to the existing behavior in the event that PRM is hosted at the root path instead. This is the same behavior used in the [TS SDK](https://github.com/modelcontextprotocol/typescript-sdk/blob/f3584f2a4162febadf39300d7f37be63a91ce05d/src/client/auth.ts#L549-L579).

Note that this fallback behavior is not actually spec-compliant, I've only implemented it this way to account for the possibility that some MCP servers are non-compliant to match the client SDK.

## Motivation and Context
Fixes compatibility with spec-compliant MCP servers.

## How Has This Been Tested?
Unit tests. **TODO: spot check against example server (putting implementation out now for early feedback)**

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
#1052

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -203,13 +203,24 @@ class OAuthClientProvider(httpx.Auth):
         )
         self._initialized = False
 
-    async def _discover_protected_resource(self) -> httpx.Request:
+    def _build_well_known_path_protected_resource(self, pathname: str) -> str:
+        """Construct well-known path for OAuth protected resource metadata discovery."""
+        well_known_path = f"/.well-known/oauth-protected-resource{pathname}"
+        if pathname.endswith("/"):
+            # Strip trailing slash from pathname to avoid double slashes
+            well_known_path = well_known_path[:-1]
+        return well_known_path
+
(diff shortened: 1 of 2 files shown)


LucaButBoring:
Closing in favor of #1071 which was just merged.

Issue #1141, opened 2025-07-14
Progress notifications cause server to hang on stdio transport


abhishekgahlot2:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

When using the low-level Server API (`mcp.server.lowlevel.Server`) with stdio transport, sending progress notifications during request handling causes the server to hang indefinitely and never send the final response.

**What I am seeing:**
- Server successfully sends progress notification via `write_stream.send()`
- Server handler returns `CallToolResult` 
- The final response is never sent to the client
- Client times out waiting for response
- Server remains blocked and time out eventually

### Example Code

```Python
import asyncio
from mcp.server.lowlevel import Server
from mcp.server.stdio import stdio_server
from mcp.types import *

server = Server("progress-bug-demo")

@server.list_tools()
async def handle_list_tools() -> list[Tool]:
    return [Tool(
        name="test_tool",
        description="Demonstrates the bug",
        inputSchema={"type": "object", "properties": {}}
    )]

@server.call_tool()
async def handle_call_tool(name: str, arguments: dict) -> list[TextContent]:
    ctx = server.request_context
    progress_token = ctx.meta.progressToken if ctx.meta else None
    
    if progress_token:
        # This causes the server to hang
        await ctx.session.send_progress_notification(
            progress_token=progress_token,
            progress=0.5,
            total=1.0,
            message="Working..."
        )
    
    # This response is never sent due to the hang
    return [TextContent(type="text", text="Task completed!")]

async def main():
    async with stdio_server() as (read_stream, write_stream):
        await server.run(
            read_stream,
            write_stream,
            server.create_initialization_options()
        )

if __name__ == "__main__":
    asyncio.run(main())
```

### Python & MCP Python SDK

```Text
MCP: 1.10.1
Python: 3.12.6
```

From deeper debugging looks like buffer size is the problem : https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/stdio.py#L57

Have to test locally if increasing it will most likely fix the issue and if its causing more problems than just fixing this.

abhishekgahlot2:
Actually using buffered streams seems to fix this problem.

ihrpr:
@abhishekgahlot2 which client are you using for testing?

abhishekgahlot2:
[...]

PR #1143, opened 2025-07-15
Increasing Buffer Size in MCP stdio Transport

abhishekgahlot2:
Fixes: https://github.com/modelcontextprotocol/python-sdk/issues/1141

When using the low-level Server API (mcp.server.lowlevel.Server) with stdio transport, sending progress notifications during request handling causes the server to hang indefinitely and never send the final response.

This issue was discovered while implementing an MCP server that returns large SQL query results and detailed analysis. The server would hang indefinitely when the output exceeded the OS pipe buffer limit.

## Motivation and Context
Provides a fix for https://github.com/modelcontextprotocol/python-sdk/issues/1141

The current stdio transport implementation uses unbuffered memory object streams (buffer size 0), which can cause deadlocks when:
1. Multiple messages are sent in quick succession
2. Progress notifications are combined with tool responses

## How Has This Been Tested?
Tested with code shared in original #1141 and fix works in it.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -54,8 +54,9 @@ async def stdio_server(
     write_stream: MemoryObjectSendStream[SessionMessage]
     write_stream_reader: MemoryObjectReceiveStream[SessionMessage]
 
-    read_stream_writer, read_stream = anyio.create_memory_object_stream(0)
-    write_stream, write_stream_reader = anyio.create_memory_object_stream(0)
+    # FIXED: Changed from 0 to 10 to create buffered streams
+    read_stream_writer, read_stream = anyio.create_memory_object_stream(10)
+    write_stream, write_stream_reader = anyio.create_memory_object_stream(10)
 
     async def stdin_reader():
         try:


src/mcp/server/stdio.py
```diff
@@ -54,8 +54,9 @@ async def stdio_server(
     write_stream: MemoryObjectSendStream[SessionMessage]
     write_stream_reader: MemoryObjectReceiveStream[SessionMessage]
 
-    read_stream_writer, read_stream = anyio.create_memory_object_stream(0)
-    write_stream, write_stream_reader = anyio.create_memory_object_stream(0)
+    # FIXED: Changed from 0 to 10 to create buffered streams
+    read_stream_writer, read_stream = anyio.create_memory_object_stream(10)
+    write_stream, write_stream_reader = anyio.create_memory_object_stream(10)
```
Kludex, line 59: I think the idea is exactly to not have this buffer.

ihrpr:
Thank you for debugging this. Lets continue the discussion in the https://github.com/modelcontextprotocol/python-sdk/issues/1141. Not having a buffer is on purpose.

PR #1147, opened 2025-07-15
fix markdown error in README in main

ihrpr:
This check was added recently, when merging old PRs we might be introducing the failure in the main. Worth merging main before merge for old PRs touching README

### DIFF
--- a/README.md
+++ b/README.md
@@ -951,12 +951,11 @@ The following example shows how to use `streamable_http_app()`, a method that re
 You can then append additional routes to that application as needed.
 
 ```python
-from starlette.routing import Route
-
 mcp = FastMCP("My App")
 
 app = mcp.streamable_http_app()
 # Additional non-MCP routes can be added like so:
+# from starlette.routing import Route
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -950,13 +950,12 @@ By default, SSE servers are mounted at `/sse` and Streamable HTTP servers are mo
...
-```
-from starlette.routing import Route
-
+```python
 mcp = FastMCP("My App")
 
 app = mcp.streamable_http_app()
 # Additional non-MCP routes can be added like so:
+# from starlette.routing import Route
 # app.router.routes.append(Route("/", endpoint=other_route_function))
```
Kludex, line 959: This is actually bad advice. No one should use `app.router.routes`...

I think the idea is to mount the ASGI app from FastMCP into an existent Starlette application.

Kludex (approved):
I understand my comment below is not related to this PR, so approving.

README.md
ihrpr: Noted, I'll be re-writing more README today/tomorrow

PR #1153, opened 2025-07-16
server: skip duplicate response on CancelledError

### LINKED ISSUE #1152: CancelledError causes duplicate response and server crash
### Initial Checks

### Description

When a client cancels a long-running tool call (e.g., user hits Escape in Claude Code), the MCP server crashes with an assertion error and becomes unresponsive.

**What I'm seeing:**
- Server crashes with "AssertionError: Request already responded to"
- All subsequent requests fail
- Server must be restarted

**What I expect:**
- Server should handle cancellation gracefully
- Server should remain responsive for subsequent requests

**Root cause:**
In `Server._handle_request`, when a CancelledError occurs:
1. RequestResponder.cancel() has already sent an error response
2. The CancelledError propagates to _handle_request
3. It tries to send another error response
4. This triggers "already responded to" assertion

### Example Code

```python
from mcp.server import Server
import asyncio

server = Server("test-server")

@server.call_tool()
async def slow_tool(name: str, arguments: dict):
    # Any long-running operation
    await asyncio.sleep(30)
    return [{"type": "text", "text": "Done"}]

# When client calls slow_tool and cancels (e.g., hits Escape in Claude Code)
# Server crashes with: AssertionError: Request already responded to
```

### Python & MCP Python SDK

Python 3.8+ (tested on 3.11)
MCP Python SDK: Latest version from main branch

Related issues: jlowin/fastmcp#508, jlowin/fastmcp#823

lukacf:
When clients cancel long-running tool calls, the server crashes with an assertion error and becomes unresponsive. This PR fixes the issue by preventing duplicate responses.

## Problem
- RequestResponder.cancel() sends an error response when cancelled
- _handle_request tries to send another error response 
- This causes "Request already responded to" assertion
- Server becomes completely unresponsive

## Solution
Catch CancelledError in _handle_request and return early without sending a second response.

## Testing
Added test that cancels a tool call and verifies the server remains responsive.

Fixes #1152
Related: jlowin/fastmcp#508, jlowin/fastmcp#823


src/mcp/server/lowlevel/server.py
```diff
@@ -647,6 +648,12 @@ async def _handle_request(
                 response = await handler(req)
             except McpError as err:
                 response = err.error
+            except asyncio.CancelledError:
```
ihrpr: we use anyio in this codebase

tests/server/test_cancel_handling.py
```diff
@@ -0,0 +1,138 @@
+"""Test that cancelled requests don't cause double responses."""
+
```
ihrpr, line 2: There is a lot of mocking happening. I'd suggest to use `test_integration` or leverage `create_connected_server_and_client_session` and test with `CancelledNotification`

src/mcp/server/lowlevel/server.py
lukacf: Fixed!

tests/server/test_cancel_handling.py
lukacf: Fixed, thanks for the directions!

tests/server/test_cancel_handling.py
```diff
@@ -0,0 +1,186 @@
...
+    CancelledNotification,
+    CancelledNotificationParams,
+    ClientNotification,
+    ClientRequest,
+    Tool,
+)
+
+
+@pytest.mark.anyio
+async def test_cancelled_request_no_double_response():
```
ihrpr: why do we need this test? it doesn't look like it's testing much and `test_server_remains_functional_after_cancel` looks complete to cover the fix?

ihrpr (changes requested):
Thanks for the quick turnaround, I think we can remove `test_cancelled_request_no_double_response` test. wdyt?

PR #1165, opened 2025-07-17
feat: add claims field to AccessToken for decoded token payload

### LINKED ISSUE #1038: MCP server: AccessToken class should have field for subject claim ("sub")
### Description

```
class AccessToken(BaseModel):
    token: str
    client_id: str
    scopes: list[str]
    expires_at: int | None = None

    # Proposed:
```

```
class MyTokenVerifier:
        try:
        except ...:
            return None

        return AccessToken(
            token=token,
            ...
            subject=token_claims["sub"],
        )
```

```

```

### References

_No response_

robertofalk:
Add optional claims dict to AccessToken class to provide access to all decoded JWT claims (both standard and custom) instead of restricting to only specific fields like scopes and expires_at.

## Motivation and Context
FastMCP provides a BearerAuthProvider class to validate bearer tokens and returns it using the AccessToken class (https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/server/auth/providers/bearer.py#L387). With the current implementation the claims from the token are limited to the ones defined so far, so any additional claim is not available and to get it currently I had to define an additional middleware that decodes the token again, which is a waste since the token is was already decoded by BearerAuthProvider.

## How Has This Been Tested?
I was not able to test it since it's just an attribute that will be used by applications relying on the python-sdk AccessToken class.

## Breaking Changes
No, it's an optional attribute.

## Types of changes
- [ ] Documentation update

## Checklist

Resolves #1038

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -517,16 +517,16 @@ class OAuthClientProvider(httpx.Auth):
                     # Step 2: Discover OAuth metadata (with fallback for legacy servers)
                     discovery_urls = self._get_discovery_urls()
                     for url in discovery_urls:
-                        request = self._create_oauth_metadata_request(url)
-                        response = yield request
+                        oauth_metadata_request = self._create_oauth_metadata_request(url)
+                        oauth_metadata_response = yield oauth_metadata_request
 
-                        if response.status_code == 200:
+                        if oauth_metadata_response.status_code == 200:
                             try:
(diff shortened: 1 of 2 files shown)


robertofalk:
@ihrpr, @ochafik Any comments? since you have been driving the `auth` milestone.

felixweinberger:
Closing this as per @ochafik's comment this is incomplete and it's been a while since the last update here.

If others find this PR and need this, feel free to submit a fresh PR.

PR #1166, opened 2025-07-17
fix flaky fix-test_streamablehttp_client_resumption test

ihrpr:
While fixing the test found an bug, we should not check for **`fix-test_streamablehttp_client_resumption`** when setting as it's used for reusability and the check is done afterwards in the code anyways.

### DIFF
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -124,10 +124,7 @@ class SimpleEventStore(EventStore):
 class ServerTest(Server):
     def __init__(self):
         super().__init__(SERVER_NAME)
-        self._lock = anyio.Event()
-        # Reset the lock for each new server instance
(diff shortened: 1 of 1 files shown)


tests/shared/test_streamable_http.py
```diff
@@ -214,6 +225,39 @@ async def handle_call_tool(name: str, args: dict) -> list[TextContent]:
...
+
+                return [TextContent(type="text", text="Completed")]
+
+            elif name == "release_lock":
+                assert self._lock is not None, "Lock must be initialized before releasing"
+
+                # Release the lock
+                self._lock.set()
+                return [TextContent(type="text", text="Lock released")]
+
```
felixweinberger, line 260: really cool idea to have the tool calls create internal locks for this!

felixweinberger (approved):

PR #1171, opened 2025-07-18
fix flaky test test_88_random_error

ihrpr:
The test is [flaky](https://github.com/modelcontextprotocol/python-sdk/actions/runs/16372438953/job/46263912823?pr=1169). 

  - Replaced sleep-based synchronization with event-based coordination: Instead of using anyio.sleep() to simulate a slow
  operation, the test now uses anyio.Event() to have deterministic control over when the slow request completes
  - Removed unnecessary timing dependencies: The "fast" operation now completes instantly without any sleep, eliminating
  timing-related flakiness
  - Added  clean up
  - Simplified the synchronization logic: Removed multiple event objects in favor of a single slow_request event that precisely
   controls the test flow

### DIFF
--- a/tests/issues/test_88_random_error.py
+++ b/tests/issues/test_88_random_error.py
@@ -28,8 +28,7 @@ async def test_notification_validation_error(tmp_path: Path):
 
     server = Server(name="test")
     request_count = 0
-    slow_request_started = anyio.Event()
-    slow_request_complete = anyio.Event()
+    slow_request = anyio.Event()
 
     @server.list_tools()
     async def list_tools() -> list[types.Tool]:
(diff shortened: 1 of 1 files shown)


tests/issues/test_88_random_error.py
```diff
@@ -28,8 +28,7 @@ async def test_notification_validation_error(tmp_path: Path):
 
     server = Server(name="test")
     request_count = 0
-    slow_request_started = anyio.Event()
-    slow_request_complete = anyio.Event()
+    slow_request = anyio.Event()
```
felixweinberger: nit: `slow_request_lock` to make it obvious what this is when reading the test further down

tests/issues/test_88_random_error.py
```diff
@@ -90,16 +82,15 @@ async def client(read_stream, write_stream, scope):
...
 
             # Second call should timeout (slow operation)
             with pytest.raises(McpError) as exc_info:
                 await session.call_tool("slow")
             assert "Timed out while waiting" in str(exc_info.value)
 
-            # Wait for slow request to complete in the background
-            with anyio.fail_after(1):  # Timeout after 1 second
-                await slow_request_complete.wait()
+            # release the slow request not to have hagning process
```
felixweinberger: nit: `hanging` typo

felixweinberger (approved):
Nice!

PR #1177, opened 2025-07-21
Add `streamable_http_client` which accepts `httpx.AsyncClient` instead of `httpx_client_factory`

### LINKED ISSUE #1370: httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1002)
### Description

I am currently using the mcp-python-sdk to set up an MCP client with Streamable HTTP as the transport layer protocol. However, since I'm working in an internal network environment, I encounter the following HTTPS error:

httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1002)
the code of me is:
```
import asyncio
from mcp import ClientSession

async def main():
    # Connect to a streamable HTTP server
#the url of my mcp server can't avail
    async with streamablehttp_client("https://mcp.higress.ai/mcp-ip-query/cmfdmaly") as (
        read_stream,
        write_stream,
        _,
    ):
        # Create a session using the client streams
            # Initialize the connection
            await session.initialize()
            # List available tools
            print(f"Available tools: {[tool.name for tool in tools.tools]}")


if __name__ == "__main__":
    asyncio.run(main())
```

To temporarily resolve this issue, I modified /.venv/Lib/site-packages/mcp/shared/_httpx_utils.py by adding "verify": False after line 66.

My request is to support passing a verify parameter when constructing the streamablehttp_client, allowing users to configure certificate verification flexibly without modifying the source code.

### References

_No response_

### LINKED ISSUE #1470: SSLV3_ALERT_HANDSHAKE_FAILURE when using private gateway
### Initial Checks

### Description

I'm calling an MCP "https://mcp.deepwiki.com/mcp" through a private gateway. Because of that, the URL I'm calling is "https://secure-gateway:8081", so far so good, but unfortunately, I get _"ConnectError('[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1000)')_". 

Unfortunately, in [_handle_post_request](https://github.com/modelcontextprotocol/python-sdk/blob/61399b386c19af9b9c9277727990a6b3c6a95d83/src/mcp/client/streamable_http.py#L254), even if I send the headers "X-Forwarded-Host", they are ignored. It seems that "sni_hostname" is only retrieved by using extensions which are not present here. 

Is this expected, or is this an issue?

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Python: 3.12
MCP Python: 1.16.0
```

Kludex:
The current name is not very intuitive considering the class name and the module.

- Closes #1370
- Closes #1470

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -23,8 +23,6 @@ from httpx_sse import EventSource, ServerSentEvent, aconnect_sse
 from typing_extensions import deprecated
 
 from mcp.shared._httpx_utils import (
-    MCP_DEFAULT_SSE_READ_TIMEOUT,
-    MCP_DEFAULT_TIMEOUT,
     McpHttpClientFactory,
     create_mcp_http_client,
 )
@@ -79,7 +77,6 @@ class RequestContext:
     """Context for a request operation."""
 
     client: httpx.AsyncClient
-    headers: dict[str, str]
     session_id: str | None
     session_message: SessionMessage
     metadata: ClientMessageMetadata | None
@@ -645,16 +642,7 @@ async def streamable_http_client(
         # Create default client with recommended MCP timeouts
         client = create_mcp_http_client()
 
-    # Extract configuration from the client to pass to transport
-    headers_dict = dict(client.headers) if client.headers else None
-    timeout = client.timeout.connect if (client.timeout and client.timeout.connect is not None) else MCP_DEFAULT_TIMEOUT
-    sse_read_timeout = (
-        client.timeout.read if (client.timeout and client.timeout.read is not None) else MCP_DEFAULT_SSE_READ_TIMEOUT
-    )
-    auth = client.auth
-
-    # Create transport with extracted configuration
-    transport = StreamableHTTPTransport(url, headers_dict, timeout, sse_read_timeout, auth)
+    transport = StreamableHTTPTransport(url)
 
     async with anyio.create_task_group() as tg:
         try:
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -1172,7 +1172,7 @@ async def test_streamable_http_client_session_termination(basic_server: None, ba
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 # Attempt to make a request after termination
                 with pytest.raises(  # pragma: no branch
                     McpError,
@@ -1239,7 +1239,7 @@ async def test_streamable_http_client_session_termination_204(
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 # Attempt to make a request after termination
                 with pytest.raises(  # pragma: no branch
                     McpError,
@@ -1368,8 +1368,8 @@ async def test_streamable_http_client_resumption(event_server: tuple[SimpleEvent
                 # We should have received the remaining notifications
                 assert len(captured_notifications) == 1
 
-            assert isinstance(captured_notifications[0].root, types.LoggingMessageNotification)
-            assert captured_notifications[0].root.params.data == "Second notification after lock"
+            assert isinstance(captured_notifications[0].root, types.LoggingMessageNotification)  # pragma: no cover
+            assert captured_notifications[0].root.params.data == "Second notification after lock"  # pragma: no cover
 
 
 @pytest.mark.anyio
@@ -1552,7 +1552,7 @@ async def test_streamablehttp_request_context_propagation(context_aware_server: 
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 result = await session.initialize()
                 assert isinstance(result, InitializeResult)
                 assert result.serverInfo.name == "ContextAwareServer"
@@ -1590,7 +1590,7 @@ async def test_streamablehttp_request_context_isolation(context_aware_server: No
                 write_stream,
                 _,
             ):
-                async with ClientSession(read_stream, write_stream) as session:
+                async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                     await session.initialize()
 
                     # Call the tool that echoes context
@@ -2305,7 +2305,7 @@ async def test_streamable_http_client_mcp_headers_override_defaults(
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 await session.initialize()
 
                 # Use echo_headers tool to see what headers the server actually received
@@ -2340,7 +2340,7 @@ async def test_streamable_http_client_preserves_custom_with_mcp_headers(
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 await session.initialize()
 
                 # Use echo_headers tool to verify both custom and MCP headers are present
@@ -2390,7 +2390,7 @@ async def test_streamablehttp_client_deprecation_warning(basic_server: None, bas
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream) as session:
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
                 await session.initialize()
                 tools = await session.list_tools()
                 assert len(tools.tools) > 0


ihrpr (changes requested):
The deprecated wrapper is missing the `httpx_client_factory` and `auth` parameters. This could break existing code using those parameters.

(it's probably a good time to replace httpx_client_factory with an object)

Kludex:
I blame copilot. 😄

Kludex:
> (it's probably a good time to replace httpx_client_factory with an object)

I think the timeouts, the auth and the httpx_client_factory CAN be replaced. The timeouts I think can be subjective, but the auth and the httpx_client_factory I think need to be replaced.

ihrpr:
> > (it's probably a good time to replace httpx_client_factory with an object)
> 
> I think the timeouts, the auth and the httpx_client_factory CAN be replaced. The timeouts I think can be subjective, but the auth and the httpx_client_factory I think need to be replaced.

should we just replace it now, since we are deprecating one method, would be good to have a nice alternative and not to deprecate it twice?

Kludex:
> > > (it's probably a good time to replace httpx_client_factory with an object)
> > 
> > 
> > I think the timeouts, the auth and the httpx_client_factory CAN be replaced. The timeouts I think can be subjective, but the auth and the httpx_client_factory I think need to be replaced.
> 
> should we just replace it now, since we are deprecating one method, would be good to have a nice alternative and not to deprecate it twice?

I didn't get what you said. I agree that it's a good moment to change the parameters of the new `streamable_http_client`.

dsp-ant:
Should we do this and deprecate the old module?

felixweinberger:
FYI for viewers: on my todo list to update this

src/mcp/client/streamable_http.py
```diff
@@ -101,9 +108,9 @@ def __init__(
         self.session_id = None
         self.protocol_version = None
         self.request_headers = {
+            **self.headers,
```
felixweinberger: Previously we would have `streamablehttp_client` create the `httpx.AsyncClient` via a factory _after_ initializing the transport. That means we could use these `transport.request_headers` when creating the `httpx.AsyncClient` to ensure the client and the transport have the same headers.

If we now accept `httpx_client` as an argument to `streamable_http_client`, that client might have custom headers! In fact by default `httpx.AsyncClient` creates headers here that need to be overriden (hence moving the ACCEPT and CONTENT_TYPE _after_), for example `accept: */*`

Therefore the transport needs to _override_ these headers now if they're present to be configured correctly.

src/mcp/client/streamable_http.py
```diff
@@ -461,30 +465,57 @@ async def streamablehttp_client(
...
+    sse_read_timeout = (
+        client.timeout.read if (client.timeout and client.timeout.read is not None) else MCP_DEFAULT_SSE_READ_TIMEOUT
+    )
+    auth = client.auth
+
+    # Create transport with extracted configuration
+    transport = StreamableHTTPTransport(url, headers_dict, timeout, sse_read_timeout, auth)
+
+    # Sync client headers with transport's merged headers (includes MCP protocol requirements)
+    client.headers.update(transport.request_headers)
```
felixweinberger: ⚠️ Important change 2/2 ⚠️

We need to sync the headers on the client back so that they match with the transport after the transport has overriden them. If the user passes in a `httpx_client` they created, that could have arbitrary headers that no longer match the headers on the transport.

Overall I find this quite horrible - but I'm not sure how else to make this work via an object instead of a factory. Open for suggestions...

felixweinberger:
@Kludex @ihrpr @dsp-ant gave this a shot, trying to replace the factory with a client object.

Results in some ugly header mangling that I don't see how to avoid given the existing structure. Keen for thoughts.

src/mcp/shared/_httpx_utils.py
```diff
@@ -68,7 +72,7 @@ def create_mcp_http_client(
 
     # Handle timeout
     if timeout is None:
-        kwargs["timeout"] = httpx.Timeout(30.0)
+        kwargs["timeout"] = httpx.Timeout(MCP_DEFAULT_TIMEOUT, read=MCP_DEFAULT_SSE_READ_TIMEOUT)
```
felixweinberger, line 75: Previously we'd rely on the transport setting this to `60 * 5` when creating our client, but we can't rely on that anymore as the client may now be created before the transport.

examples/snippets/clients/oauth_client.py
```diff
@@ -14,7 +14,8 @@
 
 from mcp import ClientSession
 from mcp.client.auth import OAuthClientProvider, TokenStorage
-from mcp.client.streamable_http import streamablehttp_client
+from mcp.client.streamable_http import streamable_http_client
+from mcp.shared._httpx_utils import create_mcp_http_client
```
Kludex: I don't think we should have private methods being exposed in examples. 🤔

src/mcp/client/streamable_http.py
```diff
@@ -442,14 +449,11 @@ def get_session_id(self) -> str | None:
...
 
 @asynccontextmanager
-async def streamablehttp_client(
+async def streamable_http_client(
     url: str,
-    headers: dict[str, str] | None = None,
-    timeout: float | timedelta = 30,
-    sse_read_timeout: float | timedelta = 60 * 5,
+    *,
+    httpx_client: httpx.AsyncClient | None = None,
```
Kludex:
```suggestion
    http_client: httpx.AsyncClient | None = None,
```

fryguy04:
FWIW I pulled this PR into my venv and was able to connect to my self-signed MCP server - thx, looking forward to the merge into main

felixweinberger:
@Kludex ready for another look

src/mcp/client/streamable_http.py
felixweinberger: This is actually not necessary because the transport sets per request headers anyway - added tests to demonstrate this behavior stays stable.

src/mcp/shared/_httpx_utils.py
```diff
@@ -4,7 +4,11 @@
 
 import httpx
 
-__all__ = ["create_mcp_http_client"]
+__all__ = ["create_mcp_http_client", "MCP_DEFAULT_TIMEOUT", "MCP_DEFAULT_SSE_READ_TIMEOUT"]
+
+# Default MCP timeout configuration
+MCP_DEFAULT_TIMEOUT = 30.0  # General operations (seconds)
+MCP_DEFAULT_SSE_READ_TIMEOUT = 300.0  # SSE streams - 5 minutes (seconds)
```
maxisbey, line 11: would we want this to be accessible to users of the library? it being in a file starting with `_` will give warnings to people who import this

README.md
```diff
@@ -2314,15 +2315,16 @@ async def main():
         callback_handler=handle_callback,
     )
 
-    async with streamablehttp_client("http://localhost:8001/mcp", auth=oauth_auth) as (read, write, _):
-        async with ClientSession(read, write) as session:
-            await session.initialize()
+    async with httpx.AsyncClient(auth=oauth_auth, follow_redirects=True) as custom_client:
```
maxisbey, line 2432: would it be worth adding a comment like `# optionally create custom httpx.AsyncClient` since by default this would work without creating the custom client here (I assume)

src/mcp/shared/_httpx_utils.py
maxisbey, line 25: (non blocking question) why do we not set the timeout default here instead of doing the check inside the function?

src/mcp/shared/_httpx_utils.py
felixweinberger: I don't really see a reason for this to be public, it's more to avoid magic numbers.

README.md
felixweinberger: Actually in this case you do need the custom client because that's the way to add `auth` now

maxisbey: Ahh ok makes sense. Hmm, well then what if you also want the default timeouts we've set for MCP (re: https://github.com/modelcontextprotocol/python-sdk/pull/1177#discussion_r2457605295)

src/mcp/shared/_httpx_utils.py
felixweinberger: Seems like a stylistic choice, I just went with the existing pattern but it's a fair point it could be done differently.

README.md
felixweinberger: You'd have to set up the timeouts yourself if you're setting up the client yourself. To avoid surprising the user with unexpected behaviors, if they create their own custom client we should assume they want to control the pieces.

There's a valid point though about potentially making `_httpx_utils` public and thus allowing people to use `create_mcp_http_client`. Not sure if that's really very valuable though because it's a very thin wrapper 🤔

src/mcp/shared/_httpx_utils.py
Kludex: No one needs to access this.

src/mcp/client/streamable_http.py
```diff
@@ -461,30 +465,54 @@ async def streamablehttp_client(
...
+    if client is None:
+        # Create default client with recommended MCP timeouts
+        client = create_mcp_http_client()
+
+    # Extract configuration from the client to pass to transport
+    headers_dict = dict(client.headers) if client.headers else None
+    timeout = client.timeout.connect if (client.timeout and client.timeout.connect is not None) else MCP_DEFAULT_TIMEOUT
+    sse_read_timeout = (
+        client.timeout.read if (client.timeout and client.timeout.read is not None) else MCP_DEFAULT_SSE_READ_TIMEOUT
+    )
```
Kludex: If the client is provided, why do we need to do this? 🤔

felixweinberger: Hm, the mechanical reason is because we changed the ordering in which things get created.

In the previous model we created the transport first with the settings (see line 473 before this change). Then we used the client factory to create the client with the same settings that we added to the transport.

In order for the two to match, we need to do the inverse now as the client can be created first. However, I see your point, there's some redundancy as the client is then used in the handlers...

felixweinberger: I guess there's a question whether `StreamableHTTPTransport` still needs all these arguments in the first place if we're supporting explicit creation of the client object?

```
headers: dict[str, str] | None = None,
timeout: float | timedelta = 30,
sse_read_timeout: float | timedelta = 60 * 5,
auth: httpx.Auth | None = None,
```

felixweinberger: I guess the decision we need to make is whether the `client` passed in is the source of truth on all configuration matters (headers, timeout, auth) or not. And if not, which things does the transport definitively need to override.

In the current model, the factory creates a client with the settings passed to `streamablehttp_client`, which creates some duplication. Then in `handle_get_stream` / `post_writer` / `_handle_post_request` the transport seems to override `headers`, `timeout`, `sse_read_timeout` again, but not always. So there's some unclear division of responsibility between which settings the client owns and which ones the transport owns.

If we follow a principle of "least surprise", I think we shouldn't be overriding anything if a client is explicitly constructed by the user of this API and passed in. So ideally the API of `StreamableHTTPTransport` becomes just `StreamableHTTPTransport(url)` and that's it, no other settings on the transport. The `client` owns all the settings and the transport takes the client as given.

And in the case where a user _doesn't_ pass in a client they configured themself, we configure a "sane" client that matches the current defaults, but we let the client fully own that configuration. 

Does that make sense?

felixweinberger: I added a commit on top of this to demonstrate what this could look like: https://github.com/modelcontextprotocol/python-sdk/pull/1177/commits/94df64b11d6b14881b4f36951cf6f1f2f8195857

Is that more the direction you were thinking @Kludex?

src/mcp/client/streamable_http.py
```diff
@@ -86,29 +86,22 @@ def __init__(
...
-            ACCEPT: f"{JSON}, {SSE}",
-            CONTENT_TYPE: JSON,
-            **self.headers,
-        }
-
-    def _prepare_request_headers(self, base_headers: dict[str, str]) -> dict[str, str]:
-        """Update headers with session ID and protocol version if available."""
-        headers = base_headers.copy()
+
+    def _prepare_headers(self, client: httpx.AsyncClient) -> dict[str, str]:
```
felixweinberger: can simplify to not pass the client as it gets merged anyway.

src/mcp/client/streamable_http.py
```diff
@@ -63,52 +70,85 @@ class RequestContext:
...
         Args:
             url: The endpoint URL.
-            headers: Optional headers to include in requests.
-            timeout: HTTP timeout for regular operations.
-            sse_read_timeout: Timeout for SSE read operations.
-            auth: Optional HTTPX authentication handler.
+            headers: DEPRECATED - Ignored. Configure headers on the httpx.AsyncClient instead.
+            timeout: DEPRECATED - Ignored. Configure timeout on the httpx.AsyncClient instead.
+            sse_read_timeout: DEPRECATED - Ignored. Configure read timeout on the httpx.AsyncClient instead.
+            auth: DEPRECATED - Ignored. Configure auth on the httpx.AsyncClient instead.
```
felixweinberger: This could be breaking if anyone was using these and relying on them for configuration.

src/mcp/client/streamable_http.py
```diff
@@ -63,52 +70,85 @@ class RequestContext:
...
 
     client: httpx.AsyncClient
-    headers: dict[str, str]
     session_id: str | None
     session_message: SessionMessage
     metadata: ClientMessageMetadata | None
     read_stream_writer: StreamWriter
-    sse_read_timeout: float
+    headers: dict[str, str] | None = None  # Deprecated - no longer used
+    sse_read_timeout: float | None = None  # Deprecated - no longer used
```
felixweinberger, line 85: could also be breaking if anyone was directly instantiating these classes and relying on the config here.

felixweinberger:
@Kludex, @maxisbey finally ready for another look after we talked about removing the `client` argument from `prepare_headers` last week.

There's _theoretically_ a potential for breakage because of the now deprecated arguments I commented on above, if anyone was using `RequestContext` or `StreamableHTTPTransport` directly for something. Like we discussed IIRC, none of our examples or documentation suggest using these classes directly. The only places they are used at the moment is in internal implementation details of `streamable_http_client` and `streamablehttp_client`.

`streamablehttp_client` should continue to work as before, as we've kept it around as `@deprecated` and convert the args.

Kludex:
I didn't think this was going to be merged already. I've been thinking about this, and I'm unsure this is the cleanest approach.

I think this API is not going to last, I think the right approach is to create the `MCPClient`/`Client` that it will manage the lifecycle of the HTTP client there. Right now, most of the LLM providers are doing this thing of accepting the HTTP client, but I think what everybody should be doing is allowing to pass all the parameters (a bit contradictory to what was the motivation to this PR, since I'm saying the completely opposite).

I think the `TestClient` from Starlette is more or less how I see that happening (doesn't need to be with inheritance, it can be with composition).

maxisbey:
> I didn't think this was going to be merged already. I've been thinking about this, and I'm unsure this is the cleanest approach.
> 
> I think this API is not going to last, I think the right approach is to create the `MCPClient`/`Client` that it will manage the lifecycle of the HTTP client there. Right now, most of the LLM providers are doing this thing of accepting the HTTP client, but I think what everybody should be doing is allowing to pass all the parameters (a bit contradictory to what was the motivation to this PR, since I'm saying the completely opposite).
> 
> I think the `TestClient` from Starlette is more or less how I see that happening (doesn't need to be with inheritance, it can be with composition).

Hmm, could you make an issue for this for v2 with some thoughts? would be good to capture it when going through the refactoring

felixweinberger:
> > I didn't think this was going to be merged already. I've been thinking about this, and I'm unsure this is the cleanest approach.
> > I think this API is not going to last, I think the right approach is to create the `MCPClient`/`Client` that it will manage the lifecycle of the HTTP client there. Right now, most of the LLM providers are doing this thing of accepting the HTTP client, but I think what everybody should be doing is allowing to pass all the parameters (a bit contradictory to what was the motivation to this PR, since I'm saying the completely opposite).
> > I think the `TestClient` from Starlette is more or less how I see that happening (doesn't need to be with inheritance, it can be with composition).
> 
> Hmm, could you make an issue for this for v2 with some thoughts? would be good to capture it when going through the refactoring

@maxisbey @Kludex made one here for discussion: https://github.com/modelcontextprotocol/python-sdk/issues/1773

tsingfei:
Will `McpHttpClientFactory` be deprecated in the future? I need to automatically recreate the connection to a streamable-http mcp srerver. But now it seems that only create the connection with an `http.AsyncClient` instance is supported. Well, I need to write the factory myself or I can reuse the `McpHttpClientFactory` to implement that logic?

richardye101:
For anyone like me who just ran into this SSL issue and is unclear about how to use this PR to get around it, this is all I had to do on the client to bypass the SSL issues:

```
async with httpx.AsyncClient(verify=False) as custom_httpx_client:
        async with streamable_http_client(url="https://my-mcp.com/mcp", http_client=custom_httpx_client) as (
                read_stream,
                write_stream,
                _,
            ):
                    await session.initialize()
                    print(tools)
```

Issue #961, opened 2025-06-16
Response ID-value not matching request value, due to default int


sProdz:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

When the client sends a request with id field value of type string, the server MUST respond with exactly the same value. When the #851 was accepted to version 1.9.4 it broke this standard causing errors for example in Microsoft Copilot Studio, because of the mismatch in the response id value.

"id
  This member is REQUIRED.
  It MUST be the same as the value of the id member in the Request Object.
  If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.
" [JSON-RPC 2.0 spec.](https://www.jsonrpc.org/specification)

### Example Code

```Python
curl -N \
  -X POST http://localhost:8000/mcp/ \                    
  -H "Accept: application/json, text/event-stream" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":"1","method":"tools/list","params":{}}' # String

Response:

{
    "jsonrpc": "2.0",
    "id": 1,  # Int
    "result": {
        "tools": [
            {
                "name": "query",
                "description": "Ask an LLM agent to answer the query using internal documents",
                "inputSchema": {
                    "properties": {
                        "query": {
                            "title": "Query",
                            "type": "string"
                        }
                    },
                    "required": [
                        "query"
                    ],
                    "title": "query_toolArguments",
                    "type": "object"
                },
                "annotations": {
                    "title": "Internal Docs Query",
                    "readOnlyHint": true,
                    "destructiveHint": false,
                    "idempotentHint": true,
                    "openWorldHint": false
                }
            }
        ]
    }
}
```

### Python & MCP Python SDK

```Text
Python 3.11
MCP Python SDK 1.9.4
```

falcowinkler:
We observed the issue that, when an initialize message is sent to a server using python SDK, it always returns a message id of integer type, even though the initial request had the type of this id be of string type. this is an example request/response pair for the integration with MS Copilot Studio.

Request:
```
{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "initialize",
    "params": {
        "capabilities": {},
        "clientInfo": {
            "agentName": "Test agent python",
            "appId": "d1296152-0c7a-4e79-8112-c1c87d5287e8",
[...]

PR #1178, opened 2025-07-21
Make sure `RequestId` is not coerced as `int`

Kludex:
- Closes https://github.com/modelcontextprotocol/python-sdk/issues/961

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -36,7 +36,7 @@ DEFAULT_NEGOTIATED_VERSION = "2025-03-26"
 ProgressToken = str | int
 Cursor = str
 Role = Literal["user", "assistant"]
-RequestId = Annotated[int | str, Field(union_mode="left_to_right")]
+RequestId = Annotated[int, Field(strict=True)] | str
 AnyFunction: TypeAlias = Callable[..., Any]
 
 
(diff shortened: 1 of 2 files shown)


src/mcp/types.py
```diff
@@ -849,7 +849,7 @@ class Tool(BaseMetadata):
     """A JSON Schema object defining the expected parameters for the tool."""
     outputSchema: dict[str, Any] | None = None
     """
-    An optional JSON Schema object defining the structure of the tool's output 
+    An optional JSON Schema object defining the structure of the tool's output
```
Kludex, line 852: Somehow there are empty spaces? This shouldn't happen.

src/mcp/types.py
```diff
@@ -36,7 +36,7 @@
 ProgressToken = str | int
 Cursor = str
 Role = Literal["user", "assistant"]
-RequestId = Annotated[int | str, Field(union_mode="left_to_right")]
+RequestId = Annotated[int, Field(strict=True)] | str
```
Kludex, line 39: This just makes sure that only `int` that is an `int` is actually an `int`. 👀

i.e. a string that looks like an int doesn't become an `int`.

ihrpr (approved):
LGTM

ns-amurugesan:
@ihrpr how do I use this fix? I am on mcp==1.12.0

Kludex:
> @ihrpr how do I use this fix? I am on mcp==1.12.0

You need to wait for a release.

PR #1179, opened 2025-07-21
[POC] Update logic for constructing protected resource metadata URL to include the resource's path component, for compatibility with Oauth RFCs

smurching:
## Motivation and Context
Update logic for constructing the protected resource metadata URL for an MCP server to include the path component of the resource being accessed, [as per section 3.1 of Oauth RFC 9728](https://datatracker.ietf.org/doc/rfc9728/#:~:text=oauth%2Dprotected%2Dresource.-,3.1.%20%20Protected%20Resource%20Metadata%20Request,-A%20protected%20resource):

>  3.1.  Protected Resource Metadata Request
> 
>    A protected resource metadata document MUST be queried using an HTTP
>    GET request at the previously specified URL.
> 
>    The consumer of the metadata would make the following request when
>    the resource identifier is https://resource.example.com and the well-
>    known URI path suffix is oauth-protected-resource to obtain the
>    metadata, since the resource identifier contains no path component:
> 
>      GET /.well-known/oauth-protected-resource HTTP/1.1
>      Host: resource.example.com
> 
>    If the resource identifier value contains a path or query component,
>    any terminating slash (/) following the host component MUST be
>    removed before inserting /.well-known/ and the well-known URI path
>    suffix between the host component and the path and/or query
>    components.  The consumer of the metadata would make the following
>    request when the resource identifier is https://resource.example.com/
>    resource1 and the well-known URI path suffix is oauth-protected-
>    resource to obtain the metadata, since the resource identifier
>    contains a path component:
> 
>      GET /.well-known/oauth-protected-resource/resource1 HTTP/1.1
>      Host: resource.example.com
> 
>    Using path components enables supporting multiple resources per host.
>    This is required in some multi-tenant hosting configurations.  This
>    use of .well-known is for supporting multiple resources per host;
>    unlike its use in [RFC8615], it does not provide general information
>    about the host.

To avoid breaking changes, we first attempt RFC-compliant behavior, e.g. for `https://api.example.com/mcp`, first try to query a PRM endpoint at `https://api.example.com/.well-known/oauth-protected-resource/mcp`. If that fails, we then fall back to querying `https://api.example.com/.well-known/oauth-protected-resource` (existing behavior, excluding the `/mcp` path component)

## How Has This Been Tested?
Updated unit tests - TODO test this against some real example servers

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -233,9 +233,16 @@ class OAuthClientProvider(httpx.Auth):
         url = self._extract_resource_metadata_from_www_auth(init_response)
 
         if not url:
-            # Fallback to well-known discovery
-            auth_base_url = self.context.get_authorization_base_url(self.context.server_url)
-            url = urljoin(auth_base_url, "/.well-known/oauth-protected-resource")
+            # Fallback to well-known discovery with path component included
+            parsed = urlparse(self.context.server_url)
+            auth_base_url = f"{parsed.scheme}://{parsed.netloc}"
+
+            if parsed.path and parsed.path != "/":
+                # Include path component in the well-known URL
+                path_component = parsed.path.rstrip("/")
+                url = urljoin(auth_base_url, f"/.well-known/oauth-protected-resource{path_component}")
+            else:
+                url = urljoin(auth_base_url, "/.well-known/oauth-protected-resource")
 
         return httpx.Request("GET", url, headers={MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION})
 
--- a/.gitignore
+++ b/.gitignore
@@ -168,3 +168,4 @@ cython_debug/
 .vscode/
 .windsurfrules
 **/CLAUDE.local.md
+.idea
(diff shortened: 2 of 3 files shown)


src/mcp/client/auth.py
```diff
@@ -510,9 +513,28 @@ async def async_auth_flow(self, request: httpx.Request) -> AsyncGenerator[httpx.
...
+                            discovery_request = httpx.Request(
+                                "GET", url, headers={MCP_PROTOCOL_VERSION: LATEST_PROTOCOL_VERSION}
+                            )
+                            discovery_response = yield discovery_request
+
+                            if discovery_response.status_code == 200:
+                                await self._handle_protected_resource_response(discovery_response)
+                                break  # Success, stop trying other URLs
+                            elif discovery_response.status_code != 404:
+                                break  # Non-404 error, stop trying
```
smurching, line 537: In practice we may want to keep retrying on other types of errors too, just to be safe

felixweinberger:
Hi @smurching thank you for this contribution - discussed internally with @pcarleton as we have another PR #1334 open that tackles the same issue.

We think #1334 is the right approach here, so closing this one for now - please feel free to chime in on that PR if you have thoughts on the approach!

PR #1181, opened 2025-07-22
Simplify code on `stdio_client`

Kludex:
This PR still doesn't solve the issue we have: if the process fails to start, it needs to raise. It doesn't. User is seeing `Client disconnected` message when running `client.initialize()`.

Maybe we can play with something like this in the task group (I'm not sure...):

```python
async def process_cancellation(cancel_scope: anyio.CancelScope):
    _ = await process.wait()
	cancel_scope.cancel()
```

---

It would be cool for `open_process` to raise an exception on `returncode != None`.

### DIFF
--- a/src/mcp/client/stdio/__init__.py
+++ b/src/mcp/client/stdio/__init__.py
@@ -1,14 +1,13 @@
 import logging
 import os
 import sys
-from contextlib import asynccontextmanager
+from contextlib import AsyncExitStack, asynccontextmanager
 from pathlib import Path
 from typing import Literal, TextIO
 
 import anyio
 import anyio.lowlevel
 from anyio.abc import Process
-from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from anyio.streams.text import TextReceiveStream
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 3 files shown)


src/mcp/client/stdio/__init__.py
```diff
@@ -107,33 +106,19 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
...
-            errlog=errlog,
-            cwd=server.cwd,
-        )
-    except OSError:
-        # Clean up streams if process creation fails
-        await read_stream.aclose()
-        await write_stream.aclose()
-        await read_stream_writer.aclose()
-        await write_stream_reader.aclose()
-        raise
```
Kludex, line 136: It actually never triggers `OSError` here. `anyio.open_process` doesn't trigger.

Maybe the Windows logic does, but even if it does... The streams don't need to be closed because they are not even open yet, so just removing the `except` is fine.

src/mcp/client/stdio/__init__.py
```diff
@@ -208,10 +192,6 @@ async def stdin_writer():
             except ProcessLookupError:
                 # Process already exited, which is fine
                 pass
-            await read_stream.aclose()
-            await write_stream.aclose()
-            await read_stream_writer.aclose()
-            await write_stream_reader.aclose()
```
Kludex: No need to close `read_stream` and `write_stream`, just add the async context manager block above, and no need for `read_stream_writer` and `write_stream_reader` because they are open and closed within the tasks above.

src/mcp/shared/session.py
Kludex, line 1: Readability changes - I can revert if wanted.

src/mcp/client/stdio/__init__.py
Kludex: Well, it seems I was wrong... 👀

src/mcp/client/stdio/__init__.py
Kludex: According to the test suite I'm wrong... I'm not sure how.

PR #1209, opened 2025-07-28
Feature: Non-Blocking call_tool and request state externalisation

davemssavage:
This patch enables a non-blocking mode of running tools and introduces a RequestStateManager plugin api - this allows for long running tasks to be rejoined on the client and across process restarts.

## Motivation and Context
Rather than introducing new protocol messages as per https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617 this patch uses the existing protocol messages with client side changes to allow the client application to submit a call request and then later join that request to get the response and/or any progress notifications that may have occured. This works on top of the existing session resume processes so a client can submit a task in one process and rejoin from another if a persistent version of RequestStateManager is used (this has been tested using a redis implementation - not included in this patch)

Long running tasks currently consume resources indefinitely on the client blocking until they return, Tool calls can currently timeout allowing control to return to the client however this leads to a messy control flow that exposes the internal details of the protocol implementaton - having to check the type of code is httpx.codes.REQUEST_TIMEOUT.

This patch instead introduces new methods request_call_tool and join_call_tool that handle various complex issues with resume tokens and via state stored in a pluggable RequestStateManager. An InMemoryRequestStateManager is provided as the default, other external state managers can be provided to the ClientSession to allow for persistence outside of process memory.

## How Has This Been Tested?
This has been unit tested and tested on a server application that managed multiple sessions without blocking the server side threads.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

In order to handle some complex logic around resumption this patch introduces a new ResumeCapability which can be returned during initialisation. It currently is hard coded to assume that if streamable_http protocol is used then server sessions are resumeable. It may be worth while extending this along the lines of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617 where servers can indicate how long clients might expect a session to be resumable for.

Testing on a non-trivial application has highlighted that at least one progress notification needs to be sent from the server prior to the start_request call returning a request_id. This is due to needing at least one event to trigger the sse.event_id to be passed back to the client for the resume token to be collected. This is a little opaque and relies on the client having passed a progress callback in the start_call_tool method. A new protocol notification sent from the server to indicate it has received the request could be a viable alternative, however one of the benefits of this is patch over https://github.com/modelcontextprotocol/modelcontextprotocol/pull/617 is it is trying to avoid new protocol messages.

### DIFF
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -882,7 +882,7 @@ async def test_streamablehttp_client_tool_invocation(initialized_client_session:
     """Test client tool invocation."""
     # First list tools
     tools = await initialized_client_session.list_tools()
-    assert len(tools.tools) == 6
+    assert len(tools.tools) == 7
     assert tools.tools[0].name == "test_tool"
 
     # Call the tool
@@ -919,7 +919,7 @@ async def test_streamablehttp_client_session_persistence(basic_server: None, bas
 
             # Make multiple requests to verify session persistence
             tools = await session.list_tools()
-            assert len(tools.tools) == 6
+            assert len(tools.tools) == 7
 
             # Read a resource
             resource = await session.read_resource(uri=AnyUrl("foobar://test-persist"))
@@ -948,7 +948,7 @@ async def test_streamablehttp_client_json_response(json_response_server: None, j
 
             # Check tool listing
             tools = await session.list_tools()
-            assert len(tools.tools) == 6
+            assert len(tools.tools) == 7
 
             # Call a tool and verify JSON response handling
             result = await session.call_tool("test_tool", {})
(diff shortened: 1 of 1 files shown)


felixweinberger:
This seems closely related to https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1391 where there's an ongoing discussion about the correct abstraction for long-running or async tools. 

In order to support this we'll likely need to align on the right path forward for implementation - if we end up adding this at the protocol layer we'll likely want to leverage any messages provided there.

felixweinberger:
Converting this to draft for now to remove it from the review queue while we wait for the outcome of https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1391

maxisbey:
Closing as the SEP this was waiting on (https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1391) has been closed.

Issue #200, opened 2025-02-10
Server url truncated by urllib in FastMCP


mconflitti-pbc:
**Describe the bug**
When using urllib url joining, this truncates server urls paths when combining `/messages/` set in the transport object.
```python
sse = SseServerTransport("/messages/")
```

For example, if joining `http://localhost:8000/some/path/to/sse` with `/messages/`, it will output: `http://localhost:8000/messages/` instead of honoring the path.

Removing the leading slash fixes this.

This was found in the situation where my server was being proxied and accessed at a subpath. afaict, there is no explicit reason to keep the leading slash. This is also something that could be addressed in #194 since a dev could then customize the server setup more easily.

**To Reproduce**
```python
from urllib.parse import urljoin

url = "http://localhost:8000/some/path/to/sse"
sse_data = "/messages/?session_id=123"

endpoint_url = urljoin(url, sse_data)
assert endpoint_url == "http://localhost:8000/messages/?session_id=123"

####

url = "http://localhost:8000/some/path/to/sse"
sse_data = "messages/?session_id=123"

endpoint_url = urljoin(url, sse_data)
assert endpoint_url == "http://localhost:8000/some/path/to/messages/?session_id=123"
```

**Expected behavior**
 I would expect this to join the route to the end of the url instead of truncating it.

SOURABHMISHRA5221:
Hi @mconflitti-pbc, @Kludex!
I've made the necessary changes to fix the issue. I'd appreciate it if you could review them.

allenfeng8:
@mconflitti-pbc For the updates in the [PR](https://github.com/modelcontextprotocol/python-sdk/pull/1211), do you think this issue should be closed now?

mconflitti-pbc:
Yes. Given that SSE is deprecated, this is no longer urgent. If someone finds this is still affecting them for streamable http, they can reopen and address it.

PR #1211, opened 2025-07-28
fix: resolve URL path truncation in SSE transport for proxied servers

SOURABHMISHRA5221:
### Summary
Fix SSE endpoint path handling to avoid `urllib.parse.urljoin()` truncating base paths when servers are behind proxies or mounted at subpaths. Endpoints are treated as relative path segments and stored verbatim.

### Motivation
When clients join a base URL with an endpoint starting with “/”, `urljoin` resets to the host root. This caused paths like `http://localhost:8000/some/path/to/sse` + `/messages/` → `http://localhost:8000/messages/`. Using a relative segment preserves the base path. Resolves [#200](https://github.com/modelcontextprotocol/python-sdk/issues/200).

```python
from urllib.parse import urljoin
urljoin("http://host/hello/world", "/messages/")  # -> http://host/messages/
urljoin("http://host/hello/world/", "messages/")  # -> http://host/hello/world/messages/
```

### What changed
- `mcp/server/sse.py`
  - Store `endpoint` verbatim (no auto-leading “/”).
  - Validate endpoint is a relative path (no scheme/host/query/fragment).
  - Add `_build_message_path(root_path)` to construct the final path robustly from the app mount and `root_path`.
  - Update comments/docstrings to emphasize relative path semantics and predictable URL construction.
- Tests
  - `tests/server/test_sse_security.py`: clarify accept/reject cases and document `urljoin` behavior.
  - `tests/shared/test_sse.py`: add/clarify param cases; ensure endpoints are stored verbatim.

### Backward compatibility
- No breaking changes.

### How it was tested
- Endpoint validation tests for relative vs absolute and query/fragment rejection.
- Subpath mounting tests to reproduce and prevent original truncation.
- All existing SSE security/functionality tests pass.
- Verified path construction under various `root_path` combinations.

### Checklist
- [x] Follows repo style guidelines
- [x] Tests added/updated and passing
- [x] Error handling and docs updated

### DIFF
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -502,7 +502,7 @@ def test_sse_message_id_coercion():
 )
 def test_sse_server_transport_endpoint_validation(endpoint: str, expected_result: str | type[Exception]):
     """Validate relative endpoint semantics and storage.
-
+    
     Context on URL joining (urllib.parse.urljoin):
     - Joining a segment starting with "/" resets to the host root:
       urljoin("http://host/hello/world", "/messages") -> "http://host/messages"


SOURABHMISHRA5221:
Hi @Kludex!... Can you please review this...

src/mcp/server/sse.py
```diff
@@ -30,6 +36,15 @@ async def handle_sse(request):
     uvicorn.run(starlette_app, host="127.0.0.1", port=port)
 ```
 
+Path behavior examples:
+- With root_path="" and endpoint="/messages/": Final path = "/messages/"
+- With root_path="" and endpoint="messages/": Final path = "/messages/"
+- With root_path="/api" and endpoint="/messages/": Final path = "/api/messages/"
+- With root_path="/api" and endpoint="messages/": Final path = "/api/messages/"
```
mconflitti-pbc: The behavior I saw was: `urllib.join("http://example.com/some/path", "/messages/")` resulted in `http://example.com/messages`. I think it would be sufficient to just drop the leading forward slash sense that leads to an unexpected behavior. I would argue that the sdk should not be forcing its route to be at any particular location and that a dev should decide where it gets mounted. Relative path joining should allow that. See comment below for test cases

tests/server/test_sse_security.py
```diff
@@ -291,3 +291,34 @@ async def test_sse_security_post_valid_content_type(server_port: int):
...
+        with pytest.raises(ValueError, match="is not a relative path"):
+            SseServerTransport(invalid_endpoint)
+
+    # These should all be valid - endpoint is stored as-is (no automatic normalization)
+    valid_endpoints_and_expected = [
+        ("/messages/", "/messages/"),  # Absolute path format
+        ("messages/", "messages/"),  # Relative path format
+        ("/api/v1/messages/", "/api/v1/messages/"),
+        ("api/v1/messages/", "api/v1/messages/"),
+    ]
```
mconflitti-pbc: Here is the behavior demonstrated with examples:
```
>>> from urllib.parse import urljoin
>>> 
>>> urljoin("http://www.google.com/hello/world", "/messages")
'http://www.google.com/messages'
>>> urljoin("http://www.google.com/hello/world", "messages")
'http://www.google.com/hello/messages'
>>> urljoin("http://www.google.com/hello/world/", "messages")
'http://www.google.com/hello/world/messages'
>>> urljoin("http://www.google.com/hello/world/", "/messages")
'http://www.google.com/messages'
>>> urljoin("http://www.google.com/hello/world/", "/messages/")
'http://www.google.com/messages/'
>>> urljoin("http://www.google.com/hello/world/", "messages/")
'http://www.google.com/hello/world/messages/'
>>> urljoin("http://www.google.com/hello/world", "messages/")
'http://www.google.com/hello/messages/'
```

urllib has some odd behavior imo.

mconflitti-pbc: I believe this is the one we want to coerce/enforce:
```
>>> urljoin("http://www.google.com/hello/world/", "messages/")
'http://www.google.com/hello/world/messages/'
```

src/mcp/server/sse.py
SOURABHMISHRA5221: Hi @mconflitti-pbc! Thanks for reviewing this PR, and apologies for the late reply.

You’re right about the behavior: auto-prepending “/” to “make it relative” actually made the endpoint origin-absolute, which caused urljoin to drop any base path in proxied/subpath setups. This change removes that normalization and stores the endpoint verbatim.

- We validate only that the endpoint is a relative path (no scheme/host/query/fragment).
- Both “/messages/” and “messages/” are accepted; we recommend the relative segment (“messages/”) for predictable joining.
- The final URL is composed from the app’s mount and ASGI root_path, so mounting remains entirely up to the developer.

Thanks again for the thoughtful feedback.

tests/server/test_sse_security.py
SOURABHMISHRA5221: I’ve made the necessary updates based on your feedback (store endpoint verbatim, remove leading “/” coercion, clarify docs/tests). When you have a moment, could you please take another look? Thank you

mconflitti-pbc:
It appears there is already some normalization logic added to the repo? https://github.com/modelcontextprotocol/python-sdk/blob/a82c69bfd91a3f9bf876657ad440ca967eb3ba06/src/mcp/server/fastmcp/server.py#L693

Also, SSE is deprecated now so it may be worth reworking this for streamable http

DigdashQuentinLandi:
> It appears there is already some normalization logic added to the repo?
> 
> https://github.com/modelcontextprotocol/python-sdk/blob/a82c69bfd91a3f9bf876657ad440ca967eb3ba06/src/mcp/server/fastmcp/server.py#L693
> 
> Also, SSE is deprecated now so it may be worth reworking this for streamable http

hi , is the issue still open on Streamable http?

SOURABHMISHRA5221:
Hi @felixweinberger  and others,
Yes, @mconflitti-pbc rightly pointed out there's already a _normalize_path method at line 693 in src/mcp/server/fastmcp/server.py that handles mount path normalization. After examining the codebase more carefully, I realize the StreamableHTTPServerTransport (current transport) doesn't have this URL joining problem , it uses a single endpoint with content negotiation via Accept headers, so no urljoin() occurs in the client. This PR only addresses the deprecated SSE transport, which sends separate "endpoint" events that trigger the URL joining issue. Would you prefer I close this PR since it only affects deprecated code, or keep it for users still transitioning from SSE transport?

felixweinberger:
> Hi @felixweinberger and others, Yes, @mconflitti-pbc rightly pointed out there's already a _normalize_path method at line 693 in src/mcp/server/fastmcp/server.py that handles mount path normalization. After examining the codebase more carefully, I realize the StreamableHTTPServerTransport (current transport) doesn't have this URL joining problem , it uses a single endpoint with content negotiation via Accept headers, so no urljoin() occurs in the client. This PR only addresses the deprecated SSE transport, which sends separate "endpoint" events that trigger the URL joining issue. Would you prefer I close this PR since it only affects deprecated code, or keep it for users still transitioning from SSE transport?

Thanks for this additional detail - in this case I'd prefer keeping things simple and focusing on `StreamableHTTP` and close this one. I don't think we'll have the bandwidth to maintain deprecated functionality.

Issue #883, opened 2025-06-04
AssertionError: Unexpected message while using middlewares


toughnoah:
**Describe the bug**
when I mount the mcp to starlette, and I use middleware to do some prechecks, I met this errors.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 187, in __call__
    raise exc
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
    await self.app(scope, receive, _send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 182, in __call__
    with recv_stream, send_stream, collapse_excgroups():
                                   ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/app/.venv/lib/python3.12/site-packages/starlette/_utils.py", line 83, in collapse_excgroups
    raise exc
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 185, in __call__
    await response(scope, wrapped_receive, send)
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 223, in __call__
    async for chunk in self.body_iterator:
  File "/app/.venv/lib/python3.12/site-packages/starlette/middleware/base.py", line 169, in body_stream
    assert message["type"] == "http.response.body", f"Unexpected message: {message}"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1201, opened 2025-07-26
Add CORS middleware to Starlette app for frontend support

SmartManoj:
Introduces CORS middleware to the Starlette app instances in FastMCP to allow requests from local web frontends. This enables compatibility with web clients running on localhost and related origins, supporting credentials and common HTTP methods and headers.

## Motivation and Context

Fixes https://github.com/modelcontextprotocol/python-sdk/issues/187#issuecomment-2994114298

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #1222, opened 2025-07-31
fix: Handle SSE Disconnects Properly When use starlette middleware

wenxuwan:
## Motivation and Context

When using the Starlette middleware capabilities, if the SSE client disconnects, the current server will return starlette.responses.Response, and the Response will send a message of type "http.response.start", which will cause the following error：

	File "/opt/anaconda3/lib/python3.12/contextlib.py", line 158, in __exit__
	    self.gen.throw(value)
	  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/_utils.py", line 83, in collapse_excgroups
	    raise exc
	  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/middleware/base.py", line 185, in __call__
	    await response(scope, wrapped_receive, send)
	  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/middleware/base.py", line 223, in __call__
	    async for chunk in self.body_iterator:
	  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/middleware/base.py", line 169, in body_stream
	    assert message["type"] == "http.response.body", f"Unexpected message: {message}"
	           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	AssertionError: Unexpected message: {'type': 'http.response.start', 'status': 200, 'headers': [(b'content-length', b'0')]}

There are several issues that should be related to this PR, such as:

https://github.com/modelcontextprotocol/python-sdk/issues/883
https://github.com/jlowin/fastmcp/issues/858

## How Has This Been Tested?

Tested manually and ensured that the route was closed when clients close the connection.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist


ochafik (changes requested):
Thanks @wenxuwan !

Tested it w/ a variant of @llamantino's [code snippet](https://github.com/modelcontextprotocol/python-sdk/issues/883#issuecomment-3105036571) (just updated import to `from mcp.server.fastmcp import FastMCP`) and it seems to fix the issue 🎉

Two questions:
- Could you try and add some test case?
- Would/should this also apply to sse streams from streamable http?

wenxuwan:
> Thanks @wenxuwan !
> 
> Tested it w/ a variant of @llamantino's [code snippet](https://github.com/modelcontextprotocol/python-sdk/issues/883#issuecomment-3105036571) (just updated import to `from mcp.server.fastmcp import FastMCP`) and it seems to fix the issue 🎉
> 
> Two questions:
> 
> * Could you try and add some test case?
> * Would/should this also apply to sse streams from streamable http?

@ochafik Thank you for your comments! , For your questions:

* Could you try and add some test case?
Currently, the errors on the server side are caught by Uvicorn, making it challenging for us to add corresponding test cases.

 * Would/should this also apply to sse streams from streamable http?
The handling logic for SSE streams in streamable HTTP is different from that of the older SSE streams. In the case of streamable HTTP, the server closes the SSE connection after responding to the client’s [DELETE request](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py#L287), so the issue we’re discussing does not apply in this context.

Kludex:
I don't think this is the solution. Why are we sending 2 `http.response.start` events? The solution should be to avoid sending that, not doing a workaround with the response class.

wenxuwan:
> I don't think this is the solution. Why are we sending 2 `http.response.start` events? The solution should be to avoid sending that, not doing a workaround with the response class.

The operation for sending the response was introduced in https://github.com/modelcontextprotocol/python-sdk/pull/612

Kludex:
> > I don't think this is the solution. Why are we sending 2 `http.response.start` events? The solution should be to avoid sending that, not doing a workaround with the response class.
> 
> The operation for sending the response was introduced in #612

Ah... I see the issue.

The problem is that we are mixing _Starlette way_ with _pure ASGI way_. 😞

---

This PR is not the solution, it just hides the conceptual issue from above. I guess the solution is to decide what to do... We probably want to leverage Starlette instead of use the ASGI Interface directly...

wenxuwan:
> > > I don't think this is the solution. Why are we sending 2 `http.response.start` events? The solution should be to avoid sending that, not doing a workaround with the response class.
> > 
> > 
> > The operation for sending the response was introduced in #612
> 
> Ah... I see the issue.
> 
> The problem is that we are mixing _Starlette way_ with _pure ASGI way_. 😞
> 
> This PR is not the solution, it just hides the conceptual issue from above. I guess the solution is to decide what to do... We probably want to leverage Starlette instead of use the ASGI Interface directly...



> > > I don't think this is the solution. Why are we sending 2 `http.response.start` events? The solution should be to avoid sending that, not doing a workaround with the response class.
> > 
> > 
> > The operation for sending the response was introduced in #612
> 
> Ah... I see the issue.
> 
> The problem is that we are mixing _Starlette way_ with _pure ASGI way_. 😞
> 
> This PR is not the solution, it just hides the conceptual issue from above. I guess the solution is to decide what to do... We probably want to leverage Starlette instead of use the ASGI Interface directly...

@Kludex Thank you for your suggestion. As you mentioned, the current legacy implementation of MCP's SSE mixes ASGI and Starlette, which poses challenges for the seamless integration of the handle_sse implementation with Starlette. I see one potential solution to address this issue.Maintain our existing implementation as is, but wrap the current handle_sse implementation in a Response, allowing handle_sse to directly return a McpResponse, For example:
	
	
		class McpResponse(Response):
		    """A response that does not send any HTTP response back to the client."""
		
		    def __init__(self, sse: SseServerTransport, mcp_server: MCPServer) -> None:
		        self.sse = sse
		        self._mcp_server = mcp_server
		        super().__init__()
		
		    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
		        async with self.sse.connect_sse(
		                scope,
		                receive,
		                send,
		        ) as streams:
		            await self._mcp_server.run(
		                streams[0],
		                streams[1],
		            )
		        return
            
		
		async def handle_sse(scope: Scope, receive: Receive, send: Send):
                # Add client ID from auth context into request context if available
                return SilentResponse(sse, self._mcp_server)	


In this way, our implementation will comply with the requirements of the Starlette framework, even though we internally use ASGI directly. At the same time, it does not require significant changes to the existing architecture.What do you think?

Kludex:
Can't we just remove those Response for now?

wenxuwan:
> Can't we just remove those Response for now?

No，if we directly remove return Response, the Starlette framework will throw an exception.

		  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/routing.py", line 290, in handle
		    await self.app(scope, receive, send)
		  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/routing.py", line 78, in app
		    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
		  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
		    raise exc
		  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
		    await app(scope, receive, sender)
		  File "/Users/wangwenxue/code/gitlab/sofapyapp/myenv/lib/python3.12/site-packages/starlette/routing.py", line 76, in app
		          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Because the Starlette framework expects the handler to return a Response, as shown below:

	    async def app(scope: Scope, receive: Receive, send: Send) -> None:
	        request = Request(scope, receive, send)
	
	        async def app(scope: Scope, receive: Receive, send: Send) -> None:
	            response = await f(request)
	
	        await wrap_app_handling_exceptions(app, request)(scope, receive, send)
	
	    return app

Therefore, the handle_sse function needs to return a Response.

Kludex:
Ah, I see the issue.

We can't really mix things here. We need to remove the way we handle with pure ASGI. Which means refactoring quite a lot here, but it needs to be done.

wenxuwan:
> Ah, I see the issue.
> 
> We can't really mix things here. We need to remove the way we handle with pure ASGI. Which means refactoring quite a lot here, but it needs to be done.

@Kludex   I completely agree with your suggestion. Refactoring the current implementation is indeed a significant undertaking. I believe we can wrap the current handle_sse implementation in a way that conforms to the Starlette framework standards, while keeping the original implementation intact, just like my current implementation. This would be a more elegant solution compared to returning an invalid Response as before.

What are your thoughts on this approach?

felixweinberger:
Couple of middleware related topics, e.g. https://github.com/modelcontextprotocol/python-sdk/pull/1201

Given this looks it would require some significant refactoring that we should think through carefully, I'm tentatively marking this for `v2` now as a placeholder for maintainers to look into proactively once we've handled inbound

felixweinberger:
Converting this to draft for now to remove it from the immediate review queue as we've already marked this for v2.

BossChaos:
## Code Review

**PR:** [fix: Handle SSE Disconnects Properly When use starlette middleware](https://github.com/modelcontextprotocol/python-sdk/pull/1222) by @wenxuwan

- ✅ Bug/security fix

*Code review*

PR #1226, opened 2025-08-01
Fix : token_endpoint_auth_signing_alg_values_supported is a json array

yannj-fr:
## Motivation and Context

Fixes the OAuthMetadata model validated by Pydantic, token_endpoint_auth_signing_alg_values_supported has to be a list of string, or validation would fail
## How Has This Been Tested?
Tested against an Auth0 oauth implementation

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -117,7 +117,7 @@ class OAuthMetadata(BaseModel):
     response_modes_supported: list[Literal["query", "fragment", "form_post"]] | None = None
     grant_types_supported: list[str] | None = None
     token_endpoint_auth_methods_supported: list[str] | None = None
-    token_endpoint_auth_signing_alg_values_supported: None = None
+    token_endpoint_auth_signing_alg_values_supported: list[str] | None = None
     service_documentation: AnyHttpUrl | None = None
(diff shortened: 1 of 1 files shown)


yannj-fr:
fixed accordingly and updates the possible values. 
I have been checking the rest for consistency

src/mcp/shared/auth.py
```diff
@@ -117,17 +117,17 @@ class OAuthMetadata(BaseModel):
     response_modes_supported: list[Literal["query", "fragment", "form_post"]] | None = None
     grant_types_supported: list[str] | None = None
     token_endpoint_auth_methods_supported: list[str] | None = None
-    token_endpoint_auth_signing_alg_values_supported: None = None
+    token_endpoint_auth_signing_alg_values_supported: list[Literal["RS256", "ES256"]] | None = None
```
Kludex: With my previous comment I didn't mean to set the `Literal`.

I think the above also had `Literal` before but we had to switch to `str` because of something I can't remember... I would have to check.

Can you change it to `list[str]`?

---

Happy to approve after it.

yannj-fr: done

Issue #1216, opened 2025-07-30
Expose _meta in ClientSession methods


samchenatti:
### Description

Hello!

I'd like to request exposing the argument `_meta` in the interface `ClientSession.call_tool`.

If approved, I could implement the feature myself.

**Use case for the feature**
When calling a tool my, MCP Client needs to send metadata to the MCP Server, such user preferences and other stuff that might affect how the tools works. The LLM shouldn't be aware of such data, so this information doesn't classify as tool arguments.

[Taking a look at the protocol documentation](https://modelcontextprotocol.io/specification/2025-06-18/basic#meta) it seems like using the `_meta` field would be appropriated for this scenario.

**Demo of how code might look when using the feature**
Currently `ClientSession.call_tool` signature is as follows:
```python
async def call_tool(
        self,
        name: str,
        arguments: dict[str, Any] | None = None,
        read_timeout_seconds: timedelta | None = None,
        progress_callback: ProgressFnT | None = None,
    ) -> types.CallToolResult:
```

I'm suggesting the following:
```python
    async def call_tool(
        self,
        name: str,
        arguments: dict[str, Any] | None = None,
        read_timeout_seconds: timedelta | None = None,
        progress_callback: ProgressFnT | None = None,
        meta: dict | None # <- Exposes meta in the public interface
    ) -> types.CallToolResult:
        """Send a tools/call request with optional progress callback support."""

        result = await self.send_request(
            types.ClientRequest(
                types.CallToolRequest(
                    method="tools/call",
                    params=types.CallToolRequestParams(
                        name=name,
                        arguments=arguments,
                        meta=meta or {} # <- include meta in the JSONRPC request arguments
                    ),
                )
            ),
            types.CallToolResult,
            request_read_timeout_seconds=read_timeout_seconds,
            progress_callback=progress_callback,
        )
```

Since `_meta` is already exposed by `CallToolRequestParams` it would be a simple change.

**Alternatives**
[...]

Issue #1197, opened 2025-07-25
Why MCP doesn't support metadata?


johnson7788:
### Description

Some mcp tool, i don't want let LLM know, So i want transfer data by metadata, How can it do it? thank you !

Server example:
```
import datetime
from fastmcp import FastMCP, Context
from mcp.types import CallToolResult

mcp = FastMCP("EchoServer")

@mcp.tool()
async def echo_tool(text: str, ctx: Context) -> CallToolResult:
    # 从 ctx 获取 metadata
    meta_in = ctx.request_meta or {}
    print("[Server] meta_in:", meta_in)

    return CallToolResult(
        content=[{"type": "text", "text": text.upper()}],
        meta={
            "user_id": meta_in.get("user_id", "unknown"),
            "server_timestamp": datetime.datetime.utcnow().isoformat() + "Z"
        }
    )

if __name__ == "__main__":
    mcp.run(transport="sse")
```

Client Example
```
# client.py
from mcp.types import CallToolRequest, CallToolRequestParams, RequestParams
import asyncio
from fastmcp import Client
from mcp.types import CallToolRequestParams, RequestParams
import json

def simulate():
    # 构造调用请求，带上 meta 信息
    req = CallToolRequest(
        method="tools/call",
        params=CallToolRequestParams(
            name="echo_tool",
            arguments={"text": "hello MCP"},
            meta=RequestParams.Meta(  # 通过 alias="_meta" 被序列化为 "_meta"
                user_id="alice",
                session_id="sess-123",
            ),
        ),
    )

    # 序列化为 JSON-RPC 并发送
    payload = req.model_dump(by_alias=True)
    print("Sending:", json.dumps(payload, indent=2))
    # 这里你可以用 WebSocket/httpx/aiohttp 发送 payload 到 MCP server


# client_sse.py
async def main():
    # 指定 SSE 地址（在 server.run(transport="sse") 中默认挂载为 http://localhost:8000/sse）
    async with Client("http://localhost:8000/sse") as client:
        # 调用 echo_tool，自动使用 SSE transport
        result = await client.call_tool(
            "echo_tool",
            {"text": "hello MCP"},
            meta=RequestParams.Meta(
                user_id="alice",
                session_id="sess-123"
            )
        )
        print("Response content:", result.content)
        print("Response meta:", result.meta)

if __name__ == "__main__":
    asyncio.run(main())
```

samchenatti:
Hello, @johnson7788 

[...]

PR #1231, opened 2025-08-03
Expose RequestParams._meta in ClientSession.call_tool

samchenatti:
## Motivation and Context
This change allow clients to send metadata through the `_meta` field [as specified by the protocol](https://modelcontextprotocol.io/specification/2025-06-18/basic#meta).

It is important to expose `_meta` since there might be some metadata required by the server to control how the tool works (such as client preferences, its localization and so on). In some cases we also don't want the Agent/LLM to be aware of such data; it should be known by the MCP Host and Server only; therefore it cannot be exposed as tool usual arguments.

This feature has already been requested [by me](https://github.com/modelcontextprotocol/python-sdk/issues/1216) and [another user](https://github.com/modelcontextprotocol/python-sdk/issues/1197) - so it might benefit another people as well.

I also [started a discussion](https://github.com/jlowin/fastmcp/issues/1315) about `_meta` being exposed by FastMCP Client, but they will wait until its implemented in here.

## How Has This Been Tested?
I tested it on my local environment using a self-made MCP Server. 

Since `_meta` is currently being used for progress report tokens, I also ensured this feature doesn't break `progress_callback`.

I'd be happy to implement the required unit-tests if someone can point the best test-file for it.

## Breaking Changes
There are no breaking changes. The `python-sdk` already considers that `_meta` can carry data other than `progressToken`.

Also `types.RequestParams.Meta` model is already configured to allow extra key-values.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
I decided to expose `meta` as a plain dict instead of `types.RequestParams.Meta` since, from my understanding, `mcp.types` shouldn't be exposed to final users.

Metadata could be exposed by another interfaces as well, such as `list_tools`, `list_prompts` and etc. I'm sticking with `tool_call` for now, but would be happy to replicate it to the other interfaces.


src/mcp/client/session.py
```diff
@@ -295,8 +296,7 @@ async def call_tool(
                 types.CallToolRequest(
                     method="tools/call",
                     params=types.CallToolRequestParams(
-                        name=name,
-                        arguments=arguments,
+                        name=name, arguments=arguments, _meta=types.RequestParams.Meta(**(meta or {}))
```
samchenatti: Handling meta separately would improve readability and avoid unnecessary call to PyDantic:
```python
        ...
        _meta: types.ReadResourceRequestParams.Meta | None = None
        if meta is not None:
            _meta = types.RequestParams.Meta(**meta)

        result = await self.send_request(
            types.ClientRequest(
                types.CallToolRequest(
                    method="tools/call",
                    params=types.CallToolRequestParams(name=name, arguments=arguments, _meta=_meta),
                )
            ),
```

src/mcp/client/session.py
```diff
@@ -286,6 +286,7 @@ async def call_tool(
         name: str,
         arguments: dict[str, Any] | None = None,
         read_timeout_seconds: timedelta | None = None,
+        meta: dict[str, Any] | None = None,
```
dsp-ant: Note that this ordering breaks BC. If we go down this route we should probably just consistently 
```suggestion
        name: str,
        arguments: dict[str, Any] | None = None,
        *,
        read_timeout_seconds: timedelta | None = None,
        meta: dict[str, Any] | None = None,
```

samchenatti: Thanks for the review!

Since `read_timeout_seconds` is already part of the public interface as a positional argument, shouldn't it be:
```python
        name: str,
        *,
        meta: dict[str, Any] | None = None,
```
?

samchenatti: I'll follow with `meta` as the solely "forced" name parameter.

@dsp-ant I can make a new change if you want.

danielabonfim:
This feature looks great and would be a valuable addition. Nice work! Would it be possible to get this approved for merge @dsp-ant

samchenatti:
Hey, @dsp-ant!

Are there any other changes I need to make?

discorev:
I'd also be interested in this, is there anything further that needs changing?

samchenatti:
Hey, @dsp-ant, @ihrpr!

Is there anything else I need to fix in order to get this PR merged?

aadarshsingh191198:
I'd also be interested in this! Requesting maintainers to merge it if there are no unwanted side effects of this change.

maeryo:
Great work @samchenatti! This feature would be really valuable for the MCP ecosystem. I hope the conflicts can be resolved soon and this PR gets merged. The community clearly supports this addition based on all the positive feedback. Keep up the excellent work! 🚀

samchenatti:
Thanks for the feedback, @maeryo and @felixweinberger!

I'll fix the conflicts by the end of the day

samchenatti:
@felixweinberger, @maeryo conflicts solved!

maeryo:
cool! I also checked today and noticed that the kotlin-sdk didn’t have this implementation yet, so I submitted a PR with a similar approach to yours. In addition to just adding the meta parameter, I paid extra attention to improving validation logic as well.

https://github.com/modelcontextprotocol/kotlin-sdk/pull/289

samchenatti:
Hey, @maeryo, thanks for the feedback!

As suggested, I've implemented the validation here as well, but with a different approach.

Instead of validating _meta in the tool call only, I've implemented the validation directly within the types.RequestParams.Meta model. This seems more appropriate from a Pydantic standpoint.

I've also added unit tests with positive and negative keys, so you can confirm that all the rules are covered.

**Note**: Please be aware that these validations could be a breaking change, since users may already have non-conformant metadata keys in use.

**Edit:** I think it is a better idea to make a new PR focusing on the metadata key validation; see https://github.com/modelcontextprotocol/python-sdk/pull/1397

samchenatti:
@felixweinberger @ihrpr let me know if there's anything else i need to change!

Jaorow:
Hey @samchenatti
This is awesome and really helps out with something im trying to build! 

Do you have an ETA on when this might be in?

samchenatti:
Hey @Jaorow!

I'm waiting on the maintainers’ feedback...

This feature has become even more important since OpenAI announced the App SDK.
They’re making heavy use of _meta to enable UI components.

By merging this PR, we’d allow other MCP Clients to do the same.

src/mcp/client/session.py
```diff
@@ -273,16 +273,20 @@ async def call_tool(
...
+            _meta = types.RequestParams.Meta(**meta)
+
         result = await self.send_request(
             types.ClientRequest(
                 types.CallToolRequest(
-                    params=types.CallToolRequestParams(
-                        name=name,
-                        arguments=arguments,
-                    ),
+                    method="tools/call",
```
maxisbey: remove this, not needed as it's the default value

src/mcp/client/session.py
```diff
@@ -273,16 +273,20 @@ async def call_tool(
         arguments: dict[str, Any] | None = None,
         read_timeout_seconds: timedelta | None = None,
         progress_callback: ProgressFnT | None = None,
+        *,
+        meta: dict[str, Any] | None = None,
     ) -> types.CallToolResult:
         """Send a tools/call request with optional progress callback support."""
 
+        _meta: types.ReadResourceRequestParams.Meta | None = None
```
maxisbey: Why is this ReadResourceRequestParams?

maxisbey (changes requested):
I'd want to see some unit tests for this. You could add them in `tests/server/test_session.py`

Jaorow:
> Hey @Jaorow!
> 
> I'm waiting on the maintainers’ feedback...
> 
> This feature has become even more important since OpenAI announced the App SDK. They’re making heavy use of _meta to enable UI components.
> 
> By merging this PR, we’d allow other MCP Clients to do the same.

@samchenatti Awesome, It will definitely open the door for more complex and structured functionality.

samchenatti:
Hey, @maxisbey!

I made the requested changes and implemented the unit test.

~I had to cheat a bit in the test and manually set `_tool_output_schemas`~

After a good night’s sleep, I was able to properly implement the unit test (commit [2b61abba](https://github.com/samchenatti/mcp-python-sdk/commit/2b61abba1368a760a42149d084291b56bcd318f2)).

Now the client and server exchange all the necessary messages just like they would in a real scenario.

samchenatti:
@maxisbey thanks for the approval!

@ihrpr, @dsp-ant can you guys also take a look on my changes?

PR #1236, opened 2025-08-04
Add test for ProtectedResourceMetadataParsing

yannj-fr:
## Motivation and Context
Add tests cases for insurring ProtectedResourceMetadata validation is correctly done

## How Has This Been Tested?
with pytest

## Types of changes

- [ ] Documentation update

## Checklist


tests/shared/test_auth.py
```diff
@@ -37,3 +38,84 @@ def test_oidc(self):
...
+                "resource": "https://example.com/resource",
+                "authorization_servers": ["https://example.com/oauth2/authorize"],
+                "scopes_supported": ["read", "write"],
+                "bearer_methods_supported": ["header"],
+                "resource_name": "Example Resource",
+                "resource_documentation": "https://example.com/resource/documentation",
+            }
+        )
+
+    def test_valid_witn_invalid_resource_documentation(self):
```
pcarleton:
```suggestion
    def test_valid_with_invalid_resource_documentation(self):
```

tests/shared/test_auth.py
```diff
@@ -37,3 +38,84 @@ def test_oidc(self):
...
+                "jwks_uri": "https://example.com/.well-known/jwks.json",
+                "scopes_supported": ["read", "write"],
+                "bearer_methods_supported": ["header"],
+                "resource_signing_alg_values_supported": ["RS256"],
+                "resource_name": "Example Resource",
+                "resource_documentation": "https://example.com/resource/documentation",
+                "resource_policy_uri": "https://example.com/resource/policy",
+                "resource_tos_uri": "https://example.com/resource/tos",
+                "tls_client_certificate_bound_access_tokens": True,
+                # authorization_details_types_supported is a complext type
```
pcarleton:
```suggestion
                # authorization_details_types_supported is a complex type
```

tests/shared/test_auth.py
```diff
@@ -37,3 +38,84 @@ def test_oidc(self):
                 "userinfo_endpoint": "https://example.com/oauth2/userInfo",
             }
         )
+
+
+class TestProtectedResourceMetadataInvalid:
```
Kludex: We need a PR to drop those classes. They are following a very old Python testing layout.

---

You can ignore this comment.

Kludex (approved):

yannj-fr:
@Kludex tests updated

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -216,6 +217,25 @@ def auth_app(mock_oauth_provider):
     return app
 
 
+@pytest.fixture
+def protected_resource_app(auth_app):
+    """Fixture to create protected resource routes for testing."""
+
+    print(auth_app.router.routes)
```
Kludex:
```suggestion
```

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -216,6 +217,25 @@ def auth_app(mock_oauth_provider):
     return app
 
 
+@pytest.fixture
+def protected_resource_app(auth_app):
```
Kludex:
```suggestion
def protected_resource_app(auth_app: Starlette):
```

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1201,3 +1221,26 @@ async def test_authorize_invalid_scope(self, test_client: httpx.AsyncClient, reg
...
+        """Test the OAuth 2.0 Protected Resource metadata endpoint."""
+        print("Sending request to protected resource metadata endpoint")
+        response = await test_client.get("/.well-known/oauth-protected-resource")
+        print(f"Got response: {response.status_code}")
+        if response.status_code != 200:
+            print(f"Response content: {response.content}")
+        assert response.status_code == 200
+
+        metadata = response.json()
+        print(f"Protected Resource Metadata: {metadata}")
```
Kludex: Remove the prints.

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1201,3 +1221,26 @@ async def test_authorize_invalid_scope(self, test_client: httpx.AsyncClient, reg
...
+    """Test the Protected Resource Metadata model."""
+
+    @pytest.mark.anyio
+    async def test_metadata_endpoint(self, protected_resource_app: Starlette, test_client: httpx.AsyncClient):
+        """Test the OAuth 2.0 Protected Resource metadata endpoint."""
+        print("Sending request to protected resource metadata endpoint")
+        response = await test_client.get("/.well-known/oauth-protected-resource")
+        print(f"Got response: {response.status_code}")
+        if response.status_code != 200:
+            print(f"Response content: {response.content}")
```
Kludex: Why would this happen in this test?

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1201,3 +1221,26 @@ async def test_authorize_invalid_scope(self, test_client: httpx.AsyncClient, reg
...
+        assert response.status_code == 200
+
+        metadata = response.json()
+        print(f"Protected Resource Metadata: {metadata}")
+        assert metadata["resource"] == "https://example.com/resource"
+        assert metadata["authorization_servers"] == ["https://auth.example.com/authorization"]
+        assert metadata["scopes_supported"] == ["read", "write"]
+        assert metadata["resource_name"] == "Example Resource"
+        assert metadata["resource_documentation"] == "https://docs.example.com/resource"
+        assert metadata["bearer_methods_supported"] == ["header"]
```
Kludex: Let's use inline-snapshot on `metadata`.

```py
assert metadata == snapshot()
```

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1201,3 +1221,26 @@ async def test_authorize_invalid_scope(self, test_client: httpx.AsyncClient, reg
         # State should be preserved
         assert "state" in query_params
         assert query_params["state"][0] == "test_state"
+
+
+class TestProtectedResourceMetadata:
+    """Test the Protected Resource Metadata model."""
+
+    @pytest.mark.anyio
+    async def test_metadata_endpoint(self, protected_resource_app: Starlette, test_client: httpx.AsyncClient):
```
Kludex: This is not the best way to organize the tests. I think the `protected_resource_app` should be its own app, instead of extending the `auth_app` on the fixture above, and we need a `test_client` for this.

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -335,11 +363,8 @@ class TestAuthEndpoints:
     @pytest.mark.anyio
     async def test_metadata_endpoint(self, test_client: httpx.AsyncClient):
         """Test the OAuth 2.0 metadata endpoint."""
-        print("Sending request to metadata endpoint")
+
         response = await test_client.get("/.well-known/oauth-authorization-server")
-        print(f"Got response: {response.status_code}")
-        if response.status_code != 200:
-            print(f"Response content: {response.content}")
```
Kludex, line 349: Ah, I see why you added it... 😅

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1201,3 +1224,25 @@ async def test_authorize_invalid_scope(self, test_client: httpx.AsyncClient, reg
...
+    """Test the Protected Resource Metadata model."""
+
+    @pytest.mark.anyio
+    async def test_metadata_endpoint(self, protected_resource_test_client: httpx.AsyncClient):
+        """Test the OAuth 2.0 Protected Resource metadata endpoint."""
+
+        response = await protected_resource_test_client.get("/.well-known/oauth-protected-resource")
+        assert response.status_code == 200
+        metadata = response.json()
+        assert metadata == snapshot(
```
Kludex:
```suggestion
        assert response.json() == snapshot(
```

yannj-fr:
@Kludex done with the requested changes

tests/shared/test_auth.py
Kludex, line 1: Are the tests you added in here necessary?

Kludex (approved):

Kludex:
Maybe we could benefit more if we move the test you added to a new file, so we can actually don't mix bad practices with good ones.

tests/shared/test_auth.py
yannj-fr: To me it prevents someone to introduce regression in the model description by checking against a realistic payload

yannj-fr:
I will create a separated file, actually could be directly in the test/server/auth/test_proctected_resource.py

tests/server/auth/test_protected_resource.py
```diff
@@ -0,0 +1,58 @@
...
+        assert metadata == snapshot(
+            {
+                "resource": "https://example.com/resource",
+                "authorization_servers": ["https://auth.example.com/authorization"],
+                "scopes_supported": ["read", "write"],
+                "resource_name": "Example Resource",
+                "resource_documentation": "https://docs.example.com/resource",
+                "bearer_methods_supported": ["header"],
+            }
+        )
```
Kludex:
```suggestion
@pytest.mark.anyio
async def test_metadata_endpoint(self, test_client: httpx.AsyncClient):
    """Test the OAuth 2.0 Protected Resource metadata endpoint."""

    response = await protected_resource_test_client.get("/.well-known/oauth-protected-resource")
    metadata = response.json()
    assert metadata == snapshot(
        {
            "resource": "https://example.com/resource",
            "authorization_servers": ["https://auth.example.com/authorization"],
            "scopes_supported": ["read", "write"],
            "resource_name": "Example Resource",
            "resource_documentation": "https://docs.example.com/resource",
            "bearer_methods_supported": ["header"],
        }
    )
```

tests/server/auth/test_protected_resource.py
```diff
@@ -0,0 +1,58 @@
...
+        resource_name="Example Resource",
+        resource_documentation=AnyHttpUrl("https://docs.example.com/resource"),
+    )
+
+    app = Starlette(routes=protected_resource_routes)
+    return app
+
+
+@pytest.fixture
+async def protected_resource_test_client(protected_resource_app: Starlette):
```
Kludex:
```suggestion
async def test_client(protected_resource_app: Starlette):
```

tests/shared/test_auth.py
Kludex: I don't think it will have that effect... People that modify the model class will just update the test... The assertion is just making sure the validation passes, there's no assertion that makes use of any further logic.

yannj-fr:
As per your advice, I removed the test that is just on the model, and fixed naming.
Just one question, it is not good practice to have test case in a class?

Kludex:
> Just one question, it is not good practice to have test case in a class?

I don't think you get anything from it, it's more often seen in old test suites and when using `unittest` (https://docs.python.org/3/library/unittest.html#basic-example).

I prefer to stick with what modern packages do: not have wrapper classes.

---

This is my opinion tho.

felixweinberger:
Looks like we also need approval from an auth CODEOWNER to land this, @pcarleton mind taking another look?

PR #1244, opened 2025-08-06
Update FastMCP constructor to respect Pydantic settings configuration

saqadri:
Fix environment variable loading in FastMCP by preserving Pydantic Settings behavior

## Motivation and Context
PR #1198 introduced a regression where environment variables (e.g., `FASTMCP_HOST=0.0.0.0`) were no longer being loaded by FastMCP. The issue occurred because the refactored code explicitly passed all settings parameters to the Pydantic Settings constructor, which overrides environment variable values.

This fix ensures that only explicitly provided parameters are passed to the Settings constructor, allowing Pydantic's intended priority order to work correctly.

## How Has This Been Tested?
Tested the following scenarios:
- Environment variables are correctly loaded when no explicit parameters are provided
- Explicit parameters override environment variables when provided
- `.env` file values are loaded with correct priority
- Default values from the Settings class are used as fallback

Example test cases:
```python
server = FastMCP()  # host="127.0.0.1" from Settings default

# With FASTMCP_HOST=0.0.0.0 in environment
server = FastMCP()  # host="0.0.0.0" from env var

server = FastMCP(host="192.168.1.1")  # host="192.168.1.1" from explicit param
```

## Breaking Changes
No breaking changes. This restores the original behavior where environment variables are properly loaded.

## Types of changes
- [ ] Documentation update

## Checklist
- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)

### Settings Priority Order
This fix preserves Pydantic Settings' intended priority order (highest to lowest):
1. **Explicit parameters** passed to `FastMCP()` constructor
2. **Environment variables** (e.g., `FASTMCP_HOST`)
3. **`.env` file** values
4. **Default values** defined in the Settings class

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -10,7 +10,7 @@ from typing import Any, Generic, Literal
 
 import anyio
 import pydantic_core
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from pydantic.networks import AnyUrl
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from starlette.applications import Starlette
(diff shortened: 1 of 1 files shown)


Kludex:
I would prefer to revert that PR as to have this one merged.

It's very annoying that this was build the way it was. We need to deprecate some of those arguments and move to `run()`.

saqadri:
> I would prefer to revert that PR as to have this one merged.
> 
> It's very annoying that this was build the way it was. We need to deprecate some of those arguments and move to `run()`.

@Kludex yeah neither approach is ideal. If you prefer I can just update this PR with the revert of that specific file. #1198 had a handful of other small changes too

saqadri:
@Kludex please let me know what you'd prefer for this change

felixweinberger:
Hi @saqadri thanks for this contribution, will bring up in the maintainer meeting to figure out a path forward.

Kludex:
We've completely refactored a lot of stuff in that code path. I'm really sorry about the delay here.

Since we are about to release V2, I don't think we should revert that PR now.

PR #1260, opened 2025-08-10
Add ValidationOptions for lenient output schema validation

lorenss-m:
## Motivation and Context

Currently, MCP clients raise `RuntimeError` when tools don't return structured content matching their `outputSchema`. This breaks integrations with servers that have schema/implementation mismatches.

This PR adds optional lenient validation that converts these errors to warnings, allowing clients to continue operating with imperfect servers.

## How Has This Been Tested?

- Added test suite in `tests/client/test_validation_options.py`
- All existing tests pass
- Tested with real MCP servers exhibiting validation issues

## Breaking Changes

None. Default behavior unchanged. Lenient validation is opt-in:

```python
from mcp.client.session import ValidationOptions

# Opt-in to lenient validation
options = ValidationOptions(strict_output_validation=False)
session = ClientSession(read, write, validation_options=options)
```

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

- Added `ValidationOptions` class following existing patterns (e.g., `TransportSecuritySettings`)
- Modified `ClientSession.call_tool()` to check options before raising validation errors
- Fixed workspace config issue with `structured_output_lowlevel.py` by creating proper package structure

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -18,6 +18,7 @@ DEFAULT_CLIENT_INFO = types.Implementation(name="mcp", version="0.1.0")
 logger = logging.getLogger("client")
 
 
+
 class SamplingFnT(Protocol):
     async def __call__(
         self,
@@ -118,6 +119,7 @@ class ClientSession(
         logging_callback: LoggingFnT | None = None,
         message_handler: MessageHandlerFnT | None = None,
         client_info: types.Implementation | None = None,
+        validate_structured_outputs: bool = True,
     ) -> None:
         super().__init__(
             read_stream,
@@ -133,6 +135,7 @@ class ClientSession(
         self._logging_callback = logging_callback or _default_logging_callback
         self._message_handler = message_handler or _default_message_handler
         self._tool_output_schemas: dict[str, dict[str, Any] | None] = {}
+        self._validate_structured_outputs = validate_structured_outputs
 
     async def initialize(self) -> types.InitializeResult:
         sampling = types.SamplingCapability() if self._sampling_callback is not _default_sampling_callback else None
@@ -324,13 +327,27 @@ class ClientSession(
 
         if output_schema is not None:
             if result.structuredContent is None:
-                raise RuntimeError(f"Tool {name} has an output schema but did not return structured content")
-            try:
-                validate(result.structuredContent, output_schema)
-            except ValidationError as e:
-                raise RuntimeError(f"Invalid structured content returned by tool {name}: {e}")
-            except SchemaError as e:
-                raise RuntimeError(f"Invalid schema for tool {name}: {e}")
+                if self._validate_structured_outputs:
+                    raise RuntimeError(f"Tool {name} has an output schema but did not return structured content")
+                else:
+                    logger.warning(
+                        f"Tool {name} has an output schema but did not return structured content. "
+                        f"Continuing without structured content validation."
+                    )
+            else:
+                try:
+                    validate(result.structuredContent, output_schema)
+                except ValidationError as e:
+                    if self._validate_structured_outputs:
+                        raise RuntimeError(f"Invalid structured content returned by tool {name}: {e}") from e
+                    else:
+                        logger.warning(
+                            f"Invalid structured content returned by tool {name}: {e}. "
+                            f"Continuing without validation."
+                        )
+                except SchemaError as e:
+                    # Schema errors are always raised - they indicate a problem with the schema itself
+                    raise RuntimeError(f"Invalid schema for tool {name}: {e}") from e
 
     async def list_prompts(self, cursor: str | None = None) -> types.ListPromptsResult:
         """Send a prompts/list request."""
--- a/src/mcp/shared/memory.py
+++ b/src/mcp/shared/memory.py
@@ -61,6 +61,7 @@ async def create_connected_server_and_client_session(
     client_info: types.Implementation | None = None,
     raise_exceptions: bool = False,
     elicitation_callback: ElicitationFnT | None = None,
+    validate_structured_outputs: bool = True,
 ) -> AsyncGenerator[ClientSession, None]:
     """Creates a ClientSession that is connected to a running MCP server."""
     async with create_client_server_memory_streams() as (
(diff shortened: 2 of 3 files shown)


src/mcp/client/session.py
```diff
@@ -324,13 +337,27 @@ async def _validate_tool_result(self, name: str, result: types.CallToolResult) -
...
+                if self._validation_options.strict_output_validation:
+                    raise RuntimeError(f"Tool {name} has an output schema but did not return structured content")
+                else:
+                    logger.warning(
+                        f"Tool {name} has an output schema but did not return structured content. "
+                        f"Continuing without structured content validation due to lenient validation mode."
+                    )
+            else:
+                try:
+                    validate(result.structuredContent, output_schema)
```
ochafik, line 338: What error would None produce here, do we need the two branches?

src/mcp/client/session.py
```diff
@@ -18,6 +18,17 @@
 logger = logging.getLogger("client")
 
 
+class ValidationOptions(BaseModel):
+    """Options for controlling validation behavior in MCP client sessions."""
+
+    strict_output_validation: bool = Field(
```
ochafik: Wondering if we wanna make this sound more explicitly about tool call output, as there's also validation happening around elicitation (https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation#security-considerations)

Maybe something like `validate_structured_outputs` (no sure we need strict in the name, if we don't raise an exception the output isn't really validated, warnings being often ignored)

src/mcp/client/session.py
```diff
@@ -324,13 +337,27 @@ async def _validate_tool_result(self, name: str, result: types.CallToolResult) -
...
+                    logger.warning(
+                        f"Tool {name} has an output schema but did not return structured content. "
+                        f"Continuing without structured content validation due to lenient validation mode."
+                    )
+            else:
+                try:
+                    validate(result.structuredContent, output_schema)
+                except ValidationError as e:
+                    if self._validation_options.strict_output_validation:
+                        raise RuntimeError(f"Invalid structured content returned by tool {name}: {e}")
```
ochafik: nit: might wanna raise ... from e

ochafik (commented):
Hi @lorenss-m, thanks for sending this!

Seems reasonable to add this option given the spec only says [clients "should" validate structured results against the output schema](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema).

I'm curious tho, which examples of non-conforming outputs did you find / how bad were they?

cc/ @bhosmer-ant since authored initial support in #993

src/mcp/client/session.py
```diff
@@ -18,6 +18,17 @@
 logger = logging.getLogger("client")
 
 
+class ValidationOptions(BaseModel):
```
Kludex: Let's not introduce parameters that are BaseModel's please. The best experience is usually to have flat parameters.

lorenss-m:
Added the validate_structured_outputs suggestion!

src/mcp/client/session.py
lorenss-m: Added the validate_structured_outputs suggestion!

dsp-ant:
We should make sure that this matches how other SDKs behave before we merge this. @modelcontextprotocol/sdk-maintainers .

jba:
In the Go SDK, currently all schema validation is done on the server side.

mikekistler:
The C# SDK delegates the validation of the structured content to the MCP Host. This is mainly because there is no standard, built-in JSON Schema validator in .NET, so doing this in the client would require pulling in an external package, one of potentially several, which adds bulk to the SDK and usurps the MCP Host's choce of validation package.

lorenss-m:
Anything else we should change here? Would love to get this rolled out, our library is forced to depend on our fork otherwise!

src/mcp/server/auth/provider.py
```diff
@@ -281,7 +281,7 @@ async def revoke_token(
 
 def construct_redirect_uri(redirect_uri_base: str, **params: str | None) -> str:
     parsed_uri = urlparse(redirect_uri_base)
-    query_params = [(k, v) for k, vs in parse_qs(parsed_uri.query) for v in vs]
+    query_params = [(k, v) for k, vs in parse_qs(parsed_uri.query).items() for v in vs]
```
felixweinberger, line 284: intentional? seems unrelated

src/mcp/client/session.py
```diff
@@ -324,13 +326,26 @@ async def _validate_tool_result(self, name: str, result: types.CallToolResult) -
...
+                    validate(result.structuredContent, output_schema)
+                except ValidationError as e:
+                    if self._validate_structured_outputs:
+                        raise RuntimeError(f"Invalid structured content returned by tool {name}: {e}") from e
+                    else:
+                        logger.warning(
+                            f"Invalid structured content returned by tool {name}: {e}. Continuing without validation."
+                        )
+                except SchemaError as e:
+                    # Schema errors are always raised - they indicate a problem with the schema itself
```
felixweinberger, line 347: nit: remove redundant comment

src/mcp/client/session.py
```diff
@@ -324,13 +326,26 @@ async def _validate_tool_result(self, name: str, result: types.CallToolResult) -
...
         if output_schema is not None:
             if result.structuredContent is None:
-                raise RuntimeError(f"Tool {name} has an output schema but did not return structured content")
-            try:
-                validate(result.structuredContent, output_schema)
-            except ValidationError as e:
-                raise RuntimeError(f"Invalid structured content returned by tool {name}: {e}")
-            except SchemaError as e:
-                raise RuntimeError(f"Invalid schema for tool {name}: {e}")
+                if self._validate_structured_outputs:
```
felixweinberger, line 329: nit: getting quite heavily nested here, could at least partially simplify with

```
if output_schema is None:
    return
    
if result.structuredContent is None:
    # raise / log depending
    
# handle the base case of try / except with raise / log depending
```

felixweinberger:
Hi @lorenss-m are you planning to come back to this one?

felixweinberger:
Closing this for now due to inactivity.

PR #1268, opened 2025-08-13
Update dependencies and fix type issues

dsp-ant:
## Summary
- Updated Click dependency to 8.2.0 across all examples and projects
- Updated httpx to 0.27.1 and uvicorn to 0.31.1 in main pyproject.toml
- Fixed type issues in tests by adding proper variable declarations and type annotations
- Improved CI matrix to test both lowest-direct and highest dependency resolution strategies

## Test plan
- [x] All tests pass with updated dependencies
- [x] Type checking passes with pyright
- [x] CI matrix tests both dependency resolution modes
- [x] Examples still work with updated Click version


ihrpr:
CI is unhappy

.github/workflows/shared.yml
```diff
@@ -48,18 +49,11 @@ jobs:
           version: 0.7.2
 
       - name: Install the project
-        run: uv sync --frozen --all-extras --python ${{ matrix.python-version }}
+        run: uv sync --frozen --all-extras --python ${{ matrix.python-version }} --resolution $${{ matrix.dep-resolution }}
```
Kludex:
```suggestion
        run: uv sync --frozen --all-extras --python ${{ matrix.python-version }} --resolution ${{ matrix.dep-resolution }}
```

tests/client/test_list_roots_callback.py
```diff
@@ -35,7 +38,7 @@ async def list_roots_callback(
         return callback_return
 
     @server.tool("test_list_roots")
-    async def test_list_roots(context: Context, message: str):  # type: ignore[reportUnknownMemberType]
+    async def test_list_roots(context: Context[ServerSession, Any, Any], message: str):
```
Kludex: Can we avoid adding `Any`?

Is the below possible?
```suggestion
    async def test_list_roots(context: Context[ServerSession, None], message: str):
```

tests/shared/test_progress_notifications.py
```diff
@@ -217,6 +217,8 @@ async def test_progress_context_manager():
 
     server = Server(name="ProgressContextTestServer")
 
+    progress_token = None
```
Kludex, line 220: What's the type error you were having on those that you added `= None`?

dsp-ant: That they are potentially unused

dsp-ant:
CI happy

tests/shared/test_progress_notifications.py
Kludex: Why it was not failing before?

Kludex: Ah, it seems you are running a new version of pyright locally, and that's why you needed this.

Can you bump the uv.lock to have the new pyright version?

ihrpr (approved):
LGTM

ihrpr:
"Require status checks to pass" needs to be updated in rulesets

.github/workflows/shared.yml
```diff
@@ -48,18 +49,11 @@ jobs:
           version: 0.7.2
 
       - name: Install the project
-        run: uv sync --frozen --all-extras --python ${{ matrix.python-version }}
+        run: uv sync --frozen --all-extras --python ${{ matrix.python-version }} --resolution ${{ matrix.dep-resolution }}
```
Kludex, line 52: You can't use `--frozen` with `--resolution`.

Kludex: If you remove the `--frozen`, the lockfile is useless.

Kludex: I created https://github.com/modelcontextprotocol/python-sdk/pull/1325 to use the matrix, but it's pretty much the same as it was before.

PR #1280, opened 2025-08-18
Replace httpx_client_factory with direct httpx_client parameter for better flexibility

### LINKED ISSUE #870: Add SSL Certificate Verification Options to FastMCP Client

When connecting to an MCP server over HTTPS with self-signed or internal certificates, the FastMCP Client fails with SSL certificate verification errors. Currently, there's no straightforward way to disable SSL verification or provide custom certificates when using the Client class.

```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)
```

Please add one or more of the following options to the FastMCP Client:

A verify parameter to disable SSL verification or to provide a custom certificate:
```
client = Client(MCP_URL, verify=False)
```

```
client = Client(MCP_URL, verify="/path/to/certificate.pem")
```

This is needed when working with internal or development environments that use self-signed certificates or private certificate authorities. 

Users need to resort to insecure workarounds like patching Python's SSL module.

**Additional context**

The issue appears to be in how the transport is inferred and created from the URL string. When a URL is provided, the infer_transport function creates a StreamableHttpTransport for HTTPS URLs, but there's no way to configure the SSL verification options for this transport.

According to FastMCP requirements, streamable-http transport is required for web deployments, and the client needs to connect to the /mcp path. The current implementation doesn't provide a way to handle SSL verification for these connections.

This would align with how other HTTP client libraries (like requests) handle SSL verification, making the API more intuitive and flexible for enterprise environments.

ruhz3:
Added support for the `verify` option to the `create_mcp_http_client` function and related client factories, allowing users to control SSL certificate verification when creating an httpx AsyncClient.

## Motivation and Context

Some users need to disable SSL verification (e.g., for testing or internal endpoints), or use a custom CA bundle. This change makes it possible to pass `verify=True`, `verify=False`, or a custom `ssl.SSLContext` when creating MCP HTTP clients.

## How Has This Been Tested?

Tested by creating AsyncClient instances with different `verify` values (`True`, `False`, and custom context) and confirming correct behavior for secure and insecure endpoints. Existing unit tests for HTTP client creation were also run.

## Breaking Changes

No breaking changes. The new `verify` parameter is optional and defaults to the previous behavior.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This change improves flexibility for users working in environments with custom or self-signed certificates, or those who need to disable SSL verification for

resolves #870 
This PR does not fully resolve the linked issue, but it applies the necessary modifications to the current repository. Further functional modifications following this PR are needed in the fastmcp repository.

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -7,13 +7,16 @@ import anyio
 import httpx
 from anyio.abc import TaskStatus
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+from httpx._config import DEFAULT_TIMEOUT_CONFIG
 from httpx_sse import aconnect_sse
 
 import mcp.types as types
-from mcp.shared._httpx_utils import McpHttpClientFactory, create_mcp_http_client
+from mcp.shared._httpx_utils import create_mcp_http_client
 from mcp.shared.message import SessionMessage
 
 logger = logging.getLogger(__name__)
+
+HTTPX_DEFAULT_TIMEOUT = DEFAULT_TIMEOUT_CONFIG
 
 
 def remove_request_params(url: str) -> str:
@@ -26,8 +29,8 @@ async def sse_client(
     headers: dict[str, Any] | None = None,
     timeout: float = 5,
     sse_read_timeout: float = 60 * 5,
-    httpx_client_factory: McpHttpClientFactory = create_mcp_http_client,
     auth: httpx.Auth | None = None,
+    httpx_client: httpx.AsyncClient | None = None,
 ):
     """
     Client transport for SSE.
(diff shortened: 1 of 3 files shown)


Kludex:
We should be passing the client, instead of allowing more and more options.

ruhz3:
Respecting the original author's intention to avoid exposing AsyncClient directly to users, the FactoryClient was modified to accept kwargs.

ruhz3:
I’d greatly appreciate any active and constructive reviews.

felixweinberger:
Hi @ruhz3 thank you for updating this PR - I don't think we'll go with this approach as we're actually intending to remove all the arguments other than the `httpx.AsyncClient`

We're currently working on this here: https://github.com/modelcontextprotocol/python-sdk/pull/1177

I'm closing this for now, feel free to chime in on the linked PR though if it's missing something you need!

Issue #1284, opened 2025-08-20
We need basic "cancel / reject / accept" elicitations requiring only confirmation / reject with no data


owengo:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

It would be nice to create very simple cancel / reject / accept elicitations like this:

# Elicitation schemas for user confirmation
class CreateDocumentConfirmation(BaseModel):
    """Schema for create document confirmation."""
    #confirm: bool   <== We don't need this, it creates  a useless box to  the click for the client

An easy fix in elicitation.py would be:

```
    if result.action == "accept" and result.content:
        # Validate and parse the content using the schema
        validated_data = schema.model_validate(result.content)
        return AcceptedElicitation(data=validated_data)
    elif result.action == "accept":
        validated_data = schema.model_validate(dict())
        return AcceptedElicitation(data=validated_data)
    elif result.action == "decline":
        return DeclinedElicitation()
    elif result.action == "cancel":
        return CancelledElicitation()
    else:
        # This should never happen, but handle it just in case
        raise ValueError(f"Unexpected elicitation action: {result.action}")
```

### Example Code

```Python
from mcp.server.fastmcp import Context, FastMCP

# Elicitation schemas for user confirmation
class CreateDocumentConfirmation(BaseModel):
    """Schema for create document confirmation."""

mcp = FastMCP("test-elicit")

@mcp.tool()
async def create_document(
    ctx: Context[ServerSession, None],
    name: str,
    html_content: str
) -> Dict[str, Any]:
    """Create a new  document"""
    user_email = await check_user_auth(ctx)
    start_time = time.time()
    
    try:
        # Request user confirmation via elicitation
        try:
            await ctx.info(f"Requesting confirmation for document creation: {name}")
            confirmation_result = await ctx.elicit(
                f"Are you sure you want to create document '{name}'?",
                CreateDocumentConfirmation
            )
            
            await ctx.info(f"Confirmation result: {confirmation_result.action}")
            if confirmation_result.action != "accept":
                return {"result": f"Document creation cancelled by user"}
                
        except Exception as elicit_error:
            await ctx.error(f"Elicitation failed: {elicit_error}")
            return {"result": f"Document creation cancelled due to confirmation error"}
 ...


[...]

PR #1285, opened 2025-08-20
fix: allow elicitations accepted without content

owengo:
This change allows to handle elicitations without any data, just the user action.

## Motivation and Context

The change allows to create very simple "Accept / Reject / Cancel" elicitations without check boxes or input data for the user.

cf: https://github.com/modelcontextprotocol/python-sdk/issues/1284

## How Has This Been Tested?

Yes we have an MCP server which allows create and update documents in a database, and we have an elicitation requiring the user to accept the action

## Breaking Changes

No, the behavior is exactly the same for all the elicitations which actually require data.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The behavior is possible with the typescript sdk, but not the python sdk:

```
   const result = await server.server.elicitInput({
       message: `${confirmationMessage}`,
        requestedSchema: {
         type: "object",
         properties: {
         },
         required: []
       }
     });  
     if (result.action === "accept") { 
       log.warn(`User accepted usage of ${toolName}`, {
         user: userEmail,
       });
       return true;
     } else {
...

```

### DIFF
--- a/src/mcp/server/elicitation.py
+++ b/src/mcp/server/elicitation.py
@@ -98,7 +98,7 @@ async def elicit_with_validation(
         related_request_id=related_request_id,
     )
 
-    if result.action == "accept" and result.content:
+    if result.action == "accept" and result.content is not None:
         # Validate and parse the content using the schema
         validated_data = schema.model_validate(result.content)
         return AcceptedElicitation(data=validated_data)


src/mcp/server/elicitation.py
```diff
@@ -102,6 +102,10 @@ async def elicit_with_validation(
         # Validate and parse the content using the schema
         validated_data = schema.model_validate(result.content)
         return AcceptedElicitation(data=validated_data)
+    elif result.action == "accept":
```
ochafik: I think the fix should be above instead: `if result.action == "accept" and result.content is not None:`

(an empty content `{}` is considered falsy in Python)

This otherwise allows for content to be omitted (`{"action": "accept"}`), which isn't consistent with [the spec](https://modelcontextprotocol.io/specification/2025-06-18/schema#elicitresult). We should just cater to the valid `{"action": "accept", "content": {}}` (use case from https://github.com/modelcontextprotocol/python-sdk/issues/1284) unless the spec is updated :-)

owengo: You are right.

```
   if result.action == "accept" and result.content:
        # Validate and parse the content using the schema
        validated_data = schema.model_validate(result.content)
        return AcceptedElicitation(data=validated_data)
    elif result.action == "accept" and result.content is not None:
        validated_data = schema.model_validate(result.content)
        return AcceptedElicitation(data=validated_data)
[...]       
```
is it ok?

owengo: You are right, I have pushed your fix.

owengo:
@ochafik I understand if you don't want to merge such a small fix but maybe you could close it and instead add the fix with another PR fixing other issues?

ochafik (approved):
Thanks @owengo !

PR #1286, opened 2025-08-20
feat: add paginated list decorators for prompts, resources, and tools

maxisbey:
## Summary
This PR adds pagination support for listing prompts, resources, and tools using cursor-based pagination as defined in the MCP spec.

The lowlevel Server decorators `list_resources`, `list_prompts`, and `list_tools` now handle two types of handlers:
- No pagination (original functionality)
    - A handler with the signature `Callable[[], Awaitable[list[<resource, prompt, or tool>]]]`
    - This matches the original way these decorators work and means the API is still backwards compatible.
- Pagination (new functionality)
    - A handler with the signature `Callable[[List<___>Request], Awaitable[List<___>Result]] `
    - The cursor for a paginated request will be in the corresponding `List<___>Request` object depending on the decorator which the handle can now optionally handle.
    - The response object is a `List<___>Response` which the handler can specify the cursor for the next page.
    - This allows future extensions to the Request or Response objects to still be supported.

### How It Works
The decorators now automatically detect whether a callback accepts a cursor parameter:
```python
# Non-paginated (existing code continues to work)
@server.list_resources()
async def list_resources() -> list[Resource]:
    return [...]

# Paginated (new capability)
@server.list_resources()
async def list_resources(cursor: Cursor | None) -> ListResourcesResult:
    return ListResourcesResult(resources=[...], nextCursor="...")
```

## Changes
- Modify `list_prompts()`, `list_resources()`, and `list_tools()` decorators to support both signatures
- Add comprehensive unit tests for pagination and signature detection
- Add examples demonstrating pagination usage
- Tool cache properly maintained for both paginated and non-paginated modes

## Test plan
- [x] Unit tests pass for all three decorators with both signatures
- [x] Tests verify cursor passthrough for both None and string values
- [x] Function inspection tests cover various callback signatures
- [x] Backward compatibility verified with existing non-paginated handlers
- [x] Manually tested via the inspector and with example client code

👨 Below generated by the human, [Max Isbey](https://github.com/maxisbey)

## Manual testing

I spun up the MCP inspector and ran the example pagination server in both stdio mode and SSE mode. Below is some testing evidence using the SSE mode:

### Clicking "List More Resources" correctly lists all 30 resources:

Note: this works for prompts and tools as well


jerome3o-anthropic (changes requested):
Can we have a some docs with examples? not sure if we have good docs for the lowlevel api but some usage examples would be great (if not just to help with reviewing).

Otherwise this looks good to me. happy with the separate functions to avoid breaking changes. keen to know how you think we should approach this in the high level interface.

maxisbey:
> Can we have a some docs with examples? not sure if we have good docs for the lowlevel api but some usage examples would be great (if not just to help with reviewing).
> 
> Otherwise this looks good to me. happy with the separate functions to avoid breaking changes. keen to know how you think we should approach this in the high level interface.

Updated the docs, added example code, and included testing in the PR desription

jerome3o-anthropic (dismissed):

.pre-commit-config.yaml
```diff
@@ -25,22 +25,22 @@ repos:
     hooks:
       - id: ruff-format
         name: Ruff Format
-        entry: uv run ruff
+        entry: uv run --frozen ruff
```
Kludex: Same for below
```suggestion
        entry: ruff
```

src/mcp/server/lowlevel/server.py
```diff
@@ -242,6 +242,19 @@ async def handler(_: Any):
...
+
+            async def handler(req: types.ListPromptsRequest):
+                result = await func(req.params.cursor if req.params else None)
+                return types.ServerResult(result)
+
+            self.request_handlers[types.ListPromptsRequest] = handler
+            return func
+
+        return decorator
+
```
Kludex: Why can't the `list_prompts` handle pagination?

maxisbey: Well there were a few paths and I wanted to just put something down to start the discussion.

There were three options I was thinking:

1. Separate decorators (current state of this PR):
    - Backwards compatible, it won't break existing uses of list_prompts where it's used on a function that takes no arguments
    - Makes implementation of this easier, rather than having to do inspection on the `func` being passed in to see whether it takes an argument or not, it's just separated into two methods
2. Single decorator but break backwards compatibility:
    - This would just be changing list_prompts to look exactly like list_prompts_paginated, all `func`s passed in need to take in an argument of `types.Cursor | None`
    - Breaks potentially a lot of existing uses of the lowlevel code
3. Single decorator, union type:
    - This would be having a single `list_prompts` that takes in a `func` that looks like `Callable[[types.Cursor | None], Awaitable[types.ListPromptsResult]] | Callable[[], Awaitable[list[types.Prompt]]]`
    - Makes it backwards compatible while allowing for an optional cursor
    - Makes implementation a bit trickier by having to do inspection on the function to make sure it fits the shapes expected.
 
Were you preferring option 2 or 3?

.pre-commit-config.yaml
maxisbey: good idea, updated

src/mcp/server/lowlevel/server.py
maxisbey: I'm probably leaning towards option 3 now though :)

.pre-commit-config.yaml
maxisbey: Doesn't play nice with the CI, gonna revert to using `uv run --frozen` for now

src/mcp/server/lowlevel/server.py
maxisbey: This has been changed now to having a single decorator (option 3)

examples/servers/simple-pagination/mcp_simple_pagination/server.py
```diff
@@ -0,0 +1,225 @@
...
+            description=prompt.description,
+            messages=[
+                types.PromptMessage(
+                    role="user",
+                    content=types.TextContent(type="text", text=message_text),
+                )
+            ],
+        )
+
+    if transport == "sse":
```
Kludex, line 195: Why are we doing this every time?

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,54 @@
...
+    except (ValueError, TypeError):
+        return False
+
+    params = dict(sig.parameters.items())
+
+    method = inspect.ismethod(func)
+
+    if method:
+        params.pop("self", None)
+        params.pop("cls", None)
```
Kludex: We shouldn't rely on parameter names.

src/mcp/server/lowlevel/server.py
```diff
@@ -381,16 +416,36 @@ async def handler(req: types.UnsubscribeRequest):
...
-                for tool in tools:
-                    self._tool_cache[tool.name] = tool
-                return types.ServerResult(types.ListToolsResult(tools=tools))
+            pass_cursor = accepts_cursor(func)
+
+            if pass_cursor:
+                cursor_func = cast(Callable[[types.Cursor | None], Awaitable[types.ListToolsResult]], func)
+
+                async def cursor_handler(req: types.ListToolsRequest):
+                    result = await cursor_func(req.params.cursor if req.params is not None else None)
```
Kludex: Would it make sense to pass the ListToolRequest instead? What does that object contains?

I'm on my phone only this week.

Kludex: Because then you, as a user don't need to type it as Cursor | None.

examples/servers/simple-pagination/mcp_simple_pagination/server.py
maxisbey: Just following the pattern of all the other sample servers

src/mcp/server/lowlevel/server.py
maxisbey: Hmm well the only thing in the ListToolsRequest is the cursor. Unless you think in the future there's a chance more would be added to it?

src/mcp/server/lowlevel/func_inspection.py
maxisbey: Oh fair, turns out this isn't needed anyway as `inspect.signature.parameters` doesn't contain these by default. Updated the unit tests to reflect this.

examples/servers/simple-pagination/mcp_simple_pagination/server.py
maxisbey: I'd prefer to leave this as is, and maybe chnage in a different PR dedicated to simplifying the example servers or something if that's alright with you?

src/mcp/server/lowlevel/server.py
Kludex: My point is more that it makes more sense to receive a request as parameter.

But answering the question... Yes, if something is added in the future, it will be there.

maxisbey: Alrighty, changed from pasing cursors to passing request objects

README.md
```diff
@@ -1737,6 +1738,116 @@ Tools can return data in three ways:
...
+                    break
+
+            print(f"Total resources: {len(all_resources)}")
+
+
+if __name__ == "__main__":
+    asyncio.run(list_all_resources())
+```
+
+_Full example: [examples/snippets/clients/pagination_client.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/clients/pagination_client.py)_
```
ochafik, line 1839: nit: prefer relative links: `[examples/snippets/clients/pagination_client.py](./examples/snippets/clients/pagination_client.py)`

README.md
```diff
@@ -1737,6 +1738,116 @@ Tools can return data in three ways:
...
+        for item in ITEMS[start:end]
+    ]
+
+    # Determine next cursor
+    next_cursor = str(end) if end < len(ITEMS) else None
+
+    return types.ListResourcesResult(resources=page_items, nextCursor=next_cursor)
+```
+
+_Full example: [examples/snippets/servers/pagination_example.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/pagination_example.py)_
```
ochafik, line 1789: nit: prefer relative link
`[examples/snippets/servers/pagination_example.py](./examples/snippets/servers/pagination_example.py)`

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,48 @@
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+
+def accepts_request(func: Callable[..., Any]) -> bool:
+    """
+    True if the function accepts a cursor parameter call, otherwise false.
+
+    `accepts_cursor` does not validate that the function will work. For
```
ochafik: update name here

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,48 @@
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+
+def accepts_request(func: Callable[..., Any]) -> bool:
+    """
+    True if the function accepts a cursor parameter call, otherwise false.
```
ochafik: Please expand here on what this means. A literal read implies this accepts a `cursor` or (from the function name) a `request` parameter, which is confusing (I had to read through entire function to understand its semantics)

src/mcp/server/lowlevel/server.py
```diff
@@ -229,12 +230,29 @@ def request_context(
...
+            | Callable[[types.ListPromptsRequest], Awaitable[types.ListPromptsResult]],
+        ):
             logger.debug("Registering handler for PromptListRequest")
+            pass_request = accepts_request(func)
 
-            async def handler(_: Any):
-                prompts = await func()
-                return types.ServerResult(types.ListPromptsResult(prompts=prompts))
+            if pass_request:
+                request_func = cast(Callable[[types.ListPromptsRequest], Awaitable[types.ListPromptsResult]], func)
```
ochafik: This feels wrong: since we're introspecting on the function, we should check it does have this signature.

If we cast anything that passed the loosely-defined accepts_request check (e.g. `def foo(x: str)`), we allow cryptic mismatched type failures down the line

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,48 @@
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+
+def accepts_request(func: Callable[..., Any]) -> bool:
```
ochafik: I think maybe the function we want is `def accepts_single_positional_arg(func, arg_type)`? (or `can_be_called_with_single_positional` to allow defaults)

ochafik (changes requested):
Exciting changes! Just some concerns about the accepts_request logic

README.md
Kludex: this is automatically generated

README.md
maxisbey: as per @Kludex 's comment, this is generated by the `update_readme_snippets.py` file

src/mcp/server/lowlevel/func_inspection.py
maxisbey: updated

src/mcp/server/lowlevel/func_inspection.py
maxisbey: updated

src/mcp/server/lowlevel/func_inspection.py
maxisbey: I've renamed the function, but didn't include an `arg_type` because it's likely users will pass in an untyped handler which would then cause validation errors if we're checking for type hints.

src/mcp/server/lowlevel/server.py
maxisbey: Hmm well I don't want to be too strict on the introspection of the function since we won't know the exact signature.

For example the following functions will all have completely different signatures while still being fully accepted/working as a handler:

- Doesn't accept request:
  - `def foo():`
  - `def foo(*, other: int = 3):`
- Accepts request: 
  - `def foo(req):`
  - `def foo(req: types.ListPromptsRequest):`
  - `def foo(req, other: int = 3):`
  - `def foo(req, *, other: int = 3):`
  - `def foo(req=None):`

But I wasn't super sure if there is a better way to accept arbitrary functions signatures while still having strict typing with pyright. Do you have any suggestions for something else that'd work better?

src/mcp/server/lowlevel/func_inspection.py
ochafik: I think a handler w/ an untyped single pos param (or untyped `*args`) is perfectly fine, but if that single pos param does have a type it's a low hanging fruit to check it.

src/mcp/server/lowlevel/server.py
maxisbey: After thinking on it I think we should leave it as is. Being more strict in my opinion brings two negatives:

1. We have to make decisions around a lot of possible function signatures and whether or not they should be accepted.
    - For example is `foo(*, other: int = 3)` accepted, or do we reject it? Currently this would work perfectly fine as a handler, but would break with us being strict
    - see other possible shapes listed above
2. There are a _lot_ of possible ways users could type a function while still being valid. We then have to potentially have complex registration time type checking, whereas all we're doing now is "can we pass a request in or not"
    - For example `foo(req)`, `foo(req: ListPromptsRequest`, `foo(req: Any)`, `foo(*args)`, `foo(*args, **kwargs)`, `foo(req: Annotated[ListPromptsRequest, "thing"]`, `foo(req: Generic[T])`, `foo(req: Generic[T] = None)`, `foo(req: SomeCustomProtocol)`, `foo(req: ListPromptsRequest | ListToolsRequest | ListResourcesRequest)`, etc. are all valid handlers which would take a request object but require different runtime type checking while still passing static type checkers. If instead we were extremely strict and said your function must _either_ be `foo(req)` or `foo(req: ListPromptsRequest)`, then static type checking will pass for the other function signatures but would raise an exception at runtime.
    - I feel people doing unpredictable things in Python is common enough that us adding this extra layer of type checking and strictness here doesn't have a big enough benefit while having definite negatives. This isn't a library like Pydantic where we're explicitly doing type checking/validation for serialization/deserialization of untyped json objects or similar.

I'd be keen to leave it as for now unless it's a serious anti pattern.

I could be totally mistaken here and am happy to redo it with type checking if that's the consensus. I'd of course prefer whatever is most idiomatic/best practice.

maxisbey: There's three options I see we could take here:

1. Use the currently implemented "best effort" and least strict approach, just checking whether it's possible to pass anything positionally or not.
    - I prefer this one as it's simple to implement, type checkers will be happy with all the weird but accepted signatures, and the rest of the above reasons
2. Do type checking to ensure the `func` conforms to the exact shape expected. Meaning the signature **must** be one of `foo()`, `foo(req)`, `foo(req: Any)`, `foo(req: ListPromptsRequest)`, otherwise raise an error for things like `foo(req, extra=None)`, `foo(req: str)`, etc.
    - Or some variation of this, such as the rule "the first argument in your function must be of type `ListPromptsRequest`, otherwise we assume it doesn't take a request object"
    - Although we then need to explicitly decide whether we accept things like `*args` or Generics, or Protocols, as this may mean type checkers pass while our code will error at runtime.
3. Do something similar to the `Context` object in FastAPI, where the rule is "either the first positional argument must be of type ListPromptsRequest, or a keyword argument is of ListPromptsRequest, otherwise we will assume `func` takes no arguments"

Let me know if you have other ideas, or whether you're happy to go with any of the options. If there's a consensus for something other than option 1 then I'm happy to switch it out although I'd want it to be fully defined with clear rules of what we do and don't accept.

Kludex: I think the way this should work is the following:
1. Check the type of the parameter, and inject based on that.
2. Deprecate the use of function without the request parameter.
3. In some months (v2?), drop the support for function without the parameter.

That was what I had in mind when I proposed to use the same methods.

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,217 @@
...
+
+    The decision logic:
+    1. If the function has no parameters -> (False, True) - old style without params, deprecate
+    2. If the function has parameters but can't accept positional args -> (False, False)
+    3. If the first parameter type accepts the request type -> (True, False) - pass request, no deprecation
+    4. If the first parameter is typed as Any -> (True, True) - pass request but deprecate (effectively untyped)
+    5. If the first parameter is typed with something incompatible -> (False, True) - old style, deprecate
+    6. If the first parameter is untyped but accepts positional args -> (True, True) - pass request, deprecate
+    """
+    can_accept_arg = accepts_single_positional_arg(func)
```
ochafik: Can you inline & simplify this function's / file logic? It is now calling inspect.signature up to twice (from 3 places). Something like this maybe?

```py
def function_accepts_single_param(func: Callable[..., Any], param_type: type) -> bool:
    sig = inspect.signature(func)
    if not sig.parameters:
        # TODO: warn about old style without params
        return False
    elif len(sig.parameters) != 1:
        raise ValueError(f"Function must have exactly one parameter of type {type} or none (legacy).")

    param = next(iter(sig.parameters.values()))
    if param.kind not in (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD):
        # TODO: warn about exotic param type
        return False
    
    if param.annotation == inspect.Parameter.empty or param.annotation == Any:
        # TODO: warn about untyped, depending on decision
        return False # or True if allow untyped

    return param.annotation == param_type
```

(not sure about swallowing the signature exceptions, removed it / better to fail loudly unless you know of legit cases that would throw?)

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,217 @@
...
+    Returns False for None or incompatible types.
+    """
+    if param_type is None:
+        return False
+
+    # Check for Any type
+    if param_type is Any:
+        return True
+
+    # Exact match
```
ochafik: I think we really only need to handle this case / can be prescriptive re/ signature of functions we allow (we don't need to allow all of valid python).

tests/server/lowlevel/test_func_inspection.py
```diff
@@ -0,0 +1,176 @@
...
+    """Function with multiple keyword-only params all with defaults"""
+    pass
+
+
+async def mixed_positional_and_keyword(request: types.ListPromptsRequest, *, extra: str = "test"):
+    """Function with positional and keyword-only params"""
+    pass
+
+
+@pytest.mark.parametrize(
```
ochafik: can you also test `def bad_pos_arg_type(request: str)` and  `def good_untyped_pos_arg(request)` ?

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,217 @@
...
+    - Generic types (basic support)
+    - Any (always returns True)
+
+    Returns False for None or incompatible types.
+    """
+    if param_type is None:
+        return False
+
+    # Check for Any type
+    if param_type is Any:
```
ochafik: @Kludex what's your take on allowing untyped `def foo(request)`?

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,217 @@
...
+        return request_type == origin
+
+    return False
+
+
+def should_pass_request(func: Callable[..., Any], request_type: type) -> tuple[bool, bool]:
+    """
+    Determine if a request should be passed to the function based on parameter type inspection.
+
+    Returns a tuple of (should_pass_request, should_deprecate):
```
ochafik: nit: maybe best use a named tuple here?

ochafik (changes requested):
Thanks Max!

src/mcp/server/lowlevel/func_inspection.py
Kludex: https://github.com/modelcontextprotocol/python-sdk/pull/1286#discussion_r2351708929

I don't think any Python package should implement logic around positional arguments - modern packages rely on proper typing.

maxisbey: Hmm fair enough.

Alright so just so we can all be on the same page:

1. `def foo()`, `def foo(thing: int = 1)`
    - no exception, pass no request object, warn as deprecated
2. `def foo(req: ListPromptsRequest)`, `def foo(req: ListPromptsRequest, thing: int = 1)`
    - no exception, will pass through the request
3. `def foo(thing: int = 1, req: ListPromptsRequest | None = None)`
    - no exception, pass no request object, warn as deprecated?
3. `def foo(req)`, `def foo(req: Any)`, `def foo(req: Generic[T])`, `def foo(req: str)`
    - no exception on registration, pass no request object, warn as deprected (even though it will fail at runtime)
4. `def foo(thing: int, req: ListPromptsRequest)`
    - do not raise exception, attempt to pass request at runtime (even though it will fail at runtime)

Note: No raising custom exception at registration time

Not sure if I'm missing anything else

maxisbey: Would be keen to hear your thought before I implement @Kludex and @ochafik

maxisbey: Had a call with @Kludex, we've agreed on the above (now updated) 5 examples for me to now implement and then get a re-review.

src/mcp/server/lowlevel/func_inspection.py
Kludex: Avoid `NamedTuple`s - rather use `TypedDict`s.

---

`NamedTuple`s were a tentative to add typing to tuples without breaking code, but nowadays, there's no need to use them.

src/mcp/server/lowlevel/func_inspection.py
maxisbey: I've update the code to match the following **except** for adding the deprecation warnings.

I initially implemented it with deprecation (as you can see from the commented out code), but this infected the rest of the code base pretty thoroughly and caused a lot of issues with failed unit tests (from deprecation warnings), and the requirement of adding a lot of boilerplate code that generally isn't needed.

e.g. if you had:

```python
@server.list_tools()
def handle_list_tools() -> list[Tool]:
    return [ <my one tool> ]
```

Then to avoid the deprecation warning you'd need to change it to:

```python
@server.list_tools()
def handle_list_tools(_: ListToolsRequest) -> ListToolsResult:
    return ListToolsResult(tools=[ <my one tool> ])
```

I feel this is too much of a "breaking" change and would prefer to move the implementation/discussion around it into a separate PR rather than this one so we can get pagination in now.

src/mcp/server/lowlevel/server.py
```diff
@@ -229,12 +230,26 @@ def request_context(
...
+            func: Callable[[], Awaitable[list[types.Prompt]]]
+            | Callable[[types.ListPromptsRequest], Awaitable[types.ListPromptsResult]],
+        ):
             logger.debug("Registering handler for PromptListRequest")
 
-            async def handler(_: Any):
-                prompts = await func()
-                return types.ServerResult(types.ListPromptsResult(prompts=prompts))
+            # Create wrapper that knows how to call func with the request
+            wrapper, _ = create_call_wrapper(func, types.ListPromptsRequest)
```
ochafik: ditch tuple since second field not used, and the corresponding dead code (+ file a bug to track & discuss follow up deprecation ideas)

tests/server/lowlevel/test_deprecation_warnings.py
ochafik, line 1: Avoid committing dead code, attach this to an issue / draft pr instead

src/mcp/server/lowlevel/func_inspection.py
```diff
@@ -0,0 +1,68 @@
...
+    3. No request parameter or parameter with default (deprecated): func()
+    """
+    try:
+        sig = inspect.signature(func)
+        type_hints = get_type_hints(func)
+    except (ValueError, TypeError, NameError):
+        # Can't inspect signature or resolve type hints, assume no request parameter (deprecated)
+        return lambda _: func(), True
+
+    # Check for positional-only parameter typed as request_type
```
ochafik, line 26: won't this try and call `def foo(x: str, req: param_name)` as `foo(req)`? Also if you wanna handle defaults `def foo(x=10, *, req: param_name)` should work, and so should `def foo(req: param_name, x=10)`

src/mcp/server/lowlevel/func_inspection.py
maxisbey: I've not really used TypedDicts much for this sort of thing and just stuck with tuples for multiple returns. Is there much benefit to using a TypedDict over a tuple? And why not just go to something like a dataclass so you get attribute access rather than key based access?

Is it worth switching to one of those, or just leave as is?

src/mcp/server/lowlevel/func_inspection.py
ochafik: I think the key is to always inject by name (i.e. you won't have req in a positional-only situation, as `def foo(x=1, req: param_name)` is invalid w/o default for req).

Claude suggested something like this instead:

```py
import inspect
import warnings
from collections.abc import Callable
from typing import Any, get_type_hints


def issue_deprecation_warning(func: Callable[..., Any], request_type: type) -> None:
    """
    Issue a deprecation warning for handlers that don't use the new request parameter style.
    """
    func_name = getattr(func, "__name__", str(func))
    warnings.warn(
        f"Handler '{func_name}' should accept a '{request_type.__name__}' parameter. "
        "Support for handlers without this parameter will be removed in a future version.",
        DeprecationWarning,
        stacklevel=4,
    )


def create_call_wrapper(func: Callable[..., Any], request_type: type) -> tuple[Callable[[Any], Any], bool]:
    """
    Create a wrapper function that knows how to call func with the request object.

    Returns a tuple of (wrapper_func, should_deprecate):
    - wrapper_func: A function that takes the request and calls func appropriately
    - should_deprecate: True if a deprecation warning should be issued

    Supports patterns like:
    - def handler(req: RequestType)  # inject req
    - def handler(*, req: RequestType)  # inject req
    - def handler(req: RequestType, x=10)  # inject req, x has default
    - def handler(x=10, *, req: RequestType)  # inject req, x has default
    - def handler()  # deprecated, no injection
    """
    try:
        sig = inspect.signature(func)
        type_hints = get_type_hints(func)
    except (ValueError, TypeError, NameError):
        # Can't inspect signature or resolve type hints, assume no request parameter (deprecated)
        return lambda _: func(), True

    # Find the request parameter
    request_param_name = None
    for param_name, param in sig.parameters.items():
        param_type = type_hints.get(param_name)
        if param_type == request_type:
            # Found a parameter with the request type
            if param.default is inspect.Parameter.empty:
                # It's required - this is our injection target
                request_param_name = param_name
                break
            # Has a default - treat as deprecated old style
            return lambda _: func(), True

    if request_param_name is None:
        # No request parameter found - deprecated
        return lambda _: func(), True

    # Validate all other parameters have defaults
    for param_name, param in sig.parameters.items():
        if param_name != request_param_name:
            if param.default is inspect.Parameter.empty and param.kind not in (
                inspect.Parameter.VAR_POSITIONAL,
                inspect.Parameter.VAR_KEYWORD,
            ):
                raise TypeError(
                    f"Handler '{getattr(func, '__name__', str(func))}' has required parameter '{param_name}' "
                    f"but only '{request_param_name}' should be required"
                )

    # Create wrapper that injects only the request parameter
    def wrapper(req: Any) -> Any:
        return func(**{request_param_name: req})

    return wrapper, False
```

maxisbey: This will intentionally fail later if you pass in `func(thing: int, req: ListPromptsRequest)`. After some discussion with @Kludex this was the most "pythonic" common pattern.

Essentially, we're not going to raise any exceptions on registration, only at run-time. If you pass in a function with the wrong number of arguments, then that's on you. The type hinting is very explicit to saying your handler must have the correct type.

maxisbey: See this comment for the list of functionality/rules https://github.com/modelcontextprotocol/python-sdk/pull/1286#discussion_r2356124766

src/mcp/server/lowlevel/server.py
maxisbey: removed

tests/server/lowlevel/test_deprecation_warnings.py
maxisbey: removed

src/mcp/server/lowlevel/func_inspection.py
maxisbey: this code has been removed now

src/mcp/server/lowlevel/func_inspection.py
maxisbey: > (i.e. you won't have req in a positional-only situation, as def foo(x=1, req: param_name) is invalid w/o default for req).

You can have `foo(req: ListPromptsRequest, /)` which is positional only, no keywords. So `foo(req=request)` will fail.

ochafik (approved):
Cool, thanks for your patience, Max! Could you update the PR to mention the return type semantic change for annotated handlers? (we might wanna generalize this to all the other handlers as a follow up for consistency)

src/mcp/server/lowlevel/func_inspection.py
maxisbey: resolving for now as we no longer return multiple things

maxisbey:
> Cool, thanks for your patience, Max! Could you update the PR to mention the return type semantic change for annotated handlers? (we might wanna generalize this to all the other handlers as a follow up for consistency)

Updated!

PR #1290, opened 2025-08-21
SDK Parity: Avoid Parsing Server Response for non-JsonRPCMessage Requests

justin-yi-wang:
Achieve parity with the TypeScript SDK by only parsing server responses to JsonRPCRequests.

## Motivation and Context

In the TypeScript SDK, server responses are only parsed if the request is a JsonRPCRequest (ie. has "method" and "id" fields) - [link](https://github.com/modelcontextprotocol/typescript-sdk/blob/3bc2235d747c320dfa0b6227cc84414c6d0add89/src/client/streamableHttp.ts#L472). Otherwise, the response is ignored. Here in the Python SDK, all responses are parsed. Concretely, we've seen in the wild that servers may try to respond to notifications with a 204 No content rather than a 202 Accepted (which is special cased in both SDKs), resulting in a Pydantic validation error. Subsequently, server developers are confused why their servers work with clients presumably using the TypeScript SDK, but not ones using the Python SDK. 

## How Has This Been Tested?

This PR includes a test.

## Breaking Changes

No. Per the MCP protocol, the server should not respond to notifications. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -279,11 +279,10 @@ class StreamableHTTPTransport:
             if is_initialization:
                 self._maybe_extract_session_id_from_response(response)
 
-            content_type = response.headers.get(CONTENT_TYPE, "").lower()
-
             # Per https://modelcontextprotocol.io/specification/2025-06-18/basic#notifications:
             # The server MUST NOT send a response to notifications.
             if isinstance(message.root, JSONRPCRequest):
+                content_type = response.headers.get(CONTENT_TYPE, "").lower()
                 if content_type.startswith(JSON):
                     await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
                 elif content_type.startswith(SSE):
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -281,6 +281,8 @@ class StreamableHTTPTransport:
 
             content_type = response.headers.get(CONTENT_TYPE, "").lower()
 
+            # Per https://modelcontextprotocol.io/specification/2025-06-18/basic#notifications:
+            # The server MUST NOT send a response to notifications.
             if isinstance(message.root, JSONRPCRequest):
                 if content_type.startswith(JSON):
                     await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -281,15 +281,16 @@ class StreamableHTTPTransport:
 
             content_type = response.headers.get(CONTENT_TYPE, "").lower()
 
-            if content_type.startswith(JSON):
-                await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
-            elif content_type.startswith(SSE):
-                await self._handle_sse_response(response, ctx, is_initialization)
-            else:
-                await self._handle_unexpected_content_type(
-                    content_type,
-                    ctx.read_stream_writer,
-                )
+            if isinstance(message.root, JSONRPCRequest):
+                if content_type.startswith(JSON):
+                    await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
+                elif content_type.startswith(SSE):
+                    await self._handle_sse_response(response, ctx, is_initialization)
+                else:
+                    await self._handle_unexpected_content_type(
+                        content_type,
+                        ctx.read_stream_writer,
+                    )
 
     async def _handle_json_response(
         self,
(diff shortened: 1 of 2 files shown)


src/mcp/client/streamable_http.py
```diff
@@ -281,15 +281,16 @@ async def _handle_post_request(self, ctx: RequestContext) -> None:
...
-            if content_type.startswith(JSON):
-                await self._handle_json_response(response, ctx.read_stream_writer, is_initialization)
-            elif content_type.startswith(SSE):
-                await self._handle_sse_response(response, ctx, is_initialization)
-            else:
-                await self._handle_unexpected_content_type(
-                    content_type,
-                    ctx.read_stream_writer,
-                )
+            if isinstance(message.root, JSONRPCRequest):
```
ochafik, line 284: Can you leave a comment, e.g. w/ link to https://modelcontextprotocol.io/specification/2025-06-18/basic#notifications

tests/client/test_notification_response.py
```diff
@@ -0,0 +1,150 @@
...
+    proc.join(timeout=2)
+
+
+@pytest.mark.anyio
+async def test_notification_with_204_response(non_sdk_server: None, non_sdk_server_port: int) -> None:
+    """Test that client handles 204 responses to notifications correctly.
+
+    This test verifies the fix for the issue where non-SDK servers
+    might return 204 No Content for notifications instead of 202 Accepted.
+    The client should handle this gracefully without trying to parse
```
ochafik: Could you mention SDK parity as as justification here?

The spec usually is the northstar used to decide what the sdk should do, here we're just out of its scope / accommodating bad server behaviour (https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server), so there's no "correctly" so to speak.

src/mcp/client/streamable_http.py
justin-yi-wang: done!

tests/client/test_notification_response.py
justin-yi-wang: Yep done! 

I'm mixed on if this is bad server behavior (the server has to respond to the HTTP request somehow, and 202 and 204 seem equally appealing to meet the spec of "No response") or if it's a footgun in the SDK. Either way this allows the client to be a tad more defensive.

ochafik: > 202 and 204 seem equally appealing to meet the spec of "No response"

The spec ([link](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server) above) is very specific about 202 + no response body for the streamable http transport:

```
If the input is a JSON-RPC response or notification:
- If the server accepts the input, the server MUST return HTTP status code 202 Accepted with no body.
- If the server cannot accept the input, it MUST return an HTTP error status code (e.g., 400 Bad Request). The HTTP response body MAY comprise a JSON-RPC error response that has no id.
```

tests/client/test_notification_response.py
```diff
@@ -0,0 +1,148 @@
...
+        pytest.fail("Server failed to start within 10 seconds")
+
+    yield
+
+    proc.kill()
+    proc.join(timeout=2)
+
+
+@pytest.mark.anyio
+async def test_notification_with_204_response(non_sdk_server: None, non_sdk_server_port: int) -> None:
```
ochafik: test_non_compliant_notification_response

tests/client/test_notification_response.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    This test verifies that the client does not parse responses to non-JsonRPCRequests, which matches the
+    behavior of the TypeScript SDK. The test uses a 204 No Content (commonly seen from servers), but in reality
+    any 2xx response should be handled the same way.
```
ochafik: "This test verifies that the client ignores unexpected responses to notifications: the spec states they should either be 202 + no response body, or 4xx + optional error body (https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server), but some servers wrongly return other 2xx codes (e.g. 204). For now we simply ignore unexpected responses (aligning behaviour w/ the TS SDK)."

src/mcp/client/streamable_http.py
```diff
@@ -281,15 +281,18 @@ async def _handle_post_request(self, ctx: RequestContext) -> None:
 
             content_type = response.headers.get(CONTENT_TYPE, "").lower()
```
ochafik: nit: move down inside the if branch

tests/client/test_notification_response.py
justin-yi-wang: Oh interesting! Thanks for pointing that out, missed that line.

ochafik (approved):
Thanks!

PR #1292, opened 2025-08-21
types: Setting default value for method: Literal

sreenaths:
method fields are literals that enforce a specific value. As the value is not set by default, we have to pass the correct value every time a model is instantiated.

## Motivation and Context
This change simplifies model instantiation, as the method would be automatically set.
For example
```python
    initialize_request = InitializeRequest(
        # method="initialize", - This wouldn't be required
        params=InitializeRequestParams(
            ...
        )
    )
```

## How Has This Been Tested?
Existing tests work as expected, and a new test was added.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
None

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -151,7 +151,6 @@ class ClientSession(
         result = await self.send_request(
             types.ClientRequest(
                 types.InitializeRequest(
-                    method="initialize",
                     params=types.InitializeRequestParams(
                         protocolVersion=types.LATEST_PROTOCOL_VERSION,
                         capabilities=types.ClientCapabilities(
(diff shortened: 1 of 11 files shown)


ochafik (changes requested):
Hi @sreenaths, thanks for your PR!

I think this makes sense, could you please also remove the now-redundant method ctor args throughout the SDK? (mostly in {client,server}/session.py and tests)

sreenaths:
Hi @ochafik,
Thank you so much for the review.
I have updated the PR to remove redundant method args.

ochafik (approved):
Thanks @sreenaths !

PR #1308, opened 2025-08-26
Catch http errors 

lorenss-m:
# Fix: Propagate HTTP errors to client

## Motivation and Context
HTTP request failures were not being properly communicated to clients, causing silent failures. This fix ensures errors are sent through the read stream so clients are notified when requests fail.

## How Has This Been Tested?
Has been tested in production environments (used by https://github.com/hud-evals/hud-python with a remote server).

## Breaking Changes
None - this is a backwards-compatible error handling improvement.

## Types of changes

## Checklist
- [x] Documentation not needed for this internal fix

## Additional context
This change wraps the `handle_request_async` function in a try-catch block and sends any exceptions to `ctx.read_stream_writer` to ensure proper error propagation in the streamable HTTP transport layer.

### Example scenario where this fix is critical:

**HTTP Error Codes (502, 503, 504, etc.)**
```python
# Without this fix: Client hangs when server returns 502 Bad Gateway
# With this fix: Client receives the HTTP error

async with streamablehttp_client(server_url) as (read_stream, write_stream):
        await session.initialize()
        try:
            result = await session.call_tool("api_operation", arguments={})
        except Exception as e:
            print(f"Error received: {e}")  # Now properly catches 502, 503, 504 errors
```

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -400,17 +400,22 @@ class StreamableHTTPTransport:
                         sse_read_timeout=self.sse_read_timeout,
                     )
 
-                    async def handle_request_async():
-                        if is_resumption:
-                            await self._handle_resumption_request(ctx)
-                        else:
-                            await self._handle_post_request(ctx)
+                    async def handle_request_async(ctx: RequestContext, is_resumption: bool) -> None:
+                        try:
+                            if is_resumption:
+                                await self._handle_resumption_request(ctx)
+                            else:
+                                await self._handle_post_request(ctx)
+                        except Exception as e:
+                            # Send error to read stream so client knows request failed
+                            logger.error("Request handler error: %s", e)
+                            await ctx.read_stream_writer.send(e)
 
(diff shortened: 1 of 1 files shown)


lorenss-m:
@Kludex bumping this -- really important for our remote server to include this, our library is forced to depend on our fork otherwise!

src/mcp/client/streamable_http.py
```diff
@@ -400,17 +400,22 @@ async def post_writer(
...
-                            await self._handle_resumption_request(ctx)
-                        else:
-                            await self._handle_post_request(ctx)
+                    async def handle_request_async(ctx: RequestContext, is_resumption: bool) -> None:
+                        try:
+                            if is_resumption:
+                                await self._handle_resumption_request(ctx)
+                            else:
+                                await self._handle_post_request(ctx)
+                        except Exception as e:
```
felixweinberger, line 567: This is too broad of an error, we should be more specific about the type of Exception we handle here.

We should at least limit to `HTTPStatusError` if your problem is with HTTP errors specifically.

PR #1310, opened 2025-08-26
types: Setting default value for jsonrpc

sreenaths:
The JSON-RPC version is fixed at 2.0 for MCP messages. However, as the value is not set by default, we have to pass the correct value every time a model is instantiated.

## Motivation and Context
This change simplifies model instantiation by automatically setting the JSON-RPC version. If the version ever needs to change in the future, it only has to be updated in a single place.

For example
```python
JSONRPCMessage(
    root=JSONRPCRequest(
        # jsonrpc="2.0", - This wouldn't be required
        id=1,
        method="ping"
    )
),
```

## How Has This Been Tested?
Tests were improved and are working as expected.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
None


sreenaths:
Hi @ochafik
When you have a chance, please review this PR.
It is similar to PR #1292 that you had helped with before.
Thank you.

sreenaths:
Hi @hesreallyhim 
Thank you so much for the suggestion. It would be nice to have a common base class for JSONRPCRequest, JSONRPCNotification, JSONRPCResponse, and JSONRPCError.

But, as the original schema is [defined in TypeScript](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts), and types.py depends on it... I am not sure if we can change the base structure in Python SDK. Additionally, we might have to use multiple inheritance as JSONRPCRequest already inherits from Request and JSONRPCNotification from Notification.

I hope one of the maintainers can guide us. 

\cc @ochafik

hesreallyhim:
> But, as the original schema is [defined in TypeScript](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts), and types.py depends on it... I am not sure if we can change the base structure in Python SDK. Additionally, we might have to use multiple inheritance as JSONRPCRequest already inherits from Request and JSONRPCNotification from Notification.

yeah... well i think they've kind of fixed this in the draft spec, but the sdk may still choose to do things a little differently.

Kludex:
We need to migrate to this: https://github.com/Kludex/mcp/blob/main/src/mcp-types/mcp_types/__init__.py

ochafik (approved):
Thanks @sreenaths !

ochafik:
> We need to migrate to this: https://github.com/Kludex/mcp/blob/main/src/mcp-types/mcp_types/__init__.py

@Kludex not sure it would support passing the literal values as defaults?

hesreallyhim:
sorry, perhaps I should have been more direct - i would need to study the codebase more closely to understand how the sdk consumes the spec schema.ts - in the case of the typescript SDK i think they had decided to declare the internal types as not including `jsonprc`, and adding this in at a "protocol" layer before sending the message over the wire. In any case, the observations made above by @sreenaths, such as:

>  It would be nice to have a common base class for JSONRPCRequest, JSONRPCNotification, JSONRPCResponse, and JSONRPCError.

was actually just an oversight in the MCP schema.ts (or so I argued) - the schema hierarchy was not type-theoretically correct. I merged a change into the draft so that all requests inherit from JSONRPCRequest, and notifications extend JSONRPCNotification - see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1026 for more info if interested. I hope this is relevant, sorry if it's not, just flagging that there is an upcoming change surrounding this part of the schema.

sreenaths:
Thank you for the review, @ochafik.

@hesreallyhim It looks like your changes are one level down the class hierarchy, so I believe our changes are independent.

Thanks, @Kludex. Will mcp_types be included in the Python SDK? As @ochafik mentioned, does mcp_types set literal values as defaults? From what I can see, it only defines the types here: [mcp_types](https://github.com/Kludex/mcp/blob/main/src/mcp-types/mcp_types/__init__.py#L403)

sreenaths:
Thank you, @felixweinberger 
Hope this can be merged soon.

Kludex:
Closing this as stale. I think this is not that important since it's for internal code.

PR #1321, opened 2025-08-28
Handle Abrupt Server Disconnect in Middle of Tool Call

EiffL:
Propagates transport disconnects (e.g. server killed mid tool call) as CONNECTION_CLOSED errors instead of hanging indefinitely. Adds regression test.

## Motivation and Context

A call_tool over the StreamableHTTP transport would hang forever if the server closed the SSE/HTTP connection mid-response (e.g. process crash/kill). The underlying RemoteProtocolError was logged but in‑flight requests never completed. This change converts that low-level failure into a deterministic JSON-RPC error for every pending request, matching expected resilience semantics.

## How Has This Been Tested?

- Added integrated test test_streamable_http_mid_call_disconnect:
  - Starts real server process.
  - Invokes long-running tool (wait_for_lock_with_notification).
  - Kills server mid-call.
  - Asserts an McpError with code CONNECTION_CLOSED (previously would hang).
- All existing tests pass locally.

## Breaking Changes

None. Behavior only changes for abnormal transport termination; normal flows unaffected. No public API changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Implementation details:

In _handle_sse_response (client transport) on exception: send the exception into the session read stream then close the writer to signal termination.

In BaseSession._receive_loop: on receiving an Exception, broadcast a synthesized JSONRPCError (code CONNECTION_CLOSED) to all pending request streams and clear them.

Integrated test placed alongside existing StreamableHTTP tests for consistency. Potential follow-up: introduce optional reconnection/resumption logic for recoverable disconnects.


src/mcp/client/streamable_http.py
```diff
@@ -337,7 +337,13 @@ async def _handle_sse_response(
...
         except Exception as e:
             logger.exception("Error reading SSE stream:")
-            await ctx.read_stream_writer.send(e)
+            # Forward the transport exception to the session read stream so it can
+            # fail any in-flight requests, then close the stream to signal that
+            # no further messages will arrive.
+            try:
+                await ctx.read_stream_writer.send(e)
+            finally:
+                await ctx.read_stream_writer.aclose()
```
felixweinberger, line 346: Closing on any type of exception seems to broad - we should be more explicit about the failure cases that definitely mean the session is unusable, there might be recoverable errors.

src/mcp/shared/session.py
```diff
@@ -336,7 +336,21 @@ async def _receive_loop(self) -> None:
...
+                        # Send error to any pending request response streams immediately
+                        for id, stream in list(self._response_streams.items()):
+                            try:
+                                await stream.send(JSONRPCError(jsonrpc="2.0", id=id, error=error))
+                                await stream.aclose()
+                            except Exception:
+                                pass
+                        self._response_streams.clear()
+                        # Break out of the receive loop; connection is no longer usable.
+                        break
```
felixweinberger, line 353: This seems like a duplication of all the logic we have at the bottom of `_receive_loop` (L440-450) (except for the `break`)

The changes here also don't seem necessary to make the tests you added pass - do we need this?

felixweinberger:
Closing this for inactivity as we're working through our backlog at the moment to reduce our review queue size - feel free to reopen if you plan to come back to this at a later time @EiffL

PR #1322, opened 2025-08-29
chore: add ability to remove tools

brandonspark:
## Motivation and Context
I am interested in being able to dynamically have a server offer only certain tools based on information received during its initialization. I could conditionally declare the tool, for instance:
```
if cond:
  @mcp.tool
  async def foo():
    ...
```
but I perceive this to be relatively gross. I would prefer to be able to remove a tool within the API.

## How Has This Been Tested?
I can achieve comparable behavior by inserting an `if` before defining a function with the `mcp.tool()` decorator.

## Breaking Changes
No.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/tools/tool_manager.py
+++ b/src/mcp/server/fastmcp/tools/tool_manager.py
@@ -72,6 +72,7 @@ class ToolManager:
         """Remove a tool by name."""
         if name not in self._tools:
             logger.warning(f"Tried to remove unknown tool: {name}")
+            return
         del self._tools[name]
 
     async def call_tool(
--- a/src/mcp/server/fastmcp/tools/tool_manager.py
+++ b/src/mcp/server/fastmcp/tools/tool_manager.py
@@ -68,6 +68,12 @@ class ToolManager:
         self._tools[tool.name] = tool
         return tool
 
+    def remove_tool(self, name: str) -> None:
+        """Remove a tool by name."""
+        if name not in self._tools:
+            logger.warning(f"Tried to remove unknown tool: {name}")
+        del self._tools[name]
+
     async def call_tool(
         self,
         name: str,
--- a/tests/server/fastmcp/test_tool_manager.py
+++ b/tests/server/fastmcp/test_tool_manager.py
@@ -633,3 +633,117 @@ class TestStructuredOutput:
         # Test converted result
         result = await manager.call_tool("get_scores", {})
         assert result == expected
+
+
+class TestRemoveTools:
+    """Test tool removal functionality in the tool manager."""
+
+    def test_remove_existing_tool(self, caplog: pytest.LogCaptureFixture):
+        """Test removing an existing tool."""
+        def add(a: int, b: int) -> int:
+            """Add two numbers."""
+            return a + b
+
+        manager = ToolManager()
+        manager.add_tool(add)
+
+        # Verify tool exists
+        assert manager.get_tool("add") is not None
+        assert len(manager.list_tools()) == 1
+
+        # Remove the tool
+        with caplog.at_level(logging.WARNING):
+            manager.remove_tool("add")
+            # Should not log a warning for removing existing tool
+            assert "Tried to remove unknown tool: add" not in caplog.text
+
+        # Verify tool is removed
+        assert manager.get_tool("add") is None
+        assert len(manager.list_tools()) == 0
+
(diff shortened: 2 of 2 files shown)


src/mcp/server/fastmcp/tools/tool_manager.py
```diff
@@ -68,6 +68,13 @@ def add_tool(
         self._tools[tool.name] = tool
         return tool
 
+    def remove_tool(self, name: str) -> None:
+        """Remove a tool by name."""
```
dsp-ant, line 72: We should probably say a word about the beahviour if the tool doesn't exists.

dsp-ant (approved):
I think this is very reasonable. @Kludex opinions?

src/mcp/server/fastmcp/tools/tool_manager.py
Kludex: I think this should raise if not found.

Kludex:
> I think this is very reasonable. @Kludex opinions?

It seems something is missing, since the user doesn't have access to the underlying `ToolManager` from the `FastMCP` object.

How is the user supposed to remove a tool?

brandonspark:
> > I think this is very reasonable. @Kludex opinions?
> 
> It seems something is missing, since the user doesn't have access to the underlying `ToolManager` from the `FastMCP` object.
> 
> How is the user supposed to remove a tool?

I was under the impression that one could obtain it from doing `mcp._tool_manager` from a `ToolManager` object. I assume your objection is because that field is supposed to be "private", but it would enable my use case. Do you have a significant objection?

It feels natural to me that this code should exist, as `add_tool` also exists.

Kludex:
That's a significant objection.

brandonspark:
> That's a significant objection.

Then, as a maintainer, do you have thoughts on how you would rather enable this behavior? I am not as familiar with your project.

It seems reasonably clear to me that a user may want to alter the tools which are available to the MCP, in a way that is not solely static. I don't see why the tool manager must be private, nor why we could not surface some of these methods through the API.

Kludex:
I think the current implementation (besides my comment about the exceptions) makes sense. It's just incomplete.

Is the idea here to disable the tool for the whole app lifecycle or for the session? Does it even makes sense to have it disable for the app lifecycle, since the app may be running in multiple machines? I guess we want to expose a method you can call from the Context object anyway.

Kludex:
I don't think ToolManager should be exposed.

brandonspark:
> I think the current implementation (besides my comment about the exceptions) makes sense. It's just incomplete.
> 
> Is the idea here to disable the tool for the whole app lifecycle or for the session? Does it even makes sense to have it disable for the app lifecycle, since the app may be running in multiple machines? I guess we want to expose a method you can call from the Context object anyway.

In my use case, it would actually be most ideal if a tool could be enabled or disabled for specific sessions (meaning, differing tools for differing consumers of a single server). I wanted to downscope this, as this is certainly more complicated than I feel I am seeking to implement right now. My understanding is that currently `ToolManager` is "global" for the entire app.

In my use case, I want users who are hosting their own servers (instances of the application) to be able to decide if they want to allow certain tools to be available to the consumers of their server. The best way to do this currently is with conditional declaration, which I view to be quite heavy handed.

If we do not expose `ToolManager`, we could expose a method which makes use of `remove_tool`, but I don't see a place in the current API (a particular class, specifically) where that makes sense.

brandonspark:
Oh, I now see that `FastMCP` exposes a thin wrapper around `add_tool`. Any objections to me doing the same for `remove_tool`?

https://github.com/modelcontextprotocol/python-sdk/blob/47d35f0b3ca2e59ddc26ddc0f2a816ac3dc8df9b/src/mcp/server/fastmcp/server.py#L121

src/mcp/server/fastmcp/tools/tool_manager.py
felixweinberger: Resolving as this seems addressed to me

maxisbey:
Hey @brandonspark  I've pushed up the changes that @felixweinberger requested to help move the PR across the line since the changes were quite simple.

felixweinberger (approved):
LGTM, thanks for updating @maxisbey

lukehsiao:
Any thoughts on when this will make it into a release?

felixweinberger:
> Any thoughts on when this will make it into a release?

Should be in the next release on Thursday.

PR #1323, opened 2025-08-29
Accept additional response_types values from OAuth servers

jonshea:
OAuth servers may return additional response_types beyond what the client requested (e.g., ["code", "none"] instead of just ["code"]). Per RFC 7591 Section 3.2.1, servers can modify registration metadata and return all registered values including server-provisioned fields. For example, Keycloak returns ["code", "none"] even when just ["code"] was requested.

- Changed response_types field from list[Literal["code"]] to list[str]
- Added validation in registration handler to ensure "code" is present
- Added tests for response_types flexibility

This fixes compatibility issues with OAuth servers that return additional response_types while maintaining MCP's requirement for the "code" flow.


pcarleton (approved):

pcarleton:
thanks for this!

PR #1324, opened 2025-08-29
[client] Implement MCP OAuth scope selection and step-up authorization

dogacancolak:
Implements the MCP specification's OAuth scope selection strategy and step-up authorization flow for handling scope challenges.

## Specification Sections Implemented

1. **[Scope Selection Strategy](https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-selection-strategy)** - Priority order for selecting scopes during authorization:
   - Use `scope` from WWW-Authenticate header (if provided)
   - Use all scopes from PRM `scopes_supported` (if available)
   - Omit scope parameter (if neither is available)

2. **[Scope Challenge Handling](https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling)** - Step-up authorization when 403 with `insufficient_scope` is received:
   - Extract updated scope from WWW-Authenticate header
   - Trigger re-authorization with new scope
   - Retry request with new tokens

## Changes Made

**Core Implementation:**
- Added `_extract_field_from_www_auth()` - Generic WWW-Authenticate header parser
- Added `_extract_scope_from_www_auth()` - Extracts scope per RFC6750
- Added `_select_scopes()` - Implements MCP scope selection priority
- Added 403 handler in `async_auth_flow()` - Step-up authorization flow
- Removed fallback to authorization server metadata scopes

## Testing

Comprehensive test coverage added in `tests/client/test_auth.py`:

**Scope Selection Tests:**
- WWW-Authenticate scope prioritized over PRM
- PRM scopes used when WWW-Authenticate has no scope
- Scope omitted when neither provides values
- Various WWW-Authenticate header formats (quoted/unquoted)

**Step-Up Authorization Tests:**
- 403 with `insufficient_scope` triggers step-up flow
- Scope correctly updated from WWW-Authenticate
- 403 without `insufficient_scope` does not trigger step-up

## Breaking Changes

- No longer falls back to authorization server metadata `scopes_supported`

---

## Checklist

- [x] Follows MCP specification
- [x] Code follows style guidelines
- [x] All tests pass locally
- [x] Added appropriate error handling

### DIFF
--- a/src/mcp/client/auth.py
+++ b/src/mcp/client/auth.py
@@ -272,14 +272,13 @@ class OAuthClientProvider(httpx.Auth):
         else:
             raise OAuthFlowError(f"Protected Resource Metadata request failed: {response.status_code}")
 
-    def _configure_scope_selection(self, init_response: httpx.Response) -> None:
+    def _select_scopes(self, init_response: httpx.Response) -> None:
         """Select scopes as outlined in the 'Scope Selection Strategy in the MCP spec."""
         # Per MCP spec, scope selection priority order:
         # 1. Use scope from WWW-Authenticate header (if provided)
         # 2. Use all scopes from PRM scopes_supported (if available)
         # 3. Omit scope parameter if neither is available
         #
-        # Step 1: Extract scope from WWW-Authenticate header
         www_authenticate_scope = self._extract_scope_from_www_auth(init_response)
         if www_authenticate_scope is not None:
             # Priority 1: WWW-Authenticate header scope
@@ -555,7 +554,7 @@ class OAuthClientProvider(httpx.Auth):
                     await self._handle_protected_resource_response(discovery_response)
 
                     # Step 2: Apply scope selection strategy
-                    self._configure_scope_selection(response)
+                    self._select_scopes(response)
 
                     # Step 3: Discover OAuth metadata (with fallback for legacy servers)
                     discovery_urls = self._get_discovery_urls()
@@ -600,7 +599,7 @@ class OAuthClientProvider(httpx.Auth):
                 if error == "insufficient_scope":
                     try:
                         # Step 2a: Update the required scopes
-                        self._configure_scope_selection(response)
+                        self._select_scopes(response)
 
                         # Step 2b: Perform (re-)authorization
                         auth_code, code_verifier = await self._perform_authorization()
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -449,7 +449,7 @@ class TestOAuthFallback:
         await oauth_provider._handle_protected_resource_response(prm_metadata_response)
 
         # Process the scope selection with WWW-Authenticate header
-        oauth_provider._configure_scope_selection(init_response_with_www_auth_scope)
+        oauth_provider._select_scopes(init_response_with_www_auth_scope)
 
         # Verify that WWW-Authenticate scope is used (not PRM scopes)
         assert oauth_provider.context.client_metadata.scope == "special:scope from:www-authenticate"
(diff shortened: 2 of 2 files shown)


jennsun:
Can we confirm if updating the Python SDK will update the Claude Client, or will we also need to make updates in the TypeScript SDK for these to reflect in Claude.ai? In the Typescript SDK, for example, we are not retrieving scopes from the PRM endpoint: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/auth.ts#L328
CC: @pcarleton

src/mcp/client/auth.py
```diff
@@ -478,9 +478,18 @@ async def _handle_oauth_metadata_response(self, response: httpx.Response) -> Non
...
+        if self.context.client_metadata.scope is None:
+            # Priority 1: Use PRM's scopes_supported if available
+            if (
+                self.context.protected_resource_metadata is not None
+                and self.context.protected_resource_metadata.scopes_supported is not None
+            ):
+                self.context.client_metadata.scope = " ".join(self.context.protected_resource_metadata.scopes_supported)
+            # Priority 2: Fall back to OAuth metadata scopes if available
+            elif metadata.scopes_supported is not None:
+                self.context.client_metadata.scope = " ".join(metadata.scopes_supported)
```
pcarleton: see this latest change to the spec:
> When implementing authorization flows, MCP clients **SHOULD** follow the principle of least privilege by requesting
only the scopes necessary for their intended operations. During the initial authorization handshake, MCP clients
**SHOULD** follow this priority order for scope selection:
>
> 1. **Use `scope` parameter** from the initial `WWW-Authenticate` header in the 401 response, if provided
> 2. **If `scope` is not available**, use all scopes defined in `scopes_supported` from the Protected Resource Metadata document, omitting the `scope` parameter if `scopes_supported` is undefined.
>
> This approach accommodates the general-purpose nature of MCP clients, which typically lack domain-specific knowledge to make informed decisions about individual scope selection. Requesting all available scopes allows the authorization server and end-user to determine appropriate permissions during the consent process.


Are you able to update this PR to:
1. Get scope from `www-authenticate`
2. Omit scope if not present in PRM? (currently impl falls back to `scopes_supported` from AS metadata)

dogacancolak: Updated. Although the complete priority order is:
1. Explicitly configured scope in the client
2. `www-authenticate` scope
3. PRM scope
4. Omit

pcarleton: Hm I don't think (1) is correct there.  If we always use what's configured on the client, there's no path for upgrading scope if the initial request didn't have the correct scope.

We had some discussion in this across a few PR's ([example](https://github.com/modelcontextprotocol/typescript-sdk/pull/983#issuecomment-3374271037)) and in [discord](https://discord.com/channels/1358869848138059966/1424202473387589706) about similar issues.

dogacancolak: According to this spec https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling the step-up only happens when the server returns 403. The changes in this PR only take effect upon a 401.

I can definitely update the 401 path by not prioritizing the configured scope on the client, but the step-up logic remains unhandled in the MCP client in this SDK. I can create a separate PR for that if that makes sense

dogacancolak: I can make it part of this PR too considering this is already a breaking change

dogacancolak: working on adding step-up auth flow

dogacancolak: done

jennsun:
Thanks for making these changes @dogacancolak and approving @pcarleton! Do we have an estimate on when these scope selection updates will be available for Claude desktop/claude.ai connectors?

PR #1325, opened 2025-08-30
Use proper sync resolution in CI

Kludex:


Kludex:
@maxisbey Do you have time to debug this?

maxisbey:
> @maxisbey Do you have time to debug this?

Pretty sure it's because of this function here: https://github.com/modelcontextprotocol/python-sdk/blob/543961968c0634e93d919d509cce23a1d6a56c21/tests/client/test_config.py#L26

It runs a uv command built here: https://github.com/modelcontextprotocol/python-sdk/blob/4fee123e72f3e01d01e2fb31282eb206e8cee308/src/mcp/cli/claude.py#L67

This is then run without `uv --frozen` and without whatever dependency resolution is being set by the changes you made in this PR. So the unit test times out since it's re-installing a bunch of dependencies due to the different dependency version resolution uv is using when launched inside the unit test.

felixweinberger:
Will pick this up from @ihrpr

felixweinberger:
Closing in favor of https://github.com/modelcontextprotocol/python-sdk/pull/1507

PR #1334, opened 2025-09-02
Add client_secret_basic authentication support

jonshea:
Add support for HTTP Basic Authentication (client_secret_basic) as a
client authentication method for the token and revoke endpoints, alongside 
the existing client_secret_post method. This improves compatibility with 
OAuth servers like Keycloak that use Basic auth.

Key changes:
- Update OAuthClientMetadata to accept "client_secret_basic" as valid
  token_endpoint_auth_method
- Return 401 status for authentication failures (was 400)
- Update metadata endpoints to advertise both auth methods
- Add tests for both auth methods and edge cases

### DIFF
--- a/src/mcp/server/auth/handlers/token.py
+++ b/src/mcp/server/auth/handlers/token.py
@@ -92,6 +92,22 @@ class TokenHandler:
 
     async def handle(self, request: Request):
         try:
+            client_info = await self.client_authenticator.authenticate_request(request)
+        except AuthenticationError as e:
+            # Authentication failures should return 401
+            return PydanticJSONResponse(
+                content=TokenErrorResponse(
+                    error="unauthorized_client",
+                    error_description=e.message,
+                ),
+                status_code=401,
+                headers={
+                    "Cache-Control": "no-store",
+                    "Pragma": "no-cache",
+                },
+            )
+
+        try:
             form_data = await request.form()
             token_request = TokenRequest.model_validate(dict(form_data)).root
         except ValidationError as validation_error:
@@ -99,19 +115,6 @@ class TokenHandler:
                 TokenErrorResponse(
                     error="invalid_request",
                     error_description=stringify_pydantic_error(validation_error),
-                )
-            )
-
-        try:
-            client_info = await self.client_authenticator.authenticate(
-                client_id=token_request.client_id,
-                client_secret=token_request.client_secret,
-            )
-        except AuthenticationError as e:
-            return self.response(
-                TokenErrorResponse(
-                    error="unauthorized_client",
-                    error_description=e.message,
                 )
             )
 
--- a/src/mcp/server/auth/middleware/client_auth.py
+++ b/src/mcp/server/auth/middleware/client_auth.py
@@ -1,5 +1,8 @@
+import base64
 import time
 from typing import Any
+
+from starlette.requests import Request
 
 from mcp.server.auth.provider import OAuthAuthorizationServerProvider
 from mcp.shared.auth import OAuthClientInformationFull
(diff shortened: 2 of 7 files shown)


src/mcp/server/auth/middleware/client_auth.py
```diff
@@ -30,19 +33,67 @@ def __init__(self, provider: OAuthAuthorizationServerProvider[Any, Any, Any]):
...
+                decoded = base64.b64decode(encoded_credentials).decode("utf-8")
+                if ":" not in decoded:
+                    raise ValueError("Invalid Basic auth format")
+                basic_client_id, request_client_secret = decoded.split(":", 1)
+
+                if basic_client_id != client_id:
+                    raise AuthenticationError("Client ID mismatch in Basic auth")
+            except AuthenticationError:
+                raise
+            except Exception:
```
ochafik: Maybe `except (ValueError, UnicodeDecodeError, base64.binascii.Error):` to be more explicit about what exceptions we should tolerate here?

src/mcp/server/auth/handlers/token.py
```diff
@@ -91,6 +91,22 @@ def response(self, obj: TokenSuccessResponse | TokenErrorResponse):
...
+                ),
+                status_code=401,
+                headers={
+                    "Cache-Control": "no-store",
+                    "Pragma": "no-cache",
+                },
+            )
+
         try:
             form_data = await request.form()
```
ochafik, line 111: Are we not reading the request.form() twice here? (once in authenticate_request, and here again? (Think starlette might complain about this)

Might wanna push the form data (maybe other request fields, e.g. auth header) to the authenticator method?

src/mcp/server/auth/middleware/client_auth.py
```diff
@@ -30,19 +33,67 @@ def __init__(self, provider: OAuthAuthorizationServerProvider[Any, Any, Any]):
...
+        if client.token_endpoint_auth_method == "client_secret_basic":
+            if not auth_header.startswith("Basic "):
+                raise AuthenticationError("Missing or invalid Basic authentication in Authorization header")
+
+            try:
+                encoded_credentials = auth_header[6:]  # Remove "Basic " prefix
+                decoded = base64.b64decode(encoded_credentials).decode("utf-8")
+                if ":" not in decoded:
+                    raise ValueError("Invalid Basic auth format")
+                basic_client_id, request_client_secret = decoded.split(":", 1)
```
ochafik, line 76: We should probably urldecode both parts, as per [RFC 6749 Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1)

> The client identifier is encoded using the 'application/x-www-form-urlencoded' encoding algorithm per Appendix B, and the encoded value is used as the username; the client password is encoded using the same algorithm and used as the password.

src/mcp/server/auth/middleware/client_auth.py
```diff
@@ -30,19 +33,67 @@ def __init__(self, provider: OAuthAuthorizationServerProvider[Any, Any, Any]):
...
+        else:
+            raise AuthenticationError(f"Unsupported auth method: {client.token_endpoint_auth_method}")
+
         if client.client_secret:
-            if not client_secret:
+            if not request_client_secret:
                 raise AuthenticationError("Client secret is required")
 
-            if client.client_secret != client_secret:
+            if client.client_secret != request_client_secret:
```
ochafik: meganit: in theory we should use `compare_digest()` here to avoid any chance of timing attack, although it's probably drowned in the noise of network and Python execution overheads.

src/mcp/server/auth/middleware/client_auth.py
jonshea: Good catch. Thank you.

src/mcp/server/auth/handlers/token.py
jonshea: My understanding is that the current version of starlette caches calls to `form()`, `json()`, and `body()`, so it is safe to call them multiple times. For example, https://github.com/Kludex/starlette/blob/main/starlette/requests.py#L254-L287 . But I am not a starlette expert and I might be misunderstanding. There seems to be an [edge case
](https://github.com/Kludex/starlette/discussions/1933) with reading `form()` and then `body()` later. But if it was ever unsafe to call `form()` multiple times, I was unable to find when it changed after a modest search.

My first implementation parsed out the form in the request handler and passed it into the `authenticate` method, as you suggest, but I thought it felt clunky to duplicate the code that parsed "form data and auth header" across both the token and revoke endpoints. It’s not the end of the world, but it turns
```
# Current version
async def handle(self, request: Request):
    try:
        client_info = await self.client_authenticator.authenticate_request(request)
```
into something like this
```
# Parse out form and auth_header
async def handle(self, request: Request):
    try:
        form_data = await request.form()
    except Exception:
        return self.response(
            TokenErrorResponse(
                error="invalid_request",
                error_description="Unable to parse request body",
            )
        )

    auth_header = request.headers.get("Authorization")

    try:
        client_info = await self.client_authenticator.authenticate_request(form_data, auth_header)
```

Or I suppose we could handle invalid for data in the authenticate method:
```
# Parse out form and auth_header, handle form error in client_authenticator
async def handle(self, request: Request):
    form_data = None
    try:
        form_data = await request.form()
    except Exception:
        pass

    auth_header = request.headers.get("Authorization")

    try:
        client_info = await self.client_authenticator.authenticate_request(form_data, auth_header)
```

Anyway, I defer to the maintainers. If you would like me to switch to one of the above implementations, I would be happy to do so.

src/mcp/client/auth/oauth2.py
```diff
@@ -291,6 +327,27 @@ async def _register_client(self) -> httpx.Request | None:
 
         registration_data = self.context.client_metadata.model_dump(by_alias=True, mode="json", exclude_none=True)
 
+        # If token_endpoint_auth_method is None, auto-select based on server support
+        if self.context.client_metadata.token_endpoint_auth_method is None:
+            preference_order = ["client_secret_basic", "client_secret_post", "none"]
```
jonshea, line 288: I am torn on whether or not we should allow auto-selecting `"none"`. It seems possibly like bad security to allow that, but I suppose if the server allows it then it is ok?

I suppose _ideally_ we should allow the user to pick a list of auth methods they want to allow to be auto-configured, but I am not sure anyone cares enough to want to use it.

jonshea:
Thank you for the thoughtful review, @ochafik. I believe I have addressed all of the provided feedback, though with a [reply / question](https://github.com/modelcontextprotocol/python-sdk/pull/1334/files#r2388425820) about the concern around calling `request.form()` twice rather than a fix.

As suggested, I also added support for `client_secret_basic` to the MCP client. I made `OAuthClientMetadata.token_endpoint_auth_method` [optional](https://github.com/modelcontextprotocol/python-sdk/pull/1334/files#diff-01350058961f3196e3373d707e2178962377dc9a85f481ade94e8f05e4701c8dR45) to support auto-configuring a `token_endpoint_auth_method`. I think this is a reasonable change, but I wanted to call it out. I have a [question here](https://github.com/modelcontextprotocol/python-sdk/pull/1334/files#r2393339173) about whether or not we should allow the `token_endpoint_auth_method` auto-configure to pick `"none"`.

I put the client-side `client_secret_basic` auth support in its own commit in case we would like to pull it into its own PR.

HenriChabert:
Hello @ochafik and @pcarleton !
Any update on this PR ? Does it seem good enough ?
`client_secret_basic` is pretty common in the industry, and it would be great if we could integrate servers using it.
Thanks 🙏

pcarleton (approved):

HenriChabert:
Thanks for handling it @pcarleton !

PR #1343, opened 2025-09-04
Add idle session cleanup to prevent unbounded session accumulation

ddworken:
## Summary
- Added automatic cleanup of idle sessions when session count exceeds threshold
- Sessions inactive longer than idle timeout are terminated to free resources
- Prevents memory exhaustion from clients that create sessions but never terminate them

## Changes
- Added configurable session idle timeout (default 30 minutes)
- Added background cleanup task that activates when session count > 10k
- Track last activity time for each session
- Double-check idle status before cleanup to prevent race conditions

## Testing
- Added tests for idle session cleanup behavior
- Added tests for threshold-based activation
- Added tests for session activity tracking

This ensures the server remains stable even when clients don't properly close their sessions.


ochafik:
> long running tool calls for instance, so 30 min w/o activity may be perfectly fine / ditching these sessions would be a breaking change.

Other suggestions:
- Define idle as "no activity since N seconds *and* no pending request": while this would allow long running tools, it would still not allow patterns of, say, a client that's just listening to resource updates all day long (and reads said updated resources once a day to index them), so I probably wouldn't default to this either
- Define idle as "no open connection / stream since N seconds" (i.e. connection was closed and not resumed for a while). Would keep track of # open streams for a given connection (maybe in the try/finally of run_server). Probably the least breaking change

WDYT?

felixweinberger:
Related: https://github.com/modelcontextprotocol/python-sdk/pull/1159

maxisbey:
Closing due to inactivity. Related discussion continuing at #1159.

PR #1348, opened 2025-09-05
fix(examples): Prevent resource leak in simple-auth-client CallbackServer

pankaj-bind:
**Description**

This update addresses a potential resource leak in the `simple-auth-client` example. The `CallbackServer`, which handles the OAuth2 callback, was not guaranteed to be shut down if an exception occurred during the connection setup.

Specifically, if an exception was raised in the `connect` method after `callback_server.start()` was called but before the main logic was entered, the `finally` block that called `callback_server.stop()` would never be reached. This would leave the HTTP server running in a background thread, consuming resources.

**The Fix**

The `connect` method in `mcp_simple_auth_client/main.py` has been refactored to use a single, top-level `try...finally` block. This ensures that `callback_server.stop()` is executed regardless of whether the connection process succeeds or fails with an exception.

This change makes the example client more robust and reliable by guaranteeing proper resource cleanup.

**Type of change**

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

### DIFF
--- a/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
+++ b/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
@@ -132,7 +132,6 @@ class CallbackServer:
             self.thread.join(timeout=1)
         print("🛑 Stopped callback server.")
 
-
     def wait_for_callback(self, timeout=300):
         """Wait for OAuth callback with timeout."""
         start_time = time.time()


examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
```diff
@@ -158,19 +159,15 @@ def __init__(self, server_url: str, transport_type: str = "streamable_http"):
...
             callback_server.start()
 
             async def callback_handler() -> tuple[str, str | None]:
                 """Wait for OAuth callback and return auth code and state."""
                 print("⏳ Waiting for authorization callback...")
-                try:
-                    auth_code = callback_server.wait_for_callback(timeout=300)
-                    return auth_code, callback_server.get_state()
-                finally:
-                    callback_server.stop()
```
pcarleton, line 173: we also want to stop it here though, so we're not holding on to the port.  Agree it's better to have it in the finally as well, but I think we want it both places since we don't need the callback server after the callback is complete.

felixweinberger:
Closing for now due to inactivity as we're managing our inbound. Feel free to ping here if you plan to come back to this!

PR #1381, opened 2025-09-19
Add comprehensive Unicode tests for streamable HTTP transport

pja-ant:
## Motivation and Context
We've seen quite a few issues related to Unicode handling and while things appear to work, I noticed that we don't actually have tests for this (there's an example server though). Thought it would be useful to have a test to point people to if they expect Unicode handling isn't working and just ensure it doesn't regress.

## How Has This Been Tested?
Run the new test

## Breaking Changes
N/A

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/tests/client/test_http_unicode.py
+++ b/tests/client/test_http_unicode.py
@@ -1,0 +1,247 @@
+"""
+Tests for Unicode handling in streamable HTTP transport.
+
+Verifies that Unicode text is correctly transmitted and received in both directions
+(server→client and client→server) using the streamable HTTP transport.
+"""
+
+from collections.abc import Generator
+
+import pytest
+
+from mcp.client.session import ClientSession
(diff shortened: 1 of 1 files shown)


tests/client/test_http_unicode.py
```diff
@@ -0,0 +1,247 @@
...
+        s.bind(("127.0.0.1", 0))
+        return s.getsockname()[1]
+
+
+@pytest.fixture
+def running_unicode_server(unicode_server_port: int) -> Generator[str, None, None]:
+    """Start a Unicode test server in a separate process."""
+    import multiprocessing
+    import socket
+    import time
```
felixweinberger: nit: I think most of these imports could be moved to the top of the file (I see the note on `run_unicode_server` though), we seem to import the same packages multiple times

PR #1389, opened 2025-09-23
Use streamable-http consistently in examples

### LINKED ISSUE #1386: Transport type naming inconsistency: `streamable_http` in example vs `streamable-http` in spec and other SDKs
### Initial Checks

### Description

In the MCP specification, the transport name is defined as **“Streamable HTTP”**.  
In the Python SDK server code (e.g., `fastmcp/server.py`), the string identifier is `streamable-http` (with a hyphen), which matches many clients.  
However, in the `simple-auth-client` example, the identifier is written as `streamable_http` (with an underscore). This discrepancy can confuse developers, and some clients (e.g., Continue) only accept the hyphenated form (`streamable-http`).  

**Problem**  
Following the example as-is leads developers to use `streamable_http`, which does not work with some clients.  

**Expected behavior**  
Please align SDK/examples with one clear approach and document it explicitly:  

- Option A (accept both): Allow both `streamable-http` and `streamable_http` as input (lenient parsing), but document `streamable-http` as the canonical form.  
- Option B (unify on one): Standardize on `streamable-http` as the only valid transport string in examples/docs, and raise an error/warning if `streamable_http` is used.  

This would reduce developer confusion and ensure consistency with the spec and major client expectations.  

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
**Environment**  
- Python: 3.13 (project requires Python >= 3.13)  
- MCP Python SDK:  
  - fastmcp 2.10.1  
  - mcp 1.10.1
```

maxisbey:
Fixes #1386

Updated simple-auth examples to use `streamable-http` (hyphenated) instead of `streamable_http` (underscored) for consistency with the MCP specification and server implementation.

### DIFF
--- a/examples/clients/simple-auth-client/README.md
+++ b/examples/clients/simple-auth-client/README.md
@@ -71,4 +71,4 @@ mcp> quit
 ## Configuration
 
 - `MCP_SERVER_PORT` - Server URL (default: 8000)
-- `MCP_TRANSPORT_TYPE` - Transport type: `streamable_http` (default) or `sse`
+- `MCP_TRANSPORT_TYPE` - Transport type: `streamable-http` (default) or `sse`
--- a/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
+++ b/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
@@ -150,7 +150,7 @@ class CallbackServer:
 class SimpleAuthClient:
     """Simple MCP client with auth support."""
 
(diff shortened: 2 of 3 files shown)


Kludex:
Why is this auth team included in most PRs... ?

felixweinberger:
cc: @pcarleton for a quick stamp :)

> Why is this auth team included in most PRs... ?

I'm guessing the CODEOWNERS rule catches anything that has `auth` in the name? I'll see if that can be improved.

PR #1392, opened 2025-09-24
fix: wrapped annotations handling in func_metadata

### LINKED ISSUE #1391: wrapped tools using annotations fail
### Initial Checks

### Description

If you try to use a tool a which is wrapped and uses `from __future__ import annotations` setup fails with and `InvalidSignature` exception e.g.
```
mcp.server.fastmcp.exceptions.InvalidSignature: Unable to evaluate type annotation ForwardRef("Literal['literal'] | None")
```

The cause of this is that the global name space used for the lookup is the wrapper namespace and not the original function namespace.

### Example Code

```Python

import logfire

@logfire.instrument
def test_tool(location_type: Literal["literal"] | None = None) -> str:
    """Test tool."""
    return f"a is {location_type}"

mcp = FastMCP(
    "Wrap test",
    tools=[
        Tool.from_function(test_tool),
    ],
)
```

### Python & MCP Python SDK

```Text
SDL: 1.14.1
Python: 3.13.7
```

stevenh:
Use the original function's __globals__ for type hint resolution when dealing with wrapped functions. This ensures that any type hints defined in the original function's module are correctly resolved.

This also includes adding common typing names for resiliency.

Fixes #1391

## Motivation and Context
Fixes tool handling for wrapped methods that use annotations.

## How Has This Been Tested?
Test is included.

## Breaking Changes
This should be fully backwards compatible.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Tries to ensure success even if standard typing types are missing.

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -1,8 +1,11 @@
+import functools
 import inspect
 import json
+import sys
+import typing as t
 from collections.abc import Awaitable, Callable, Sequence
 from itertools import chain
-from types import GenericAlias
+from types import GenericAlias, MethodType, ModuleType
 from typing import Annotated, Any, ForwardRef, cast, get_args, get_origin, get_type_hints
 
 import pydantic_core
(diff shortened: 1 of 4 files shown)


stevenh:
Thanks @maxisbey will do that.

stevenh:
Is this what you had in mind @maxisbey ?

maxisbey:
> Is this what you had in mind @maxisbey ?

Hmm actually I was thinking that to avoid having to eval stuff.

But, I now realise that also calls `eval`. So, I think we could actually replace all of this with just `inspect.signature(call, eval_str=True)`, and we can then delete the original `_get_typed_signature` function. Sorry for the back and forth but at least this is much simpler :)

maxisbey:
Closing due to inactivity.

stevenh:
I can only guess this is a bot?

maxisbey:
Sorry for the confusion, I'm not a bot! Was just running through all the old PRs with Claude Code to close as many stale ones as we could to make the queue a bit more manageable before the holidays. If you're wanting to continue this then please feel free to re-open it and make the changes I suggested above and rebase onto main.

PR #1397, opened 2025-09-24
Validate metadata keys

samchenatti:
This PR validates `types.RequestParams.Meta` to enforce the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta) for the `_meta` field.

## Motivation and Context
This change ensures that users won't rely on invalid `key` names when using the `_meta` field.

## How Has This Been Tested?
It was tested in a local client/server setup.

Unit-tests were also added to ensure the correct behavior.

## Breaking Changes
There are breaking changes in the case some users are already using invalid `_meta` keys.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
It is the result of discussion in https://github.com/modelcontextprotocol/python-sdk/pull/1231

Although **I'm not entirely sure we should enforce this validation through the SDK**, I'm offering this approach as a possible option.

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -1,7 +1,8 @@
+import re
 from collections.abc import Callable
 from typing import Annotated, Any, Generic, Literal, TypeAlias, TypeVar
 
-from pydantic import BaseModel, ConfigDict, Field, FileUrl, RootModel
+from pydantic import BaseModel, ConfigDict, Field, FileUrl, RootModel, model_validator
 from pydantic.networks import AnyUrl, UrlConstraints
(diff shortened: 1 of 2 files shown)


Kludex:
It seems to be not backwards compatible to actually validate this.

Kludex:
I've commented it on discord: https://discord.com/channels/1358869848138059966/1399986970851282954/1471149499379613790, but I'm inclined to close this. Let's see.

Kludex:
I think the code complexity doesn't justify the addition. We can revisit this if someone disagrees (but I'm pretty sure we can simplify the code).

PR #1398, opened 2025-09-25
Asynchronous Tool Execution

LucaButBoring:
This PR implements the required changes for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1391, which adds asynchronous tool execution.

**Not yet fully-implemented (WIP)**
- Elicitation/sampling support
- Configuring immediate return values
- Cancelled status
- Documentation

This is a large PR, and I expect that if the associated SEP is accepted, we might want to break this down into several smaller PRs for SDK reviewers. I tried to generally have separate commits for each step of the implementation, to try and make this easier to review in its current form.

## Motivation and Context
Today, most applications integrate with tools in a straightforward but naive manner, choosing to have agents invoke tools synchronously with the conversation instead of allowing agents to multitask where possible. There are a few reasons why we believe this is the case, including the lack of clarity around tool interfaces (single tool or multiple for job tracking), model failures when manually polling on operations, and not having a way to retrieve results with a well-defined TTL, among other problems (described in more detail in the linked issue). Here, we introduce an alternative API that establishes a clear integration path for async job-style use cases that are typically on the order of minutes to hours.

The ultra high-level overview is as follows:
- Tools now support synchronous or asynchronous invocation modes
- A single tool only advertises itself as _either_ sync or async to a given client, controlled by protocol version
- Sync tools behave just like they always did
- Async tools are split into start/poll/retrieve stages:
  - Starting a call:
    - `tools/call` begins an async tool call
    - The result is a `CallToolResult` containing an **operation token**, which is used to interact with the async tool call across multiple RPC calls
  - Polling:
    - The operation token is used to call `tools/async/status`, which returns the current operation status
    - The client should poll this method until the status reaches a terminal value
  - Result retrieval:
    - The operation token is used to call `tools/async/result`, which has the final tool output

Whether a tool is sync, async, or both (on old/new protocol versions) is defined by tool implementors. This enables remote server operators to control this based on how long each tool is expected to take to execute, rather than potentially serving HTTP requests with widely varying execution times on the same endpoint. This also makes it much more clear to client applications what the "time contract" of a tool is, so that fast tools can still be executed synchronously while allowing long-running tools to be immediately backgrounded.

## How Has This Been Tested?
Unit tests, integration tests, and new example snippets.

## Breaking Changes
Existing users will not need to update their applications to continue using synchronous tool calls. Asynchronous tool calls will require minor code changes that will be documented.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
There were a bunch of decisions in the implementation we may want to discuss further, some of which were due to ambiguity in the proposal (which will be revised again) and some of which were due to working things into the SDK implementation.

- I added a faux-version called `next` to deal with the requirement that sessions on the current-latest version always advertise as sync-only. The tests and examples explicitly set the advertised client protocol version to `next` when calling async-only tools.
- Reusing the existing `tool/call` method creates some ambiguities in how `outputSchema` should be handled, as the immediate tool call result (communicating an accepted state) would no longer have meaningful `structuredContent`. The output that should be validated is actually the result of `GetOperationPayloadResult`, so for now I'm skipping validation of the immediate `CallToolResult` (only in async execution) and only validating `GetOperationPayloadResult` (sync tool executions are always validated, just like before).
- `keepAlive` should have a sentinel value representing "no expiration," and I'm leaning towards `None`. However, in SDK implementations, that becomes somewhat ambiguous with sync tool calls, which also implicitly have a `keepAlive` of `None` already. For now, I default it to 1 hour if not specified/`None`, but this should probably be changed before this is merged.
- In sHTTP, the SDK has behavior to send tool-related server messages on the same SSE stream that the server used as a response to the client's `CallToolRequest`, by attaching a `related_request_id` to the stream for fast lookups and session resumption. To support sampling and elicitation, we'll need some alternative way to key a stream by an async operation token for async tool calls (WIP).
- The client session needs to cache a mapping of in-flight operation tokens to tool names for validating `structuredContent` in async tool calls, as it otherwise has no way to look up the cached `outputSchema`. We could consider including a `toolName` in `GetOperationPayloadResult` to avoid the inconvenience, but in this draft I'm using a cache expiry based on `keepAlive` to avoid holding that mapping forever.


LucaButBoring:
Latest [commit](https://github.com/modelcontextprotocol/python-sdk/pull/1398/commits/b802dc423bc74458830bfdcb3626ac55ca77bcaf) implements working stream binding to support elicitation and sampling. There's definitely room for refactoring that one 😔

LucaButBoring:
Added support for configuring the immediate result and made some changes to avoid smuggling parameters through `_meta` (now we do it through unserialized fields).

src/mcp/server/lowlevel/server.py
```diff
@@ -498,46 +507,77 @@ async def handler(req: types.CallToolRequest):
...
+
+                                # Process results using shared logic
+                                result = self._process_tool_result(results, tool)
+                                self.async_operations.complete_operation(operation.token, result)
+                                logger.debug(f"Completed async operation {operation.token}")
+                            except Exception as e:
+                                logger.exception(f"Async execution failed for {tool_name}")
+                                self.async_operations.fail_operation(operation.token, str(e))
+
+                        asyncio.create_task(execute_async())
```
Kludex: Let's stop using `asyncio` in this code source. Use `anyio` please.

Kludex:
Why would a tool have 2 flavors (async & sync)? 

Is there a hint to the client when it should try to retrieve the result, e.g. what `retry-after` is for HTTP.

LucaButBoring:
> Is there a hint to the client when it should try to retrieve the result, e.g. what `retry-after` is for HTTP.

As in for how often it should poll? There is not, but that's a good callout - I'll amend the proposal to include that.

LucaButBoring:
> Why would a tool have 2 flavors (async & sync)?

In the case where a tool is being migrated from sync to async, and may support clients that haven't yet supported the latest (or next, in this case) protocol version, it's useful to support both on the same tool - so clients that don't support the latest version will have a single long request, while clients that do support it will use async tool call semantics.

There will be some use cases where that's desirable, and some where it's not, so it's optional behavior.

src/mcp/server/lowlevel/server.py
LucaButBoring: Fixed here, still need to fix `AsyncOperationManager`. Still trying to figure out how to do that properly with cancellation scopes.

Kludex: I can help if you need help.

Kludex:
If the choice of using async/sync is made by the client, why should we add flavors when defining a tool on the server side?

Kludex:
> If the choice of using async/sync is made by the client, why should we add flavors when defining a tool on the server side?

Answering my own question... You need to signal in the tool definition which flavor it supports.

Kludex:
Okay. I don't think we should be adding multiple flavors to a tool, I think it makes things more complicated. Also, the spec anyway supports the `invocationMode` keyword, and not `invocationModes`, which I think it makes sense.

I think a better API is to have a new method e.g. `async_tool` (although "async" is not the best keyword for this in Python). Is it still possible to find a synonym to "async"? "long-run"? 😅

src/mcp/types.py
```diff
@@ -898,11 +938,81 @@ class ListToolsResult(PaginatedResult):
...
+    """Number of seconds the client wants the result to be kept available upon completion."""
+    model_config = ConfigDict(extra="allow")
+
+
+class AsyncResultProperties(BaseModel):
+    """Properties for async tool execution results."""
+
+    token: str
+    """Server-generated token to use for checking status and retrieving results."""
+    keepAlive: int
```
Kludex, line 954:
```suggestion
    keep_live: int = Field(alias="keepAlive")
```

src/mcp/types.py
```diff
@@ -898,11 +938,81 @@ class ListToolsResult(PaginatedResult):
...
+    result: "CallToolResult"
+    """The result of the tool call."""
+
+
 class CallToolRequestParams(RequestParams):
     """Parameters for calling a tool."""
 
     name: str
     arguments: dict[str, Any] | None = None
+    operation_params: AsyncRequestProperties | None = Field(serialization_alias="operation", default=None)
```
Kludex, line 1014: What's the problem with the real name "operation"?

src/mcp/server/lowlevel/server.py
```diff
@@ -498,46 +506,87 @@ async def handler(req: types.CallToolRequest):
...
+                                # Process results using shared logic
+                                result = self._process_tool_result(results, tool)
+                                self.async_operations.complete_operation(operation.token, result)
+                                logger.debug(f"Completed async operation {operation.token}")
+                            except Exception as e:
+                                logger.exception(f"Async execution failed for {tool_name}")
+                                self.async_operations.fail_operation(operation.token, str(e))
+
+                        async with anyio.create_task_group() as tg:
+                            tg.start_soon(execute_async)
```
Kludex: I think we need to make this implementation a bit more abstract.

What if I want to execute my tasks in another machine/node/process than the MCP server? The world seems to like a lot durable execution products like Temporal as well.

I think we need something like what this design proposes: https://github.com/pydantic/fasta2a/tree/main?tab=readme-ov-file#design

src/mcp/types.py
LucaButBoring: The base `RequestParams` type also has a `_operation` used for association metadata, which needs to be aliased to `operation` to not be treated as a private/protected field by pyright wherever it gets used.

```
error: "_operation" is protected and used outside of the class in which it is declared (reportPrivateUsage)
```

I didn't want to ignore pyright just because of a naming conflict, but now that I'm looking at this again there are only 3 places where we'd need to do so.

src/mcp/server/lowlevel/server.py
LucaButBoring: I made `AsyncOperationManager` a `Server` parameter with the intention that consumers would implement something like that for that use case, but that's not enough due to missing a broker to enable that pattern, yeah. Will adjust this accordingly.

LucaButBoring:
> I think a better API is to have a new method e.g. `async_tool` (although "async" is not the best keyword for this in Python). Is it still possible to find a synonym to "async"? "long-run"? 😅

On the SEP, we're going to be moving to referring to these uniformly as long-running operations, so an `@mcp.long_running` would work.

Do we want to force tool implementors to make a clean break from existing tools, though? Or are you suggesting that in the case where both modes are desired on a tool in an interim state, they would add both `@mcp.tool` and `@mcp.long_running` to the same function?

If so, that then requires the tool manager to support adding the same tool more than once, and updating the parameters if `long_running` is present and a tool is already registered (since the decorators could be applied in either order).

Kludex:
> Do we want to force tool implementors to make a clean break from existing tools, though?

As for what I understood, since the `invokeMode` is a string, you can only have a tool that is either "long-running" or "short-running". If that's the case, then a short running tool is different from a long-running tool. So... I'm not sure if I'm missing something, or the spec reflects something different.

> Or are you suggesting that in the case where both modes are desired on a tool in an interim state, they would add both @mcp.tool and @mcp.long_running to the same function?

I wouldn't like to add two decorators. 🤔

---

If a tool can be short, and long-running, why the `invokeMode` is not a boolean? `supportsAsync`?

LucaButBoring:
> As for what I understood, since the `invokeMode` is a string, you can only have a tool that is either "long-running" or "short-running". If that's the case, then a short running tool is different from a long-running tool. So... I'm not sure if I'm missing something, or the spec reflects something different.

A tool is exactly one or the other from the perspective of a single client. ListTools must show exactly one execution mode. In addition, version negotiation is used to hide long-running tools from clients that don't support them, yet.

However, we can still support hybrid tools for backwards-compatibility purposes if the server operator allows it. Essentially, because the core tool implementation is the same `@mcp.tool`, we can wrap a tool in the functionality it needs to behave in either way within the server SDK, but only allow a client to use one or the other depending on its version.

There's a few sections of the SEP which discuss this, but pulling one I think is useful:

```
Old Clients (pre-async support):
// tools/list response (filtered)
{
  tools: [
    // only supports sync execution
    { name: "search_web", description: "Search the web" },
    // only supports sync execution
    { name: "quick_calc", description: "Fast calculation" },
    // supports both sync and async - invocationMode field is hidden from old clients
    { name: "get_weather", description: "Get weather" }
    // async-capable tools hidden from old clients
  ]
}

New Clients (async support):
// tools/list response (complete)
{
  tools: [
    // explicitly only supports sync execution
    { name: "search_web", description: "Search the web", invocationMode: "sync" },
    // implicitly only supports sync execution
    { name: "quick_calc", description: "Fast calculation" },
    // new clients see the async invocation mode and should assume async-only execution
    { name: "get_weather", description: "Get weather", invocationMode: "async" },
    // tool is async-only, and is not shown to old clients
    { name: "deep_analysis", description: "Complex analysis", invocationMode: "async" }
    // All tools visible, async capabilities declared
  ]
}
```

Note that some tools are present in the "old clients" list despite presenting as LRO-only in the "new clients" list. When an old client invokes a tool represented like this, they won't get the `immediate_result` response, or even an operation token - the server will simply not return a response until the tool implementation completes.

> If a tool can be short, and long-running, why the `invokeMode` is not a boolean? `supportsAsync`?

This was the original idea, actually, but for futureproofing it's been changed to an enum.

src/mcp/server/lowlevel/server.py
LucaButBoring: Finally dealt with this, resolving thread.

src/mcp/server/lowlevel/server.py
LucaButBoring: Split this into an `AsyncOperationStore` and `AsyncOperationBroker` in the latest commit - a dedicated distributable worker would be nice, but I don't think it makes much sense here, given that the server itself is the only thing that can meaningfully execute tool calls right now (or alternatively: the server fulfills all requirements of a worker as of now).

Open to further feedback on this - I'm particularly unhappy with `RequestContext`. It isn't entirely serializable, so in the example implementation with SQLite, I split off the serializable pieces into a base class and restored what I could, leaving the server session, request object, and lifespan out. I'm not sure if there's some way to hydrate those at runtime, but for the example, they weren't strictly necessary.

LucaButBoring: I'm working on filling out `RequestContext` support further, to handle elicitation and sampling across restarts with sHTTP. That's turning out to be considerably more complicated as the entire session assumes there are request/response streams established.

I'm chipping away at a solution involving creating mock streams in the task dispatcher that feed into `EventStore` to deal with this, and "reattaching" the mock streams to real HTTP streams when the client reconnects after a server restart. It'll likely take a bit to get that working, but just a heads-up.

I haven't tried a similar resumption example on non-LRO tools yet, but I actually suspect there are similar problems with resumability of elicitation/sampling on `main`, too - there are a few maps of request IDs to streams in the core session logic and the sHTTP transport which I don't think would get restored properly today for server-to-client requests, if someone just plugged in a persistent `EventStore` right now. Once I get this working here, I'll likely extract those changes to a separate PR with a non-LRO SQLite example to both confirm the problem and simplify this PR a bit.

LucaButBoring: Nevermind, I wound up not using `EventStore` at all for this, I instead introduced a dedicated `OperationEventQueue` type which is more suitable, given that this part is transport-agnostic.

This was quite a lot of work, but this PR now has support for a complete `RequestContext`, with elicitation and sampling working across restarts of both the client and server.

LucaButBoring:
Alright, there's a lot of small remaining things to address here, particularly around naming (and lots of debugging cruft to clean up), but the SQLite example with both server and client restarts with elicitation now works. Regardless of the exact SEP accepted for LROs, I think there's a useful core implementation here that is likely generalizable. The operation token is essentially used as a routing key for task messages, but as long as there's some globally-unique key, I think something like this would work in several LRO models.

Issue #1405, opened 2025-09-26
It is not possible to define a Resource that takes only the Context parameter as an argument.


0Delta:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

## Problem

Currently, it is not possible to define a Resource that takes only the Context parameter as an argument. When attempting to create such a resource, the registration logic incorrectly classifies it as a template resource instead of a regular resource.

## Expected Behavior

Resources with only a Context parameter should be registered as regular resources, not template resources.

## Current Behavior

Resources with only a Context parameter are incorrectly registered as template resources, causing them to not appear in the resource list and fail when accessed.

## Root Cause

The parameter validation logic in the `FastMCP.resource()` decorator doesn't properly distinguish between effective function parameters and Context parameters. It treats any function with parameters as a potential template resource, even when the only parameter is a Context.

## Solution

1. Fix the parameter validation logic to exclude Context parameters when determining if a resource should be a template
2. Update the `Resource.read()` method signature to accept an optional context parameter
3. Update all resource implementations to support the new signature

This change maintains full backward compatibility while enabling context-only resources.

### Example Code

```Python
@mcp.resource("resource://user_profile")
def get_user_profile(ctx: Context) -> str:
    # Something Code ...
```

### Python & MCP Python SDK

```Text
Python 3.13.5
MCP Python SDK v1.15.0
```

0Delta:
I have created a PR to fix this issue, so please review it.

hackthemarket:
agree this would be a very nice addition to the sdk, thank you for raising it and providing a pr, @0Delta !

PR #1406, opened 2025-09-26
Add support for context-only resources (#1405)

0Delta:
Add support for context-only resources 

## Motivation and Context
#1405

Currently, it is not possible to define a Resource that takes only the Context parameter as an argument. When attempting to create such a resource, the registration logic incorrectly classifies it as a template resource instead of a regular resource.

This limitation prevents developers from creating resources that need access to server session context but don't require dynamic URI components, reducing flexibility in resource design.

## How Has This Been Tested?

- [x] Added comprehensive unit tests for context-only resource functionality
- [x] Added integration tests to verify correct resource registration
- [x] All existing resource tests continue to pass (22/22 tests passed)
- [x] Tested with example context-only resource implementation
- [x] Verified backward compatibility with existing resource definitions

## Breaking Changes

No breaking changes. The `context` parameter in `Resource.read()` method is optional with default `None`. All existing resource implementations continue to work unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/snippets/servers/context_resource.py
+++ b/examples/snippets/servers/context_resource.py
@@ -1,0 +1,11 @@
+from mcp.server.fastmcp import Context, FastMCP
+from mcp.server.session import ServerSession
+
+mcp = FastMCP(name="Context Resource Example")
+
+
+@mcp.resource("resource://only_context")
+def resource_only_context(ctx: Context[ServerSession, None]) -> str:
+    """Resource that only receives context."""
+    assert ctx is not None
+    return "Resource with only context injected"
--- a/src/mcp/server/fastmcp/prompts/base.py
+++ b/src/mcp/server/fastmcp/prompts/base.py
@@ -99,12 +99,12 @@ class Prompt(BaseModel):
 
         # Find context parameter if it exists
         if context_kwarg is None:  # pragma: no branch
-            context_kwarg = find_context_parameter(fn)
+            context_kwarg = find_context_parameter(fn) or ""
 
         # Get schema from func_metadata, excluding context parameter
         func_arg_metadata = func_metadata(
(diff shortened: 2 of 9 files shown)


felixweinberger (changes requested):
Hi @0Delta thank you for this contribution - are you running into this limitation with a concrete application you're trying to implement? If so would you be able to provide some additional details as a motivating example for this change?

0Delta:
Hi @felixweinberger.
Specifically, I wanted to define a simple resource like the one below and display logs within it.

```python
mcp = FastMCP()


@mcp.resource("resource://user_profile")
def get_user_profile(ctx: Context) -> str:
    ctx.info("Fetching user profile")
    return "Profile data for user..."
```

When I tried to display this in Inspector, Inspector mistakenly thought it needed arguments and wouldn't run.

image

0Delta:
This resource should be recognized as just a Resource, not a template, right?

0Delta:
@felixweinberger 
Do you have any plans for when it will be merged?
Please let me know if there is anything else I can do to help.

bskinn:
I've had success working around this by:

1. Adding an `{ignored}` template parameter as a URI path element.
2. Adding an `ignored` function parameter to match.
3. Putting a note in the docstring that the parameter is currently needed due to an implementation quirk and that MCP clients should just always pass zero to it.

Cline has consistently followed that instruction in the docstring, and it's been behaving just like a concrete resource.

maxisbey:
Thanks for this — closing as stale. This PR targets the old `fastmcp/` directory structure which was renamed, and the same fix is being tracked in #1666 (which is rebased on the current code and fixes #1405).

PR #1404, opened 2025-09-26
fix: set default to none for optional field without default value

vincent0426:
## Motivation and Context

Closes https://github.com/modelcontextprotocol/python-sdk/issues/1402

pydantic treats Optional type as required if there is no default value
```python
import json
from pydantic import BaseModel, Field
from typing import Optional


class TestModel(BaseModel):
    param_1: str | None = Field(description="Optional without default")
    param_2: Optional[str] = Field(description="Optional without default")

schema = TestModel.model_json_schema()
print("Pydantic schema:")
print(json.dumps(schema, indent=2))
```

output
```
Pydantic schema:
{
  "properties": {
    "param_1": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "description": "Optional without default",
      "title": "Param 1"
    },
    "param_2": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "description": "Optional without default",
      "title": "Param 2"
    }
  },
  "required": [
    "param_1",
    "param_2"
  ],
  "title": "TestModel",
  "type": "object"
}                
```

## How Has This Been Tested?

Unit tested on 5 scenarios
- required
- optional with `type | none`
- optional with `Optional[type]`

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1408, opened 2025-09-27
Fix optional parameters in tools handling

### LINKED ISSUE #1402: SDK does not handle optional parameters in tools properly
### Initial Checks

### Description

When I run the example code below there are couple of issues I see in MCP inspector regarding the tools/list call
Pasting the response here for clarity
```
{
  "tools": [
    {
      "name": "sample_tool_with_params",
      "description": "A tool with parameters",
      "inputSchema": {
        "type": "object",
        "properties": {
          "param1": {
            "description": "Normal Required Param",
            "title": "Param1",
            "type": "string"
          },
          "param2": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "description": "Normal Param Defined as Optional without a default value in Pydantic Field",
            "title": "Param2"
          },
          "param3": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "description": "Normal Param Defined as Optional with a default value in Pydantic Field",
            "title": "Param3"
          }
        },
        "required": [
          "param1",
          "param2"
        ],
        "title": "sample_tool_with_paramsArguments"
      }
    }
  ]
}
```

Issues:
* If I mark a tool param as optional(like param2 in the example provided), but do not put default value as none in pydantic field, it still ads it to the required fields list indicating the SDK is using the default value to decide which parameters are required and optional instead of the parameter type
* For parameters which are marked as optional, the type is set as part of an "anyOf" array with type being either string or null in this case. This behaviour is different from expectation as per spec and also the typescript sdk where we rely only on the required array to decide the required params. For example param3 has both type set as Optional[str] and has default value set to none, so it does not come in the required list, but when making the actual tool call it forces the client to send the value of param3 as `null` instead of ommiting the param all together. Attaching tool\call request and response below:
Request:
```
{
  "method": "tools/call",
  "params": {
    "name": "sample_tool_with_params",
    "arguments": {
      "param1": "abc",
      "param2": "def",
      "param3": null
    },
    "_meta": {
      "progressToken": 1
    }
  }
}
```

Response:
```
{
  "content": [
    {
      "type": "text",
      "text": "{\n  \"param1\": \"abc\",\n  \"param2\": \"def\",\n  \"param3\": null\n}"
    }
  ],
  "isError": false
}
```

I have verifed this behaviour with the typescript sdk specifically with the exaple server [here](https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem)
Here you can look at `read_text_file` tool for how optional parameters are handled. Attaching an excerpt from tools/list to show how it behaves
```
{
      "name": "read_text_file",
      "description": "Read the complete contents of a file from the file system as text. Handles various text encodings and provides detailed error messages if the file cannot be read. Use this tool when you need to examine the contents of a single file. Use the 'head' parameter to read only the first N lines of a file, or the 'tail' parameter to read only the last N lines of a file. Operates on the file as text regardless of extension. Only works within allowed directories.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "path": {
            "type": "string"
          },
          "tail": {
            "type": "number",
            "description": "If provided, returns only the last N lines of the file"
          },
          "head": {
            "type": "number",
            "description": "If provided, returns only the first N lines of the file"
          }
        },
        "required": [
          "path"
        ],
        "additionalProperties": false,
        "$schema": "http://json-schema.org/draft-07/schema#"
      }
    }
```

### Example Code

```Python
from pydantic import Field
from typing import Optional



# Create MCP server instance
mcp = FastMCP()

@mcp.tool()
def sample_tool_with_params(
    param1: str = Field(description="Normal Required Param"),
    param2: Optional[str] = Field(description="Normal Param Defined as Optional without a default value in Pydantic Field"),
    param3: Optional[str] = Field(default=None, description="Normal Param Defined as Optional with a default value in Pydantic Field")
) -> dict:
    "A tool with parameters"
    return {
        "param1": param1,
        "param2": param2,
        "param3": param3
    }

    


def main():
    """Run the MCP server"""

if __name__ == "__main__":
    main()
```

### Python & MCP Python SDK

```Text
Python: 3.13.5
MCP SDK: 1.14.1
```

PreistlyPython:
## Summary
Fixes #1402 - Optional parameters without explicit defaults are now correctly treated as optional in JSON schemas, aligning Python SDK behavior with the TypeScript SDK.

## Problem
Pydantic treats `Optional[T]` and `T | None` as **required** if no default value is set, even though the type annotation indicates the parameter should be optional. This caused:
- Optional parameters incorrectly marked as required in JSON schemas
- Inconsistency with TypeScript SDK behavior
- Confusing developer experience when defining tools

## Solution
This PR uses a **Pydantic-native approach** (inspired by PR #1404) that checks if a parameter's type annotation is a Union containing None, and sets `default=None` on the FieldInfo if no explicit default exists.

### Implementation (6 lines of code)
```python
# If it's Optional[SomeType] or SomeType | None and has no explicit default, set default=None
origin = get_origin(annotation)
if (origin is Union or origin is UnionType) and type(None) in get_args(annotation):
    if isinstance(param.default, FieldInfo) and param.default.default is PydanticUndefined:
        param.default.default = None
```

This approach:
- ✅ Uses Pydantic's native Union/UnionType detection
- ✅ Avoids complex type introspection
- ✅ Minimal code changes (6 lines vs 61 in previous version)
- ✅ Aligns with maintainer feedback

### Before Fix
```python
def my_tool(
    required: str = Field(description="Required param"),
    optional: str | None = Field(description="Optional param"),
) -> str:
    return f"{required}|{optional}"

# Generated schema (WRONG):
{
    "required": ["required", "optional"],  # ❌ optional is marked required!
    "properties": {
        "optional": {"anyOf": [{"type": "string"}, {"type": "null"}]}
    }
}
```

### After Fix
```python
# Generated schema (CORRECT):
{
    "required": ["required"],  # ✅ Only required params
    "properties": {
        "optional": {
            "anyOf": [{"type": "string"}, {"type": "null"}],
            "default": None  # ✅ Proper default
        }
    }
}
```

## Changes
- **Modified**: `src/mcp/server/fastmcp/utilities/func_metadata.py` (+8 lines)
  - Import `Union` and `UnionType` 
  - Add 6-line check for optional types
- **New**: `tests/server/fastmcp/test_func_metadata_optional.py` (+132 lines)
  - 5 comprehensive test cases
  - Coverage for all optional patterns

## Testing
All tests pass:
- ✅ `test_optional_parameters_not_required` - Verifies optional params not in required list
- ✅ `test_required_parameters_still_required` - Ensures required params unchanged
- ✅ `test_optional_with_explicit_none_default` - Tests explicit None defaults
- ✅ `test_mixed_optional_types` - Various Union patterns (str | None, str | int | None)
- ✅ `test_runtime_behavior_with_optional_params` - Runtime validation works correctly
- ✅ All existing func_metadata tests pass (no regressions)

## Backward Compatibility
✅ **Fully backward compatible**:
- Parameters with explicit defaults unchanged
- Required parameters remain required
- No breaking changes to function semantics
- All existing tests pass

## Comparison with PR #1404
This PR uses the **exact same approach** as PR #1404 (by @vincent0426), addressing the maintainer's feedback about avoiding unnecessary type introspection. Both PRs:
- Use Pydantic-native Union detection
- Set `default=None` for FieldInfo with PydanticUndefined
- Minimal code changes
- Same test coverage goals

The key difference is that this PR was updated to align with the simpler approach after reviewer feedback.

**Response**: Agreed! The updated implementation removes all complex introspection and uses only Pydantic's native `get_origin()` and `get_args()` functions, which is the correct way to work with Pydantic types.

> "I also think we don't use Optional[T] anywhere in this code base"

**Response**: While the codebase may not use `Optional[T]` internally, many users do, and the modern `T | None` syntax (used in this codebase) has the same Pydantic behavior. This fix ensures both patterns work correctly.

Closes #1402

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -2,8 +2,8 @@ import inspect
 import json
 from collections.abc import Awaitable, Callable, Sequence
 from itertools import chain
-from types import GenericAlias
-from typing import Annotated, Any, ForwardRef, cast, get_args, get_origin, get_type_hints
+from types import GenericAlias, UnionType
+from typing import Annotated, Any, ForwardRef, Union, cast, get_args, get_origin, get_type_hints
 
 import pydantic_core
 from pydantic import (
@@ -231,6 +231,14 @@ def func_metadata(
                 # 🤷
                 WithJsonSchema({"title": param.name, "type": "string"}),
             ]
+
+        # If it's Optional[SomeType] or SomeType | None and has no explicit default, set default=None
+        # This ensures optional parameters are not marked as required in the JSON schema
+        origin = get_origin(annotation)
+        if (origin is Union or origin is UnionType) and type(None) in get_args(annotation):
+            # Check if param.default is a FieldInfo with no default set
+            if isinstance(param.default, FieldInfo) and param.default.default is PydanticUndefined:
+                param.default.default = None
 
         field_info = FieldInfo.from_annotated_attribute(
             _get_typed_annotation(annotation, globalns),
--- a/tests/server/fastmcp/test_func_metadata_optional.py
+++ b/tests/server/fastmcp/test_func_metadata_optional.py
@@ -1,0 +1,130 @@
+"""Tests for Optional parameter handling in func_metadata.
+
+This module tests the fix for issue #1402 where Optional[T] parameters
+without explicit defaults were incorrectly marked as required in JSON schemas.
+"""
+
+
+from pydantic import Field
+
+from mcp.server.fastmcp.utilities.func_metadata import func_metadata
+
+
+def test_optional_parameters_not_required():
+    """Test that Optional parameters without defaults are not marked as required."""
+
+    def func_with_optional_params(
+        required: str = Field(description="This should be required"),
+        optional: str | None = Field(description="This should be optional"),
+        optional_with_default: str | None = Field(default="hello", description="This should be optional with default"),
+        optional_with_union_type: str | None = Field(description="This should be optional"),
+        optional_with_union_type_and_default: str | None | None = Field(
+            default="hello", description="This should be optional with default"
+        ),
+    ) -> str:
+        return (
+            f"{required}|{optional}|{optional_with_default}|"
+            f"{optional_with_union_type}|{optional_with_union_type_and_default}"
+        )
+
+    meta = func_metadata(func_with_optional_params)
+    schema = meta.arg_model.model_json_schema()
+
+    # Only 'required' parameter should be in the required list
+    assert schema["required"] == ["required"]
+
+    # All optional parameters should have default=None
+    assert schema["properties"]["optional"]["default"] is None
+    assert schema["properties"]["optional_with_default"]["default"] == "hello"
+    assert schema["properties"]["optional_with_union_type"]["default"] is None
+    assert schema["properties"]["optional_with_union_type_and_default"]["default"] == "hello"
+
+
+def test_required_parameters_still_required():
+    """Test that non-optional parameters remain required."""
+
+    def func_with_required_params(
+        required_str: str = Field(description="Required string"),
+        required_int: int = Field(description="Required int"),
+        optional: str | None = Field(description="Optional string"),
+    ) -> str:
+        return f"{required_str}|{required_int}|{optional}"
+
+    meta = func_metadata(func_with_required_params)
+    schema = meta.arg_model.model_json_schema()
+
+    # Both required parameters should be in the required list
+    assert set(schema["required"]) == {"required_str", "required_int"}
+
+    # Optional parameter should have default=None
+    assert schema["properties"]["optional"]["default"] is None
+
+
+def test_optional_with_explicit_none_default():
+    """Test Optional parameters with explicit None default."""
+
+    def func_with_explicit_none(
+        optional_explicit: str | None = Field(default=None, description="Explicit None default"),
+        optional_implicit: str | None = Field(description="Implicit None default"),
+    ) -> str:
+        return f"{optional_explicit}|{optional_implicit}"
+
+    meta = func_metadata(func_with_explicit_none)
+    schema = meta.arg_model.model_json_schema()
+
+    # No parameters should be required
+    assert schema.get("required", []) == []
+
+    # Both should have None defaults
+    assert schema["properties"]["optional_explicit"]["default"] is None
+    assert schema["properties"]["optional_implicit"]["default"] is None
+
+
+def test_mixed_optional_types():
+    """Test various Optional type patterns."""
+
+    def func_with_mixed_optionals(
+        union_pipe: str | None = Field(description="str | None"),
+        union_optional: str | None = Field(description="Optional[str]"),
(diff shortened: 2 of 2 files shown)


ISSUE_1402_FIX.md
```diff
@@ -0,0 +1,220 @@
+# Fix for Issue #1402: Optional Parameters in Tools Handling
+
+## Summary
+
+This fix addresses Issue #1402 in the MCP Python SDK, which involved incorrect handling of optional parameters in tools. The SDK now properly detects optional parameters using type annotations and generates correct JSON schemas, aligning with TypeScript SDK behavior.
```
felixweinberger: Please remove verbose documentation about individual fixes and PRs - I assume this may have been LLM generated?

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -232,9 +275,24 @@ def func_metadata(
...
+        if param.default is not inspect.Parameter.empty:
+            # Parameter has an explicit default value
+            default_value = param.default
+        elif _is_optional_type(typed_annotation):
+            # Parameter is Optional[T] or Union[T, None] without explicit default
+            # Set default to None to make it truly optional in the schema
+            default_value = None
+        else:
+            # Required parameter
+            default_value = PydanticUndefined
```
felixweinberger: I might be wrong, but this feels like we might be using Pydantic incorrectly if we need to do this kind of type introspection.

Any thoughts on this @Kludex?

felixweinberger (changes requested):
Hi @PreistlyPython thanks for this contribution.

This seems to attempt to solve the same issue as https://github.com/modelcontextprotocol/python-sdk/pull/1404

Doing all this introspection and type based logic feels like something we shouldn't have to do if we're using the Pydantic tools correctly - maybe we are simply using them wrong?

I also think we don't use `Optional[T]` anywhere in this code base, so I'm not sure we want to add logic to support it specifically.

src/mcp/server/fastmcp/utilities/func_metadata.py
Kludex: I don't think adding a default value is needed here...

ISSUE_1402_FIX.md
PreistlyPython: Will do

PreistlyPython:
I've updated this PR to address your feedback @felixweinberger.

**Changes made**:
1. ✅ Removed all complex type introspection logic (deleted 43 lines of  function)
2. ✅ Adopted the simpler Pydantic-native approach from PR #1404
3. ✅ Now uses only  and  - standard Pydantic tools
4. ✅ Reduced from 61 lines to just 6 lines of core logic
5. ✅ Same test coverage with cleaner implementation

The updated approach is identical to PR #1404's method - checking if the type is a Union containing None, and setting  on the FieldInfo if needed. This feels much cleaner and aligns with how we should be using Pydantic.

All tests pass, including backward compatibility checks. Ready for re-review!

Kludex:
Same as https://github.com/modelcontextprotocol/python-sdk/pull/1404#issuecomment-3354955968.

PR #1413, opened 2025-09-29
Delete CODEOWNERS to eliminate notification overload

felixweinberger:
## Summary
Removes CODEOWNERS file that was causing unnecessary reviewer notification spam without providing actual merge protection.

## Problem
- Every PR triggered automatic review requests to 6+ team members
- No branch protection rules exist, so CODEOWNERS approval wasn't required
- Team members were getting spammed with notifications that didn't affect merge ability

## Solution
Delete CODEOWNERS entirely. Reviews will be handled through manual review requests based on PR content.


felixweinberger:
@pcarleton @ochafik FYI - based on offline conversation it sounds like we weren't gaining much value from these notifs in the first place, so going to try a different approach

Kludex (approved):

PR #1417, opened 2025-10-01
Stateful Streamable-HTTP: send 404 Not Found on unknown Mcp-Session-Id

### LINKED ISSUE #1004: Session ID not present in memory store should return 404, not 400
### Initial Checks

### Description

When a session ID is provided by a client, but it is not present in the in-memory session store for the server, the server returns a 400 https://github.com/modelcontextprotocol/python-sdk/blob/d0443a18328a2fc9e87da6feee9130f95c0b37a7/src/mcp/server/streamable_http_manager.py#L250-L256

This is a problem, because as noted in https://github.com/modelcontextprotocol/python-sdk/issues/880 sessions are not persisted between restarts

This library should return a 404, as in https://github.com/modelcontextprotocol/python-sdk/blob/d0443a18328a2fc9e87da6feee9130f95c0b37a7/src/mcp/server/streamable_http.py#L371 So that the client can re-initialize and get a new session ID as needed (as in https://github.com/mark3labs/mcp-go/blob/8f5b048218f6d044c3322f16b8cb0b08e10bf5d0/client/transport/streamable_http.go#L257). 

400 isn't applicable because the request isn't _bad_, but the server simply doesn't persist sessions between restarts.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
3.10.12 and 2.3.4
```

Zalewa:
## Motivation and Context
Fixes #1004

MCP protocol spec states that when Mcp-Session-Id arrives for a session that has been removed, the server **MUST** reply with a "404 Not Found" HTTP status code. It was replying with a "400 Bad Request" instead. The relevant part of the spec is here in point 2.5.3: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management

> The server MAY terminate the session at any time, after which it MUST respond to requests containing that session ID with HTTP 404 Not Found.

## How Has This Been Tested?
I wrote a unit test.

## Breaking Changes
User apps may be relying on the incorrect behavior of sending "400 Bad Request".
https://xkcd.com/927/

## Types of changes
- [x] Bug fix (~non-~potentially-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

## Additional context
https://github.com/modelcontextprotocol/python-sdk/pull/641/files#r2394152091

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -273,7 +273,7 @@ class StreamableHTTPSessionManager:
         else:
             # Invalid session ID
             response = Response(
-                "Bad Request: No valid session ID provided",
-                status_code=HTTPStatus.BAD_REQUEST,
+                "Not Found: Unknown session ID",
+                status_code=HTTPStatus.NOT_FOUND,
             )
             await response(scope, receive, send)
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -1,5 +1,6 @@
 """Tests for StreamableHTTPSessionManager."""
 
+from http import HTTPStatus
 from typing import Any
 from unittest.mock import AsyncMock, patch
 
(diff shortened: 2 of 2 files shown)


Kludex:
Thanks for the PR.

Closing this since it was already fixed.

PR #1426, opened 2025-10-03
Add documentation about testing

Kludex:


src/mcp/shared/memory.py
```diff
@@ -63,10 +59,13 @@ async def create_connected_server_and_client_session(
...
     """Creates a ClientSession that is connected to a running MCP server."""
-    async with create_client_server_memory_streams() as (
-        client_streams,
-        server_streams,
-    ):
+
+    # TODO(Marcelo): we should have a proper `Client` that can use this "in-memory transport",
+    # and we should expose a method in the `FastMCP` so we don't access a private attribute.
+    if isinstance(server, FastMCP):
+        server = server._mcp_server  # type: ignore[reportPrivateUsage]
```
Kludex, line 66: I've also introduced support for `FastMCP` here.

PR #1432, opened 2025-10-06
Update bug.yaml w/ fully-self-contained client+server example

ochafik:
We chatted about updating the bug template to get users to provide full repro test cases.

This might be a bit too much but shows a single-file, no-dep (besides uv) client+server example as an easy starting point.

@felixweinberger @maxisbey @Kludex wdyt?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yaml
+++ b/.github/ISSUE_TEMPLATE/bug.yaml
@@ -39,9 +39,50 @@ body:
         demonstrating the bug.
 
       placeholder: |
+        #!/usr/bin/env uv run
+        # /// script
+        # requires-python = ">=3.11"
+        # dependencies = [
+        #     "anyio",
+        #     "mcp",
+        # ]
+        # ///
+        import threading
+
+        import anyio
+
+        from mcp.client.session import ClientSession
+        from mcp.client.streamable_http import streamablehttp_client
         from mcp.server.fastmcp import FastMCP
 
-        ...
+
+        async def run_server():
+            mcp = FastMCP()
+
+            @mcp.tool()
+            def add(a: int, b: int) -> int:
+                """Add two numbers."""
+                return a + b
+            
+            # ...
+
+            await mcp.run_streamable_http_async()
+
+        async def run_client():
+            async with streamablehttp_client("http://localhost:8000/mcp") as (read_stream, write_stream, _):
+                async with ClientSession(read_stream, write_stream) as session:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -39,9 +39,50 @@ body:
...
 
       placeholder: |
+        #!/usr/bin/env uv run
+        # /// script
+        # requires-python = ">=3.11"
+        # dependencies = [
+        #     "anyio",
+        #     "mcp",
+        # ]
+        # ///
```
felixweinberger, line 49: Does this do anything? It's not obvious to me whether this needs to be included in the file or not or create a `pyproject.toml`file for this or something. Is this just a listing to explicitly list dependencies you might have to add?

If it does more than just list dependencies in plaintext it could be cool to have a comment like `# list the dependencies needed to run your example here` or something.
Otherwise I think this is a 10x improvement over the nonexistent example right now.

Makes it much more obvious what a good code example is!

felixweinberger (changes requested):

.github/ISSUE_TEMPLATE/bug.yaml
Kludex: > Does this do anything?

`uv run` installs what is in those headers. `uv run --help | grep script` locally, and you can read about it.

Kludex:
I would find it annoying to do it. 🤷‍♂️

maxisbey:
> I would find it annoying to do it. 🤷‍♂️

Which part? Asking people to provide code snippets, or is the one here too verbose or something?

.github/ISSUE_TEMPLATE/bug.yaml
maxisbey: I'm not entirely convinced we need this.

At least from most of the examples I've run into they've not needed third party dependencies, and I think it's reasonable to assume the maintainers will have mcp and anyio installed as it's the requirements of the repo.

Although it's supported by uv, it's a rather obscure part of uv and may confuse users.

Kludex: It's a python thing: https://peps.python.org/pep-0723/

maxisbey: Oh huh I totally thought it was just uv.

I'm not fully against adding it if others agree, but I'm not convinced we need it.

PR #1453, opened 2025-10-09
fix: send params as empty object for list methods without cursor

felixweinberger:
## Motivation and Context

Some external MCP servers implement strict JSON-RPC validation requiring the params field to always be present. Previously, the Python SDK omitted params entirely when `cursor=None`, causing validation errors. There is no way to force `params={}` with the current implementation of `list_tools()`, making it not possible to interact with these servers.

This change allows sending `params: {}` by allowing the caller to specify a full `types.PaginatedRequestParams` rather than just `cursor`, matching TypeScript SDK flexibility and allowing more compatibility with strict servers. We achieve this by introducing a new optional param `params: types.PaginatedRequestParams | None = None` which will be the new way to interact with this API.

## How Has This Been Tested?

Added a test demonstrating the specific behavior as well the new param.
Refactored the existing tests to be parameterized.

## Breaking Changes

This change is fully backwards compatible as existing behavior stays the same and the new param is completely optional.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/resources/base.py
+++ b/src/mcp/server/fastmcp/resources/base.py
@@ -13,7 +13,7 @@ from pydantic import (
     field_validator,
 )
 
-from mcp.types import Icon
+from mcp.types import Annotations, Icon
 
 
 class Resource(BaseModel, abc.ABC):
@@ -31,6 +31,7 @@ class Resource(BaseModel, abc.ABC):
         pattern=r"^[a-zA-Z0-9]+/[a-zA-Z0-9\-+.]+$",
     )
     icons: list[Icon] | None = Field(default=None, description="Optional list of icons for this resource")
+    annotations: Annotations | None = Field(default=None, description="Optional annotations for the resource")
 
     @field_validator("name", mode="before")
     @classmethod
--- a/src/mcp/server/fastmcp/resources/resource_manager.py
+++ b/src/mcp/server/fastmcp/resources/resource_manager.py
@@ -10,7 +10,7 @@ from pydantic import AnyUrl
 from mcp.server.fastmcp.resources.base import Resource
 from mcp.server.fastmcp.resources.templates import ResourceTemplate
 from mcp.server.fastmcp.utilities.logging import get_logger
-from mcp.types import Icon
+from mcp.types import Annotations, Icon
 
 if TYPE_CHECKING:
     from mcp.server.fastmcp.server import Context
@@ -63,6 +63,7 @@ class ResourceManager:
         description: str | None = None,
         mime_type: str | None = None,
         icons: list[Icon] | None = None,
+        annotations: Annotations | None = None,
     ) -> ResourceTemplate:
         """Add a template from a function."""
         template = ResourceTemplate.from_function(
@@ -73,6 +74,7 @@ class ResourceManager:
             description=description,
             mime_type=mime_type,
             icons=icons,
+            annotations=annotations,
         )
         self._templates[template.uri_template] = template
         return template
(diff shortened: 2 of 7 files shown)


felixweinberger:
Re-evaluating the approach here, as we're basically making the `params={}` the default behavior with this when no cursor is supplied. The TS SDK allows both via:

```ts
// explicit {} param
const result = await client.listTools({});

// undefined param, gets ommited by JSON.stringify()
const result = await client.listTools();
```

Whereas with this change we would now always force `list_tools(params={})` in Python. Maybe the better approach is to allow both `{}` and `None` in Python on `list_tools()` somehow.

The spec defines `ListToolRequest` with both fields as optional:

```ts
interface ListToolsRequest {
  method: “tools/list”;
  params?: { cursor?: string };
}
```

https://modelcontextprotocol.io/specification/2025-06-18/schema#listtoolsrequest

An alternative solution could be to change the signature of `list_tools` to something like:

```python
  async def list_tools(
      self,
      cursor: str | None = None, # deprecated
      *,
      params: types.PaginatedRequestParams | None = None,
  ) -> types.ListToolsResult:
```

Which would be more complete and equivalent to TS capabilities here: https://github.com/modelcontextprotocol/typescript-sdk/blob/f4b8a48ded019a54a38d3d150a013427d6cbdbc6/src/client/index.ts#L500-L514

felixweinberger:
Updated this PR with the approach described in the above comment.

This PR is probably easier to review looking at each commit, as I also took the chance to refactor tests for less repetition in the last commit.

src/mcp/client/session.py
```diff
@@ -212,25 +213,55 @@ async def set_logging_level(self, level: types.LoggingLevel) -> types.EmptyResul
...
+            params: Pagination parameters. Defaults to None (omits params field).
+        """
+        if cursor is not None:
+            warnings.warn(
+                "cursor parameter is deprecated, use params=PaginatedRequestParams(cursor=...) instead",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            if params is None:
+                params = types.PaginatedRequestParams(cursor=cursor)
```
felixweinberger: Not sure this is the "right" way to deprecate args like this, I couldn't find many examples in the code base to go on.

Keen for thoughts if there are better ways to do this!

Kludex: you'd need to check the message to see if the stack level is correct, did u?

maxisbey: I was going to do this with the pagination PR but realised the deprecation can be kinda aggressive with log spam. Should we worry about this?

felixweinberger: I found one other deprecation using `stacklevel=2` here: https://github.com/modelcontextprotocol/python-sdk/blob/71889d7387f070cd872cab7c9aa3d1ff1fa5a5d2/src/mcp/server/lowlevel/server.py#L344

The output when I run the tests while capturing warnings:

```
  /tmp/test_stacklevel.py:20: DeprecationWarning: cursor parameter is deprecated, use params=PaginatedRequestParams(cursor=...) instead
      await client_session.list_tools(cursor="test_cursor")
```

So I believe that looks correct?

Kludex: yes

felixweinberger:
@maxisbey implemented the `@overload` approach in commit on top.

However I think the decorator was in the wrong place - it should only be on the old signature, i.e. `list_tools(self, cursor: str | None)`.

I.e. I think the type errors should look like this (what it looks like with the current branch):

CleanShot 2025-10-13 at 17 45 24@2x

What this looks like on hover:

CleanShot 2025-10-13 at 18 03 37@2x

src/mcp/client/session.py
felixweinberger: I've replaced this with the `@deprecated` decorator + overloads as suggested by @maxisbey below.

The @deprecated decorator adds both warning & type checking support according to the docs: https://github.com/python/typing_extensions/blob/420f8f2e1b9a2ea6c63597856ecc78455724614f/src/typing_extensions.py#L2921

PR #1459, opened 2025-10-09
Allow CallToolResult to be returned directly to support _meta field for OpenAI Apps

BrandonShar:
This PR allows tools to directly return `CallToolResult` instead of having it cast automatically by the lowlevel server.

## Motivation and Context
[OpenAI's Apps SDK](https://developers.openai.com/apps-sdk/build/mcp-server#structure-the-data-your-tool-returns) uses the `_meta` field of a tool call result to allow passing data to app components without them being seen by the model. This behavior is supported by the [protocol](https://modelcontextprotocol.io/specification/2025-06-18/basic#meta), but isn't implemented in the SDK. 

In order to maximize discoverability and future change-abililty, I added the ability to directly return the `CallToolResult`, which already supports the _meta field, rather than add additional tuple fields.

~~The drawback to this method is that output schemas can't be used since they're based on the method's return type. I think that's a fair tradeoff though because the user is basically saying they want to take control themselves.~~ We can use `Annotated` types to still support output schemas with `CallToolResult` returns.

Users will have a choice of either:

1. No return typehint, no output schema (same behavior as now)
2. `CallToolResult` return typehint -> no output schema. 
3. `Annotated[CallToolResult, SomeClass]` -> `SomeClass` output schema 

## How Has This Been Tested?

MCP Inspector with a new example file
Screenshot 2025-10-09 at 9 40 07 PM

Screenshot 2025-10-09 at 9 43 32 PM

Screenshot 2025-10-09 at 9 44 05 PM

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,61 @@ causes the tool to be classified as structured _and this is undesirable_,
 the  classification can be suppressed by passing `structured_output=False`
 to the `@tool` decorator.
 
+##### Advanced: Direct CallToolResult
+
+For full control over tool responses including the `_meta` field (for passing data to client applications without exposing it to the model), you can return `CallToolResult` directly:
+
+<!-- snippet-source examples/snippets/servers/direct_call_tool_result.py -->
+```python
+"""Example showing direct CallToolResult return for advanced control."""
+
+from typing import Annotated
+
+from pydantic import BaseModel
+
+from mcp.server.fastmcp import FastMCP
+from mcp.types import CallToolResult, TextContent
+
+mcp = FastMCP("CallToolResult Example")
+
+
+class ValidationModel(BaseModel):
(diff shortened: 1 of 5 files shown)


BrandonShar:
@felixweinberger could you add this to the Meta improvements milestone?

maxisbey:
Related:
- https://github.com/modelcontextprotocol/python-sdk/pull/1463
- https://github.com/modelcontextprotocol/python-sdk/issues/1465
- https://github.com/modelcontextprotocol/python-sdk/issues/1462

maxisbey:
Thank you for the submission! I've added some code to error on registration if the function has a union type with CallToolResults as that would kinda break stuff. Also added some documentation for using this.

PR #1463, opened 2025-10-10
feat: add tool metadata in FastMCP.tool decorator

mat-octave:
Adds a feature to add metadata to a tool using the `FastMCP.tool` decorator, passing it to the [Tool](https://modelcontextprotocol.github.io/python-sdk/api/#mcp.Tool) object.

## Motivation and Context

ChatGPT's App SDK expects information about UI resources to be in a tool's `_meta` field. The decorator `FastMCP.tool` didn't accept metadata but the field exists on the lower-level [mcp.Tool](https://modelcontextprotocol.github.io/python-sdk/api/#mcp.Tool) object.

## How Has This Been Tested?

I added metadata to `mcp.tool(...)` and started the server successfully. I used the Model Inspector to list the tools, and the metadata was visible in the schema.

## Breaking Changes

No, this just adds an optional field `meta`.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #1476, opened 2025-10-13
feat: Allow ResourceContents objects to be returned directly from read_resource handlers

alexanderreiff:
This PR allows resources to directly return `TextResourceContents` or `BlobResourceContents` instead of having it cast automatically by the lowlevel server.

Inspired by #1459, but for resources.

## Motivation and Context
[OpenAI's Apps SDK](https://developers.openai.com/apps-sdk/reference) uses the `_meta` field of a read resource result representing a "widget" to configure Content Security Policy and other behaviors.

This use of `_meta` is supported by the [protocol](https://modelcontextprotocol.io/specification/2025-06-18/basic#meta), but isn't implemented in the SDK.

In order to maximize discoverability and future change-abililty, I added the ability to directly return the two `ResourceContents` classes, which already supports the `_meta` field, rather than add additional tuple fields.

## How Has This Been Tested?
Test suite additions and example snippets are provided. Running the example shows `_meta` in the response.
```
npx @modelcontextprotocol/inspector uv run examples/snippets/servers/resource_contents_direct.py
```
image

image

## Breaking Changes
N/A

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A

### DIFF
--- a/tests/issues/test_141_resource_templates.py
+++ b/tests/issues/test_141_resource_templates.py
@@ -61,11 +61,11 @@ async def test_resource_template_edge_cases():
     if isinstance(content, ReadResourceContents):
         assert content.content == "Post 456 by user 123"
         assert content.mime_type == "text/plain"
-    elif isinstance(content, TextResourceContents):
+    elif isinstance(content, TextResourceContents):  # pragma: no cover
         # If it's TextResourceContents (direct return)
         assert content.text == "Post 456 by user 123"
         assert content.mimeType == "text/plain"
-    else:
+    else:  # pragma: no cover
         # Should not happen for string resources
         raise AssertionError(f"Unexpected content type: {type(content)}")
 
--- a/tests/server/fastmcp/servers/test_file_server.py
+++ b/tests/server/fastmcp/servers/test_file_server.py
@@ -100,10 +100,10 @@ async def test_read_resource_dir(mcp: FastMCP):
     if isinstance(res, ReadResourceContents):
         assert res.mime_type == "text/plain"
         files = json.loads(res.content)
-    elif isinstance(res, TextResourceContents):
+    elif isinstance(res, TextResourceContents):  # pragma: no cover
         assert res.mimeType == "text/plain"
         files = json.loads(res.text)
-    else:
+    else:  # pragma: no cover
         raise AssertionError(f"Unexpected content type: {type(res)}")
 
     assert sorted([Path(f).name for f in files]) == [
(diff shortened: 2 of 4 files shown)


alexanderreiff:
@felixweinberger here is one more related to META Improvements milestone, for resources.

Similar to #1459 for tool call results.

BrandonShar:
I think this could be implemented the same way as #1463 by just exposing the `meta` field in the resource decorator instead of altering the return type.

alexanderreiff:
Adding the metadata as part of the decorator would certainly work for the OpenAI use case, since the widget metadata is static.

But I was thinking that, because the `_meta` is embedded within the `List[Resource]` object in the response, it conceptually has the potential to be dynamic based on the resource requested and thus it would be more convenient to be part of the read resource output.

andrewg-oai:
A workaround whilst unmerged is to define a custom handler that returns _meta on read.
```
def _setup_read_resource_handler(self) -> None:
    """
    Override to include resource meta in read_resource responses.
    og
    https://github.com/jlowin/fastmcp/blob/2f561ec98d0b5f078221acf6a2126b1152966365/src/fastmcp/server/server.py#L679
    """
    super()._setup_read_resource_handler()
    original_handler = self._mcp_server.request_handlers[ReadResourceRequest]

    async def wrapper(req: ReadResourceRequest) -> ServerResult:
        uri = req.params.uri

        # Look up resource first to get meta
        resource = None
        resource_meta = None
        async with Context(fastmcp=self):
            resource = await self._get_resource_or_template_or_none(str(uri))
            if resource and self._should_enable_component(resource):
                resource_meta = resource.get_meta(
                    include_fastmcp_meta=self.include_fastmcp_meta
                )

        # Use parent handler for task routing and reading
        result = await original_handler(req)

        # Rebuild contents with meta (Pydantic models are immutable)
        read_result = result.root
        if (
            resource_meta
            and isinstance(read_result, ReadResourceResult)
            and read_result.contents
        ):
            new_contents = []
            for content in read_result.contents:
                existing_meta = getattr(content, "_meta", None) or {}
                merged_meta = {**resource_meta, **existing_meta}
                if hasattr(content, "text"):
                    new_contents.append(
                        TextResourceContents(
                            uri=content.uri,
                            text=content.text,
                            mimeType=content.mimeType,
                            _meta=merged_meta,
                        )
                    )
                elif hasattr(content, "blob"):
                    new_contents.append(
                        BlobResourceContents(
                            uri=content.uri,
                            blob=content.blob,
                            mimeType=content.mimeType,
                            _meta=merged_meta,
                        )
                    )
                else:
                    new_contents.append(content)
            return ServerResult(ReadResourceResult(contents=new_contents))

        return result

    self._mcp_server.request_handlers[ReadResourceRequest] = wrapper
```

from: [Conrad_Frame](https://community.openai.com/u/conrad_frame) on https://community.openai.com/t/issue-fastmcp-resources-cannot-return-meta-in-read-responses-but-openai-client-expects-it/1368892

PR #1478, opened 2025-10-14
fix: Set the Server session initialization state immediately after respond…

daamitt:
Set the Server session initialization state immediately after responding to InitializeRequest to fix the race condition that gives the following warnings

```
WARNING  Failed to validate request: Received request before initialization was complete              session.py:363
DEBUG    Message that failed validation: method='tools/list' params={} jsonrpc='2.0' id=2              session.py:364
```

## Motivation and Context
For Stateful http mcp servers (when stateless_http=False) this change is required for most clients to work correctly. I have tested this with Claude desktop and VSC. Mcp Inspector does not show the issue as it does not automatically call tools/list after initialize. 

This change brings the behaviour of the mcp python-sdk inline with the TS sdk
[TS SDK link] https://github.com/modelcontextprotocol/typescript-sdk/blob/caa25503cdfc449d116c204e866bccc2617d7037/src/server/streamableHttp.ts#L501

## How Has This Been Tested?
A test case is attached to the PR
[main.py](https://github.com/user-attachments/files/22900144/main.py)

uv run pytest : 
Results (19.93s):
       633 passed
         2 skipped
         1 xfailed

## Breaking Changes
No

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [ x] My code follows the repository's style guidelines
- [ x] New and existing tests pass locally
- [ x] I have added appropriate error handling

### DIFF
--- a/tests/server/test_session_race_condition.py
+++ b/tests/server/test_session_race_condition.py
@@ -1,0 +1,155 @@
+"""
+Test for race condition fix in initialization flow.
+
+This test verifies that requests can be processed immediately after
+responding to InitializeRequest, without waiting for InitializedNotification.
+
+This is critical for HTTP transport where requests can arrive in any order.
+"""
+
+import anyio
+import pytest
+
+import mcp.types as types
+from mcp.server.models import InitializationOptions
+from mcp.server.session import ServerSession
+from mcp.shared.message import SessionMessage
+from mcp.shared.session import RequestResponder
+from mcp.types import ServerCapabilities, Tool
+
+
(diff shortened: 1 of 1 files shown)


daamitt:
Description mentions the warning msg but actually most MCP clients fail to list the prompts and tools due to this.

daamitt:
Thank you for the test case, it captures the race condition well. The race condition is quite common when there are async awaits in the HTTP request handling - like custom middleware for auth.

PR #1490, opened 2025-10-17
Add tests for resource leak in streamable_http SSE handlers

### LINKED ISSUE #1450: Response leak in SSE handlers
I found a resource leak in the streamable HTTP client. When SSE streaming fails with an exception, the HTTP response isn't closed.

Location:
`src/mcp/client/streamable_http.py`:
- `_handle_sse_response` (line 336)
- `_handle_resumption_request` (line 251)

The Issue:
python
async def _handle_sse_response(self, response: httpx.Response, ...):
    try:
        event_source = EventSource(response)
            if is_complete:
                await response.aclose()  # Only closed here
                break
    except Exception as e:
        # response leaked!

If the SSE iteration raises an exception (malformed JSON, network error, etc.), the response is never closed.

Impact:
Connection pool gets exhausted in long-running clients, eventually causing new requests to hang or fail.

Fix:
try:
    ...
except Exception as e:
    ...
finally:
    await response.aclose()

Both methods need this fix.

Env:
Python SDK version: 1.1.2 (or main branch)
Python: 3.12
Transport: StreamableHTTP

certainly-param:
Fixes #1450

I found that HTTP responses don't get closed properly when SSE streaming fails with exceptions in the streamable HTTP client.

The Problem:
When the async for loop in `_handle_sse_response` and `_handle_resumption_request` throws an exception (malformed JSON, network error, etc.), the response is never closed because `response.aclose()` is only called in the success path.

Impact:
This can cause connection pool exhaustion in long-running clients, eventually causing new requests to hang or fail.

What I Added:
- `test_resource_leak_reproduction.py` - Standalone script that demonstrates the issue
- `tests/client/test_streamable_http_resource_leak.py` - Pytest tests that verify the resource leak and show how the fix should work

Both test files show that when SSE parsing fails, the response doesn't get closed, confirming the resource leak.

The Fix:
The methods need `finally` blocks to ensure `response.aclose()` is always called:

```python
try:
    # ... existing code ...
except Exception as e:
    # ... existing exception handling ...
finally:
    await response.aclose()
```

This ensures resources are properly cleaned up regardless of whether exceptions occur.

### DIFF
--- a/resource_leak_reproduction.py
+++ b/resource_leak_reproduction.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python3
+"""
+Reproduction script for the resource leak I found in streamable_http.py
+
+I noticed that when SSE streaming fails, the HTTP response doesn't get closed properly.
+This happens in both _handle_sse_response and _handle_resumption_request methods.
+
+The problem: if the async for loop throws an exception (like malformed JSON or network issues),
+the response.aclose() call never happens because it's only in the success path.
+
+Files affected:
+- src/mcp/client/streamable_http.py (lines 336 and 251)
+
+This can cause connection pool exhaustion over time in production.
+"""
+
+import asyncio
+import sys
+from pathlib import Path
+
+# Add the mcp module to the path
+sys.path.insert(0, str(Path(__file__).parent / "src"))
+
+from mcp.client.streamable_http import StreamableHTTPTransport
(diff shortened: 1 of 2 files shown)


maxisbey:
Marking as a draft for now since this is only a reproduction and not a completed fix

PR #1491, opened 2025-10-17
fix: Update spec links to new modelcontextprotocol.io location

maxisbey:
## Summary

The MCP specification has moved from `spec.modelcontextprotocol.io` to `modelcontextprotocol.io/specification/`. This PR updates all references in the README to point to the new location.

## Changes

- Updated spec badge link to point to `modelcontextprotocol.io/specification/2025-06-18`
- Updated documentation section spec link to use the same versioned URL

## Discussion Point

This PR uses the versioned spec URL (`/specification/2025-06-18`) for stability. We should discuss whether to:
- Keep the versioned link (ensures consistent documentation even as spec evolves)
- Switch to `/specification/draft` (always points to latest, but may change)


maxisbey:
> Good find!
> 
> Can we point to https://modelcontextprotocol.io/specification/latest instead potentially?
> 
> I think that should always point to the latest released spec.

ah yep perfect, I was wondering what that link would be. Updated

maxisbey:
failed test is flaky, fixed here: https://github.com/modelcontextprotocol/python-sdk/pull/1525

PR #1496, opened 2025-10-18
Refactor `func_metadata()` implementation

Viicos:
## Motivation and Context

This is a first step in adding support for Pydantic 2.12 and Python 3.14. It avoids relying on patterns that we discourage using in Pydantic (mostly around the `FieldInfo` class), and removes usage of private utilities. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -219,7 +219,7 @@ def func_metadata(
     """
     try:
         sig = inspect.signature(func, eval_str=True)
-    except NameError as e:
+    except NameError as e:  # pragma: no cover
         # This raise could perhaps be skipped, and we (FastMCP) just call
         # model_rebuild right before using it 🤷
         raise InvalidSignature(f"Unable to evaluate type annotations for callable {func.__name__!r}") from e
@@ -298,7 +298,9 @@ def func_metadata(
                 # Reconstruct the original annotation, by preserving the remaining metadata,
                 # i.e. from `Annotated[CallToolResult, ReturnType, Gt(1)]` to
                 # `Annotated[ReturnType, Gt(1)]`:
-                original_annotation = Annotated[(return_type_expr, *inspected_return_ann.metadata[1:])]
+                original_annotation = Annotated[
+                    (return_type_expr, *inspected_return_ann.metadata[1:])
+                ]  # pragma: no cover
             else:
                 # We only had `Annotated[CallToolResult, ReturnType]`, treat the original annotation
                 # as beging `ReturnType`:
--- a/tests/server/fastmcp/test_func_metadata.py
+++ b/tests/server/fastmcp/test_func_metadata.py
@@ -1188,7 +1188,7 @@ def test_disallowed_type_qualifier():
     from mcp.server.fastmcp.exceptions import InvalidSignature
 
     def func_disallowed_qualifier() -> Final[int]:  # type: ignore
-        pass
+        pass  # pragma: no cover
 
     with pytest.raises(InvalidSignature) as exc_info:
         func_metadata(func_disallowed_qualifier)
@@ -1196,7 +1196,7 @@ def test_disallowed_type_qualifier():
 
 
 def test_preserves_pydantic_metadata():
-    def func_with_metadata() -> Annotated[int, Field(gt=1)]: ...
+    def func_with_metadata() -> Annotated[int, Field(gt=1)]: ...  # pragma: no branch
 
     meta = func_metadata(func_with_metadata)
 


pyproject.toml
```diff
@@ -33,6 +33,8 @@ dependencies = [
     "uvicorn>=0.31.1; sys_platform != 'emscripten'",
     "jsonschema>=4.20.0",
     "pywin32>=310; sys_platform == 'win32'",
+    "typing-extensions>=4.9.0",
+    "typing-inspection>=0.4.1",
```
Viicos, line 38: `typing-extensions` is already used in the project, but not declared as an explicit dependency. 4.9.0 is the oldest version with (proper) support for `typing_extensions.deprecated`.

`typing-inspection` is already a Pydantic dependency, so nothing new added here.

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -298,58 +303,58 @@ def _try_create_model_and_schema(
...
+            model = _create_wrapped_model(func_name, type_annotation)
             wrap_output = True
 
         # Case 4: Other class types (dataclasses, regular classes with annotations)
         else:
             type_hints = get_type_hints(type_annotation)
             if type_hints:
                 # Classes with type hints can be converted to Pydantic models
-                model = _create_model_from_class(type_annotation)
+                model = _create_model_from_class(type_annotation, type_hints)
```
Viicos, line 399: `_create_model_from_class()` used to compute the type hints again, pass them to avoid unnecessarily computing them again.

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -449,43 +451,6 @@ class DictModel(RootModel[dict_annotation]):
...
-            name=param.name,
-            kind=param.kind,
-            default=param.default,
-            annotation=_get_typed_annotation(param.annotation, globalns),
-        )
-        for param in signature.parameters.values()
-    ]
-    typed_return = _get_typed_annotation(signature.return_annotation, globalns)
-    typed_signature = inspect.Signature(typed_params, return_annotation=typed_return)
-    return typed_signature
```
Viicos, line 515: Removed, we can just use `inspect.get_annotations()` with `eval_str=True`.

It also enables more use cases to be supported, e.g.:

```python

def func[T](a: T): ...

from_metadata(func)  # Used to raise
```

Let me know if I should add a test for this.

Kludex (dismissed):
This PR keeps the support for Pydantic 2.11.

@Viicos is maintaining Pydantic, so he knows stuff.

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -272,35 +262,50 @@ def func_metadata(
...
         raise InvalidSignature(f"Function {func.__name__}: return annotation required for structured output")
 
-    output_info = FieldInfo.from_annotation(_get_typed_annotation(sig.return_annotation, globalns))
-    annotation = output_info.annotation
+    inspected_return_ann = inspect_annotation(sig.return_annotation, annotation_source=AnnotationSource.FUNCTION)
+    return_type_expr = inspected_return_ann.type
+    if return_type_expr is UNKNOWN and structured_output is True:
+        # `return_type_expr` is `UNKNOWN` when a bare type qualifier is used (unlikely to happen as a return annotation
+        # because it doesn't make any sense, but technically possible).
+        raise InvalidSignature(f"Function {func.__name__}: return annotation required for structured output")
```
maxisbey: do we already have a unit test that covers this? otherwise would be good to add one

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -272,35 +262,50 @@ def func_metadata(
...
-        if output_info.metadata:
-            annotation = output_info.metadata[0]
+    if isinstance(return_type_expr, type) and issubclass(return_type_expr, CallToolResult):
+        if inspected_return_ann.metadata:
+            return_type_expr = inspected_return_ann.metadata[0]
+            if len(inspected_return_ann.metadata) >= 2:
+                # Reconstruct the original annotation, by preserving the remaining metadata,
+                # i.e. from `Annotated[CallToolResult, ReturnType, Gt(1)]` to
+                # `Annotated[ReturnType, Gt(1)]`:
+                original_annotation = Annotated[(return_type_expr, *inspected_return_ann.metadata[1:])]
```
maxisbey: if I'm reading this right, does that mean the current code on `main` would strip the extra metadata and cause different Pydantic models to be created? if so a unit test for this seems like a good idea

maxisbey (changes requested):

src/mcp/server/fastmcp/utilities/func_metadata.py
Viicos: I'll add one.

src/mcp/server/fastmcp/utilities/func_metadata.py
Viicos: The code on `main` behaves as expected (in most cases, see [here](https://github.com/modelcontextprotocol/python-sdk/pull/1496/files#diff-7ab569bc3087a1aab22a2cac2a25598f907b2faa0d83708b7b8eaf04218bbebeR353-R354) where it doesn't on `main` and on this PR), because it is constructing a `FieldInfo` instance from the annotation, and passing it again as `Annotated[ann, field_info_instance]`, which is an undocumented pattern although kind of works.

I'll add tests related to this.

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -212,56 +217,47 @@ def func_metadata(
...
-            # Use a prefixed internal name
-            internal_name = f"field_{param.name}"
-            dynamic_pydantic_model_params[internal_name] = (field_info.annotation, field_info)
+            field_kwargs["alias"] = field_name
+            # Use a prefixed field name
+            field_name = f"field_{field_name}"
+
+        if param.default is not inspect.Parameter.empty:
+            dynamic_pydantic_model_params[field_name] = (
+                Annotated[(annotation, *field_metadata, Field(**field_kwargs))],
```
maxisbey, line 251: Should this be without paranthesis?

e.g. `Annotated[annotation, *field_metadata, Field(**field_kwargs)]`

maxisbey (dismissed):

maxisbey (approved):

Viicos:
Thanks @maxisbey (sorry I'm out for some days couldn't rebase).

PR #1499, opened 2025-10-20
docs: Update examples to use stateless HTTP with JSON responses

domdomegg:
## Summary

Updates README and example code to consistently recommend and demonstrate stateless HTTP with JSON responses as the recommended pattern for production deployments.

## Changes

- Update quickstart example to use `stateless_http=True, json_response=True` and make it executable
- Update all streamable HTTP mounting examples with recommended config
- Update OAuth server example with recommended config  
- Reorder `streamable_config.py` to show recommended option first
- Change quickstart integration instructions to use Claude Code with HTTP transport
- Clarify that stateless HTTP + JSON is optimal for scalability

## Rationale

This encourages users away from stdio/SSE/stateful HTTP patterns that have extra complexity and makes servers less reliable for most use cases. Stateless HTTP with JSON responses provides:

- Better scalability (no session state management)
- Simpler deployment (works across multiple instances)
- More reliability (simpler error recovery without SSE stream parsing)
- Lower complexity (no connection/session management overhead)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Run from the repository root:
 from mcp.server.fastmcp import FastMCP
 
 # Create an MCP server
-mcp = FastMCP("Demo", stateless_http=True, json_response=True)
+mcp = FastMCP("Demo", json_response=True)
 
 
 # Add an addition tool
@@ -901,7 +901,6 @@ class SimpleTokenVerifier(TokenVerifier):
 # Create FastMCP instance as a Resource Server
 mcp = FastMCP(
     "Weather Service",
-    stateless_http=True,
     json_response=True,
     # Token verifier for authentication
     token_verifier=SimpleTokenVerifier(),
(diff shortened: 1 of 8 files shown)


domdomegg:
Thanks! I've updated the PR to remove `stateless_http=True` from all the examples where it was newly added - see https://github.com/modelcontextprotocol/python-sdk/pull/1499/commits/8dbdfb27da77d8f0aac6f716f293bd2f341a2980.

Also will note that the `stateless_http` implementation is already supported by the spec, see [2.5.1](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports). "A server using the Streamable HTTP transport MAY assign a session ID at initialization time" - `stateless_http` in the Python SDK = just not doing this 'MAY' component. But can see there may be other transports changes soon anyways so won't push this further for now.

PR #1500, opened 2025-10-20
Relax Accept header requirement for JSON-only responses

domdomegg:
## Summary

This PR relaxes the Accept header validation for servers with `is_json_response_enabled=True`. These servers only return `application/json` responses and never use SSE, so they should only require `application/json` in the Accept header rather than requiring both `application/json` and `text/event-stream`.

## Changes

- Modified `_handle_post_request` to conditionally validate the Accept header based on response mode
- Extracted validation logic into a new `_validate_accept_header` method to reduce complexity
- Added test coverage for JSON-only Accept header validation
- Maintained backward compatibility - requests with both content types still work

## Benefits

- Easier to test JSON-only MCP servers with tools like `curl`
- Follows the principle that servers should only require what they actually need to return
- All existing tests pass

### DIFF
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -708,6 +708,36 @@ def test_json_response_accept_json_only(json_response_server: None, json_server_
     assert response.headers.get("Content-Type") == "application/json"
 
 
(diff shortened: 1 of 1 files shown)


domdomegg:
@felixweinberger sounds good, have added tests for those

PR #1504, opened 2025-10-22
feat: Add filtering(include/exclude params) for listing tools

jamesev15:
## Motivation and Context

Add `include` and `exclude` parameters to the `list_tools` request to enable programmatic filtering of tools on the server side. This enhancement addresses a key limitation where MCP clients must retrieve all available tools from a server and then filter them client side, leading to unnecessary data transfer and larger context windows for LLMs.

While MCP compliant UIs like Claude Desktop or Continue(VS Code) provide manual tool selection, programmatic use cases (such as LangChain integrations) would benefit from server side filtering capabilities. This is particularly important for LLM applications where reducing the number of tools in the context helps minimize token usage and improves model performance by avoiding overwhelming the LLM with non used tools.

For example, when using MCP servers with LangChain:
```python
# Current approach - gets ALL tools
# Create server parameters for stdio connection

from langchain_mcp_adapters.tools import load_mcp_tools
from langgraph.prebuilt import create_react_agent

server_params = StdioServerParameters(...)

async with stdio_client(server_params) as (read, write):
        # Initialize the connection
        await session.initialize()

        # Get tools
        tools = await load_mcp_tools(session)

        # Create and run the agent
        agent = create_react_agent("openai:gpt-4.1", tools)
        ....


# Desired approach, get only specific tools
# This would be possible with server side filtering, but it could be done with this PR as well.
```

## How Has This Been Tested?
- Added comprehensive unit tests for the new filtering functionality in `ToolManager`
- Tested both `include` and `exclude` parameter scenarios
- Verified error handling for non existent tool names
- Tested mutual exclusivity of `include` and `exclude` parameters
- Validated backward compatibility with existing `list_tools` implementations

## Breaking Changes
None. This is a backward compatible addition. Existing clients that don't use the new parameters will continue to work unchanged, receiving all available tools as before.

## Types of changes
- [x] New feature (non breaking change which adds functionality)

## Checklist

## Additional context
The implementation follows the existing MCP patterns:
- Uses `PaginatedRequestParams` as the base for `ListToolsRequest` to include the new filtering parameters
- Provides clear error messages when specified tools don't exist
- Ensures mutual exclusivity between `include` and `exclude` to prevent ambiguous requests

### DIFF
--- a/src/mcp/server/fastmcp/server.py
+++ b/src/mcp/server/fastmcp/server.py
@@ -61,7 +61,7 @@ from mcp.server.streamable_http import EventStore
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from mcp.server.transport_security import TransportSecuritySettings
 from mcp.shared.context import LifespanContextT, RequestContext, RequestT
-from mcp.types import Annotations, AnyFunction, ContentBlock, GetPromptResult, Icon, ToolAnnotations
+from mcp.types import Annotations, AnyFunction, ContentBlock, GetPromptResult, Icon, ListToolsRequest, ToolAnnotations
 from mcp.types import Prompt as MCPPrompt
 from mcp.types import PromptArgument as MCPPromptArgument
 from mcp.types import Resource as MCPResource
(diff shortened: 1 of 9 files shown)


maxisbey:
Thank you for submitting the PR, I appreciate the effort to try and get some new functionality in! Unfortunately, as this changes the underlying MCP specification itself it would first require an SEP to be submitted and accepted into the protocol specification before we can make any changes to the SDK.

There is an existing SEP with a proposal for tool filtering being discussed which you're welcome to add your opinions on if you'd like here: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1300

As the changes in this PR would require a whole new SEP to be submitted/accepted before it could be merged, and are already quite different to the current leading SEP (linked above) I'm gong to close the PR for now.

PR #1524, opened 2025-10-28
Add name and title fields to Tool class

maxisbey:
Adds the `name` and `title` fields to the `Tool` class that were missing from the type definition.

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -871,6 +871,19 @@ class ToolAnnotations(BaseModel):
 class Tool(BaseMetadata):
     """Definition for a tool the client can call."""
 
+    name: str
+    """
+    Intended for programmatic or logical use, but used as a display name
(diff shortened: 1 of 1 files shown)


src/mcp/types.py
```diff
@@ -871,6 +871,19 @@ class ToolAnnotations(BaseModel):
 class Tool(BaseMetadata):
```
felixweinberger, line 871: Isn't this already inherited from `BaseMetadata`? Why do we need to duplicate it?

https://github.com/modelcontextprotocol/python-sdk/blob/3e798bfc01856df711b921f7f3b55379fbd087e4/src/mcp/types.py#L199-L213

maxisbey: omg you're so right, I read the inheritance as "BaseModel", not BaseMetadata. my bad, closing this 🙃

PR #1528, opened 2025-10-28
fix: Eliminate port allocation race condition in all affected test files

maxisbey:
This PR extends the port allocation race condition fix to all test files affected by the same TOCTOU issue, eliminating intermittent test failures when running with pytest-xdist parallel workers.

## Motivation and Context

When running tests with pytest-xdist parallel workers, multiple workers would receive the same "free" port from `socket.bind(("127.0.0.1", 0))`, causing a TOCTOU (Time-of-Check-Time-of-Use) race condition. This manifested as:
- **"address already in use"** errors when multiple workers tried to bind the same port
- **Tests connecting to wrong servers** (e.g., expecting one server but getting another)
- **Flaky tests in CI** that would pass locally but fail intermittently in parallel runs

This builds on the initial fix in tests/server/fastmcp/test_integration.py by applying the same solution to 7 additional test files that had the identical issue.

## How Has This Been Tested?

- ✅ Ran full test suite with 14 parallel workers: `PYTEST_DISABLE_PLUGIN_AUTOLOAD="" uv run --frozen pytest tests/ -v`
- ✅ All 703 tests pass consistently
- ✅ Verified no "address already in use" errors occur
- ✅ Confirmed no test cross-contamination between workers
- ✅ Tested multiple consecutive runs to ensure stability

Each pytest-xdist worker now receives an exclusive port range from 40000-59999 based on its worker ID, eliminating port conflicts entirely.

## Breaking Changes

None. This is an internal test infrastructure change with no impact on the public API or user-facing functionality.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Files modified:**
- `tests/server/test_streamable_http_security.py` - Updated `server_port` fixture
- `tests/server/test_sse_security.py` - Updated `server_port` fixture
- `tests/shared/test_streamable_http.py` - Updated 3 port fixtures (basic, json, event server)
- `tests/shared/test_sse.py` - Updated `server_port` fixture
- `tests/shared/test_ws.py` - Updated `server_port` fixture
- `tests/client/test_http_unicode.py` - Updated `unicode_server_port` fixture
- `tests/client/test_notification_response.py` - Updated `non_sdk_server_port` fixture
- `tests/test_test_helpers.py` - Fixed socket cleanup to prevent ResourceWarning

**Implementation details:**
All affected fixtures now use `get_worker_specific_port(worker_id)` instead of the race-prone `socket.bind(("127.0.0.1", 0))` pattern. The helper function allocates worker-specific port ranges ensuring no overlap between parallel test workers.

### DIFF
--- a/tests/client/test_http_unicode.py
+++ b/tests/client/test_http_unicode.py
@@ -14,6 +14,7 @@ import pytest
 
 from mcp.client.session import ClientSession
 from mcp.client.streamable_http import streamablehttp_client
+from tests.test_helpers import get_worker_specific_port
 
 # Test constants with various Unicode characters
 UNICODE_TEST_STRINGS = {
@@ -145,11 +146,9 @@ def run_unicode_server(port: int) -> None:
 
 
 @pytest.fixture
-def unicode_server_port() -> int:
+def unicode_server_port(worker_id: str) -> int:
     """Find an available port for the Unicode test server."""
-    with socket.socket() as s:
-        s.bind(("127.0.0.1", 0))
-        return s.getsockname()[1]
+    return get_worker_specific_port(worker_id)
 
 
 @pytest.fixture
--- a/tests/client/test_notification_response.py
+++ b/tests/client/test_notification_response.py
@@ -22,6 +22,7 @@ from mcp import ClientSession, types
 from mcp.client.streamable_http import streamablehttp_client
 from mcp.shared.session import RequestResponder
 from mcp.types import ClientNotification, RootsListChangedNotification
+from tests.test_helpers import get_worker_specific_port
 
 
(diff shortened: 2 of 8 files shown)


maxisbey:
actually, I don't like this approach anymore.

It's a bit late at night so I'm going to leave it as draft for now, but some other options I was thinking are:

- Adding more port randomisation - although this only just _reduces_ the likelihood of port collisions.
- Using a sub process with a mutex or shared memory or something, have that process hand out port numbers - although this means we have a limited number of ports, so we'd need a way to release them over time
- Using a FileLock to allocate ports per test

Although none of them really are clean IMO. So going to leave this for now

PR #1494, opened 2025-10-18
Fix: Add missing lifespan parameter to StreamableHTTP mounting examples

ranjitodedra:
Fixes #1484

Fixes the RuntimeError that occurs when mounting StreamableHTTP servers to existing ASGI applications. The error "Task group is not initialized. Make sure to use run()" was caused by missing session manager initialization.

## Motivation and Context

`RuntimeError: Task group is not initialized. Make sure to use run().` when trying to mount StreamableHTTP servers to existing ASGI applications (like Starlette). This was caused by missing session manager initialization. The documentation examples were incomplete and didn't show the required `lifespan` parameter needed to properly initialize the session manager.

## How Has This Been Tested?

- ✅ Tested all 4 mounting examples individually
- ✅ Verified HTTP 200 responses instead of RuntimeError
- ✅ Tested with PowerShell Invoke-WebRequest on Windows

## Breaking Changes

No breaking changes. This is a documentation fix that provides the correct implementation pattern.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

N/A

Issue #1484, opened 2025-10-16
Documentation on mounting to existing ASGI server is missing information


lei-lei-shanda:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

# Overview

PR #1172 is missing information. 

## Minimal Working Example

take a slightly modified script from [here](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/streamable_http_path_config.py) and run it with `uv run --script main.py`

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "mcp>=1.12",
#     "starlette",
#     "uvicorn",
# ]
# ///
"""
Basic example showing how to mount StreamableHTTP server in Starlette.

Run from the repository root:
    uvicorn examples.snippets.servers.streamable_http_basic_mounting:app --reload
"""

import asyncio

from mcp.server.fastmcp import FastMCP
from starlette.applications import Starlette
from starlette.routing import Mount

# Create MCP server
mcp = FastMCP("My App", stateless_http=True)


@mcp.tool()
def hello() -> str:
    """A simple hello tool"""
    return "Hello from MCP!"


# Mount the StreamableHTTP server to the existing ASGI server
app = Starlette(
    routes=[
        Mount("/", app=mcp.streamable_http_app()),
    ]
)

if __name__ == "__main__":
    import asyncio
    import uvicorn

    asyncio.create_task(mcp.run(transport="streamable-http"))
    asyncio.run(uvicorn.run(app, host="0.0.0.0", port=8000))
```

and test it with the following bash script (`bash test_mcp_server.sh`)

```shell
#!/usr/bin/env bash

MCP_SERVER_PORT=8000
MCP_SERVER_HOST=0.0.0.0

# for mcp service
# first initialize it
curl -X POST "http://${MCP_SERVER_HOST}:${MCP_SERVER_PORT}/mcp" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json,text/event-stream" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
      "protocolVersion": "2025-03-26",
      "capabilities": {
        "tools": {},
        "resources": {},
        "prompts": {}
      },
      "clientInfo": {
        "name": "test-client",
        "version": "1.0.0"
[...]

PR #1534, opened 2025-10-29
fix: Lifespan pr cleanup/Extending ASGI lifespan examples

FloorIsGround:
Removed the stateless http additions by @ranjitodedra as they weren't necessary for the issue and may add confusion as stateless_http isn't necessary for a streamable http mcp service.

## Motivation and Context
Fixing #1494 and also added an example with more detailed explanation of attaching ASGI services through lifespan.

## How Has This Been Tested?
I tested the new example with information from #1484. The main issue was attempting to run two separate ASGI services at the same time rather than mounting them to a main one to call other ASGI services.

## Breaking Changes
No there are no breaking changes just documentation and examples.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context
As was said in Issue #1494 they found they could mount existing server routes onto an mcp server but were struggling on the configuration to run multiple services asynchronously. It is possible to do so with Starlette examples, the mcp service, as well as other asynchronous services can be started and stopped by a context manager lifespan function, on a Starlette app call.

The confusion with stateless_http is probably that a streamable_http mcp server requires stateless_http due to its name, unless you look deeper you won't find that it is only tied to if requests track sessions or carry persistence between requests. I removed the added stateless_http since it wasn't necessary for the examples in question.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1325,7 +1325,7 @@ from starlette.routing import Mount
 from mcp.server.fastmcp import FastMCP
 
 # Create MCP server
-mcp = FastMCP("My App", stateless_http=True)
+mcp = FastMCP("My App")
 
 
 @mcp.tool()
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for working on this! The lifespan fixes were merged via #1669, so this PR is now superseded. Closing, but appreciate the contribution!

Issue #1537, opened 2025-10-30
Implement SEP-986: Tool Name Guidance


felixweinberger:
This is a tracking issue for implementation of [SEP-986](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986).

### Summary

This SEP establishes clear guidelines and format specifications for tool naming conventions within the MCP protocol. The Python SDK needs to implement validation for tool name formats, provide clear error messages when tool names don't follow the specified patterns, and ensure consistency with the standardized naming conventions to improve cross-implementation compatibility and prevent naming conflicts.

### Related Issues & PRs

- **Implementation PRs**: n/a
- **Related PRs**: [#972](https://github.com/modelcontextprotocol/python-sdk/pull/972)
- **Related Issues**: [#97](https://github.com/modelcontextprotocol/python-sdk/issues/97), [#74](https://github.com/modelcontextprotocol/python-sdk/issues/74)

PR #1550, opened 2025-10-30
feat: Implement SEP-986: Tool Name Guidance

vincent0426:
### Changes
- Added a length check that raises an error when the tool name is not within the allowed range:
  ```python
  raise ValueError(f"Invalid tool name length: {len(value)}. Tool name must be between 1 and 128 characters.")
  ```
- Added a regex validation using `re.compile(r"^[A-Za-z0-9_.-]+$")` that raises an error for invalid characters:
  ```python
  raise ValueError("Invalid tool name characters. Allowed: A-Z, a-z, 0-9, underscore (_), dash (-), dot (.).")
  ```

## Motivation and Context

Related issues: SEP-986 https://github.com/modelcontextprotocol/python-sdk/issues/1537

Implements the guidance from https://github.com/modelcontextprotocol/modelcontextprotocol/issues/986
```
- Tool names SHOULD be between 1 and 128 characters in length (inclusive).
- Tool names SHOULD be considered case-sensitive.
- The following SHOULD be the only allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits
  (0-9), underscore (\_), dash (-), and dot (.)
- Tool names SHOULD NOT contain spaces, commas, or other special characters.
- Tool names SHOULD be unique within a server.
- Example valid tool names:
  - getUser
  - DATA_EXPORT_v2
  - admin.tools.list
```

### Valid Values
- `getUser`
- `DATA_EXPORT_v2`
- `admin.tools.list`
- `a`
- `Z9_.-`
- `"x" * 128`  (max length)

### Invalid Values
- `""`  (empty string)
- `"x" * 129`  (exceeds max length)
- `has space`
- `comma,name`
- `not/allowed`
- `name@`
- `name#`
- `name$`

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -1,7 +1,8 @@
 from collections.abc import Callable
+import re
 from typing import Annotated, Any, Generic, Literal, TypeAlias, TypeVar
 
-from pydantic import BaseModel, ConfigDict, Field, FileUrl, RootModel
+from pydantic import BaseModel, ConfigDict, Field, FileUrl, RootModel, field_validator
 from pydantic.networks import AnyUrl, UrlConstraints
 from typing_extensions import deprecated
 
@@ -38,6 +39,10 @@ Cursor = str
 Role = Literal["user", "assistant"]
 RequestId = Annotated[int, Field(strict=True)] | str
 AnyFunction: TypeAlias = Callable[..., Any]
+
+# Tool name validation pattern (ASCII letters, digits, underscore, dash, dot)
+# Pattern ensures entire string contains only valid characters by using ^ and $ anchors
+TOOL_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_.-]+$")
 
 
 class RequestParams(BaseModel):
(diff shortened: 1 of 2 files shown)


src/mcp/types.py
```diff
@@ -891,6 +896,22 @@ class Tool(BaseMetadata):
...
+
+        if not TOOL_NAME_PATTERN.fullmatch(value):
+            raise ValueError("Invalid tool name characters. Allowed: A-Z, a-z, 0-9, underscore (_), dash (-), dot (.).")
+
+        return value
+
+    """
+    See [MCP specification](https://modelcontextprotocol.io/specification/draft/server/tools#tool-names)
+    for more information on tool naming conventions.
+    """
```
Kludex: @maxisbey This is a less verbose way to write the same thing:

```py
from typing import Annotated

from pydantic import StringConstraints

tool: Annotated[str, StringConstraints(max_length=128, pattern=r"^[A-Za-z0-9_.-]+$")]
```

---

I'm not sure if this should be applied on `BaseMetadata`, but if not, please drop the inheritance, and apply this here.

tests/test_types.py
Kludex, line 1: I personally don't think there's a need to test that Pydantic is working as expected.

src/mcp/types.py
Kludex: Btw, most or all the `field_validator`s in this code source can be replaced by proper annotated versions - that you can see are cuter.

src/mcp/types.py
```diff
@@ -891,6 +896,22 @@ class Tool(BaseMetadata):
...
+    @field_validator("name")
+    @classmethod
+    def _validate_tool_name(cls, value: str) -> str:
+        if not (1 <= len(value) <= 128):
+            raise ValueError(f"Invalid tool name length: {len(value)}. Tool name must be between 1 and 128 characters.")
+
+        if not TOOL_NAME_PATTERN.fullmatch(value):
+            raise ValueError("Invalid tool name characters. Allowed: A-Z, a-z, 0-9, underscore (_), dash (-), dot (.).")
+
+        return value
```
maxisbey: We don't want it to actually fail, instead we'd want it to raise a deprecation warning

maxisbey (changes requested):

BossChaos:
## Code Review

**PR:** [feat: Implement SEP-986: Tool Name Guidance](https://github.com/modelcontextprotocol/python-sdk/pull/1550) by @vincent0426

- ✅ Feature

*Code review*

PR #1551, opened 2025-10-30
Make test dependencies required instead of optional

maxisbey:
## Summary

This PR adds `rich`, `typer`, `python-dotenv`, and `websockets` to the required dev dependencies.

## Motivation and Context

Previously, these dependencies were only available when specific optional features were enabled (e.g., `[cli]`, `[sse]`). This meant running `uv sync` without `--all-extras` wouldn't install them, which was problematic because:

1. Test suites that import these libraries would fail when developers ran `uv sync` without `--all-extras`
2. The `--all-extras` flag had to be remembered and used consistently across development workflows
3. CI workflows required `--all-extras` to function properly

This change makes development more ergonomic by ensuring all dependencies needed for testing are always present.

## How Has This Been Tested?

- Verified that `uv sync` now includes rich, typer, python-dotenv, and websockets
- Confirmed CI workflow changes (removed `--all-extras`) still work correctly

## Breaking Changes

None - this only affects development workflow, not the published package.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The main benefit is that `uv sync` now "just works" without needing to remember `--all-extras`. The CI workflow has been updated to remove the `--all-extras` flag since it's no longer needed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,11 @@ dev = [
     "pytest-pretty>=1.2.0",
     "inline-snapshot>=0.23.0",
     "dirty-equals>=0.9.0",
+    # Optional dependencies needed for testing
+    "rich>=13.9.4",
+    "typer>=0.16.0",
+    "python-dotenv>=1.0.0",
+    "websockets>=15.0.1",
 ]
 docs = [
     "mkdocs>=1.6.1",
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -20,7 +20,7 @@ jobs:
           enable-cache: true
           version: 0.9.5
       - name: Install dependencies
-        run: uv sync --frozen --all-extras --python 3.10
+        run: uv sync --frozen --python 3.10
 
       - uses: pre-commit/action@v3.0.1
         with:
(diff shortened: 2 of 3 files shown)


maxisbey:
Marked this as draft as I'm not actually sure this is the best approach here. Basically, I was annoyed having to remember to type --all-extras, and was thinking it just makes more sense to include the optional dependencies in the dev dependencies by default.

Otherwise, maybe would it be better to just mark the tests that rely on those optional dependencies to skip if they're not installed? But what if that just masks issues down the road?

pyproject.toml
```diff
@@ -60,6 +60,11 @@ dev = [
     "pytest-pretty>=1.2.0",
     "inline-snapshot>=0.23.0",
     "dirty-equals>=0.9.0",
+    # Optional dependencies needed for testing
+    "rich>=13.9.4",
+    "typer>=0.16.0",
+    "python-dotenv>=1.0.0",
+    "websockets>=15.0.1",
```
Kludex, line 67: This is the same as: https://github.com/Kludex/starlette/pull/2999/files#r2327181201

```suggestion
    # Optional dependencies needed for testing
    "mcp[rich, cli, ws]"
```

PR #1554, opened 2025-10-31
SEP-1686: Tasks

LucaButBoring:
This PR implements the required changes for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686, which adds task augmentation to requests.

## Motivation and Context
The current MCP specification supports tool calls that execute a request and eventually receive a response, and tool calls can be passed a progress token to integrate with MCP’s progress-tracking functionality, enabling host applications to receive status updates for a tool call via notifications. However, there is no way for a client to explicitly request the status of a tool call, resulting in states where it is possible for a tool call to have been dropped on the server, and it is unknown if a response or a notification may ever arrive. Similarly, there is no way for a client to explicitly retrieve the result of a tool call after it has completed — if the result was dropped, clients must call the tool again, which is undesirable for tools expected to take minutes or more. This is particularly relevant for MCP servers abstracting existing workflow-based APIs, such as AWS Step Functions, Workflows for Google Cloud, or APIs representing CI/CD pipelines, among other applications.

This proposal (and implementation) solves this by introducing the concept of Tasks, which are pending work objects that can augment any other request. Clients generate a task ID and augment their request with it — that task ID is both a reference to the request and an idempotency token. If the server accepts the task augmentation request, clients can then poll for the status and eventual result of the task with the `tasks/get` and `tasks/result` operations.

## How Has This Been Tested?
Unit tests and updated sHTTP example.

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Simplified version of #1398 implementing only the bare minimum logic for task execution as an initial pass. Ported as closely as possible from https://github.com/modelcontextprotocol/typescript-sdk/pull/1041.

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -7,7 +7,6 @@ from uuid import uuid4
 
 import anyio.lowlevel
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
-from jsonschema import SchemaError, ValidationError, validate
 from pydantic import AnyUrl, BaseModel, TypeAdapter
 from typing_extensions import deprecated
 
@@ -451,6 +450,8 @@ class ClientSession(
             logger.warning(f"Tool {name} not listed by server, cannot validate any structured content")
 
         if output_schema is not None:
+            from jsonschema import SchemaError, ValidationError, validate
+
             if result.structuredContent is None:
                 raise RuntimeError(
                     f"Tool {name} has an output schema but did not return structured content"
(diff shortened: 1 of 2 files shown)


LucaButBoring:
FYI, this PR still needs to be revised for last week's batch of spec changes, as does its [TS counterpart](https://github.com/modelcontextprotocol/typescript-sdk/pull/1041). I'll bring this out of draft state as soon as it's updated.

Kludex:
I think I mentioned on the old PR about the broker/worker architecture. Was this discussed in the SEP?

LucaButBoring:
Closing this in favor of https://github.com/modelcontextprotocol/python-sdk/pull/1645 (@maxisbey will be handling the Python SDK).

Issue #1557, opened 2025-10-31
Security: Update Starlette dependency to address vulnerabilities


ColeMurray:
## Issue

The project currently uses Starlette `>=0.27`, which includes versions with known security vulnerabilities. The dependency should be updated to `>=0.49.1` to address these issues.

## Impact

- Main project: `pyproject.toml`
- Example servers:
  - `examples/servers/simple-streamablehttp/pyproject.toml`
  - `examples/servers/simple-streamablehttp-stateless/pyproject.toml`

## Resolution

A pull request has been created to update the Starlette dependency across all affected files:
- PR: #1556

## References

- Updated version: Starlette 0.49.1
- Minimum required version: `>=0.49.1`

ColeMurray:
@Kludex the vulnerable starlette version is pinned in the uv.lock and is causing downstream consumers to pull it in as a transitive dependency.

Surely we aren't suggesting that everyone downstream of MCP should explicitly add the updated starlette version to resolve this since we're not willing to update the lock file, right?

I agree with the non-pinned update in the pyproject.toml, but not updating the lock seems not ideal

Kludex:
What are you talking about?

Who are the downstream consumers?

---

Yeah, it's okay to update the lock. But that shouldn't matter anyway as well.

PR #1556, opened 2025-10-31
Update Starlette dependency to >=0.49.1

ColeMurray:
## Summary

Updates the Starlette dependency constraint from `>=0.27` to `>=0.49.1` across the main project and example servers to address security vulnerabilities in earlier versions.

## Changes

- Updated `pyproject.toml` to require Starlette `>=0.49.1`
- Updated example server dependencies in:
  - `examples/servers/simple-streamablehttp/pyproject.toml`
  - `examples/servers/simple-streamablehttp-stateless/pyproject.toml`
- Updated `uv.lock` to reflect the new dependency version (0.49.1)

## Test Plan

- [x] Formatting checks pass (`ruff format`)
- [x] Linting checks pass (`ruff check`)
- [x] Type checking passes (`pyright`)
- [ ] CI tests will run automatically

PR #1559, opened 2025-10-31
Update Starlette to 0.49.1 in uv.lock

ColeMurray:
## Summary

Updates the locked version of Starlette from 0.47.3 to 0.49.1 to address security vulnerabilities in earlier versions.

This PR updates only the `uv.lock` file without modifying the dependency constraints in `pyproject.toml`, avoiding any potential breaking changes for downstream consumers while still providing a secure default version for direct users of this repository.

## Changes

- Updated `uv.lock` to pin Starlette to version 0.49.1
- No changes to `pyproject.toml` files (maintains `>=0.27` constraint)

## Rationale

This approach follows the principle of not pinning specific versions in library code to avoid breaking downstream consumers, while still ensuring that users who clone and install this repository directly will get the secure version by default.

## Test Plan

- [x] Formatting checks pass (`ruff format`)
- [x] Linting checks pass (`ruff check`)
- [x] Type checking passes (`pyright`)
- [ ] CI tests will run automatically

## Related

- Issue: #1557
- Alternative PR with constraint updates: #1556

### DIFF
--- a/src/mcp/cli/cli.py
+++ b/src/mcp/cli/cli.py
@@ -67,7 +67,7 @@ def _build_uv_command(
     with_editable: Path | None = None,
     with_packages: list[str] | None = None,
 ) -> list[str]:
-    """Build the uv run command that runs a MCP server through mcp run."""
+    """Build the uv run command that runs an MCP server through mcp run."""
     cmd = ["uv"]
 
     cmd.extend(["run", "--with", "mcp"])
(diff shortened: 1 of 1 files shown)


Kludex:
Same as https://github.com/modelcontextprotocol/python-sdk/pull/1552.

ColeMurray:
@Kludex the vulnerable starlette version is pinned in the uv.lock and is causing downstream consumers to pull it in as a transitive dependency.

Surely we aren't suggesting that everyone downstream of MCP should explicitly add the updated starlette version to resolve this since we're not willing to update the lock file, right?

Kludex:
> @Kludex the vulnerable starlette version is pinned in the uv.lock and is causing downstream consumers to pull it in as a transitive dependency.

Who are the downstream consumers?

Kludex:
I'm merging this to avoid spending time here.

Issue #1495, opened 2025-10-18
Protected Resource Metadata request failed: 404


daltonnyx:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

In python-sdk v1.18.0, the PR https://github.com/modelcontextprotocol/python-sdk/pull/1324 introduced new scopes selection and step-up authorization. but in the step 1 https://github.com/modelcontextprotocol/python-sdk/blob/40acbc596c727e8ccc58c69df8e93e9cf6e56a4e/src/mcp/client/auth.py#L272-L273
It raise an error if the `oauth-protected-resource` url is not found on server. this cause issue for old mcp server that does not support it yet. for example `https://mcp.linear.app/sse` or `https://mcp.atlassian.com/v1/sse`

Since we still allows discover fallback url in step 4. I suggest we should remove this exception and let the flow fall through.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
python=3.12
mcp=v1.18.0
```

dogacancolak:
Both the latest and the draft [MCP specs](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#overview) say that:

> MCP servers MUST implement OAuth 2.0 Protected Resource Metadata ([RFC9728](https://datatracker.ietf.org/doc/html/rfc9728)). MCP clients MUST use OAuth 2.0 Protected Resource Metadata for authorization server discovery.

However, I agree that we should minimize breaking changes. I checked if the lack of PRM would break anything in the later stages if we just let it fall through. Probably the most important things that PRM gives us is the Auth Server Metadata URL and which scopes to request.

Following the examples of the two MCP servers you provided, both [Linear](https://mcp.linear.app/.well-known/oauth-authorization-server) and [Atlassian](https://mcp.atlassian.com/.well-known/oauth-authorization-server) expose the Auth Server Metadata endpoint in one of the well-known fallback endpoints, so that part is good.

The scope selection strategy depends on the WWW-Auth response header, or the PRM endpoint as a fallback. Otherwise no scopes are requested. It seems that in Atlassian's `/authorize` endpoint, the user is able to manually select which scopes they want without the MCP client dynamically deciding them: https://support.atlassian.com/atlassian-rovo-mcp-server/docs/authentication-and-authorization/

[...]

PR #1562, opened 2025-10-31
fall through if PRM endpoint is not implemented

dogacancolak:
## Motivation and Context

As raised in https://github.com/modelcontextprotocol/python-sdk/issues/1495, one of the changes in https://github.com/modelcontextprotocol/python-sdk/pull/1324 included raising an error if PRM endpoint request failed. This caused backwards compatibility issues with MCP servers that haven't yet added support for PRM

## Types of changes

- [ ] Documentation update

## Checklist


dogacancolak:
> Thanks for the PR! Could you please add a unit test for this which would fail before the fix and pass with your changes

@maxisbey Thanks for the feedback! I added a unit test that closely mimics another test named `test_oauth_discovery_fallback_conditions` (right above the one I added). I have confirmed that the new unit test fails without the change in the PR, and succeeds with the change.

dogacancolak:
Hi @maxisbey, seems like you've made some extensive changes that also cover this bug. I can fix the PR checks if you think this unit test will be useful. Otherwise we can close this PR.

PR #1576, opened 2025-11-04
feat: Pass through and expose additional parameters in `ClientSessionGroup.call_tool` and `.connect_to_server`

inaku-Gyan:
Exposed additional optional parameters in the `call_tool` and `connect_to_server` methods so that they can fully pass through supported arguments to the underlying session-level methods.  

Replace the parameter `args` in `call_tool`  with `arguments` so that it's aligned with the original `ClientSession.call_tool` method.
Use `@overload` and `@deprecated` to mark `args` as deprecated while still supporting its usage.

## Motivation and Context
Previously, `call_tool` and `connect_to_server` ignored several optional parameters supported by the lower-level `ClientSession` and session methods.  
As a result, certain advanced use cases (e.g., custom timeouts, callbacks, or client metadata) could only be supported by directly accessing lower-level APIs, which was cumbersome and inconsistent.  

This change:  
- Adds new parameters to `call_tool` for timeouts, progress tracking, and metadata.  
- Adds `session_params` to `connect_to_server` to configure `ClientSession` creation directly.  
- Introduces `ClientSessionParameters` dataclass for clear typing.

## How Has This Been Tested?
Pyright static type check and pytest.

## Breaking Changes
- Not strictly breaking, as old usage patterns are still supported.  
- Minor deprecation warning: The `args` parameter in `call_tool` is deprecated in favor of `arguments`.  

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
`ClientSessionParameters` uses `dataclasses.dataclass` instead of `pydantic.BaseModel` because `BaseModel` does not support `Protocol`-typed fields. Attempts with `SkipValidation` and `arbitrary_types_allowed=True` did not resolve the issue.

### DIFF
--- a/src/mcp/client/session_group.py
+++ b/src/mcp/client/session_group.py
@@ -14,11 +14,11 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, TypeAlias
+from typing import Any, TypeAlias, overload
 
 import anyio
 from pydantic import BaseModel
-from typing_extensions import Self
+from typing_extensions import Self, deprecated
 
 import mcp
 from mcp import types
@@ -190,21 +190,45 @@ class ClientSessionGroup:
         """Returns the tools as a dictionary of names to tools."""
         return self._tools
 
+    @overload
     async def call_tool(
         self,
         name: str,
-        args: dict[str, Any],
+        arguments: dict[str, Any],
         read_timeout_seconds: timedelta | None = None,
         progress_callback: ProgressFnT | None = None,
         *,
         meta: dict[str, Any] | None = None,
+    ) -> types.CallToolResult: ...
+
+    @overload
+    @deprecated("The 'args' parameter is deprecated. Use 'arguments' instead.")
+    async def call_tool(
+        self,
+        name: str,
+        *,
+        args: dict[str, Any],
+        read_timeout_seconds: timedelta | None = None,
+        progress_callback: ProgressFnT | None = None,
+        meta: dict[str, Any] | None = None,
+    ) -> types.CallToolResult: ...
+
+    async def call_tool(
+        self,
+        name: str,
+        arguments: dict[str, Any] | None = None,
+        read_timeout_seconds: timedelta | None = None,
+        progress_callback: ProgressFnT | None = None,
+        *,
+        meta: dict[str, Any] | None = None,
+        args: dict[str, Any] | None = None,
     ) -> types.CallToolResult:
         """Executes a tool given its name and arguments."""
         session = self._tool_to_session[name]
         session_tool_name = self.tools[name].name
         return await session.call_tool(
             session_tool_name,
-            args,
+            arguments if args is None else args,
             read_timeout_seconds=read_timeout_seconds,
             progress_callback=progress_callback,
             meta=meta,
--- a/tests/client/test_session_group.py
+++ b/tests/client/test_session_group.py
@@ -67,7 +67,7 @@ class TestClientSessionGroup:
         # --- Test Execution ---
         result = await mcp_session_group.call_tool(
             name="server1-my_tool",
-            args={
+            arguments={
                 "name": "value1",
                 "args": {},
             },


BetterAndBetterII:
we need this :)

kilvil:
I need this function to refresh tools.

src/mcp/client/session_group.py
```diff
@@ -172,11 +191,49 @@ def tools(self) -> dict[str, types.Tool]:
...
+        name: str,
+        arguments: dict[str, Any],
+        read_timeout_seconds: timedelta | None = None,
+        progress_callback: ProgressFnT | None = None,
+        *,
+        meta: dict[str, Any] | None = None,
+    ) -> types.CallToolResult: ...
+
+    @overload
+    @deprecated("The 'args' parameter is deprecated. Use 'arguments' instead.")
```
kylestratis, line 205:
```suggestion
```

I'd avoid adding conventions that aren't used elsewhere in the SDK.

src/mcp/client/session_group.py
```diff
@@ -172,11 +191,49 @@ def tools(self) -> dict[str, types.Tool]:
         """Returns the tools as a dictionary of names to tools."""
         return self._tools
 
-    async def call_tool(self, name: str, args: dict[str, Any]) -> types.CallToolResult:
+    @overload
```
kylestratis, line 193:
```suggestion
```

src/mcp/client/session_group.py
```diff
@@ -225,13 +282,16 @@ async def connect_with_session(
     async def connect_to_server(
         self,
         server_params: ServerParameters,
+        session_params: ClientSessionParameters | None = None,
     ) -> mcp.ClientSession:
         """Connects to a single MCP server."""
-        server_info, session = await self._establish_session(server_params)
+        server_info, session = await self._establish_session(server_params, session_params)
```
kylestratis: session_params should follow the pattern of server_params and be required or at least default to an empty session_params object (I prefer the former for consistency's sake). This will make `_establish_session()` able to take `session_params` as a required argument, removing the need to check if `session_params()` exists and having different calls to `enter_async_context()` in lines 322-328.

kylestratis (commented):
Some feedback on the structure of this compared to the rest of the codebase and a proposed simplification.

src/mcp/client/session_group.py
inaku-Gyan: The default value is changed into `ClientSessionParameters()` as required in commit [`4ab9757` (#1576)](https://github.com/modelcontextprotocol/python-sdk/pull/1576/commits/4ab9757109931ac820425da92175ab749c8f1a6a).

src/mcp/client/session_group.py
inaku-Gyan: Thank you for your feedback!
Personally, I felt it might be more consistent to rename `args` to `arguments`, since the original method `ClientSession.call_tool` uses `arguments` as its parameter name, so I added the deprecation warning. But per your suggestion, I reverted the changes in commit [`4b9546d` (#1576)](https://github.com/modelcontextprotocol/python-sdk/pull/1576/commits/4b9546d055b63a9840fd7a1f03af9d08d05c24d3).

maxisbey: I agree with the original code to be honest. It makes more sense to be consistent here with ClientSession and rename it to `arguments`. Similarly, the convention of overloading with deprecation decorators is already in the repo for example here: https://github.com/modelcontextprotocol/python-sdk/blob/89e9c43acf7e23cf766357d776ec1ce63ac2c58e/src/mcp/client/session.py#L227

This means we're not breaking backwards compatibility while improving consistency.

src/mcp/client/session_group.py
```diff
@@ -225,13 +258,16 @@ async def connect_with_session(
     async def connect_to_server(
         self,
         server_params: ServerParameters,
+        session_params: ClientSessionParameters = ClientSessionParameters(),
     ) -> mcp.ClientSession:
         """Connects to a single MCP server."""
-        server_info, session = await self._establish_session(server_params)
+        server_info, session = await self._establish_session(server_params, session_params)
```
maxisbey: Better to use None as the default rather than a mutable object.

```suggestion
        session_params: ClientSessionParameters | None = None,
) -> mcp.ClientSession:
"""Connects to a single MCP server."""
        server_info, session = await self._establish_session(server_params, session_params or ClientSessionParameters())
```

kylestratis: @maxisbey Agreed, for some reason I was thinking that the warning against this only applied to the mutable Python primitives

src/mcp/client/session_group.py
```diff
@@ -25,6 +26,9 @@
 from mcp.client.stdio import StdioServerParameters
 from mcp.client.streamable_http import streamablehttp_client
 from mcp.shared.exceptions import McpError
+from mcp.shared.session import ProgressFnT
+
+from .session import ElicitationFnT, ListRootsFnT, LoggingFnT, MessageHandlerFnT, SamplingFnT
```
maxisbey: change this to be an absolute import, not a relative path

src/mcp/client/session_group.py
inaku-Gyan: Appreciate the heads-up! Fixed in [`dd8410f` (#1576)](https://github.com/modelcontextprotocol/python-sdk/pull/1576/commits/dd8410ff128b22847b7ee7cc2704a72348b0b27b).

src/mcp/client/session_group.py
inaku-Gyan: Changed in [`4296f76` (#1576)](https://github.com/modelcontextprotocol/python-sdk/pull/1576/commits/4296f7684580b7f9ef22fba095a9e8e9b8b1e0d4).

src/mcp/client/session_group.py
inaku-Gyan: Glad to hear that! I'll reapply it.

kylestratis: @inaku-Gyan apologies for leading you astray here

inaku-Gyan: @kylestratis No worries at all! Glad we got it sorted in the end.

PR #1586, opened 2025-11-06
refactor: extract OAuth helper functions and simplify provider state

maxisbey:
Refactors OAuth 2.0 authentication implementation to expose reusable utilities.

## Motivation and Context

The OAuth authentication logic was tightly coupled to the httpx auth provider, making it difficult for external tools (like MCP proxies or custom clients) to reuse core OAuth functionality. This refactor extracts the OAuth primitives into standalone utilities that can be used independently.

Key improvements:
- Core OAuth utilities (discovery URL generation, PKCE, token handling) are now accessible without the full auth provider
- Simplified the main OAuth2Auth class by removing internal state management
- Better separation of concerns between protocol logic and HTTP transport

## How Has This Been Tested?

Existing test suite updated to reflect new structure. All tests pass.

## Breaking Changes

None - this is a pure refactor with no API changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

New utilities added:
- `client/auth/utils.py`: OAuth protocol helpers (discovery URLs, WWW-Authenticate parsing, scope selection, response handlers)
- `shared/auth_utils.py`: PKCE generation and token expiry calculation

The main OAuth2Auth provider (`client/auth/oauth2.py`) now delegates to these utilities, reducing from 221 lines to 93 lines of core logic.

### DIFF
--- a/src/mcp/shared/auth_utils.py
+++ b/src/mcp/shared/auth_utils.py
@@ -1,9 +1,5 @@
 """Utilities for OAuth 2.0 Resource Indicators (RFC 8707) and PKCE (RFC 7636)."""
 
-import base64
-import hashlib
-import secrets
-import string
 import time
 from urllib.parse import urlparse, urlsplit, urlunsplit
 
@@ -72,38 +68,6 @@ def check_resource_allowed(requested_resource: str, configured_resource: str) ->
         configured_path += "/"
 
     return requested_path.startswith(configured_path)
-
-
-def generate_pkce_parameters(verifier_length: int = 128) -> tuple[str, str]:
-    """Generate PKCE verifier and challenge per RFC 7636.
-
-    Generates cryptographically secure code_verifier and code_challenge
-    for OAuth 2.0 PKCE (Proof Key for Code Exchange).
-
-    Args:
-        verifier_length: Length of code_verifier (43-128 chars per RFC 7636, default 128)
-
-    Returns:
-        Tuple of (code_verifier, code_challenge)
-
-    Raises:
-        ValueError: If verifier_length is not between 43 and 128
-    """
-    if not 43 <= verifier_length <= 128:
-        raise ValueError("verifier_length must be between 43 and 128 per RFC 7636")  # pragma: no cover
-
-    # Generate code_verifier using unreserved characters per RFC 7636 Section 4.1
-    # unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
-    code_verifier = "".join(
-        secrets.choice(string.ascii_letters + string.digits + "-._~") for _ in range(verifier_length)
-    )
-
-    # Generate code_challenge using S256 method per RFC 7636 Section 4.2
-    # code_challenge = BASE64URL(SHA256(ASCII(code_verifier)))
-    digest = hashlib.sha256(code_verifier.encode("ascii")).digest()
-    code_challenge = base64.urlsafe_b64encode(digest).decode("ascii").rstrip("=")
-
-    return code_verifier, code_challenge
 
 
 def calculate_token_expiry(expires_in: int | str | None) -> float | None:
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -4,8 +4,11 @@ OAuth2 Authentication implementation for HTTPX.
 Implements authorization code flow with PKCE and automatic token refresh.
 """
 
+import base64
+import hashlib
 import logging
 import secrets
+import string
 import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from dataclasses import dataclass, field
@@ -42,7 +45,6 @@ from mcp.shared.auth import (
 from mcp.shared.auth_utils import (
     calculate_token_expiry,
     check_resource_allowed,
-    generate_pkce_parameters,
     resource_url_from_server_url,
 )
 
@@ -57,8 +59,10 @@ class PKCEParameters(BaseModel):
 
     @classmethod
     def generate(cls) -> "PKCEParameters":
-        """Generate new PKCE parameters using shared util function."""
-        code_verifier, code_challenge = generate_pkce_parameters(verifier_length=128)
+        """Generate new PKCE parameters."""
+        code_verifier = "".join(secrets.choice(string.ascii_letters + string.digits + "-._~") for _ in range(128))
+        digest = hashlib.sha256(code_verifier.encode()).digest()
+        code_challenge = base64.urlsafe_b64encode(digest).decode().rstrip("=")
         return cls(code_verifier=code_verifier, code_challenge=code_challenge)
 
 


src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Extract protected resource metadata URL from WWW-Authenticate header as per RFC9728.
+
+    Returns:
+        Resource metadata URL if found in WWW-Authenticate header, None otherwise
+    """
+    if not response or response.status_code != 401:
+        return None
+
+    return extract_field_from_www_auth(response, "resource_metadata")
```
shao-anthropic, line 64: These might be too detailed functions with straight forward extraction. Do not think it would be that helpful if we adopt those.

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def build_protected_resource_discovery_urls(www_auth_url: str | None, server_url: str) -> list[str]:
```
shao-anthropic, line 67: for the PRM discovery, we can either:

1. Have an e2e flow that: hit url with expected 401 (not in the utils.py right now); construct PRM discovery URL; in a for loop to call and validate PRM (handle_protected_resource_response semi done that, but without the for loop step)
2. Or, we can have those steps as individual functions. 

I do not have strong preference which way to proceed. Choose the way that would also work for the Python SDK.

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,191 @@
...
+        return www_authenticate_scope
+    elif protected_resource_metadata is not None and protected_resource_metadata.scopes_supported is not None:
+        # Priority 2: PRM scopes_supported
+        return " ".join(protected_resource_metadata.scopes_supported)
+    else:
+        # Priority 3: Omit scope parameter
+        return None
+
+
+def get_discovery_urls(auth_server_url: str) -> list[str]:
```
shao-anthropic, line 129: this is great! Love this priority logic here.

As we discussed, we should decide should we incorporate the each oauth url discovery logic (in a for loop) in the SDK, or the caller just loop over those URLs. I do not have a strong preference, please proceed the one that can best work for the Python SDK.

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,191 @@
...
+        urls.append(urljoin(base_url, oidc_path))
+
+    # OIDC 1.0 fallback (appends to full URL per OIDC spec)
+    oidc_fallback = f"{auth_server_url.rstrip('/')}/.well-known/openid-configuration"
+    urls.append(oidc_fallback)
+
+    return urls
+
+
+async def handle_protected_resource_response(
```
shao-anthropic, line 156: need to rethink for the return here.

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,265 @@
...
+        content = await response.aread()
+        client_info = OAuthClientInformationFull.model_validate_json(content)
+        return client_info
+        # self.context.client_info = client_info
+        # await self.context.storage.set_client_info(client_info)
+    except ValidationError as e:
+        raise OAuthRegistrationError(f"Invalid registration response: {e}")
+
+
+async def handle_token_response_scopes(
```
shao-anthropic, line 229: based on the discussion, we’ll want to remove that check from the SDK

maxisbey: removed

maxisbey: cc: @pcarleton

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,250 @@
...
+    Parses token response JSON. Callers should check response.status_code before calling.
+
+    Args:
+        response: HTTP response from token endpoint (status already checked by caller)
+
+    Returns:
+        Validated OAuthToken model
+
+    Raises:
+        OAuthTokenError: If response JSON is invalid or contains unauthorized scopes
```
pcarleton:
```suggestion
        OAuthTokenError: If response JSON is invalid
```

src/mcp/shared/auth_utils.py
```diff
@@ -67,3 +72,50 @@ def check_resource_allowed(requested_resource: str, configured_resource: str) ->
         configured_path += "/"
 
     return requested_path.startswith(configured_path)
+
+
+def generate_pkce_parameters(verifier_length: int = 128) -> tuple[str, str]:
```
pcarleton: do we ever need to support different verifier lengths? seems like we can just always do 128, and cut out some complexity here

src/mcp/shared/auth_utils.py
```diff
@@ -67,3 +72,50 @@ def check_resource_allowed(requested_resource: str, configured_resource: str) ->
...
+    Raises:
+        ValueError: If verifier_length is not between 43 and 128
+    """
+    if not 43 <= verifier_length <= 128:
+        raise ValueError("verifier_length must be between 43 and 128 per RFC 7636")  # pragma: no cover
+
+    # Generate code_verifier using unreserved characters per RFC 7636 Section 4.1
+    # unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
+    code_verifier = "".join(
+        secrets.choice(string.ascii_letters + string.digits + "-._~") for _ in range(verifier_length)
```
pcarleton: optional: can use this: 
```
code_verifier = secrets.token_urlsafe(96)
```

(borrowed from : https://github.com/RomeoDespres/pkce/blob/master/pkce/__init__.py#L40C5-L40C55)

src/mcp/client/auth/utils.py
```diff
@@ -0,0 +1,250 @@
...
+    urls: list[str] = []
+    parsed = urlparse(auth_server_url)
+    base_url = f"{parsed.scheme}://{parsed.netloc}"
+
+    # RFC 8414: Path-aware OAuth discovery
+    if parsed.path and parsed.path != "/":
+        oauth_path = f"/.well-known/oauth-authorization-server{parsed.path.rstrip('/')}"
+        urls.append(urljoin(base_url, oauth_path))
+
+    # OAuth root fallback
```
pcarleton, line 140: jfyi: we need to change this, but do it carefully (see https://github.com/modelcontextprotocol/typescript-sdk/issues/1103) likely better in a follow-up, but wanted to flag as you're rolling this out.

tl;dr having this root-based above path-based OIDC means we'll get the root-level metadata when there's a path-based one we should use.

src/mcp/shared/auth_utils.py
maxisbey: Hmm not sure, I was trying to change as little as possible and just pull the functions out. Leaving it here does mean others could use custom lengths, do you think there's ever a reason to do that? Happy to remove if you think we should

src/mcp/shared/auth_utils.py
maxisbey: Would this change functionality since it's always 96 bytes whereas the current code does it up to verifier_length. I see in the code you linked it always uses 96 but also limits it to the custom length passed in.

src/mcp/client/auth/utils.py
maxisbey: made an issue here: https://github.com/modelcontextprotocol/python-sdk/issues/1623

src/mcp/client/auth/utils.py
maxisbey: removed

src/mcp/shared/auth_utils.py
pcarleton: I think this is a change:
https://github.com/modelcontextprotocol/python-sdk/blob/be730674adae118ad89806cd3011dc699fef8801/src/mcp/client/auth/oauth2.py#L56-L61

previous one had it hardcoded to 128.

also assumed that this was extracted b/c we can't use the class for some reason, but is that right? or rather what is the reason to have this separate from the `PKCEParameters` class?

maxisbey: Oh yea totally fair, this doesn't need to be separate. I'll move it back and remove the custom size.

src/mcp/shared/auth_utils.py
maxisbey: ended up reverting this anyway

pcarleton: probably best to leave the method as is.

for posterity if we want to change this in a follow-up:

96 bytes + urlsafe base64 encoding = 128 characters.

The slight difference here is that:
* current method: alphabet size is 66   (bonus chars over b64url: `~` and `.`)
* base64url encoding alphabet size: 64

So we have a slightly smaller alphabet, so slightly less entropy (8*96 = 768 vs. (log2(66)*128) = ~773), but faster to generate.

Issue #1401, opened 2025-09-25
ClientSession Error Handling


Unshure:
### Question

I work on the Strands SDK, and we have an integration with MCP where we invoke tools through an MCP client. 

Our design spins up a thread, creates an asyncio event loop, and schedules a task to create a session with an MCP server using the python-sdk mcp ClientSession: https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/mcp/mcp_client.py#L391

Then, when an LLM decides to call a tool, or multiple tools, we schedule additional tasks on this thread's event loop to call the mcp server: https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/mcp/mcp_client.py#L324-L330

We ran into an issue where when the mcp `streamablehttp_client` sse_read_timeout was lower than the time it took for a tool to return, the tool invocation tasks would hang due to an exception that was not propagated out of the ClientSession. We see an exception message (included in the additional context section), but the stack trace never enters our code, it stops in the mcp code.

After investigating further, I found that this exception is sent through the [MCP ClientSession `message_handler`](https://github.com/modelcontextprotocol/python-sdk/blob/71889d7387f070cd872cab7c9aa3d1ff1fa5a5d2/src/mcp/client/session.py#L119), and in the default implementation of this message handler, an exception is never raised: https://github.com/modelcontextprotocol/python-sdk/blob/71889d7387f070cd872cab7c9aa3d1ff1fa5a5d2/src/mcp/client/session.py#L57-L60

To work around this issue, I have introduced my own message_handler to raise an exception passed to it: https://github.com/strands-agents/sdk-python/pull/922

I wanted to know why these exceptions are not raised in the MCP ClientSession? This silent failure took a long time to debug, and I want able to find any documentation on this behavior. As a user of the ClientSession, I would expect these exceptions to be raised so that my code can handle them.

### Additional Context

Error stack trace:
```
Error reading SSE stream:
Traceback (most recent call last):
  File "/Users/ncclegg/Library/Application Support/hatch/env/virtual/strands-agents/X7vsTQrp/strands-agents/lib/python3.13/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File "/Users/ncclegg/Library/Application Support/hatch/env/virtual/strands-agents/X7vsTQrp/strands-agents/lib/python3.13/site-packages/httpx/_transports/default.py", line 271, in __aiter__
    async for part in self._httpcore_stream:
[...]

PR #1595, opened 2025-11-08
fix: Raise exceptions in default ClientSession message handler (#1401)

matthew-gries:
Fixes #1401.
 - Logs and raises exceptions in the default ClientSession message handler. Includes test that verifies exceptions passed to the default handler are properly raised.
 - Makes session more resilient to stale server responses by logging warnings (instead of raising exceptions) when responses with unknown request IDs are received. This handles the case where the client times out and cleans up a pending request, but the server responds late (see `test_notification_validation_error()`). Previously, a RuntimeError was raised for unknown IDs, but it was swallowed by the old default message handler. With the new exception-raising behavior, keeping the RuntimeError would crash the session on benign timeout scenarios.

## Motivation and Context
It solves the issue described in #1401 where exceptions in the default message handler for the ClientSession were being swallowed, which leads to silent failures that are difficult to debug.

## How Has This Been Tested?

Added unit test for this, also `test_notification_validation_error()` in `test_88_random_error.py`

## Breaking Changes
Technically yes because those who use the default message handler and were not receiving silent errors will now start seeing them. Those who override the message handler are not affected.

## Types of changes

- [ ] Documentation update

## Checklist


maxisbey:
Thanks for your PR! Closing this in favour of #2122 as it fixes the issue at the transport level and also addresses the related reconnection-loop bug identified in the issue discussion.

PR #1607, opened 2025-11-10
Fix compatibility with pydantic 2.12+ Field defaults in Annotated types

maxisbey:
This PR fixes a compatibility issue with pydantic 2.12+ where fields using `Annotated[T, Field(default)]` without explicit parameter defaults were incorrectly marked as required in JSON schemas.

## Motivation and Context
Pydantic 2.12.0 introduced a breaking change in how `FieldInfo.from_annotated_attribute()` handles defaults. When passing `PydanticUndefined` as the second parameter, it now overwrites any default value specified inside `Field()` within an `Annotated` type. This causes fields that should be optional (with defaults from `Field()`) to be incorrectly marked as required in the generated JSON schema.

The issue affects function parameters like:
```python
def my_function(
    param: Annotated[int, Field(1, description="Has default")]
):
    ...
```

With pydantic 2.12+, this parameter would be marked as required in the JSON schema, even though `Field(1)` specifies a default value.

## How Has This Been Tested?
- Existing test `test_complex_function_json_schema` validates the fix (tests fields with `Field(default)` in Annotated types)
- All 31 tests in `test_func_metadata.py` pass
- The fix maintains backward compatibility with pydantic 2.11.x

The solution detects when an `Annotated` type contains a `FieldInfo` with a default and uses `FieldInfo.from_annotation()` to preserve that default, while continuing to use `from_annotated_attribute()` for standard cases where parameter defaults should take precedence.

## Breaking Changes
None. This is a bug fix that maintains backward compatibility with pydantic 2.11 while adding forward compatibility with pydantic 2.12+.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
This fix prepares the codebase for pydantic 2.12+, which will be included in dependency updates. The change is minimal and focused on preserving the intended behavior: defaults specified in `Field()` within `Annotated` types should be respected in the JSON schema.

The implementation checks for the presence of a `FieldInfo` with a non-undefined default in the `Annotated` metadata and chooses the appropriate `FieldInfo` construction method accordingly.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,5 @@
 # Applied 120 line-length rule to all files: https://github.com/modelcontextprotocol/python-sdk/pull/856
 543961968c0634e93d919d509cce23a1d6a56c21
+
+# Added 100% code coverage baseline with pragma comments: https://github.com/modelcontextprotocol/python-sdk/pull/1553
+89e9c43acf7e23cf766357d776ec1ce63ac2c58e
--- a/examples/servers/everything-server/mcp_everything_server/server.py
+++ b/examples/servers/everything-server/mcp_everything_server/server.py
@@ -166,6 +166,38 @@ async def test_elicitation(message: str, ctx: Context[ServerSession, None]) -> s
         return f"Elicitation not supported or error: {str(e)}"
 
 
+class SEP1034DefaultsSchema(BaseModel):
+    """Schema for testing SEP-1034 elicitation with default values for all primitive types"""
+
+    name: str = Field(default="John Doe", description="User name")
+    age: int = Field(default=30, description="User age")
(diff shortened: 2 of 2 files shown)


Viicos:
@maxisbey I'm on mobile but I think https://github.com/modelcontextprotocol/python-sdk/pull/1496 actually fixes it, by avoiding relying on the `FieldInfo` methods that are not meant to be used externally/are not behaving as users would expect.

maxisbey:
> @maxisbey I'm on mobile but I think #1496 actually fixes it, by avoiding relying on the `FieldInfo` methods that are not meant to be used externally/are not behaving as users would expect.

Oh perfect, I'd be keen to get that in then! Do you have time to fix the merge conflicts?

PR #1609, opened 2025-11-10
Fix CI highest resolution test to actually test highest versions

maxisbey:
This PR fixes the CI dependency resolution testing to properly test against the highest available dependency versions, which would have caught the pydantic 2.12 breaking change before it reached main.

## Motivation and Context
The CI test matrix includes a "highest" resolution test intended to catch compatibility issues with newer dependency versions. However, it was using the `--frozen` flag, which prevented uv from resolving to the highest available versions. Instead, it just used whatever was locked in `uv.lock` from main, defeating the purpose of testing against latest versions.

This meant breaking changes in newer dependency versions (like pydantic 2.12's `FieldInfo` refactor) that were compatible with our version constraints but not yet in the lockfile would not be caught by CI.

## How Has This Been Tested?
The changes are to CI configuration only. The fix:
- Uses `--resolution highest` during `uv sync` for the highest resolution test
- Uses `--resolution lowest-direct` for the lowest resolution test (unchanged behavior)
- Simplifies `uv run` to use `--frozen --no-sync` since resolution already happened during sync
- Removes the now-unnecessary `UV_RESOLUTION` environment variable

This ensures the highest test matrix actually resolves to and tests against the highest available versions within our constraints.

## Breaking Changes
None. This is a CI configuration improvement that makes the tests work as originally intended.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
This fix is related to the pydantic 2.12 compatibility issue discovered in the investigation. With this change, future dependency updates that introduce breaking changes will be caught by the "highest" resolution CI tests before merging to main.

The fix clarifies the intent of each step:
1. `uv sync` with resolution flags → resolve and install dependencies
2. `uv run --frozen` → run tests with what was synced (no re-resolution)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,8 @@ dependencies = [
     "jsonschema>=4.20.0",
     "pywin32>=310; sys_platform == 'win32'",
     "pyjwt[crypto]>=2.10.1",
+    "typing-extensions>=4.9.0",
+    "typing-inspection>=0.4.1",
 ]
 
 [project.optional-dependencies]
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -1,10 +1,9 @@
 import inspect
 import json
-import types
 from collections.abc import Awaitable, Callable, Sequence
 from itertools import chain
 from types import GenericAlias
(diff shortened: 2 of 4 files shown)


maxisbey:
The tests are currently failing, which is intended. This PR will fix the tests: https://github.com/modelcontextprotocol/python-sdk/pull/1607

PR #1624, opened 2025-11-13
Fix OAuth discovery fallback and URL ordering

### LINKED ISSUE #1495: Protected Resource Metadata request failed: 404

### LINKED ISSUE #1623: Review OAuth discovery URL priority order for path-based vs root-based OIDC
## Problem

The OAuth authorization server discovery URL priority in `get_discovery_urls()` (`src/mcp/client/auth/utils.py:129-153`) may not correctly prioritize path-based OIDC discovery over root-based discovery, which can cause issues with certain OAuth providers.

## Context

This issue was flagged during review of PR #1586 (comment: https://github.com/modelcontextprotocol/python-sdk/pull/1586#discussion_r2519218950) by @pcarleton:

> jfyi: we need to change this, but do it carefully (see https://github.com/modelcontextprotocol/typescript-sdk/issues/1103) likely better in a follow-up, but wanted to flag as you're rolling this out.
> 
> tl;dr having this root-based above path-based OIDC means we'll get the root-level metadata when there's a path-based one we should use.

The TypeScript SDK issue (#1103 in typescript-sdk) shows that changing discovery URL priority in v1.21.1 broke Jira MCP OAuth because the root-level well-known URLs returned 404 when path-based URLs should have been tried first.

## Current Behavior

The current discovery URL order in `get_discovery_urls()` is:

1. Path-aware OAuth: `/.well-known/oauth-authorization-server{path}`
2. **Root OAuth**: `/.well-known/oauth-authorization-server` 
3. Path-aware OIDC: `/.well-known/openid-configuration{path}`
4. OIDC fallback: `{server_url}/.well-known/openid-configuration`

For a server like `https://mcp.atlassian.com/v1/sse`, this tries root-based OAuth discovery (#2) before path-based OIDC discovery (#3-4), which may fetch root-level metadata when path-based metadata should be preferred.

## Desired Behavior

The discovery order should prioritize path-based OIDC appropriately, potentially before root-based OAuth discovery, to ensure we use the correct metadata for path-scoped OAuth providers.

## Why This Needs Careful Handling

- This affects the core OAuth discovery flow
- Incorrect ordering can break OAuth with certain providers (as seen in TS SDK)
- The RFC 8414 specification needs to be carefully reviewed for correct priority
- Changes should be tested against multiple OAuth provider configurations

## References

- PR #1586: OAuth helper extraction refactor
- TypeScript SDK issue: https://github.com/modelcontextprotocol/typescript-sdk/issues/1103
- RFC 8414 (OAuth 2.0 Authorization Server Metadata): https://www.rfc-editor.org/rfc/rfc8414.html
- RFC 8414 Section 5 (Path-aware OIDC): https://www.rfc-editor.org/rfc/rfc8414.html#section-5
- MCP Auth Spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization

## Related Code

`src/mcp/client/auth/utils.py:129-153` - `get_discovery_urls()` function

maxisbey:
## Motivation and Context

This PR addresses two related OAuth discovery issues that affect backward compatibility and correctness:

### Issue #1495: Legacy Server Compatibility
Version 1.18.0 broke compatibility with legacy MCP servers (like Linear and Atlassian) that don't yet support Protected Resource Metadata (PRM) endpoints. The implementation raised an exception when PRM discovery failed, preventing fallback to the March 2025 spec behavior.

### Issue #1623: Incorrect OAuth Discovery URL Priority
The OAuth discovery URL ordering was incorrect, causing the system to try root-level URLs before path-based OIDC URLs. This broke authentication with services like Jira MCP where tenant-specific authorization servers should take precedence over root-level servers.

### 1. Enable Fallback to March 2025 Spec (Issue #1495)
- **Removed exception** when PRM discovery fails completely
- **Added fallback logic** to use root-based OAuth discovery for backward compatibility
- When PRM unavailable, only check `/.well-known/oauth-authorization-server` at the root
- Maintains compatibility with legacy servers like Linear (`https://mcp.linear.app/sse`) and Atlassian (`https://mcp.atlassian.com/v1/sse`)

### 2. Fix OAuth Discovery URL Ordering (Issue #1623)
- **Path-based discovery only when needed**: Only check path-based URLs when the authorization server URL (from PRM) contains a path
- **Prevents incorrect root discovery**: Avoids mistakenly discovering root AS when tenant-specific AS exists (e.g., Okta multi-tenant setups)
- **Follows RFC 8414 priority**: Path-aware OAuth → Path-aware OIDC → OIDC appended format → Root (only when no path)
- **Fixes TypeScript SDK parity issue**: Aligns with the fix made in TypeScript SDK v1.21.1

### Implementation Details

**Function Renaming:**
- `build_protected_resource_discovery_urls` → `build_protected_resource_metadata_discovery_urls`
- `get_discovery_urls` → `build_oauth_authorization_server_metadata_discovery_urls`

**New Behavior:**
- `build_oauth_authorization_server_metadata_discovery_urls(auth_server_url: str | None, server_url: str)`
  - **Legacy path** (`auth_server_url` is `None`): Only try `{root}/.well-known/oauth-authorization-server`
  - **Path-aware** (`auth_server_url` has path like `/tenant1`): Only try path-based URLs in priority order
  - **Root-only** (`auth_server_url` is root like `https://auth.example.com`): Only try root URLs

### Example 1: Legacy Server (No PRM)
```
Server: https://mcp.linear.app/sse
1. Try PRM URLs (all fail):
   - /.well-known/oauth-protected-resource/sse
   - /.well-known/oauth-protected-resource
2. Fall back to legacy OAuth discovery:
   - https://mcp.linear.app/.well-known/oauth-authorization-server ✓
```

### Example 2: Path-Aware Discovery
```
Server: https://api.example.com/v1/mcp
PRM returns: https://auth.example.com/tenant1
1. Try path-aware URLs only:
   - https://auth.example.com/.well-known/oauth-authorization-server/tenant1
   - https://auth.example.com/.well-known/openid-configuration/tenant1
   - https://auth.example.com/tenant1/.well-known/openid-configuration
2. Does NOT try root (prevents wrong AS discovery)
```

### Example 3: Root Authorization Server
```
Server: https://api.example.com/mcp
PRM returns: https://auth.example.com
1. Try root URLs only:
   - https://auth.example.com/.well-known/oauth-authorization-server
   - https://auth.example.com/.well-known/openid-configuration
```

## Breaking Changes

Some **invalid server configurations** will no longer work (these violate RFC specifications):
1. No PRM available AND OASM at a path other than root
2. PRM returns auth server URL with path, but OASM only exists at root

These configurations are not expected to exist in practice.

## Testing

- Updated existing tests to use new function signatures
- Added test for path-aware discovery behavior
- All pre-commit hooks pass (formatting, linting, type checking)

## Additional Context

- Aligns with SEP-985 (Align OAuth 2.0 Protected Resource Metadata with RFC 9728)
- Follows RFC 8414 discovery priorities
- Matches TypeScript SDK behavior after their similar fix
- Community consensus from issue discussions supports this approach

## Related Issues

Fixes #1495
Fixes #1623

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -130,6 +130,7 @@ class OAuthMetadata(BaseModel):
     introspection_endpoint_auth_methods_supported: list[str] | None = None
     introspection_endpoint_auth_signing_alg_values_supported: list[str] | None = None
     code_challenge_methods_supported: list[str] | None = None
+    client_id_metadata_document_supported: bool | None = None
 
 
 class ProtectedResourceMetadata(BaseModel):


pcarleton (approved):
great, LGTM

PR #1626, opened 2025-11-13
Revert "Fix CI highest resolution test to actually test highest versions"

lucianaliorarountree-cmd:
Reverts modelcontextprotocol/python-sdk#1609


maxisbey:
Why revert this?

PR #1631, opened 2025-11-15
fix get_client_metadata_scopes on 401

### LINKED ISSUE #1630: OAuth2: Protected-resource URL passed as scope instead of parsed scop
### Initial Checks

### Description

I noticed that the scope parameter is being set to the OAuth protected-resource URL instead of the parsed scope (or None when no scope is provided).

After inspecting the code, it seems the URL is passed directly as the scope here:

https://github.com/modelcontextprotocol/python-sdk/blob/9724ad1ce66c2a4f18c7539ef589d4d3b9aebd82/src/mcp/client/auth/oauth2.py#L508-L512

This results in an incorrect scope value being propagated during OAuth2 authentication.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
1.21.1
```

abliznyuk:
This PR fixes an issue where the OAuth2 client passed the protected-resource URL as the `scope` value during token requests. The implementation now correctly uses the parsed `scope` (or `None` when no scope is provided).
Fix #1630

## Motivation and Context

The previous behavior incorrectly substituted the OAuth protected-resource URL for the `scope` parameter, leading to malformed token requests and potential authorization failures. By ensuring that the correct scope is passed, OAuth2 flows behave as expected and remain compliant with spec and provider expectations.

## How Has This Been Tested?

The fix was tested against an OAuth2 provider in a real application environment. Token requests were validated to ensure that:
- the correct scope value is included,
- requests without an explicit scope omit the field rather than using the resource URL,
- no regressions occur in authenticated MCP interactions.

## Breaking Changes

No breaking changes. This is a bug fix that restores correct behavior and should not require user code changes.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The problematic logic was located around the OAuth2 request-building code near lines 508–512, where the URL was incorrectly forwarded as the scope. The updated implementation ensures correct scope derivation and preserves compatibility with standard OAuth2 servers.

### DIFF
--- a/tests/client/test_scope_bug_1630.py
+++ b/tests/client/test_scope_bug_1630.py
@@ -1,0 +1,166 @@
+"""
+Regression test for issue #1630: OAuth2 scope incorrectly set to resource_metadata URL.
+
+This test verifies that when a 401 response contains both resource_metadata and scope
+in the WWW-Authenticate header, the actual scope is used (not the resource_metadata URL).
+"""
+
+from unittest import mock
+
+import httpx
+import pytest
+from pydantic import AnyUrl
+
+from mcp.client.auth import OAuthClientProvider
+from mcp.shared.auth import OAuthClientInformationFull, OAuthClientMetadata, OAuthToken
+
+
+class MockTokenStorage:
+    """Mock token storage for testing."""
+
+    def __init__(self) -> None:
+        self._tokens: OAuthToken | None = None
+        self._client_info: OAuthClientInformationFull | None = None
+
+    async def get_tokens(self) -> OAuthToken | None:
+        return self._tokens  # pragma: no cover
+
+    async def set_tokens(self, tokens: OAuthToken) -> None:
+        self._tokens = tokens
+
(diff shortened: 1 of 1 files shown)


src/mcp/client/auth/oauth2.py
```diff
@@ -506,7 +506,7 @@ async def async_auth_flow(self, request: httpx.Request) -> AsyncGenerator[httpx.
 
                     # Step 3: Apply scope selection strategy
                     self.context.client_metadata.scope = get_client_metadata_scopes(
-                        www_auth_resource_metadata_url,
+                        extract_scope_from_www_auth(response),
```
maxisbey, line 509: Please add a unit test for this

maxisbey (changes requested):

src/mcp/client/auth/oauth2.py
maxisbey: will add myself to unblock and get this merged :)

maxisbey:
Thank you for this fix! I've gone ahead and added a unit test which fails on main just to confirm the fix.

PR #1654, opened 2025-11-23
Add SSE polling support (SEP-1699)

felixweinberger:
## Summary
Implements SEP-1699 which enables servers to disconnect SSE connections at will by sending priming events and retry fields.

## Motivation and Context
[SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1783) introduces SSE polling behavior that allows servers to control client reconnection timing and close connections gracefully. This enables more efficient resource management on the server side while maintaining resumability.

We implement this on the `POST` SSE stream as implied by the SEP language linked above. I.e. when a server establishes an SSE stream:

1. It's first message will be an event including no data, only an event ID.
2. After that, it may call `close_sse_stream` to close the stream while still gathering the events.
3. The client can start "polling" the SSE stream based on the `retryInterval` supplied by the server before disconnection.

## How Has This Been Tested?
- Added e2e integration tests demonstrating the initial priming message, disconnection initiated by the server, and subsequent reconnection via polling by the client.
- Example server & client demonstrating polling
- [currently WIP] upcoming conformance tests in https://github.com/modelcontextprotocol/conformance/pull/47

Example server and client:

CleanShot 2025-11-24 at 21 33 16

## Breaking Changes
None. Client falls back to exponential backoff if no retry field is provided.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -1,7 +1,8 @@
 import logging
+from collections.abc import Callable
 from contextlib import asynccontextmanager
 from typing import Any
-from urllib.parse import urljoin, urlparse
+from urllib.parse import parse_qs, urljoin, urlparse
 
 import anyio
 import httpx
@@ -21,6 +22,11 @@ def remove_request_params(url: str) -> str:
     return urljoin(url, urlparse(url).path)
 
 
+def _extract_session_id_from_endpoint(endpoint_url: str) -> str | None:
+    query_params = parse_qs(urlparse(endpoint_url).query)
+    return query_params.get("sessionId", [None])[0] or query_params.get("session_id", [None])[0]
+
+
 @asynccontextmanager
 async def sse_client(
     url: str,
@@ -29,6 +35,7 @@ async def sse_client(
     sse_read_timeout: float = 60 * 5,
     httpx_client_factory: McpHttpClientFactory = create_mcp_http_client,
     auth: httpx.Auth | None = None,
+    on_session_created: Callable[[str], None] | None = None,
 ):
     """
     Client transport for SSE.
@@ -42,6 +49,7 @@ async def sse_client(
         timeout: HTTP timeout for regular operations.
         sse_read_timeout: Timeout for SSE read operations.
         auth: Optional HTTPX authentication handler.
+        on_session_created: Optional callback invoked with the session ID when received.
     """
     read_stream: MemoryObjectReceiveStream[SessionMessage | Exception]
     read_stream_writer: MemoryObjectSendStream[SessionMessage | Exception]
@@ -88,6 +96,11 @@ async def sse_client(
                                             )
                                             logger.error(error_msg)  # pragma: no cover
                                             raise ValueError(error_msg)  # pragma: no cover
+
+                                        if on_session_created:
+                                            session_id = _extract_session_id_from_endpoint(endpoint_url)
+                                            if session_id:
+                                                on_session_created(session_id)
 
                                         task_status.started(endpoint_url)
 
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -4,6 +4,7 @@ import socket
 import time
 from collections.abc import AsyncGenerator, Generator
 from typing import Any
+from unittest.mock import Mock
 
 import anyio
 import httpx
@@ -16,9 +17,10 @@ from starlette.requests import Request
 from starlette.responses import Response
 from starlette.routing import Mount, Route
 
+import mcp.client.sse
 import mcp.types as types
 from mcp.client.session import ClientSession
-from mcp.client.sse import sse_client
+from mcp.client.sse import _extract_session_id_from_endpoint, sse_client
 from mcp.server import Server
 from mcp.server.sse import SseServerTransport
 from mcp.server.transport_security import TransportSecuritySettings
@@ -182,6 +184,57 @@ async def test_sse_client_basic_connection(server: None, server_url: str) -> Non
             # Test ping
             ping_result = await session.send_ping()
             assert isinstance(ping_result, EmptyResult)
+
+
+@pytest.mark.anyio
+async def test_sse_client_on_session_created(server: None, server_url: str) -> None:
+    captured_session_id: str | None = None
+
+    def on_session_created(session_id: str) -> None:
+        nonlocal captured_session_id
+        captured_session_id = session_id
+
+    async with sse_client(server_url + "/sse", on_session_created=on_session_created) as streams:
+        async with ClientSession(*streams) as session:
+            result = await session.initialize()
+            assert isinstance(result, InitializeResult)
+
+    assert captured_session_id is not None
+    assert len(captured_session_id) > 0
+
+
+@pytest.mark.parametrize(
+    "endpoint_url,expected",
+    [
+        ("/messages?sessionId=abc123", "abc123"),
+        ("/messages?session_id=def456", "def456"),
+        ("/messages?sessionId=abc&session_id=def", "abc"),
+        ("/messages?other=value", None),
+        ("/messages", None),
+        ("", None),
+    ],
+)
+def test_extract_session_id_from_endpoint(endpoint_url: str, expected: str | None) -> None:
+    assert _extract_session_id_from_endpoint(endpoint_url) == expected
+
+
+@pytest.mark.anyio
+async def test_sse_client_on_session_created_not_called_when_no_session_id(
+    server: None, server_url: str, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    callback_mock = Mock()
+
+    def mock_extract(url: str) -> None:
+        return None
+
+    monkeypatch.setattr(mcp.client.sse, "_extract_session_id_from_endpoint", mock_extract)
+
+    async with sse_client(server_url + "/sse", on_session_created=callback_mock) as streams:
+        async with ClientSession(*streams) as session:
+            result = await session.initialize()
+            assert isinstance(result, InitializeResult)
+
+    callback_mock.assert_not_called()
 
 
 @pytest.fixture


src/mcp/client/streamable_http.py
```diff
@@ -415,12 +415,12 @@ async def _handle_sse_response(
                 await ctx.read_stream_writer.send(e)
 
         # Auto-reconnect if stream ended without completion and we have priming event
-        if not is_complete and has_priming_event and last_event_id:
```
felixweinberger: For the life of me I couldn't get these code paths covered - I'm not sure why, as the example client and server demonstrate this works and the conformance tests pass too:

CleanShot 2025-11-24 at 22 05 09

Will try again in a follow-up but would prefer not to block on this...

src/mcp/client/streamable_http.py
```diff
@@ -150,16 +162,54 @@ def _maybe_extract_protocol_version_from_message(
...
+            Tuple of (is_complete, has_event_id) where:
+            - is_complete: True if the response stream is complete (got response/error)
+            - has_event_id: True if this event had an ID (indicating resumability)
+        """
+        event_id = sse.id  # httpx_sse defaults to "" for missing ID
+        has_event_id = bool(event_id)  # True if non-empty string
+
+        # Capture server-provided retry value for reconnection timing
+        if sse.retry is not None:  # pragma: no cover
+            self._server_retry_seconds = sse.retry / 1000.0  # Convert ms to seconds
```
crondinini-ant: is there any specific reason for converting this to seconds instead of storing it in ms?

src/mcp/client/streamable_http.py
```diff
@@ -150,16 +162,54 @@ def _maybe_extract_protocol_version_from_message(
...
+            Time to wait in seconds before next reconnection attempt
+        """
+        # Use server-provided retry value if available
+        if self._server_retry_seconds is not None:
+            return self._server_retry_seconds
+
+        # Fall back to exponential backoff
+        opts = self.reconnection_options
+        delay = opts.initial_reconnection_delay * (opts.reconnection_delay_grow_factor**attempt)
+        return min(delay, opts.max_reconnection_delay)
```
crondinini-ant: have you considered using a callable instead of/in addition to these options? I think different clients might want to use different strategies.

src/mcp/client/streamable_http.py
crondinini-ant: also, would it make sense to double check this is not greater than the max_reconnection_delay?

src/mcp/client/streamable_http.py
crondinini-ant: I'm saying this because when I had to implement backoff strategies, there were different strategies to consider, i.e.: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

src/mcp/client/streamable_http.py
```diff
@@ -321,25 +375,98 @@ async def _handle_sse_response(
...
                     ctx.read_stream_writer,
                     resumption_callback=(ctx.metadata.on_resumption_token_update if ctx.metadata else None),
                     is_initialization=is_initialization,
                 )
-                # If the SSE event indicates completion, like returning respose/error
+
+                # Track priming events
+                if has_event_id:
+                    has_priming_event = True
+                    last_event_id = sse.id
```
crondinini-ant: wonder if you need both? if I understand this correctly, they represent the same thing, or at least they are always true at the same time.

src/mcp/client/streamable_http.py
```diff
@@ -321,25 +375,98 @@ async def _handle_sse_response(
...
+            logger.exception("Error reading SSE stream:")
+            # Don't send exception if we can reconnect
+            if not (has_priming_event and last_event_id):
+                await ctx.read_stream_writer.send(e)
+
+        # Auto-reconnect if stream ended without completion and we have priming event
+        if not is_complete and has_priming_event and last_event_id:  # pragma: no cover
+            await self._attempt_sse_reconnection(ctx, last_event_id, attempt)
+
+        return has_priming_event, last_event_id
```
crondinini-ant: I'm curious why you need to return, I'm prob missing something but I can't see where they are used.

src/mcp/client/streamable_http.py
```diff
@@ -321,25 +375,98 @@ async def _handle_sse_response(
...
+            error_msg = f"Max reconnection attempts ({max_retries}) exceeded"
+            logger.error(error_msg)
+            await ctx.read_stream_writer.send(StreamableHTTPError(error_msg))
+            return
+
+        # Calculate delay (uses server retry if available, else exponential backoff)
+        delay = self._get_next_reconnection_delay(attempt)
+        logger.info(f"SSE stream closed, reconnecting in {delay:.1f}s (attempt {attempt + 1}/{max_retries})")
+
+        await anyio.sleep(delay)
```
crondinini-ant: I wonder if the consumer might also want to choose the sleep function or if this is an universal way that all python codebases do sleep?

src/mcp/client/streamable_http.py
```diff
@@ -442,6 +569,61 @@ def get_session_id(self) -> str | None:
...
+
+        try:
+            async with aconnect_sse(
+                client,
+                "GET",
+                self.url,
+                headers=headers,
+                timeout=httpx.Timeout(self.timeout, read=self.sse_read_timeout),
+            ) as event_source:
+                event_source.response.raise_for_status()
```
crondinini-ant: can you read the response body before raising? this allows the consumer to access the error response, whereas just raising swallows the response.

src/mcp/client/streamable_http.py
```diff
@@ -442,6 +569,61 @@ def get_session_id(self) -> str | None:
...
+
+                async for sse in event_source.aiter_sse():  # pragma: no cover
+                    _is_complete, has_event_id = await self._handle_sse_event(
+                        sse,
+                        read_stream_writer,
+                        resumption_callback=on_resumption_token,
+                    )
+
+                    # Call resumption callback if we have a new event ID
+                    if has_event_id and sse.id and on_resumption_token:
```
crondinini-ant: in which case would we have a new event ID? is it a valid use case to have a new event ID while resuming? how should consumers interpret this?

src/mcp/client/streamable_http.py
```diff
@@ -442,6 +569,61 @@ def get_session_id(self) -> str | None:
...
+                timeout=httpx.Timeout(self.timeout, read=self.sse_read_timeout),
+            ) as event_source:
+                event_source.response.raise_for_status()
+                logger.debug(f"Resumed SSE stream from event ID: {last_event_id}")  # pragma: no cover
+
+                async for sse in event_source.aiter_sse():  # pragma: no cover
+                    _is_complete, has_event_id = await self._handle_sse_event(
+                        sse,
+                        read_stream_writer,
+                        resumption_callback=on_resumption_token,
```
crondinini-ant: what does token mean in this context?

src/mcp/server/streamable_http.py
```diff
@@ -669,6 +716,27 @@ async def terminate(self) -> None:
...
+        """
+        request_id_str = str(request_id)
+        if request_id_str in self._request_streams:
+            try:
+                sender, receiver = self._request_streams[request_id_str]
+                await sender.aclose()
+                await receiver.aclose()
+            except Exception:  # pragma: no cover
+                # Stream might already be closed
+                logger.debug(f"Error closing SSE stream {request_id_str} - may already be closed")
```
crondinini-ant: if we don't re-raise here, does this exception get lost?

src/mcp/client/streamable_http.py
felixweinberger: > is there any specific reason for converting this to seconds instead of storing it in ms?

main reason is that's what the spec uses + in Python stdlib durations are generally stored as floats in seconds

> also, would it make sense to double check this is not greater than the max_reconnection_delay?

yeah good idea, added in `__post_init__`

src/mcp/client/streamable_http.py
felixweinberger: interesting idea, seems like something we could do in a follow-up if people want it?

src/mcp/client/streamable_http.py
felixweinberger: Good point - I originally had these separate because technically it's

> The server SHOULD immediately send an SSE event consisting of an event ID and an empty data field in order to prime the client to reconnect (using that event ID as Last-Event-ID).

I.e. there might not be a priming event and last event id could come from something that's not a priming event. Will rework this a bit.

src/mcp/client/streamable_http.py
felixweinberger: refactoring left-over, good catch 🫠

src/mcp/client/streamable_http.py
felixweinberger: The SDK uses `anyio`, so I wouldn't introduce the complexity of supporting other sleep functions here - not sure what the benefit would be?

src/mcp/client/streamable_http.py
felixweinberger: it's meant like a "pagination token" basically. We could also call it `on_last_event_id` I guess, as that's basically what we're passing in

src/mcp/server/streamable_http.py
felixweinberger: It does - I was going off the `terminate()` implementation above. Arguably for the `close_sse_stream` though this should behave differently and actually throw if this fails 🤔 because we're not actually in a "shutdown phase". Going to amend.

felixweinberger:
Reworking this from scratch.

src/mcp/client/streamable_http.py
felixweinberger: I actually decided to simplify this instead.

src/mcp/client/streamable_http.py
felixweinberger: Changed to match sse spec which is in ms

src/mcp/client/streamable_http.py
felixweinberger: simplified

src/mcp/client/streamable_http.py
felixweinberger: approach has changed, this is now different

src/mcp/client/streamable_http.py
felixweinberger: this is simplified in the new version

src/mcp/server/streamable_http.py
felixweinberger: made this catch a ClosedResource instead in the new version - that's expected because we explicitly `close_sse_stream` now

felixweinberger:
Reworked this from the ground by going from tests first.

The best way to review this is probably look at the tests in `tests/shared/test_streamable_http.py` which describe the expected behavior introduced by SEP-1699

Then looking through `server/streamable_http.py` where we introduced `close_sse_stream` and `streamable_http.py` where we actually perform the automatic reconnection via GET stream in `_handle_reconnection`

The `close_sse_stream` is available via the `RequestContext` so the server doesn't have to reach down into the transport.

felixweinberger:
I updated the examples as well to clearly demonstrate how a stream can be disconnected by the server, triggering a reconnection & finally completing the request:

CleanShot 2025-11-27 at 18 33 05

examples/servers/sse-polling-demo/mcp_sse_polling_demo/server.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @app.call_tool()
+    async def call_tool(name: str, arguments: dict[str, Any]) -> list[types.ContentBlock]:
+        """Handle tool calls."""
+        ctx = app.request_context
+
+        if name == "process_batch":
+            items = arguments.get("items", 10)
+            checkpoint_every = arguments.get("checkpoint_every", 3)
+
```
crondinini-ant, line 65: I've ran this example with a proxy in the middle, saved the request/results and I checked the event IDs being used. I noticed that the priming event ID is not used in this scenario because the connection is closed after sending messages, so the client ends up using the event ID from the last message, not from the priming event.
I'm not sure if the intention was to test the usage of the priming event, but if we wanted to add a test to this, then my local test with Claude indicates you'd have to close a connection before sending the first message:
```
            if test_priming and ctx.close_sse_stream:
                logger.info("Testing priming: closing SSE stream immediately (before any events)")
                await ctx.close_sse_stream()
```

it does work successfully btw, this is more of a comment in case we wanted to add a specific flag to test this.

src/mcp/server/streamable_http.py
```diff
@@ -459,10 +529,16 @@ async def _handle_post_request(self, scope: Scope, request: Request, receive: Re
...
 
+                # Store writer reference so close_sse_stream() can close it
+                self._sse_stream_writers[request_id] = sse_stream_writer
+
                 async def sse_writer():
                     # Get the request ID from the incoming request message
                     try:
                         async with sse_stream_writer, request_stream_reader:
+                            # Send priming event for SSE resumability
+                            await self._send_priming_event(request_id, sse_stream_writer)
```
crondinini-ant, line 563: [future improvement idea] not something for now, but as I understand it, a context manager would be nice here to make the code even cleaner and ensure we don't forget to clean up the writers. i.e.:
```python
  @asynccontextmanager
  async def _managed_sse_stream(self, request_id, sse_writer, request_reader):
      self._sse_stream_writers[request_id] = sse_writer
      try:
          async with sse_writer, request_reader:
              yield
      finally:
          self._sse_stream_writers.pop(request_id, None)
          await self._clean_up_memory_streams(request_id)

  Usage:
  async with self._managed_sse_stream(request_id, sse_stream_writer, request_stream_reader):
      ...
```

src/mcp/server/streamable_http.py
```diff
@@ -502,8 +582,7 @@ async def sse_writer():
                     async with anyio.create_task_group() as tg:
                         tg.start_soon(response, scope, receive, send)
                         # Then send the message to be processed by the server
-                        metadata = ServerMessageMetadata(request_context=request)
-                        session_message = SessionMessage(message, metadata=metadata)
+                        session_message = self._create_session_message(message, request, request_id)
```
crondinini-ant, line 608: _create_session_message  was a nice solution for passing around the close stream

src/mcp/server/streamable_http.py
```diff
@@ -778,6 +857,13 @@ async def send_event(event_message: EventMessage) -> None:
 
                         # If stream ID not in mapping, create it
                         if stream_id and stream_id not in self._request_streams:
+                            # Register SSE writer so close_sse_stream() can close it
+                            self._sse_stream_writers[stream_id] = sse_stream_writer
```
crondinini-ant, line 884: [related to other comment] just tying it back to the contextmanager idea, if we had that, then we would not have to remember to close it here

src/mcp/server/streamable_http.py
```diff
@@ -787,6 +873,9 @@ async def send_event(event_message: EventMessage) -> None:
                                     event_data = self._create_event_data(event_message)
 
                                     await sse_stream_writer.send(event_data)
+                except anyio.ClosedResourceError:
+                    # Expected when close_sse_stream() is called
```
crondinini-ant, line 900: [q] I'm reading this relatively fast, so might be an incorrect question: 
do we need to pop self._sse_stream_writers.pop(request_id, None) here?

src/mcp/client/streamable_http.py
```diff
@@ -339,10 +357,78 @@ async def _handle_sse_response(
                 # break the loop
                 if is_complete:
                     await response.aclose()
-                    break
-        except Exception as e:
-            logger.exception("Error reading SSE stream:")  # pragma: no cover
-            await ctx.read_stream_writer.send(e)  # pragma: no cover
+                    return  # Normal completion, no reconnect needed
+        except Exception as e:  # pragma: no cover
+            logger.debug(f"SSE stream ended: {e}")
```
crondinini-ant, line 389: I'm bit confused about the change in wording here. Is this now related to the stream ending? I would expect this only happens on error reading stream?

src/mcp/client/streamable_http.py
```diff
@@ -339,10 +357,78 @@ async def _handle_sse_response(
...
+
+        # Stream ended without response - reconnect if we received an event with ID
+        if last_event_id is not None:  # pragma: no branch
+            logger.info("SSE stream disconnected, reconnecting...")
+            await self._handle_reconnection(ctx, last_event_id, retry_interval_ms)
```
crondinini-ant, line 394: [future] I think this is aligned with how we current do things in the SDK, but would love to eventually revisit these things to make it easier to handle this sort of things in a distributed manner. It's related to my previous comment here, but essentially I'd like to allow client developers to be able handle reconnection in different ways. This is 100% out of scope for this feature, specially because we'd need to change so many other things in the parent level, but just leaving a breadcrumb for the future.

src/mcp/client/streamable_http.py
```diff
@@ -339,10 +357,78 @@ async def _handle_sse_response(
...
+
+        try:
+            async with aconnect_sse(
+                ctx.client,
+                "GET",
+                self.url,
+                headers=headers,
+                timeout=httpx.Timeout(self.timeout, read=self.sse_read_timeout),
+            ) as event_source:
+                event_source.response.raise_for_status()
```
crondinini-ant, line 429: [curious] was there a challenge with reading the response before raising?

src/mcp/client/streamable_http.py
```diff
@@ -339,10 +357,78 @@ async def _handle_sse_response(
...
+                        original_request_id,
+                        ctx.metadata.on_resumption_token_update if ctx.metadata else None,
+                    )
+                    if is_complete:
+                        await event_source.response.aclose()
+                        return
+
+                # Stream ended again without response - reconnect again (reset attempt counter)
+                logger.info("SSE stream disconnected, reconnecting...")
+                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, 0)
```
crondinini-ant, line 454: [q] why do we need to reset the attempt counter here?

src/mcp/client/streamable_http.py
```diff
@@ -339,10 +357,78 @@ async def _handle_sse_response(
...
+    async def _handle_reconnection(
+        self,
+        ctx: RequestContext,
+        last_event_id: str,
+        retry_interval_ms: int | None = None,
+        attempt: int = 0,
+    ) -> None:
+        """Reconnect with Last-Event-ID to resume stream after server disconnect."""
+        # Bail if max retries exceeded
+        if attempt >= MAX_RECONNECTION_ATTEMPTS:  # pragma: no cover
```
pja-ant, line 405: Am I reading this right that we fail the stream after a couple of reconnects? Does this apply to the GET stream, and does that mean that you'll just never get unsolicited notifications ever after a couple of seconds?

pja-ant: Oh, is this only if the connection fails entirely? (rather than successful reconnect + no messages)?

felixweinberger: This should only affect the POST response reconnection logic in `_handle_sse_response`, not the standalone GET stream. Continuing a POST request via SSE polling does happen via a GET, but the standalone GET stream for unsolicited notifications isn't affected by this code path. It's also important to say that the `attempt`s only increment on actual _failures_ (e.g. network failures) - if we attempt to reconnect but the stream ends again without completion, the `attempt` counter resets to 0.

The standalone get stream is handled by `handle_get_stream` here: https://github.com/modelcontextprotocol/python-sdk/blob/f54b18a5f91b5cbab501668bd145a47245fe842e/src/mcp/client/streamable_http.py#L204 - IIUC it doesn't actually have any reconnection logic at the moment? But it likely should, given this SEP was clarified to encompass both standalone GET and POST request SSE streams.

Will update.

felixweinberger: > Oh, is this only if the connection fails entirely? (rather than successful reconnect + no messages)?

That's correct, see explanation above.

felixweinberger: @pja-ant pushed a commit on top here that adds support for closing / reconnecting the standalone stream as well.

PR #1663, opened 2025-11-24
Support client_credentials flow with JWT and Basic auth

pcarleton:
## Summary

Implements SEP-1046 OAuth client_credentials flow support with simplified providers for machine-to-machine authentication.

### New OAuth Providers

- **`ClientCredentialsOAuthProvider`**: For `client_credentials` grant with client_id + client_secret
  - Supports `client_secret_basic` (default) and `client_secret_post` auth methods
  - Sets `client_info` directly, bypassing dynamic client registration

- **`PrivateKeyJWTOAuthProvider`**: For `client_credentials` grant with `private_key_jwt` authentication (RFC 7523 Section 2.2)
  - Takes an `assertion_provider` callback that receives the audience (authorization server's issuer identifier per RFC 7523bis) and returns a JWT
  - Designed for workload identity federation (GCP, AWS IAM, Azure AD)

- **`SignedJWTParameters`**: Helper class for SDK-signed JWT assertions
  - `create_assertion_provider()` returns a callback for use with `PrivateKeyJWTOAuthProvider`

- **`static_assertion_provider()`**: Helper for pre-built JWTs that don't need the audience parameter

### Deprecation

**`RFC7523OAuthClientProvider`** is now deprecated with a `DeprecationWarning`.

The original implementation incorrectly used RFC 7523 Section 2.1 (jwt-bearer authorization grant where the JWT itself is the authorization) instead of the intended Section 2.2 (private_key_jwt client authentication with `grant_type=client_credentials`).

Use `ClientCredentialsOAuthProvider` or `PrivateKeyJWTOAuthProvider` instead.

### Example Usage

```python
# Simple client credentials with client_id + secret
provider = ClientCredentialsOAuthProvider(
    server_url="https://api.example.com",
    storage=my_token_storage,
    client_id="my-client-id",
    client_secret="my-client-secret",
)

# Private key JWT with workload identity federation
async def get_workload_identity_token(audience: str) -> str:
    return await fetch_token_from_identity_provider(audience=audience)

provider = PrivateKeyJWTOAuthProvider(
    server_url="https://api.example.com",
    storage=my_token_storage,
    client_id="my-client-id",
    assertion_provider=get_workload_identity_token,
)

# Private key JWT with SDK-signed assertions
jwt_params = SignedJWTParameters(
    issuer="my-client-id",
    subject="my-client-id",
    signing_key=private_key_pem,
)
provider = PrivateKeyJWTOAuthProvider(
    server_url="https://api.example.com",
    storage=my_token_storage,
    client_id="my-client-id",
    assertion_provider=jwt_params.create_assertion_provider(),
)
```

### Testing

- Unit tests with 100% coverage for new providers
- Conformance tests pending


src/mcp/client/auth/extensions/client_credentials.py
```diff
@@ -30,43 +345,49 @@ class JWTParameters(BaseModel):
...
+        return jwt.encode(
+            claims,
+            self.jwt_signing_key,
+            algorithm=self.jwt_signing_algorithm or "RS256",
+        )
 
 
 class RFC7523OAuthClientProvider(OAuthClientProvider):
-    """OAuth client provider for RFC7532 clients."""
+    """OAuth client provider for RFC 7523 jwt-bearer grant.
```
KKonstantinov, line 391: Hi, should this say `for RFC 7523 client-assertion-type` instead

pcarleton: I'm deprecating this one, not removing it yet. it does JWT based client authentication + jwt-bearer grant-type, where the extension specifies JWT-based client authentication + client_credentials grant-type

src/mcp/client/auth/extensions/client_credentials.py
```diff
@@ -1,14 +1,338 @@
...
+    async def _exchange_token_client_credentials(self) -> httpx.Request:
+        """Build token exchange request for client_credentials grant."""
+        token_data: dict[str, Any] = {
+            "grant_type": "client_credentials",
+        }
+
+        headers: dict[str, str] = {"Content-Type": "application/x-www-form-urlencoded"}
+
+        # Use standard auth methods (client_secret_basic, client_secret_post, none)
+        token_data, headers = self.context.prepare_token_auth(token_data, headers)
```
felixweinberger, line 98: Not an issue with this PR, but something pre-existing @claude pointed out here as a discrepancy between TS & Python: 

`prepare_token_auth` doesn't include `client_id` for the `client_secret_post` auth method. According to [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) both `_id` and `_secret` should be in the request body for this method.

TS does that as it does different things for `applyBasicAuth` and `applyPostAuth`: https://github.com/modelcontextprotocol/typescript-sdk/blob/de41e4703c937753de5299b11255656efae59bc4/src/client/auth.ts#L310

While Python seems to use `prepare_token_auth` in both cases.

Might be worth a follow-up?

felixweinberger (approved):
Left a comment regarding something that's pre-existing, otherwise LGTM

src/mcp/client/auth/extensions/client_credentials.py
pcarleton: huh... yea we should fix that! i've got a draft conformance test for it here which should catch this:
https://github.com/modelcontextprotocol/conformance/pull/48

the client credentials one uses `client_secret_basic` which would be why this didn't catch it

Issue #1635, opened 2025-11-16
Using `Context` in resource definitions results in inccorect resource template classification


aarondiel:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

## expected behaivor / actual behavior

looking at the provided sample code i would assume `current_time` would show up as a regular resource, because it needs no arguments from the llm / user.

however, is is only being listed as a resource template and cannot be invoked via the inspector.

<img width="3471" height="1521" alt="Image" src="https://github.com/user-attachments/assets/3596e348-79c3-4840-9944-b3bba5fcebae" />

### Example Code

```Python
from contextlib import asynccontextmanager
from mcp import ServerSession
from mcp.server import FastMCP
from mcp.server.fastmcp import Context

class AppContext:
    pass

@asynccontextmanager
async def lifespan(server: FastMCP):
    yield AppContext()

mcp = FastMCP("Demo", lifespan=lifespan)

@mcp.resource("time://current")
async def current_time(ctx: Context[ServerSession, AppContext]):
    return "11:41"
```

### Python & MCP Python SDK

```Text
`uv run python` shows `3.13.8`
`uv pip freeze` shows `mcp==1.21.1`
```

maxisbey:
Closing as duplicate of #1405, which is tracking the same bug (older issue).

PR #1666, opened 2025-11-25
fix: Fix logic that determines standard resource vs. resource template to account for context param (#1635)

matthew-gries:
- Fix template detection logic to exclude context  when determining if a resource is a template or not
- Add support for FunctionResources to be read with a context param, allowing resources registered with a context param to read from ccontext objects
- Use ` func_metadata.call_fn_with_arg_validation()` to validate arguments in resources while ignoring injected context param, to avoid Pydantic model validation issues 
- Extracted `is_async_callable()` as a public utility 

## Motivation and Context

This solves the issue of context not being properly injected into regular resources. Per #1635, when registering a resource with `@mcp.resource()`, the logic that determines whether to create a standard resource or a resource template was incorrectly counting the Context parameter as a function parameter. This caused functions with only a Context parameter (no URI parameters) to be incorrectly treated as templates.

## How Has This Been Tested?

I added unit tests plus used the following test application to ensure the fix worked as expected and didn't affect existing functionality around resources and resource templates. I tested reading resources and resource templates with and without context params

```
from datetime import datetime

from mcp import ServerSession


class AppContext:
    """Application context with some shared state."""

    def __init__(self):
        self.server_start_time = datetime.now()
        self.request_count = 0

    def increment_requests(self):
        self.request_count += 1


@asynccontextmanager
async def lifespan(server: FastMCP):
    """Initialize application context."""
    print("Server starting up...")
    ctx = AppContext()
    yield ctx
    print(f"Server shutting down. Total requests: {ctx.request_count}")


mcp = FastMCP("Context Test Demo", lifespan=lifespan)


@mcp.resource("time://current")
async def current_time(ctx: Context[ServerSession, AppContext]) -> str:
    return f"Current time: {datetime.now().isoformat()}, request {ctx.request_id}"


@mcp.resource("time://format")
async def time_format() -> str:
    return "Supported formats: ISO8601, RFC3339, Unix timestamp"


@mcp.resource("greeting://{name}")
async def personalized_greeting(name: str, ctx: Context[ServerSession, AppContext]) -> str:
    uptime = (datetime.now() - ctx.request_context.lifespan_context.server_start_time).total_seconds()
    return f"Hello {name}! Server uptime: {uptime:.1f}s"


@mcp.resource("math://{operation}/{a}/{b}")
async def math_operation(operation: str, a: int, b: int) -> str:
    operations = {
        "add": a + b,
        "subtract": a - b,
        "multiply": a * b,
        "divide": a / b if b != 0 else "Error: Division by zero"
    }
    result = operations.get(operation, "Error: Unknown operation")
    return f"{operation}({a}, {b}) = {result}"


if __name__ == "__main__":
    import asyncio

    mcp.run()
```

## Breaking Changes

Technically yes, because the signature of the `read()` function on the `Resource` type now includes an additional optional `context` param, but this shouldn't break at runtime since the arg is optional 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/mcpserver/resources/base.py
+++ b/src/mcp/server/mcpserver/resources/base.py
@@ -1,7 +1,9 @@
 """Base classes and interfaces for MCPServer resources."""
 
+from __future__ import annotations
+
 import abc
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from pydantic import (
     BaseModel,
@@ -12,6 +14,10 @@ from pydantic import (
 )
 
 from mcp.types import Annotations, Icon
+
+if TYPE_CHECKING:
+    from mcp.server.context import LifespanContextT, RequestT
+    from mcp.server.mcpserver.server import Context
 
 
 class Resource(BaseModel, abc.ABC):
@@ -43,6 +49,9 @@ class Resource(BaseModel, abc.ABC):
         raise ValueError("Either name or uri must be provided")
 
     @abc.abstractmethod
-    async def read(self) -> str | bytes:
+    async def read(
+        self,
+        context: Context[LifespanContextT, RequestT] | None = None,
+    ) -> str | bytes:
         """Read the resource content."""
         pass  # pragma: no cover
--- a/src/mcp/server/mcpserver/resources/templates.py
+++ b/src/mcp/server/mcpserver/resources/templates.py
@@ -2,17 +2,16 @@
 
 from __future__ import annotations
 
-import inspect
 import re
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 from urllib.parse import unquote
 
-from pydantic import BaseModel, Field, validate_call
+from pydantic import BaseModel, Field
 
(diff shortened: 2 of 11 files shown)


pepijndevos:
I can confirm that this fixes the problem for me

jlowin:
I think this is a good improvement 👍

src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -531,3 +532,13 @@ def _convert_to_content(
         result = pydantic_core.to_json(result, fallback=str, indent=2).decode()
 
     return [TextContent(type="text", text=result)]
+
+
+def is_async_callable(obj: Any) -> bool:
+    """Check if an object is an async callable."""
+    while isinstance(obj, functools.partial):  # pragma: no cover
```
maxisbey: Please add a unit test for this instead of having a pragma.

matthew-gries:
I rebased to main and made a couple of updates to make sure test coverage was updated

I think the issue in the windows-latest CI check is unrelated

PR #1677, opened 2025-11-27
feat(fastmcp): add include_in_context parameter to resource decorator

### LINKED ISSUE #1657: [FastMCP] Add 'include_in_context' argument to @resource decorator
## Is your feature request related to a problem? Please describe.

Currently, in FastMCP, if a developer wants to hint that a resource is critical and should be included in the model's context window by default (e.g., operational guidelines or system prompts), they need to manually understand and set the `annotations` field with a priority score.

While the protocol spec defines priority (0.0 - 1.0), it is not immediately obvious to a Python developer how to assert "Always include this." The current method is slightly verbose and less discoverable:

```python
# Current way
@mcp.resource("uri", annotations=Annotations(priority=1.0)) 
def my_resource(): 
    return "..."
```

## Describe the solution you'd like

I propose adding a boolean argument `include_in_context` (defaulting to `False`) to the `FastMCP.resource` decorator.

This would provide a high-level, Pythonic abstraction over the priority annotation. Setting this to `True` would automatically set the underlying resource priority to 1.0 (indicating maximum priority for context inclusion).

### Proposed Usage

```python

mcp = FastMCP("Demo")

@mcp.resource(
    "guidelines://ops", 
    include_in_context=True  # Developer intent is clear here
)
def operational_guidelines():
    return "..."
```

### Implementation Details

When `include_in_context=True`:
- Automatically creates `Annotations(priority=1.0)` if no annotations exist
- Overrides the priority field to 1.0 if annotations are explicitly provided
- Preserves other annotation fields (like `audience`)

## Why this is valuable

1. **Developer Experience (DX)**: It reduces cognitive load. Developers don't need to look up the specific float values for priority; they can just use a boolean flag.

2. **Readability**: `include_in_context=True` explicitly states the intent of the resource code-side.

3. **Discoverability**: It makes the concept of "default context" more accessible to new MCP users.

4. **Consistency**: Similar to how other frameworks provide high-level abstractions for common patterns.

## Additional context

I have reviewed `src/mcp/server/fastmcp/server.py` and believe this can be implemented by modifying the resource decorator to check this boolean and update the annotations before registration.

I am happy to submit a PR for this feature. As this would be my first contribution to the project, I wanted to validate the naming convention (`include_in_context` vs `auto_load` vs `priority_high`) and approach before pushing code.

This complements the work being done in #1475 for dynamic annotations and #1468 for annotation support.

rinaldofesta:
## Summary

Adds a user-friendly `include_in_context` boolean parameter to the `@resource` decorator that automatically sets `priority=1.0` in annotations. This simplifies marking resources for context inclusion without requiring users to manually create and manage `Annotations` objects.

Closes #1657

## Motivation

Currently, users must create an `Annotations` object to set resource priority for context inclusion:

```python
@mcp.resource("resource://data", annotations=Annotations(priority=1.0))
def get_data():
    return "important data"
```

This PR introduces a simpler API:

```python
@mcp.resource("resource://data", include_in_context=True)
def get_data():
    return "important data"
```

## Implementation

- **Parameter behavior**: When `include_in_context=True`, automatically sets `priority=1.0`
- **Override logic**: Always overrides explicit priority values when enabled
- **Field preservation**: Preserves other annotation fields like `audience`
- **Broad support**: Works with both regular and template resources, sync and async functions

## Changes

- Modified `FastMCP.resource()` decorator to accept `include_in_context` parameter
- Added logic to create/update annotations with `priority=1.0` when enabled
- Updated docstring with parameter documentation
- Added 6 comprehensive tests covering various scenarios

## Test Plan

- [x] `include_in_context=True` sets priority to 1.0
- [x] `include_in_context=False` doesn't affect annotations
- [x] Overrides explicit priority values
- [x] Preserves audience annotations
- [x] Works with template resources
- [x] Works with async functions
- [x] All existing annotation tests still pass (backward compatibility)


jlowin:
I responded on #1657 (not sure where the discussion should accrue) but I see this as introducing a new problem (two kwargs for one outcome, one of which has misleading effect) rather than solving an existing one.

maxisbey:
Thanks for the contribution! Unfortunately I'll be closing the PR as per https://github.com/modelcontextprotocol/python-sdk/issues/1657#issuecomment-3638627915

Basically, this is already supported in both the spec and framework as is, I don't think we need to have it hardcoded as an arg in the decorators.

PR #1608, opened 2025-11-10
feat: propagate session_id from transport to tool context

cbcoutinho:
## Summary

This PR enables tools to access the transport-level session ID via `Context.session_id`, supporting session-level tracing and observability use cases.

## Changes

- **InitializationOptions**: Add optional `session_id` field
- **ServerSession**: Store and expose `session_id` via property
- **RequestContext**: Add `session_id` field to dataclass
- **StreamableHTTPSessionManager**: Pass `session_id` when creating initialization options
- **FastMCP Context**: Add `session_id` property accessible in tools
- **Tests**: Comprehensive test suite for session_id propagation

## Usage

Tools can now access the session ID like this:

```python
@mcp.tool()
async def my_tool(ctx: Context) -> dict:
    session_id = ctx.session_id  # Transport's session ID (streamable-http)
    # Use for tracing, logging, correlation, etc.
    return {"session_id": session_id}
```

## Behavior

- **Streamable-HTTP (stateful)**: `ctx.session_id` returns the transport's session ID (UUID)
- **Streamable-HTTP (stateless)**: `ctx.session_id` returns `None` (no session tracking)
- **stdio/SSE transports**: `ctx.session_id` returns `None` (session_id not set during initialization)

## Use Cases

1. **Distributed Tracing**: Add session_id as a span attribute in OpenTelemetry traces to correlate all tool calls within a session
2. **Session Analytics**: Track per-session metrics and usage patterns
3. **Debugging**: Correlate logs and errors across multiple requests in the same session
4. **Audit Trails**: Associate all operations with a session identifier

## Testing

- All 418 existing server tests pass ✓
- New test suite covers session_id propagation scenarios
- Verified no regressions in FastMCP, auth, resource tests

## Fixes

- #485
- #942

## Related

Related to #421 (OpenTelemetry tracing integration)

## Breaking Changes

None. This is a backward-compatible addition:
- New optional field in `InitializationOptions`
- New property on `Context` (returns None if not available)
- Existing code continues to work without modifications

PR #1710, opened 2025-12-01
Add on_session_created callback option

crondinini-ant:
## Motivation and Context
  Currently, there's no way to retrieve the session ID from the SSE client. The session ID is embedded in the
  endpoint URL sent by the server, but it's not exposed to callers. Users who need the session ID (e.g., for
  reconnection with `mcp-session-id` header) would have to resort to intercepting raw HTTP bytes via custom transports.

  This change adds a simple callback mechanism that's invoked when the endpoint event is received, passing the
  extracted session ID.

  ## How Has This Been Tested?
  - Added unit test `test_sse_client_on_session_created` that verifies the callback is invoked with a valid session
   ID

  ## Breaking Changes
  None. The new `on_session_created` parameter is optional and defaults to `None`.

  ## Types of changes
  - [ ] Documentation update

  ## Checklist

  ## Additional context
  Usage example:
  ```python
  session_id = None

  def capture_session(sid: str):
      nonlocal session_id
      session_id = sid

  async with sse_client(url, on_session_created=capture_session) as (r, w):
      print(session_id)  # available after connection established

### DIFF
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -4,6 +4,7 @@ import socket
 import time
 from collections.abc import AsyncGenerator, Generator
 from typing import Any
+from unittest.mock import Mock
 
 import anyio
 import httpx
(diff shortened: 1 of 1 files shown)


felixweinberger:
~~What do you think about this alternative approach? https://github.com/modelcontextprotocol/python-sdk/pull/1608~~

~~Seems potentially more intuitive to have this available on the request context than to pass it back via a callback somehow?~~

Edit: not relevant, as this is on client, that makes it available on server tools

Issue #1367, opened 2025-09-15
Mounting a Streamable HTTP MCP endpoint on existing FastAPI app does not work


vemonet:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

Hi,

We are trying to create a web server with 2 endpoints:

- `/mcp` : the MCP server endpoint
- `/hello` : whatever custom endpoint getting JSON returning JSON

This is important because it allows to build more advanced app than just a MCP server, and avoid having 35 micro services to manage.

We use a really popular solution for doing so: FastAPI, which claims full support for mounting ASGI apps, such as the `Starlette` object returned by `mcp.streamable_http_app()`

```python
from fastapi import FastAPI
from mcp.server.fastmcp import FastMCP

app = FastAPI(title="Hello MCP")


mcp = FastMCP(
    "SIB BioData MCP",
    debug=True,
    dependencies=["mcp"],
)

@mcp.tool()
async def just_for_test(question: str) -> str:
    """Nothing special

    Args:
        question: The question to be answered

    Returns:
        The question
    """
    return question

mcp_app = mcp.streamable_http_app()
app.mount("/mcp", mcp_app)

@app.post("/hello")
async def chat():
    print("hello")
```

Run with:

```sh
uv run uvicorn main:app --port 8000
```

Going directly to http://localhost:8000/mcp gives plain text `Not Found`

Going to any actually not defined path gives the usual FastAPI `{"detail":"Not Found"}`

So this shows something has been mounted on `/mcp`, but your`Starlette` object returns an incorrect error (it can't be not found, maximum it could be method not allowed, or something else, but not found does not seems correct here)

Try to access it with the official MCP client using the official documentation:

```python
import asyncio
from mcp import ClientSession
from mcp.client.streamable_http import streamablehttp_client

async def main():
    mcp_url = "http://localhost:8000/mcp"
    async with streamablehttp_client(mcp_url) as (read_stream, write_stream, _):
        async with ClientSession(read_stream, write_stream) as session:
            await session.initialize()
            tools = await session.list_tools()
            print(f"Available tools: {[tool.name for tool in tools.tools]}")

if __name__ == "__main__":
    asyncio.run(main())
```

We get a redirect and a 404 in the server logs:

```
[...]

PR #1712, opened 2025-12-01
Add working example to easily integrate a FastMCP endpoint to a FastAPI app (which is not trivial)

vemonet:
This PR adds an example for integrating a FastMCP endpoint in a FastAPI app (which is not straightforward, and could be better documented). 

I have added an `examples/servers/fastapi/` folder with a minimal, focused, working example: `main.py`, `pyproject.toml`, and `README.md` (with basic uvicorn command).

I have also added a line to point to this folder in the README, this way devs searching for "fastapi" with ctrl+F in the readme will find the docs easily.

## Motivation and Context

For now only Starlette is documented, but as far as I know, most developers build APIs directly with FastAPI (which is a wrapper around Starlette) and not Starlette directly:

- FastAPI provides out of the box OpenAPI specs and docs from type annotations
- FastAPI enables devs to build well documented APIs easily, exactly the same way as FastMCP does: by using decorators and type annotations
- FastMCP took a huge inspiration from FastAPI (which is even reflected in its name!), but it is not clear how to make the 2 interoperable, current documentation encourages to use Starlette when many developers want to use the same dev friendly API for the MCP tools and other API endpoints: "Decorators and type annotations for me (MCP), not for thee (every other endpoints on the API)" :)

I already discussed the issue in https://github.com/modelcontextprotocol/python-sdk/issues/1367 without fully solving the problem (and I was already not the only one to face this problem according to the reactions on the issue)

The main problem when building a FastAPI app with a MCP endpoint is the missing `run()` lifespan (it is predefined for Starlette, but needs to be manually defined for FastAPI). It is not much, but it is not documented, and increases friction when wanting to integrate FastMCP in FastAPI.

Trying to simply mount a Starlette `mcp.streamable_http_app()` on a FastAPI `app` will end up throwing this error at runtime:

```
  File ".venv/lib/python3.13/site-packages/mcp/server/streamable_http_manager.py", line 138, in handle_request
    raise RuntimeError("Task group is not initialized. Make sure to use run().")
RuntimeError: Task group is not initialized. Make sure to use run().
```

Meanwhile it works out of the box when mounting on a Starlette app, so all existing Starlette example are a bit misleading when you want to use FastAPI (which is a large amount of the users)

## How Has This Been Tested?

I have ran the command documented in the `README.md` of the folder `examples/servers/fastapi/` locally and it worked

## Breaking Changes

No, it's just an example added to the existing docs

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

I have added `exclude = ["examples/servers/fastapi"]` to the pyright config in the root `pyproject.toml`, otherwise pyright checks fails because monorepo is still not supported by pyright. I also tried to add `[tool.pyright] exclude = ["."]` to the `examples/servers/fastapi/pyproject.toml`, but it does not seems like pyright understand the concept of nested config files.

It would be good for the sake of diversity of provided examples that the examples can use dependencies not defined in the main `pyproject.toml`

Other solutions to fix this pyright fumble could be:
- Adding `fastapi` to the dev dependency group in the root `pyproject.toml`
- Telling pyright in the root `pyproject.toml` to ignore the folder

> I am used to `mypy` and waiting for `ty`, so my knowledge of `pyright` config is limited, let me know if there is a better approach to this 

We could also make this a snippet directly integrated in the root `README.md` through `examples/snippets/` folder if you prefer

### DIFF
--- a/README.md
+++ b/README.md
@@ -676,7 +676,7 @@ The Context object provides the following capabilities:
 - `ctx.session` - Access to the underlying session for advanced communication (see [Session Properties and Methods](#session-properties-and-methods))
 - `ctx.request_context` - Access to request-specific data and lifespan resources (see [Request Context Properties](#request-context-properties))
 - `await ctx.debug(message)` - Send debug log message
-- `await ctx.info(message)` - Send info log message  
+- `await ctx.info(message)` - Send info log message
 - `await ctx.warning(message)` - Send warning log message
 - `await ctx.error(message)` - Send error log message
 - `await ctx.log(level, message, logger_name=None)` - Send log with custom level
@@ -1106,13 +1106,13 @@ The session object accessible via `ctx.session` provides advanced control over c
 async def notify_data_update(resource_uri: str, ctx: Context) -> str:
     """Update data and notify clients of the change."""
     # Perform data update logic here
-    
+
     # Notify clients that this specific resource changed
     await ctx.session.send_resource_updated(AnyUrl(resource_uri))
-    
+
     # If this affects the overall resource list, notify about that too
     await ctx.session.send_resource_list_changed()
-    
+
     return f"Updated {resource_uri} and notified clients"
 ```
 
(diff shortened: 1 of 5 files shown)


maxisbey:
Closing this due to v2 release and the docs being fully reworked

Issue #1253, opened 2025-08-09
Python SDK v2


Kludex:
I'll use this issue to add notes about what I think should happen for V2. 

- [ ] Use snake case instead of camel case on `types.py` (#97)
- [ ] Don't use `RootModel` on `types.py` - It makes hard to reason about types.
- [ ] Revisit generic classes, and add more default values for them #1207
- [ ] Rethink `ClientSessionGroup`. It doesn't have a user-friendly API.
- [ ] Rename `FastMCP` to `MCPServer`.
- [ ] https://github.com/modelcontextprotocol/python-sdk/issues/12

I'll lock the issue, so only maintainers can comment, and I can focus on the things that I really think should be here.

felixweinberger:
I'll use this issue to add things I see as well.

- [x] https://github.com/modelcontextprotocol/python-sdk/pull/113
- [x] https://github.com/modelcontextprotocol/python-sdk/pull/1222
- [x] https://github.com/modelcontextprotocol/python-sdk/pull/1713

Kludex:
# Checklist

- [x] Low level parameter callbacks (@maxisbey is working on it)
- [ ] We do need the idea that is presented here: https://github.com/modelcontextprotocol/python-sdk/pull/1591, but with a better pattern. I did explain this on the discord - details should be in the python and transport channel.
- [ ] We need to improve the documentation. This needs to be a continuous effort, which we didn't create the habit yet.
   - [ ] Making the README simple - all the examples should be in the documentation.
- [ ] We need to recheck the constructor parameters of the the MCPServer class. I did remove and refactor a bit, but we need to audit it a bit better.
- [ ] MCP Proxy, as mentioned above.
- [ ] Refactor the auth "way of using it".
   - [ ] Drop the use of `Field(description=...)`, and use proper docstrings. As far as I know, we are not creating json schemas from it, so it's useless.
- [ ] Simplify/remove/(move to docs) the current examples. My suggestion here would be to take the opportunity on docs write, to actually check the examples, and try to move them to the docs.
    - [ ] Check how FastAPI does it - they have a `docs_src` where they put all their code examples, and they actually run tests on every snippet in tests.
    - [ ] Check other projects that do inline markdown snippet, like pydantic-ai.
- [ ] Create dependency injection in `MCPServer`.

PR #1713, opened 2025-12-02
feat: backwards-compatible create_message overloads for SEP-1577

felixweinberger:
## Summary
Introduce method overloading for `create_message` to preserve backwards compatibility while supporting the new tools feature from SEP-1577.

- When called without `tools`, returns `CreateMessageResult` with single content (backwards compatible)
- When called with `tools`, returns `CreateMessageResultWithTools` which allows array content

## Motivation and Context
SEP-1577 changes `CreateMessageResult.content` from single to `single | list`. This would be a breaking change for existing SDK users. To avoid requiring a major version bump (v2.0), this PR uses `@overload` decorators to return different types based on whether tools are provided.

Existing code that doesn't use tools continues to work unchanged:
```python
result = await session.create_message(messages=..., max_tokens=100)
# result.content is single content - no changes needed
if result.content.type == "text":
    return result.content.text
```

## How Has This Been Tested?
- Type tests pass with pyright
- 469 tests pass
- Updated test for tool use to use `CreateMessageResultWithTools`
- Added test for basic `CreateMessageResult`

## Breaking Changes
None for v1.x users. Existing code that doesn't use tools continues to work unchanged.

**Note:** v2.0 will align `CreateMessageResult` with the schema (array content).

## Types of changes

## Checklist

### DIFF
--- a/tests/client/test_sampling_callback.py
+++ b/tests/client/test_sampling_callback.py
@@ -100,6 +100,7 @@ async def test_create_message_backwards_compat_single_content():
     async with create_session(server._mcp_server, sampling_callback=sampling_callback) as client_session:
         result = await client_session.call_tool("test_backwards_compat", {"message": "Test"})
         assert result.isError is False
+        assert isinstance(result.content[0], TextContent)
         assert result.content[0].text == "true"
 
 


felixweinberger:
Also added this to the list of things to clean up for v2: https://github.com/modelcontextprotocol/python-sdk/issues/1253

src/mcp/server/session.py
```diff
@@ -244,10 +245,47 @@ async def create_message(
         stop_sequences: list[str] | None = None,
         metadata: dict[str, Any] | None = None,
         model_preferences: types.ModelPreferences | None = None,
-        tools: list[types.Tool] | None = None,
```
pcarleton, line 247: note: this confused me (why not leave it out?), but its a kwarg, and is needed to discriminate between the overloads

pcarleton (approved):

pcarleton (approved):

PR #1719, opened 2025-12-04
fix: skip priming events and close_sse_stream for old protocol versions

felixweinberger:
Fixes backwards compatibility issue where SEP-1699 priming events break old clients.

## Motivation and Context
SEP-1699 introduced "priming events" - SSE events with an ID but empty data (`data: \n`). Old clients (SDK versions before the empty-data fix) try to `JSON.parse("")` and crash with "Unexpected end of JSON input".

Additionally, the `close_sse_stream` callbacks should not be provided to old clients because they can't resume if the stream is closed early (they never received a priming event with an event ID).

## How Has This Been Tested?
- Existing tests continue to pass (42 tests)
- The implementation gates priming events and callbacks on protocol version >= 2025-11-25

## Breaking Changes
None. This is a backwards-compatible fix that makes new servers work correctly with old clients.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
The fix gates on protocol version >= 2025-11-25:
- For initialize requests: extracts version from `params.protocolVersion`
- For other requests: uses `mcp-protocol-version` header

Changes:
- `_maybe_send_priming_event`: renamed from `_send_priming_event`, now checks protocol version
- `_create_session_message`: only provides `close_sse_stream` callbacks when protocol version supports resumability

String comparison works for ISO date format versions (e.g., `'2025-03-26' < '2025-11-25'` → true).

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -238,29 +238,49 @@ class StreamableHTTPServerTransport:
         message: JSONRPCMessage,
         request: Request,
         request_id: RequestId,
+        protocol_version: str,
     ) -> SessionMessage:
-        """Create a session message with metadata including close_sse_stream callback."""
+        """Create a session message with metadata including close_sse_stream callback.
 
-        async def close_stream_callback() -> None:
-            self.close_sse_stream(request_id)
+        The close_sse_stream callbacks are only provided when the client supports
+        resumability (protocol version >= 2025-11-25). Old clients can't resume if
+        the stream is closed early because they didn't receive a priming event.
+        """
+        # Only provide close callbacks when client supports resumability
+        if self._event_store and protocol_version >= "2025-11-25":
 
-        async def close_standalone_stream_callback() -> None:
-            self.close_standalone_sse_stream()
+            async def close_stream_callback() -> None:
+                self.close_sse_stream(request_id)
 
-        metadata = ServerMessageMetadata(
-            request_context=request,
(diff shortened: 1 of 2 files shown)


src/mcp/server/streamable_http.py
```diff
@@ -560,7 +589,7 @@ async def sse_writer():
                     try:
                         async with sse_stream_writer, request_stream_reader:
                             # Send priming event for SSE resumability
-                            await self._send_priming_event(request_id, sse_stream_writer)
+                            await self._maybe_send_priming_event(request_id, sse_stream_writer, protocol_version)
```
maxisbey, line 592: would it be better to put the `if protocol_version` here instead? then can keep the _send_priming_event the same as before

tests/shared/test_streamable_http.py
```diff
@@ -1761,6 +1761,120 @@ async def test_handle_sse_event_skips_empty_data():
...
+        await write_stream.aclose()
+        await read_stream.aclose()
+
+
+@pytest.mark.anyio
+async def test_close_sse_stream_callback_not_provided_for_old_protocol_version():
+    """Test that close_sse_stream callbacks are NOT provided for old protocol versions."""
+    from unittest.mock import MagicMock
+
+    from mcp.types import JSONRPCMessage, JSONRPCRequest
```
maxisbey: move these to top of file

felixweinberger: Fixed, thanks!

PR #1720, opened 2025-12-04
Fix JSON-RPC error response ID matching

crondinini-ant:
## Motivation and Context

 

When an MCP server returns a JSON-RPC response with the request ID as a string (e.g., `"id": "0"`) but the client sent the request with an integer ID (e.g., `"id": 0`), the response fails to match the pending request. This causes the client to timeout waiting for a response that was already received but couldn't be correlated.

 

The root cause is in [`src/mcp/shared/session.py:256-260`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/shared/session.py#L256-L260) where the client stores pending requests using an integer counter:

 

```python

request_id = self._request_id  # always int

self._request_id = request_id + 1

self._response_streams[request_id] = response_stream  # key is int

```

 

When a response arrives, the lookup `self._response_streams.pop(response_id, None)` fails silently if `response_id` is `"0"` (string) because Python dict lookups are type-sensitive (`0 != "0"`).

 

This is valid per the JSON-RPC spec and the SDK's own type definition at [`src/mcp/types.py:40`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/types.py#L40):

 

```python

RequestId = Annotated[int, Field(strict=True)] | str

```

 

## How Has This Been Tested?

 

Added three new unit tests in `tests/shared/test_session.py`:

- `test_response_id_type_mismatch_string_to_int` - success responses with string IDs match integer request IDs

- `test_error_response_id_type_mismatch_string_to_int` - error responses with string IDs match integer request IDs

- `test_response_id_non_numeric_string_no_match` - non-numeric string IDs (like `"abc"`) don't incorrectly match

 

## Breaking Changes

 

None. This is a backwards-compatible fix that makes the client more tolerant of server responses.

 

## Types of changes

 

- [ ] Documentation update

 

## Checklist

 

## Additional context

 

The fix adds a `_pop_response_stream()` helper method that:

1. First tries an exact match lookup

2. If the ID is a string, tries converting to int and looks up again

3. If the ID is an int, tries converting to string and looks up again

 

This handles the common case where servers echo back IDs in a different but semantically equivalent format.

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -473,7 +473,7 @@ class BaseSession(
             try:
                 return int(response_id)
             except ValueError:
-                pass
+                logging.warning(f"Response ID {response_id!r} cannot be normalized to match pending requests")
         return response_id
 
     async def _handle_response(self, message: SessionMessage) -> None:


src/mcp/shared/session.py
```diff
@@ -455,6 +455,27 @@ async def _receive_loop(self) -> None:
...
+            response_id: The response ID from the incoming message.
+
+        Returns:
+            The normalized ID (int if possible, otherwise original value).
+        """
+        if isinstance(response_id, str):
+            try:
+                return int(response_id)
+            except ValueError:
+                pass
```
felixweinberger: Should we log a warning for the exception here? 

If we hit this it means the assumption of integer `RequestId`s is violated right? Seems worth bubbling that up. Not sure if that's sufficiently obvious from wherever else this surfaces.

felixweinberger: Something like this maybe?

```py
logger.warning(
      f"Response ID {response_id!r} cannot be normalized to match pending requests"
  )
```

felixweinberger (approved):
Seems reasonable and matches the approach we take in the Typescript SDK. Might be worth adding that warning log though when we hit the unexpected case.

felixweinberger (dismissed):

PR #1594, opened 2025-11-07
Implement SEP-1577 - Sampling With Tools

ochafik:
Implement types from https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577 (accepted)

Spec PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796 (merged, and follow up https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1842 pending review)

cc/ @bhosmer-ant 

Add support for tool calling during sampling requests, enabling MCP servers to execute agentic workflows using client LLM capabilities.

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

Key changes:
- Add ToolUseContent type for assistant tool invocation requests
- Add ToolResultContent type for tool execution results
- Add ToolChoice type to control tool usage behavior
- Add UserMessage and AssistantMessage types for role-specific messages
- Extend SamplingMessage to support tool content (backward compatible)
- Add SamplingToolsCapability & SamplingContextCapability
- Update CreateMessageRequestParams with tools and toolChoice fields
- Update CreateMessageResult to support tool use content
- Update StopReason to include "toolUse" value
- Add comprehensive unit tests for all new types

The implementation maintains backward compatibility by keeping SamplingMessage as a flexible BaseModel while adding more specific UserMessage and AssistantMessage types for type-safe tool interactions.

All new types follow existing patterns:
- Use Pydantic V2 BaseModel
- Allow extra fields with ConfigDict(extra="allow")
- Include proper docstrings and field descriptions
- Support optional fields where appropriate

PR #1722, opened 2025-12-04
feat: client-side support for SEP-1577 sampling with tools

jlowin:
[SEP-1577](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577) added tool calling support to sampling (via #1594). This PR adds the necessary client-side support:

1. **Capability advertisement**: The spec requires servers to error when `tools` or `toolChoice` are provided but `clientCapabilities.sampling.tools` is missing. The SDK's types already support this, but `ClientSession` had no way to advertise it—servers would always see the capability as missing and error. This adds a `sampling_capabilities` parameter. If not provided, the previous logic applies: the presence of a `sampling_callback` enables basic sampling support without tools.

```python
session = ClientSession(
    read_stream, write_stream,
    sampling_callback=my_callback,
    sampling_capabilities=SamplingCapability(
        tools=SamplingToolsCapability()
    ),
)
```

2. **Return type for tool responses**: When tools are provided in a sampling request, the LLM may respond with tool calls, requiring `CreateMessageResultWithTools` (which supports array content). The `SamplingFnT` callback and `ClientResultType` didn't include this type, so the client's response validation would reject it before it could be sent back to the server. This adds `CreateMessageResultWithTools` to both.

FWIW, with these changes we have full agentic sampling wired up in a branch of FastMCP, which is extremely exciting!

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -115,13 +115,13 @@ class ClientSession(
         write_stream: MemoryObjectSendStream[SessionMessage],
         read_timeout_seconds: timedelta | None = None,
         sampling_callback: SamplingFnT | None = None,
-        sampling_capabilities: types.SamplingCapability | None = None,
         elicitation_callback: ElicitationFnT | None = None,
         list_roots_callback: ListRootsFnT | None = None,
         logging_callback: LoggingFnT | None = None,
         message_handler: MessageHandlerFnT | None = None,
         client_info: types.Implementation | None = None,
         *,
+        sampling_capabilities: types.SamplingCapability | None = None,
         experimental_task_handlers: ExperimentalTaskHandlers | None = None,
     ) -> None:
         super().__init__(


src/mcp/client/session.py
```diff
@@ -115,6 +115,7 @@ def __init__(
         write_stream: MemoryObjectSendStream[SessionMessage],
         read_timeout_seconds: timedelta | None = None,
         sampling_callback: SamplingFnT | None = None,
+        sampling_capabilities: types.SamplingCapability | None = None,
```
maxisbey: thoughts on moving to the bottom of kwargs or keyword only to avoid breaking positional uses of the constructor? I'm probably leaning towards ~~positional~~ keyword only

jlowin: Good call! (assuming you meant leaning "keyword" only, and I agree - will update)

maxisbey: woops yea meant keyword only haha

maxisbey (approved):

Issue #1751, opened 2025-12-09
OAuthClientProvider add prefix url for MCP Starlette Mounted endpoints to correctly route .well-known metadata discovery


abneetwats24:
### Description

I am using Starlette mounting
```python
hr_policy_mcp.settings.streamable_http_path = "/hr-policy"
math_mcp.settings.streamable_http_path = "/math"
main_server = Starlette(
        routes=[
            Mount("/hr-policy", app=hr_policy_mcp.streamable_http_app()),
            Mount("/math", app=math_mcp.streamable_http_app()),
        ],
        lifespan=combined_lifespan,
    )
main_server = CORSMiddleware(
        main_server,
        allow_origins=["*"],  # Configure appropriately for production
        allow_methods=["GET", "POST", "DELETE"],  # MCP streamable HTTP methods
        expose_headers=["Mcp-Session-Id"],
    )
```

With this I am getting 404 for Metadata Discovery

```bash
mcp-server  | INFO:     172.18.0.1:54902 - "GET /hr-policy/hr-policy HTTP/1.1" 401 Unauthorized
mcp-server  | INFO:     172.18.0.1:54902 - "GET /favicon.ico HTTP/1.1" 404 Not Found
mcp-server  | INFO:     172.18.0.1:33592 - "OPTIONS /hr-policy/hr-policy HTTP/1.1" 200 OK
mcp-server  | INFO:     172.18.0.1:33592 - "POST /hr-policy/hr-policy HTTP/1.1" 401 Unauthorized
mcp-server  | INFO:     172.18.0.1:42520 - "OPTIONS /.well-known/oauth-protected-resource/hr-policy/hr-policy HTTP/1.1" 400 Bad Request
mcp-server  | INFO:     172.18.0.1:42520 - "GET /.well-known/oauth-protected-resource/hr-policy/hr-policy HTTP/1.1" 404 Not Found
mcp-server  | INFO:     172.18.0.1:42520 - "OPTIONS /.well-known/oauth-protected-resource HTTP/1.1" 400 Bad Request
mcp-server  | INFO:     172.18.0.1:42520 - "GET /.well-known/oauth-protected-resource HTTP/1.1" 404 Not Found
mcp-server  | INFO:     172.18.0.1:42520 - "OPTIONS /.well-known/oauth-authorization-server HTTP/1.1" 400 Bad Request
[...]

PR #1762, opened 2025-12-10
Enhancement: enhance resource metadata discovery URLs for mounted starlette app

abneetwats24:
**Summary**: Add a discovery client and related support for mounted application resource metadata to enable discovery of mounted app resources by clients. This change implements the client-side helper utilities and wiring needed to query and surface metadata for mounted app resources, plus accompanying tests and examples demonstrating usage.

## Motivation and Context

- **Motivation**: Consumers of the SDK need a reliable way to discover metadata for mounted application resources so they can enumerate, inspect, and interact with resources provided by mounted apps. Previously this metadata was not discoverable through the client surface in a consistent way, which made building integrations and tooling harder.
- **Context**: This change implements a discovery client and related helpers that expose mounted app resource metadata in a consistent, typed form. It is driven by the feature work tracked in issue #1751.

## How Has This Been Tested?

- **Unit tests**: Updated unit tests covering the new discovery client behavior, response parsing, and error handling.
- **CI**: Updated tests are compatible with the existing CI; all tests pass locally prior to this PR.

## Breaking Changes

- **Breaking changes**: None — this is an additive change. Existing client APIs are preserved; the discovery client and helpers are new additions and do not change existing method signatures or runtime behavior for current consumers.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- **Related issue**: https://github.com/modelcontextprotocol/python-sdk/issues/1751
- **Implementation notes**: The change adds a small, focused client API for discovery and keeps response models typed and validated to help downstream code rely on stable shapes. Error cases are surfaced as specific exceptions to make retry and fallback strategies easier for integrators.

### DIFF
--- a/src/mcp/client/auth/utils.py
+++ b/src/mcp/client/auth/utils.py
@@ -71,6 +71,7 @@ def build_protected_resource_metadata_discovery_urls(www_auth_url: str | None, s
     Per SEP-985, the client MUST:
     1. Try resource_metadata from WWW-Authenticate header (if present)
     2. Fall back to path-based well-known URI: /.well-known/oauth-protected-resource/{path}
+        or /{mount path}/.well-known/oauth-protected-resource for starlete mounted servers
     3. Fall back to root-based well-known URI: /.well-known/oauth-protected-resource
 
     Args:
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR! I don't think this is the right fix though - the issue is actually on the server side, not the client.

Per RFC 9728, the `.well-known` URI should be inserted between the host and the path - so for `https://example.com/hr-policy/mcp` the metadata should be at `https://example.com/.well-known/oauth-protected-resource/hr-policy/mcp`, not `https://example.com/hr-policy/.well-known/oauth-protected-resource`.

The problem is that when you mount a Starlette app, it prefixes all routes including the `.well-known` ones, which puts them in the wrong place. That's a server-side issue we'd need to fix there rather than having the client guess at non-spec-compliant URLs.

Going to close this but happy to discuss further on #1751 if you want to explore server-side fixes!

Issue #1768, opened 2025-12-10
UrlElicitationRequiredError not thrown as an McpError exception


gopitk:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

I was testing the snippets in the repo corresponding to the URL mode elicitation. When I run connect-service tool the server seems to be raising UrlElicitationRequiredError. However on the client I dont see this being translated into the McpError exception and instead just seeing as a result error. 

### Example Code

```Python
Using the standard snippets. 

cd examples/snippets
# Server
uv run server elicitation sse
# Client
uv run elicitation-client
```

### Python & MCP Python SDK

```Text
Python: 3.11.13
MCP Python SDK: 1.23.3
```

gopitk:
Filed a PR that passes the above case with the snippets and a new unit test.

felixweinberger:
Landed #1780 thanks!

PR #1780, opened 2025-12-11
Fix for Url Elicitation issue 1768

gopitk:
Fix for https://github.com/modelcontextprotocol/python-sdk/issues/1768

## Motivation and Context
Change is needed because UrlElicitationRequiredError was not being propagated as an exception 

## How Has This Been Tested?
Tested with the examples/snippets indicated in the original issue. Also added a unit test for this case. 

## Breaking Changes
N/A

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x ] My code follows the repository's style guidelines
- [ x] I have added appropriate error handling

### DIFF
--- a/src/mcp/server/fastmcp/tools/base.py
+++ b/src/mcp/server/fastmcp/tools/base.py
@@ -11,6 +11,7 @@ from pydantic import BaseModel, Field
 from mcp.server.fastmcp.exceptions import ToolError
 from mcp.server.fastmcp.utilities.context_injection import find_context_parameter
 from mcp.server.fastmcp.utilities.func_metadata import FuncMetadata, func_metadata
+from mcp.shared.exceptions import UrlElicitationRequiredError
 from mcp.shared.tool_name_validation import validate_and_warn_tool_name
 from mcp.types import Icon, ToolAnnotations
 
@@ -108,6 +109,10 @@ class Tool(BaseModel):
                 result = self.fn_metadata.convert_result(result)
 
             return result
+        except UrlElicitationRequiredError:
+            # Re-raise UrlElicitationRequiredError so it can be properly handled
+            # as an MCP error response with code -32042
+            raise
         except Exception as e:
             raise ToolError(f"Error executing tool {self.name}: {e}") from e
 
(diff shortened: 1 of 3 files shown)


src/mcp/server/lowlevel/server.py
```diff
@@ -569,6 +569,10 @@ async def handler(req: types.CallToolRequest):
                             isError=False,
                         )
                     )
+                except UrlElicitationRequiredError:
+                    # Re-raise UrlElicitationRequiredError so it can be properly handled
+                    # by _handle_request, which converts it to an error response with code -32042
+                    raise
```
felixweinberger: same as above, this seems like the wrong place to introduce this concern

src/mcp/server/fastmcp/tools/base.py
```diff
@@ -108,6 +109,10 @@ async def run(
                 result = self.fn_metadata.convert_result(result)
 
             return result
+        except UrlElicitationRequiredError:
```
felixweinberger, line 112: Feels slightly awkward to have this feature specific exception at such a generic level - but can't think of a better way to achieve the intended design of 1036 right now (and also matches what TS SDK does).

felixweinberger: Created an issue to potentially find a better way to raise this error, but don't want to block on this: https://github.com/modelcontextprotocol/python-sdk/issues/1788

gopitk: Thanks @felixweinberger . I considered raising it as McpError (with the specific url elicitation error code info) but thought that could break existing clients.

felixweinberger: Yeah that wouldn't be the right approach - I was thinking more something along the lines of having a `propagate_error: bool` on `McpError` for example and then have some gating inside that on whether to raise or pass, but that's definitely overengineering for one exception case imo.

Issue #1785, opened 2025-12-14
Discover OIDC-only authorization servers when PR metadata is absent


sonmaximum:
### Summary
The client only discovers authorization servers that expose RFC 8414 OAuth metadata when no `resource_metadata` link is present.  In the no-metadata case the code currently stops after probing `/.well-known/oauth-authorization-server` and never tries `/.well-known/openid-configuration`, so OIDC-only providers fail to authenticate.

### Details
- Relevant code in method `build_oauth_authorization_server_metadata_discovery_urls` in file `src/mcp/client/auth/utils.py`
- When `auth_server_url` is missing the funtcion builds a single URL (`/.well-known/oauth-authorization-server`) derived from the MCP server's origin.
- The fallback that probes `/.well-known/openid-configuration` only executes when `authorization_server_url` exists and has a path component, so most deployments without RFC 9728 metadata cannot reach OIDC discovery, even if they have valid urls existing.

### Steps to Reproduce
1. Configure the MCP client to talk to a protected resource that does **not** emit `resource_metadata` in `WWW-Authenticate`.
2. Ensure the linked authorization server only publishes `/.well-known/openid-configuration` (no OAuth metadata document).
3. Run the client: discovery fails immediately because the only attempted URL is `/.well-known/oauth-authorization-server`.

### Expected
The client should probe both OAuth and OIDC well-known endpoints (including path-aware variants) regardless of whether the URL came from protected-resource metadata or static config.

### Actual
Only the OAuth well-known endpoind is checked, so OIDC-only servers cannot be discovered and authentication stops.

### Impact
MCP servers that rely on an OpenID Provider without RFC 8414 metadata (which is allowed by the OpenID spec) are inaccessible with the current library.

sonmaximum:
created PR #1786 to potentially address the issue

sonmaximum:
Hi @maxisbey (sorry, this is my first time trying to contribute to this probject so I don't know the exact protocol, but) I saw that you had commented on a seemingly similar issue recently  in #1751/#1762 so I thought you might want to look at this (I think what I'm getting at it related but slightly different)

pcarleton:
Hi @sonmaximum, the fallback behavior you're describing on this line is support for the 2025-03-26 MCP specification version which is very specific about what path to use to find the metadata:
[...]

PR #1786, opened 2025-12-14
Broaden authorization server discovery to include OIDC fallbacks

sonmaximum:
### Summary
Extend `build_oauth_authorization_server_metadata_discovery_urls` so that the MCP client always probes both OAuth and OpenID discovery endpoints, with path-aware  variants and deduplication.  This lets the client interoperate with authorization servers that only expose `/.well-known/openid-configuration`.

## Change Details
- Use `auth_server_url or server_url` as a single source of truth, so we always consider the configured base when no metadata URL is advertised.
- Populate discovery attempts with:
  - Path-aware OAuth metadata
  - Path-aware OIDC metadata (`/.well-known/openid-configuration{path}`)
  - Nested-path OIDC (`{path}/.well-known/openid-configuration`)
  - Root-level OAuth and OIDC fallbacks
- Deduplicate the list to avoid redundant fetches when different branches yield the same URL.

## Motivation and Context
- Some IdPs only publish OIDC discovery.  Without trying those endpoints, the OAuth flow fails even though the IdP is spec compliant.
- RFC 8414 explicitly recommends path-aware fallbacks for both OAuth and OIDC discovery; the existing implementation only covers part of that matrix.
- Deduping avoids duplicated network calls with `path == "/"` or when config already points at a well-known URL.

## How Has This Been Tested?
- Verified URL list generation manually for:
  -  No `resource_metadata`, base URL `https://idp.example.com` -> probes both OAuth and OIDC root well-known endpoints.
  - Path-based `auth_server_url` like `https://idp.example.com/tenantA` -> probes path-aware OAuth/OIDC endpoints plus root fallbacks without duplicates.
- Updated existing tests with updated behavior and added new tests to covered broadened case coverage.  Ensured pytest tests pass. 

## Types of change
(either depending on perspective?)

## Checklist

## Additional context

See related issue #1785

### DIFF
--- a/src/mcp/server/fastmcp/tools/base.py
+++ b/src/mcp/server/fastmcp/tools/base.py
@@ -11,6 +11,7 @@ from pydantic import BaseModel, Field
 from mcp.server.fastmcp.exceptions import ToolError
 from mcp.server.fastmcp.utilities.context_injection import find_context_parameter
 from mcp.server.fastmcp.utilities.func_metadata import FuncMetadata, func_metadata
+from mcp.shared.exceptions import UrlElicitationRequiredError
 from mcp.shared.tool_name_validation import validate_and_warn_tool_name
 from mcp.types import Icon, ToolAnnotations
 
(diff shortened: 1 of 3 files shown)


pcarleton:
see comment:
https://github.com/modelcontextprotocol/python-sdk/issues/1785#issuecomment-3710134033

PR #1787, opened 2025-12-15
docs: Annotate unused parameters for structured output examples

xtzie:
## Summary
Update the README and example code to clarify that parameters are defined but not used in the examples.

## Rationale
All other example snippets consistently use passed parameters. It was briefly confusing when I read the structured output snippets and had to double back to ensure they were indeed placeholder variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,7 +475,7 @@ class WeatherData(BaseModel):
 
 
 @mcp.tool()
-def get_weather(city: str) -> WeatherData:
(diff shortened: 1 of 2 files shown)


maxisbey:
Not really needed. These are more pseudo code examples rather than production code, it doesn't make much sense to show that the params aren't used since they clearly wouldn't be with dummy data. They're there for demonstration purposes rather than actual functinoality.

PR #1793, opened 2025-12-17
feat: support setting title and description for server

maxisbey:
Add support for the `title` and `description` fields in the `Implementation` type, allowing servers to provide human-readable metadata during initialization.

Changes:
- Add `description` field to `types.Implementation`
- Add `server_title` and `server_description` to `InitializationOptions`
- Wire up title/description through `Server` constructor and session
- Add test for title and description passthrough

Github-Issue:#1783

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

Issue #1783, opened 2025-12-12
Support setting title and description for server and client


JulianJvn:
### Description

The MCP spec schema for [`Implementation`](https://modelcontextprotocol.io/specification/2025-11-25/schema#implementation), which is used for `clientInfo` and `serverInfo` in [`InitializeRequestParams`](https://modelcontextprotocol.io/specification/2025-11-25/schema#initializerequestparams) and [`InitializeResult`](https://modelcontextprotocol.io/specification/2025-11-25/schema#initializeresult), features the two fields

* `title` for a human-readable title of the MCP client/server, and
* `description` for a human-readable description of the MCP client/server.

Curiously, it is not possible to set these basic fields for server and client with the Python SDK.

To be more precise, [`types.Implementation`](https://github.com/modelcontextprotocol/python-sdk/blob/65b36de4eb930a471a5eeeddf078dc8b240a90f2/src/mcp/types.py#L263-L274) only supports setting the title (via the superclass [`BaseMetadata`](https://github.com/modelcontextprotocol/python-sdk/blob/65b36de4eb930a471a5eeeddf078dc8b240a90f2/src/mcp/types.py#L231-L245)), the description is missing altogether and therefore can't be set for both client and server.

In addition, while the client accepts an `Implementation`-type [`client_info`](https://github.com/modelcontextprotocol/python-sdk/blob/65b36de4eb930a471a5eeeddf078dc8b240a90f2/src/mcp/client/session.py#L122) object during initialization (so it at least permits setting the title), the server only accepts a `server.models.InitializationOptions`-type [`init_options`](https://github.com/modelcontextprotocol/python-sdk/blob/65b36de4eb930a471a5eeeddf078dc8b240a90f2/src/mcp/server/session.py#L92) object, which [doesn't support neither title nor description](https://github.com/modelcontextprotocol/python-sdk/blob/65b36de4eb930a471a5eeeddf078dc8b240a90f2/src/mcp/server/models.py#L14-L20).

The feature request is to support setting the title and the description for both clients and servers.

The use case is as the MCP specification says: I'd like to provide human-readable titles and descriptions to MCP hosts.

the-ayyi:
Hi Max, I was planning on tackling the client-side changes. Are you already working on that part, or would you be okay with me picking it up?

maxisbey:
Hey @woshiyixinke, yea happy for you to take it on if you'd like!

[...]

PR #1634, opened 2025-11-15
Server initialize response update to last spec (add title, description)

yannj-fr:
To match the last protocol spec, the Implementation schema must be updated and new attribute propagated on server constructor.
## Motivation and Context
To match the last protocol specification update.

## How Has This Been Tested?
End To End test, also added to the test suite
## Breaking Changes
n/a
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context
Also fix the version attribute which was not flowing to the _mcp_server object so was always defaulted to package version.

PR #1799, opened 2025-12-17
feat: support setting title and description for server

the-ayyi:
This PR is based on another draft [PR](https://github.com/modelcontextprotocol/python-sdk/pull/1793). 
#### Core Type Changes
- **`src/mcp/types.py`**: Added `description: str | None` field to `Implementation` class

#### Server Model Changes  
- **`src/mcp/server/models.py`**: Added `title` and `description` fields to `InitializationOptions`

#### Low-Level Server Changes
- **`src/mcp/server/lowlevel/server.py`**: 
  - Updated `Server.__init__()` to accept `title` and `description` parameters
  - Updated `create_initialization_options()` to include these fields

#### Session Changes
- **`src/mcp/server/session.py`**: Updated to pass `title` and `description` to `serverInfo` in `InitializeResult`

#### FastMCP Changes
- **`src/mcp/server/fastmcp/server.py`**: 
  - Added `title` and `description` parameters to `__init__()`
  - Added `@property` methods for read access
  - Passes values through to underlying `MCPServer`

Github-Issue:https://github.com/modelcontextprotocol/python-sdk/issues/1783

## Motivation and Context

The MCP specification defines an `Implementation` schema (used for both `clientInfo` and `serverInfo` in initialization) that includes both `title` and `description` fields for providing human-readable metadata to MCP hosts. However, the Python SDK had several gaps:

1. **Missing `description` field**: While `Implementation` inherited `title` from `BaseMetadata`, it was missing the `description` field entirely, even though the MCP spec supports it.

2. **Server initialization limitations**: The server only accepted `InitializationOptions` which didn't support setting `title` or `description`, making it impossible for servers to provide this metadata to clients.

3. **Inconsistent support**: Clients could at least set `title` via `clientInfo` (an `Implementation` object), but servers had no way to set either field.

This PR addresses the server-side implementation, allowing MCP servers to provide human-readable titles and descriptions that can be displayed by MCP hosts in their user interfaces, improving the user experience when working with multiple MCP servers.

#### Tests
- **`tests/server/test_session.py`**: Added test for low-level server title/description
- **`tests/server/fastmcp/test_server.py`**: Added tests for FastMCP with and without title/description

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/types.py
+++ b/src/mcp/types.py
@@ -244,15 +244,6 @@ class BaseMetadata(BaseModel):
     if present).
     """
 
-    description: str | None = None
-    """
-    An optional human-readable description of what this implementation does.
-
-    This can be used by clients or servers to provide context about their purpose
-    and capabilities. For example, a server might describe the types of resources
-    or tools it provides, while a client might describe its intended use case.
-    """
-
 
 class Icon(BaseModel):
     """An icon for display in user interfaces."""
@@ -281,6 +272,15 @@ class Implementation(BaseMetadata):
     """An optional list of icons for this implementation."""
 
     model_config = ConfigDict(extra="allow")
+
+    description: str | None = None
+    """
+    An optional human-readable description of what this implementation does.
+
+    This can be used by clients or servers to provide context about their purpose
+    and capabilities. For example, a server might describe the types of resources
+    or tools it provides, while a client might describe its intended use case.
+    """
 
 
 class RootsCapability(BaseModel):


src/mcp/server/fastmcp/server.py
```diff
@@ -150,6 +150,8 @@ def __init__(  # noqa: PLR0913
         instructions: str | None = None,
         website_url: str | None = None,
         icons: list[Icon] | None = None,
+        title: str | None = None,
+        description: str | None = None,
```
maxisbey: could you move these to the end of the params, just so we don't break anyone using positional parameter passing

src/mcp/server/lowlevel/server.py
```diff
@@ -142,6 +142,8 @@ def __init__(
         instructions: str | None = None,
         website_url: str | None = None,
         icons: list[types.Icon] | None = None,
+        title: str | None = None,
+        description: str | None = None,
```
maxisbey: same as other comment, move to end of params

tests/server/test_session.py
```diff
@@ -83,6 +83,62 @@ async def run_server():
...
+@pytest.mark.anyio
+async def test_server_session_initialize_with_title_and_description():
+    """Test that server_title and server_description are passed through to serverInfo."""
+    server_to_client_send, server_to_client_receive = anyio.create_memory_object_stream[SessionMessage](1)
+    client_to_server_send, client_to_server_receive = anyio.create_memory_object_stream[SessionMessage](1)
+
+    async def message_handler(
+        message: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
+    ) -> None:
+        if isinstance(message, Exception):  # pragma: no cover
```
maxisbey: remove all pragmas from the test suite, instead use asserts

the-ayyi: Since those pragmas are already used extensively throughout this file, removing them all would significantly expand the scope of this PR. I'm happy to stick to the 'no new pragmas' rule for this change to keep the focus on the title and description support.

maxisbey: Sorry yea that is what I mean, just removing the ones you added, not the existing one!

src/mcp/types.py
```diff
@@ -244,6 +244,15 @@ class BaseMetadata(BaseModel):
     if present).
     """
 
+    description: str | None = None
```
maxisbey: This shouldn't have a description field, only Implementation should. Check the spec schema here: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6e924e6f587f30aa810325c0985e2da0b257d4eb/schema/draft/schema.ts#L526

maxisbey (changes requested):

src/mcp/types.py
the-ayyi: Thanks for catching that, Max. I've moved the description field to the Implementation class to align with the [MCP spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6e924e6f587f30aa810325c0985e2da0b257d4eb/schema/draft/schema.ts#L526). I initially thought they belonged together based on their purpose, but it's helpful to know how this is handled in other SDK implementations.

Kludex:
- Closed by https://github.com/modelcontextprotocol/python-sdk/pull/1634

PR #1808, opened 2025-12-20
fix: return HTTP 404 for unknown session IDs instead of 400

### LINKED ISSUE #1727: StreamableHTTPSessionManager returns 400 instead of 404 for unknown session ID
## Description

`StreamableHTTPSessionManager._handle_stateful_request()` returns HTTP 400 Bad Request when a session ID is provided but not found in the server instances. According to the MCP spec, this should be HTTP 404 Not Found.

## Spec Reference

From the [Streamable HTTP transport spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management):

> Servers that require a session ID SHOULD respond to requests without an `Mcp-Session-Id` header (other than initialization) with **HTTP 400 Bad Request**.

> The server MAY terminate the session at any time, after which it MUST respond to requests containing that session ID with **HTTP 404 Not Found**.

> When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it MUST start a new session by sending a new `InitializeRequest` without a session ID attached.

## Current Behavior

In `src/mcp/server/streamable_http_manager.py`, the `_handle_stateful_request` method:

```python
if request_mcp_session_id is not None and request_mcp_session_id in self._server_instances:
    # Handle existing session
    ...
elif request_mcp_session_id is None:
    # Create new session
    ...
else:
    # Session ID provided but not found
    response = Response(
        "Bad Request: No valid session ID provided",
        status_code=HTTPStatus.BAD_REQUEST,  # ← Should be NOT_FOUND (404)
    )
```

## Expected Behavior

The `else` branch should return 404 Not Found:

```python
else:
    response = Response(
        "Not Found: Session not found",
    )
```

## Why This Matters

The distinction matters for client behavior:
- **400** indicates a malformed request - clients may retry or report an error
- **404** indicates the session doesn't exist - clients know to re-initialize

The transport layer (`streamable_http.py`) correctly implements this in `_validate_session()`, but the manager layer has its own session lookup that doesn't follow the same pattern.

## Note

The TypeScript SDK's `StreamableHTTPServerTransport` correctly returns 404 for session mismatch in its `validateSession` method.

max-rousseau:
## Summary

Fixes incorrect HTTP status code returned by `StreamableHTTPSessionManager` when receiving requests with unknown or expired session IDs.

Per the MCP Streamable HTTP transport specification, servers MUST respond with HTTP 404 (Not Found) when the session ID is not recognized. The previous implementation returned HTTP 400 (Bad Request), which incorrectly signals a malformed request rather than a missing resource.

This change aligns the Python SDK behavior with:
- The MCP specification requirements
- The TypeScript SDK implementation

## Changes

- Return HTTP 404 instead of 400 for unknown session IDs
- Use JSON-RPC error format in the response body, matching TypeScript SDK behavior

Fixes #1727

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -279,6 +279,8 @@ class StreamableHTTPSessionManager:
                 await http_transport.handle_request(scope, receive, send)
         else:
             # Unknown or expired session ID - return 404 per MCP spec
+            # TODO: Align error code once spec clarifies
+            # See: https://github.com/modelcontextprotocol/python-sdk/issues/1821
             error_response = JSONRPCError(
                 jsonrpc="2.0",
                 id="server-error",


maxisbey (approved):
Thanks for the fix! I've gone ahead and used the existing type objects and added a unit test so we can get this merged

maxisbey (approved):

Issue #1801, opened 2025-12-18
Implement server-side support for Client ID Metadata Documents (CIMD)


maxisbey:
## Summary

PR #1652 implemented client-side support for Client ID Metadata Documents (CIMD) per SEP-991, but the server-side implementation is missing. Authorization servers built with the Python SDK cannot currently support CIMD.

## Background

CIMD ([draft-ietf-oauth-client-id-metadata-document-00](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00)) allows OAuth clients to use HTTPS URLs as client identifiers, where the URL points to a JSON document containing client metadata. This is the **recommended** registration approach per the MCP spec (ahead of DCR).

From the [MCP Authorization spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents):

> Authorization servers and MCP clients **SHOULD** support OAuth Client ID Metadata Documents

### Client-side (implemented in #1652)
- ✅ `is_valid_client_metadata_url()` - validates HTTPS URLs with path component
- ✅ `should_use_client_metadata_url()` - checks if server advertises CIMD support
- ✅ `create_client_info_from_metadata_url()` - uses URL as client_id
- ✅ OAuth flow integration with DCR fallback

### Server-side (missing)
- ❌ Never advertises `client_id_metadata_document_supported=true` in OAuth metadata
- ❌ No detection of URL-formatted client_ids
- ❌ No metadata document fetching
- ❌ No validation (client_id matching, redirect_uri verification, document structure)
- ❌ No caching infrastructure
- ❌ No SSRF protection for fetching

## Spec Requirements for Authorization Servers

From the MCP spec and CIMD RFC:

| Requirement | Level |
|-------------|-------|
| Fetch metadata documents when encountering URL-formatted client_ids | SHOULD |
| Validate that fetched document's `client_id` matches the URL exactly | MUST |
| Validate redirect URIs against those in the metadata document | MUST |
| Validate document structure is valid JSON with required fields | MUST |
| Cache metadata respecting HTTP cache headers | SHOULD |
| Protect against SSRF attacks | SHOULD |
| Limit document size (~5KB recommended) | SHOULD |
| Display client_id hostname during consent | SHOULD |
| Never cache error responses | MUST NOT |

## References

- MCP Spec: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents
[...]

PR #1810, opened 2025-12-20
feat(cmid-server): Implement Server-Side Support for Client ID Metadata Documents (CIMD)

punitmahes:
Implement Server-Side Support for Client ID Metadata Documents (CIMD)

## Motivation and Context
This PR addresses issue #1801 by implementing server-side support for Client ID Metadata Documents (CIMD), as defined in SEP-991 and the MCP authorization specification.
Previously, the Authorization Server logic in the Python SDK could not resolve or validate clients that identify themselves via a URL (CIMD). This change enables the server to:

- **Advertise Support**: Tells clients the server supports CIMD by adding client_id_metadata_document_supported: True to the server metadata.
- **Dynamic Resolution**: Detects when a client_id is an HTTPS URL during the authorization flow.
- **Fetch & Validate:** Fetches the metadata document from the provided URL and validates that the client_id inside the document matches the request.
- **Fallback Mechanism**: If a client is not found in the static provider registry, it attempts to resolve it as a CIMD before returning an error.

## How Has This Been Tested?
Added Unit Test Cases for the CIMD Server authorization flow - 
- test_cimd_authorization_flow
- test_cimd_authorization_invalid_cimd_url
- test_cimd_authorization_invalid_client_id
- test_cimd_authorization_metadata_fetch_error

## Breaking Changes
No

## Types of changes

## Checklist

### DIFF
--- a/src/mcp/server/auth/handlers/authorize.py
+++ b/src/mcp/server/auth/handlers/authorize.py
@@ -2,6 +2,7 @@ import logging
 from dataclasses import dataclass
 from typing import Any, Literal
 
+import httpx
 from pydantic import AnyUrl, BaseModel, Field, RootModel, ValidationError
 from starlette.datastructures import FormData, QueryParams
 from starlette.requests import Request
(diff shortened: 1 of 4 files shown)


src/mcp/server/auth/handlers/authorize.py
```diff
@@ -166,6 +171,29 @@ async def error_response(
             client = await self.provider.get_client(
                 auth_request.client_id,
             )
+            if not client:
+                # Check if `client_id` is a valid URL for Metadata Document
+                if auth_request.client_id.startswith("https://"):
```
maxisbey: We should be doing more validation here as per the RFC for all the MUST and MUST NOTs: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00#section-3

punitmahes:
@maxisbey Any updates on this PR?

PR #1815, opened 2025-12-24
Split _perform_authorization_code_grant code

HenriChabert:
## Description
The goal of this PR is to be able to reuse the function to create authorization url. Currently, it is not possible as the url is built in the same function that also handles callback functions.

## Motivation and Context
It is pretty comfortable to inherit from `OAuthClientProvider` when we want to create a specific client for a MCP server. However, we may not want to use the workflow that uses callbacks to authenticate, but rather build the authorization URL, send it to frontend, have our own `callback` url, and then proceed to code exchange.
For the moment, we have to reimplement the URL creation as it the code is located inside the same function that uses callbacks.
This change also brings more separation into a long function.

## How Has This Been Tested?
These changes are minimal, and is just separating a function into 2 parts. I don't think some specific new tests are needed, but if it is needed, I would be happy to test this new function 

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -12,7 +12,7 @@ import string
 import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from dataclasses import dataclass, field
-from typing import Any, Protocol
+from typing import Any, NewType, Protocol
 from urllib.parse import quote, urlencode, urljoin, urlparse
 
 import anyio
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for the PR! Per our [CONTRIBUTING.md](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md), PRs require a corresponding issue so we can agree on scope before reviewing code. Closing for now — please open an issue first and re-submit once it's labeled `ready for work`.

PR #1816, opened 2025-12-26
perf: optimize hot paths with caching and O(1) operations

herniqeu:
- Replace list.pop(0) with deque.popleft() for O(1) queue dequeue
- Cache compiled regex patterns in ResourceTemplate for URI matching
- Cache field info mapping in FuncMetadata via lazy property
- Throttle expired task cleanup with interval-based execution

These optimizations target high-frequency operations in message queuing, resource lookups, tool calls, and task store access.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -70,6 +70,18 @@ class FuncMetadata(BaseModel):
     output_schema: dict[str, Any] | None = None
     output_model: Annotated[type[BaseModel], WithJsonSchema(None)] | None = None
     wrap_output: bool = False
+    _key_to_field_info: dict[str, FieldInfo] | None = None
+
+    @property
+    def key_to_field_info(self) -> dict[str, FieldInfo]:
+        if self._key_to_field_info is None:
+            mapping: dict[str, FieldInfo] = {}
+            for field_name, field_info in self.arg_model.model_fields.items():
+                mapping[field_name] = field_info
+                if field_info.alias:
+                    mapping[field_info.alias] = field_info
+            self._key_to_field_info = mapping
+        return self._key_to_field_info
 
     async def call_fn_with_arg_validation(
         self,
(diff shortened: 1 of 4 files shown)


src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -17,14 +17,13 @@
 from mcp.shared.experimental.tasks.store import TaskStore
 from mcp.types import Result, Task, TaskMetadata, TaskStatus
 
+CLEANUP_INTERVAL_SECONDS = 1.0
```
dchou1618, line 20: I guess this would be a breaking change? Setting a throttling limit on the cleanup may not resolve all expired tasks and is timing-dependent, but this makes sense if the tasks are accessed very frequently (though I'm not sure what the benchmarks on this look like). How often does this happen/are there use cases that would be addressed with throttling cleanup?

src/mcp/server/fastmcp/utilities/func_metadata.py
dchou1618, line 1: nit: iiuc, these changes optimize repeated calls to pre_parse_json. I'm not sure what the protocol is for optimization-related tests, but it could be useful to add a small benchmark in the description on a ~20 field pydantic model and call pre_parse_json repeatedly (50k-100k before vs after).

maxisbey:
Thanks for the contribution, but closing this PR for now:

1. **No linked issue** - There's no evidence these are actual performance problems. Please open an issue first describing the bottleneck you're seeing.

2. **Multiple unrelated changes** - This bundles 4 separate optimizations (deque, regex caching, field info caching, cleanup throttling). These could potentially be separate PRs, or combined into one if there's a coherent issue describing the problem - but either way, an issue first would help.

3. **Unrelated modifications** - The diff includes changes unrelated to the optimizations: removing docstrings, removing explanatory comments, and code style changes. These shouldn't be mixed in.

Happy to reconsider if you open an issue with benchmarks showing the problem.

Issue #1811, opened 2025-12-21
client's `read_stream_writer` open after SSE disconnection hanging `.receive()`


ivanbelenky:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

This issue MAY be related to the following
- https://github.com/modelcontextprotocol/python-sdk/issues/1805
- https://github.com/modelcontextprotocol/python-sdk/issues/1764
- https://github.com/modelcontextprotocol/python-sdk/issues/262

Experiencing deadlocks on `streamable_http` transport. In order to reproduce the issue the following can be run.

```python
import asyncio
import logging
import threading
import time

from fastmcp import Client, FastMCP
from fastmcp.client.transports import StreamableHttpTransport

logging.basicConfig(
    level=logging.DEBUG,
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
    handlers=[logging.StreamHandler()],
)

logger = logging.getLogger(__name__)

HOST = "127.0.0.1"
PORT = 8765
SERVER_URL = f"http://{HOST}:{PORT}/mcp"

mcp = FastMCP(name="timeout issue")

SSE_TIMEOUT = 0.1
SLEEP = 60


@mcp.tool
def blocking_call() -> str:
    time.sleep(SLEEP)
    return "42"


def run_server():
    mcp.run(transport="streamable-http", host=HOST, port=PORT)


async def test_blocking_sse():
    transport = StreamableHttpTransport(SERVER_URL, sse_read_timeout=SSE_TIMEOUT)
    async with Client(transport) as client:
        tools = await client.list_tools()
        print(f"available tools: {[t.name for t in tools]}")
        result = await client.call_tool("blocking_call", {})
        print(f"blocking result: {result}")


if __name__ == "__main__":
    server_thread = threading.Thread(target=run_server, daemon=True)
    server_thread.start()

    time.sleep(2)

    try:
        asyncio.run(test_blocking_sse())
    finally:
        logger.info(f"{time.strftime('%Y-%m-%d %H:%M:%S')} [CLIENT] Shutting down...")
```

Execution context := locally built fastmcp at this commit [790ea92](https://github.com/jlowin/fastmcp/tree/790ea92eb59256da68c83097321ebde8f8819bcf)

[...]

Issue #262, opened 2025-03-12
cannot get response from  await session.call_tool()


herrrX:
hello, I'm sorry to bother you. I've recently encountered a blocking issue with the MCP client.
I've set up an MCP server, and it runs in Claude successfully. However, when I use the client I set up, the call to the tool result = await session.call_tool() gets stuck and doesn't return a response. But when I use tools = await session.list_tools(), I can retrieve the list of tools.

Through logging within the client file, I'm able to confirm that the tool call has been initiated and executed successfully, and the result is obtained, but the client side can't receive the result. It seems there's an issue with the transmission layer of the protocol. I've been struggling with this for two days and have looked for solutions without success, even after upgrading the system version.

I removed the other logic and only kept the tool invocation. The code is as follows.

Looking forward to your reply.

```

from mcp import ClientSession, StdioServerParameters, types
from mcp.client.stdio import stdio_client


server_params = StdioServerParameters(  # Create server parameters for stdio connection
    command="uvx",
    args=[
        "--from",
        "git+ssh://git@git.sankuai.com/~xuxianhui02/test-get-api-info-from-sy.git",
        "test-get-api-info-from-sy"
    ],
    env=None
)

# Optional: create a sampling callback
async def handle_sampling_message(message: types.CreateMessageRequestParams) -> types.CreateMessageResult:
    return types.CreateMessageResult(
        role="assistant",
        content=types.TextContent(
            type="text",
            text="Hello, world! from model",
        ),
        model="gpt-3.5-turbo",
        stopReason="endTurn",
    )

async def run():
    async with stdio_client(server_params) as (read, write):
        async with ClientSession(read, write, sampling_callback=handle_sampling_message) as session:
            # Initialize the connection
            await session.initialize()
            # List available tools
            tools = await session.list_tools()
            # Call a tool
            result = await session.call_tool("query-api-infos", arguments={"api_info_id": "8768555"})
            print(result)

if __name__ == "__main__":
    import asyncio
    asyncio.run(run())
```

taha-yassine:
I'm experiencing the same issue.
When stepping into `session.call_tool()` with a debugger and going through `BaseSession.send_request()` line by line, it seems to work, but I have no idea why.

STZhuang:
[...]

PR #1817, opened 2025-12-26
fix: cleanup resources properly on `BaseSession::_receive_loop` cleanup

challenger71498:
This PR fixes Session cleanup is being failed, caused by task cancellation via `__aexit__` not handled properly.

## Motivation and Context
Session cleanups the receive streams via `finally` clause of `BaseSession::_receive_loop`.
- When the session is being closed, or leaving the async context(which calls `__aexit__`), it cancels the TaskGroup.
- When a TaskGroup is being cancelled, it cancels every task children.
- The parent of task we are using in `finally` clause, such as `MemoryObjectStream::send` is its base coroutine `_receive_loop`, **IS** the TaskGroup which is being closed.
- Therefore the cleanup process is being cancelled while closing, leaving all receive streams not being closed, causing stream leak.

I fixed this problem by adding `CancelScope` with `shield=True` on `finally` clause, ensures the cleanup logic to be executed properly even TaskGroup cancel is requested.

Also, while cleaning up receive streams, we loop streams, therefore stream dictionary must not change.
- If you change dictionary while looping, it causes runtime error, which should not happen in production.
- `BaseSession` AS-IS does not protect dictionary change on cleanup.
- To handle this minor issue, I added a simple flag to signal that the session is cleaning up the resources.

## How Has This Been Tested?
I found this issue while debugging the problem, using ADK AgentTool with MCP causes an abtruse error:
```plaintext
```

After a week of debugging, I realised the ultimate fix to this problem is quite huge and need fixes both MCP SDK and Google ADK, and this fix is one of them.

I added a simple test which triggers this issue.
- In this test..
  - Server never responses
  - Creates the ClientSession and request 2 ping requests to server asynchronously
  - While 2 tasks are waiting (since the server never responses), leaves the async with block.
  - ClientSession is being closed
  - Assert that total count of response stream is equal to 0, not 2
  - Assert that the closed ClientSession fails to send ping
- BaseSession AS-IS fails this test, response stream count is still 2, since the cleanup logic is being cancelled.

## Breaking Changes
- No breaking changes at all since this is just a minor bugfix.
- There might be a side-effect, since we are now execute `finally` block until every stream is being closed, the `__aexit__` will be blocked until `finally` block finishes its execution.
  - If the `finally` block is being executed too long, user might think it is hanging.
  - But I don't think this really be a problem, stream is running on memory not an actual network, also the amount of the stream won't be that huge, expecting ~1000, then it will be fine.
  - If it IS being a problem, we could just run closing logics in parallel. (I don't want that much change in this PR though.)

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] I have added or updated documentation as needed `No docs needed to be updated`

## Additional context
This PR **MAY** resolve issues below:
- https://github.com/modelcontextprotocol/python-sdk/issues/1811

I think this problem is related to task leak (or indefinite-waiting) at Google ADK, since in ADK the tool call await forever until the receive stream responds.


egerev:
Hi! We're experiencing the same `RuntimeError: Attempted to exit cancel scope in a different task` issue when using `MCPServerStreamableHttp` with Linear's MCP server (`https://mcp.linear.app/mcp`).

The error occurs in `streamable_http.py` during the task group cleanup — specifically at `tg.cancel_scope.cancel()` in the finally block.

Your fix with `CancelScope(shield=True)` for `session.py` looks like the right approach. Would a similar fix be needed for `streamable_http.py` as well?

This is a blocking issue for us — we had to disable HTTP-based MCP servers entirely. Any timeline on when this might be reviewed/merged?

Happy to help test or contribute if needed.

challenger71498:
Hi @egerev, it seems like you are developing an MCP client!

### TL; DR
Wrap the client with a dedicated task. Reference [here](https://github.com/google/adk-python/pull/4025) for impl.

### Details
The error you've mentioned, `RuntimeError: Attempted to exit cancel scope in a different task`, occurs because of the violation of AnyIO's structured concurrency. When using AnyIO's `TaskGroup`, the task which *initiates* the `TaskGroup` with `with` clause should *exit* the `TaskGroup`.

If your project utilizes AnyIO for asynchronous programming, this may be not an issue. But if the project does not, you need an adapter, which wraps the client with a task. A dedicated task will prevent exiting the cancel scope in a different task.

You may reference my implementation at https://github.com/google/adk-python/pull/4025. Google ADK also had this issue, and I resolved it via the method I've mentioned above.

### Shall this PR fix the RuntimeError?
I am afraid that `CancelScope(shield=True)` will not resolve the issue, since it only protects the execution of the code from the `CancelledError`, not guaranteeing the same task will exit the cancel scope.

---
Please let me know if there is a correction needed.
Your opinions are greatly appreciated :)
Thank you.

PR #1818, opened 2025-12-31
fix: auto-reinitialize client session on HTTP 404

### LINKED ISSUE #1676: MCP client doesn't not initialize new session when getting 404 session not found
### Initial Checks

### Description

Per https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management
```
When a client receives HTTP 404 in response to a request containing an MCP-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.
```
This is not the case with the python client. It just sends a session terminated error in [streamable_http.py](https://github.com/modelcontextprotocol/python-sdk/blob/5983a650cc07d2dc6c6ba098e99d3545889157a9/src/mcp/client/streamable_http.py#L277)

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
mcp: 1.22.0
python: 3.14.0
```

jayhemnani9910:
## Summary

  Per MCP spec, when the server returns HTTP 404 indicating the session has expired, the client MUST start a new session by sending a new InitializeRequest without a session ID attached.

  This PR implements automatic session recovery:

  - Add `SESSION_EXPIRED` error code (-32002) to types.py
  - Modify transport 404 handling to clear `session_id` and signal `SESSION_EXPIRED` for non-initialization requests
  - Override `send_request` in `ClientSession` to catch `SESSION_EXPIRED`, re-initialize the session, and retry the original request
  - Prevent infinite loops with `_session_recovery_attempted` flag

  ## Test plan

  - [x] `test_session_recovery_on_expired_error` - Verify client re-initializes on SESSION_EXPIRED
  - [x] `test_no_infinite_retry_loop_on_repeated_session_expired` - Verify max retry limit prevents infinite loops
  - [x] `test_non_session_expired_error_not_retried` - Verify other errors don't trigger recovery
  - [x] `test_session_recovery_preserves_request_data` - Verify original request data is preserved through recovery
  - [x] All 66 related tests pass (pytest)
  - [x] Type checking passes (pyright)
  - [x] Linting passes (ruff)

  Fixes #1676


jayhemnani9910:
Testing note: The transport-level 404 → SESSION_EXPIRED conversion is covered by pragma: no cover as these tests had async cleanup issues when running without pytest-xdist (which CI disables for coverage collection). The session-level tests adequately verify the recovery behavior end-to-end.

jayhemnani9910:
Closing this. The approach here is stale against the current architecture and it is not worth rebasing.

Since this branch was opened, `BaseSession` and `src/mcp/shared/session.py` are gone in favour of the dispatcher, `ClientSession.send_request` is concrete and goes through `self._dispatcher.send_raw_request`, `McpError` became `MCPError` with a different constructor, `mcp.types` became a package with the error codes living in `mcp_types/jsonrpc.py`, and the transport's 404 branch was rewritten. Nothing in this diff applies any more.

The bigger problem is the mechanism: this PR allocates `SESSION_EXPIRED = -32002`, and `mcp_types/jsonrpc.py` now marks `-32002` as reserved-never-reused. Any correct fix needs a different code out of the `-32000..-32019` SDK band.

#1676 is still open and the bug is still real, main turns a post-session 404 into `INVALID_REQUEST` "Session terminated" and never clears the session id. @nankingjing has already scoped a v2 fix there and reached the same conclusion about this PR and about `-32002`, so I would rather not have two competing branches. Leaving the issue to them.

Thanks @maxisbey and @felixweinberger for the reproductions along the way.

PR #1824, opened 2026-01-04
Add custom content support to ToolError for isError responses

### LINKED ISSUE #348: No way to set isError=True for arbitrary tool result content

hanzili:
## Summary

This PR addresses #348 by adding an optional `content` parameter to `ToolError`, allowing tools to return arbitrary content blocks (images, multiple text blocks, etc.) with `isError=True`.

**Problem**: Previously, `ToolError` only accepted a string message, which was converted to a single `TextContent`. There was no way to return rich error content like images or structured data with the error flag set.

**Solution**: Following the approach suggested in the issue discussion, `ToolError` now accepts an optional `content` parameter:

```python
# Simple text error (existing behavior, unchanged)
raise ToolError("Something went wrong")

# Error with custom content (new capability)
raise ToolError(
    "Image processing failed",
    content=[
        ImageContent(type="image", data="...", mimeType="image/png"),
        TextContent(type="text", text="Additional error details"),
    ],
)
```

## Changes

- **`src/mcp/server/fastmcp/exceptions.py`**: Add optional `content` parameter to `ToolError.__init__`
- **`src/mcp/server/lowlevel/server.py`**: Handle `ToolError` with custom content in the low-level server's `call_tool` decorator
- **`src/mcp/server/fastmcp/tools/base.py`**: Re-raise `ToolError` as-is in `Tool.run()` to preserve custom content (without this, `ToolError` would be caught by the generic `except Exception` and wrapped in a new `ToolError`, losing the `content`)

## Test Plan

Added tests in `tests/issues/test_348_tool_error_content.py` that verify:

1. Raising `ToolError("message")` still works as before (backwards compatibility)
2. Raising `ToolError("msg", content=[...])` returns the custom content with `isError=True`
3. Works with both low-level `Server` and high-level `FastMCP` APIs
4. Generic exceptions (non-ToolError) are still handled correctly

Closes #348

### DIFF
--- a/src/mcp/server/fastmcp/exceptions.py
+++ b/src/mcp/server/fastmcp/exceptions.py
@@ -1,5 +1,12 @@
 """Custom exceptions for FastMCP."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from mcp.types import ContentBlock
+
 
 class FastMCPError(Exception):
     """Base error for FastMCP."""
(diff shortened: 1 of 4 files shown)


maxisbey:
Thanks for the PR. Tracking this in #2984 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

PR #1826, opened 2026-01-04
fix: add explicit type annotation for `call_tool` decorator

jayhemnani9910:
This pull request introduces a type alias to clarify and simplify the function signature of handlers registered with `call_tool` in the `server.py` module. The main change is the introduction and use of the `CallToolHandler` type alias, which improves code readability and maintainability.

**Type hinting and handler signature improvements:**

* Added a `CallToolHandler` type alias to explicitly define the expected function signature for call tool handler functions in `server.py`.
* Updated the `call_tool` decorator to use the new `CallToolHandler` type alias, replacing the previous inline type definition for better clarity and consistency.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -126,7 +126,7 @@ plugins:
             group_by_category: false
             # 3 because docs are in pages with an H2 just above them
             heading_level: 3
-          import:
+          inventories:
             - url: https://docs.python.org/3/objects.inv
             - url: https://docs.pydantic.dev/latest/objects.inv
             - url: https://typing-extensions.readthedocs.io/en/latest/objects.inv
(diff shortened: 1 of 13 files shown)


src/mcp/server/lowlevel/server.py
```diff
@@ -105,6 +105,14 @@ async def main():
 UnstructuredContent: TypeAlias = Iterable[types.ContentBlock]
 CombinationContent: TypeAlias = tuple[UnstructuredContent, StructuredContent]
 
+# type alias for call_tool handler function signature
+CallToolHandler: TypeAlias = Callable[
```
maxisbey: Why the alias rather than just updating the type annotation? This isn't reused anywhere

PR #1827, opened 2026-01-05
fix: raise clear error for server-to-client requests in stateless mode

### LINKED ISSUE #1097: Roots/List Request Hangs in Client-Server Communication
### Initial Checks

### Description

When implementing the roots capability in a client-server setup, the communication hangs indefinitely after the server makes a `list_roots()` request. The client successfully initializes with roots capability, but the server's request for roots never completes.

Observed behavior:
1. Client initializes successfully with roots capability
2. Server receives tool call and requests roots via `ctx.session.list_roots()`
3. Communication hangs at this point - no error, no timeout, just stuck
4. Both client and server remain running but no further progress

Expected behavior:
1. Client should receive the roots/list request
2. Client should respond with the configured roots
3. Server should receive the response and continue processing

Let me know if you need any additional information or would like me to try specific debugging steps.

### Example Code

```Python
Example Code:


# server.py

mcp = FastMCP(
    name="mcp-roots-server",
    stateless_http=True
)

@mcp.tool()
async def analyze_project(ctx: Context) -> TextContent:
    print("-> Server: Requesting project roots from client...")
    roots = await ctx.session.list_roots()  # Hangs here
    
    if not roots or not roots.roots:
        return TextContent(text="No project roots found", type="text")
    
    # Never reaches this point
    return TextContent(text="Analysis complete", type="text")

mcp_app = mcp.streamable_http_app()

# client.py
import asyncio
from pathlib import Path
from typing import Any
import mcp.types as types
from mcp import ClientSession

async def mock_roots_list(context: RequestContext["ClientSession", Any]) -> types.ListRootsResult | types.ErrorData:
    print(f"<- Client: Received roots/list request")
    root_uri = f"file://{Path.cwd().absolute()}"
    return types.ListRootsResult(roots=[types.Root(
        uri=types.FileUrl(root_uri),
        name="Current Project"
    )])

async def main():
    server_url = "http://localhost:8000/mcp/"
    async with streamablehttp_client(server_url) as (read_stream, write_stream, get_session_id):
        async with ClientSession(read_stream, write_stream, list_roots_callback=mock_roots_list) as session:
            await session.initialize()
            result = await session.call_tool("analyze_project")

if __name__ == "__main__":
    asyncio.run(main())


Server logs:

INFO:     Started server process [36096]
INFO:     Application startup complete.
INFO:     127.0.0.1:51609 - "POST /mcp/ HTTP/1.1" 200 OK
INFO:     127.0.0.1:51611 - "POST /mcp/ HTTP/1.1" 202 Accepted
INFO:     127.0.0.1:51613 - "POST /mcp/ HTTP/1.1" 200 OK
Processing request of type CallToolRequest
-> Server: Requesting project roots from client...
# Hangs here


Client logs:

🚀 Connecting to MCP server at http://localhost:8000/mcp/
✅ Connected. Initializing session...
🛠️ Session initialized with roots capability.
-> Client: Calling analyze_project tool...
# Hangs here
```

### Python & MCP Python SDK

```Text
Python & MCP Python SDK:
- Python version: 3.11.7
- MCP Python SDK version: 1.10.1
- OS: macOS 24.5.0
- Environment: Development setup using uvicorn for server

Additional Notes:
- The issue occurs consistently across multiple runs
- No error messages or exceptions are thrown
- Both client and server processes remain running
- Network requests complete successfully (200 OK responses)
- The hang occurs specifically during the roots/list request
```

maxisbey:
## Motivation and Context

Fixes #1097

Server-to-client requests (list_roots, create_message, elicit_form, elicit_url) require bidirectional communication which is not available in stateless HTTP mode. Previously, these requests would hang indefinitely. Now they raise an immediate `RuntimeError` with a clear, actionable message:

```
RuntimeError: Cannot use list_roots in stateless HTTP mode. Stateless mode does not support server-to-client requests. Use stateful mode (stateless_http=False) to enable this feature.
```

### The Fundamental Problem

In stateless mode (`stateless_http=True`), each HTTP request creates a completely fresh transport with no session tracking:

```python
# streamable_http_manager.py:168-173
http_transport = StreamableHTTPServerTransport(
    mcp_session_id=None,    # NO session tracking
    event_store=None,       # NO resumability
)
```

When a server tool handler calls `session.elicit()` or `session.create_message()`:

1. **`send_request()` blocks waiting for a response** - It creates a response stream and waits:
   ```python
   response_stream, response_stream_reader = anyio.create_memory_object_stream[...]()
   await self._write_stream.send(SessionMessage(...))
   response_or_error = await response_stream_reader.receive()  # Blocks forever
   ```

2. **In JSON response mode**: The server-initiated request is **discarded** before reaching the client:
   ```python
   # streamable_http.py:548-555
   async for event_message in request_stream_reader:
       if isinstance(event_message.message.root, JSONRPCResponse | JSONRPCError):
           response_message = event_message.message
           break
       else:
           logger.debug(f"received: {event_message.message.root.method}")  # Just logs it\!
   ```

3. **In SSE mode**: The request IS sent to the client via SSE, but:
   - Client receives the elicitation request
   - Client needs to send back a JSONRPCResponse via POST
   - **In stateless mode, this POST creates a NEW transport** with no knowledge of the pending request
   - The original `send_request()` blocks forever waiting on a response stream that will never receive anything

### Why `related_request_id` Doesn't Help

The `related_request_id` metadata helps route server messages to the correct SSE stream (the one handling the original tool call), but it does NOT solve the fundamental problem:
- In JSON mode: Messages are discarded before reaching the client
- In SSE mode: The client's response creates a new transport with no correlation mechanism

### What This Looks Like in Practice

| Mode | Server→Client Request Sent? | Client Can Respond? | Result |
|------|---------------------------|---------------------|--------|
| JSON Response + Stateless | NO (discarded) | N/A | **Hangs** |
| SSE + Stateless | YES | NO (no session) | **Hangs** |
| SSE + Stateful | YES | YES | Works |

### Future Direction (SEP-1442)

The upcoming transport changes will properly address this with:
- Intermediate result responses for "pausing" requests
- Cookie mechanism for correlating resumed requests
- Sessions at the data model layer, not protocol layer

Until then, failing fast with a clear error is the best we can do.

## How Has This Been Tested?

- Reproduced the original issue with the reporter's code - confirmed hang behavior
- Added 7 new unit tests covering all affected methods
- Verified existing tests still pass

## Breaking Changes

Yes, but intentionally so. Code that previously hung indefinitely will now raise a `RuntimeError`. This is a strict improvement - failing fast with a clear error is better than hanging.

## Types of changes

## Checklist

### DIFF
--- a/src/mcp/server/session.py
+++ b/src/mcp/server/session.py
@@ -93,6 +93,7 @@ class ServerSession(
         stateless: bool = False,
     ) -> None:
         super().__init__(read_stream, write_stream, types.ClientRequest, types.ClientNotification)
+        self._stateless = stateless
         self._initialization_state = (
             InitializationState.Initialized if stateless else InitializationState.NotInitialized
         )
@@ -155,6 +156,26 @@ class ServerSession(
                 return False
 
         return True
+
+    def _require_stateful_mode(self, feature_name: str) -> None:
+        """Raise an error if trying to use a feature that requires stateful mode.
+
+        Server-to-client requests (sampling, elicitation, list_roots) are not
+        supported in stateless HTTP mode because there is no persistent connection
+        for bidirectional communication.
+
+        Args:
+            feature_name: Name of the feature being used (for error message)
+
+        Raises:
+            RuntimeError: If the session is in stateless mode
+        """
+        if self._stateless:
+            raise RuntimeError(
+                f"Cannot use {feature_name} in stateless HTTP mode. "
+                "Stateless mode does not support server-to-client requests. "
+                "Use stateful mode (stateless_http=False) to enable this feature."
+            )
 
     async def _receive_loop(self) -> None:
         async with self._incoming_message_stream_writer:
@@ -311,7 +332,9 @@ class ServerSession(
         Raises:
             McpError: If tools are provided but client doesn't support them.
             ValueError: If tool_use or tool_result message structure is invalid.
+            RuntimeError: If called in stateless HTTP mode.
         """
+        self._require_stateful_mode("sampling")
         client_caps = self._client_params.capabilities if self._client_params else None
         validate_sampling_tools(client_caps, tools, tool_choice)
         validate_tool_use_result_messages(messages)
@@ -349,6 +372,7 @@ class ServerSession(
 
     async def list_roots(self) -> types.ListRootsResult:
         """Send a roots/list request."""
+        self._require_stateful_mode("list_roots")
         return await self.send_request(
             types.ServerRequest(types.ListRootsRequest()),
             types.ListRootsResult,
@@ -391,7 +415,11 @@ class ServerSession(
 
         Returns:
             The client's response with form data
+
+        Raises:
+            RuntimeError: If called in stateless HTTP mode.
         """
+        self._require_stateful_mode("elicitation")
         return await self.send_request(
             types.ServerRequest(
                 types.ElicitRequest(
@@ -425,7 +453,11 @@ class ServerSession(
 
         Returns:
             The client's response indicating acceptance, decline, or cancellation
+
+        Raises:
+            RuntimeError: If called in stateless HTTP mode.
         """
+        self._require_stateful_mode("elicitation")
         return await self.send_request(
             types.ServerRequest(
                 types.ElicitRequest(
--- a/tests/server/test_stateless_mode.py
+++ b/tests/server/test_stateless_mode.py
@@ -1,0 +1,263 @@
+"""Tests for stateless HTTP mode limitations.
+
+Stateless HTTP mode does not support server-to-client requests because there
+is no persistent connection for bidirectional communication. These tests verify
+that appropriate errors are raised when attempting to use unsupported features.
+
+See: https://github.com/modelcontextprotocol/python-sdk/issues/1097
+"""
+
+import anyio
+import pytest
+
+import mcp.types as types
+from mcp.server.models import InitializationOptions
+from mcp.server.session import ServerSession
+from mcp.shared.message import SessionMessage
+from mcp.types import ServerCapabilities
+
+
+def create_test_streams():
+    """Create memory streams for testing."""
+    server_to_client_send, server_to_client_receive = anyio.create_memory_object_stream[SessionMessage](1)
+    client_to_server_send, client_to_server_receive = anyio.create_memory_object_stream[SessionMessage | Exception](1)
+    return (
+        server_to_client_send,
+        server_to_client_receive,
+        client_to_server_send,
+        client_to_server_receive,
+    )
+
+
+def create_init_options():
+    """Create default initialization options for testing."""
+    return InitializationOptions(
+        server_name="test",
+        server_version="0.1.0",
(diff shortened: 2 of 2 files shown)


tests/server/test_stateless_mode.py
```diff
@@ -0,0 +1,263 @@
...
+        server_to_client_send,
+        server_to_client_receive,
+    ):
+        async with ServerSession(
+            client_to_server_receive,
+            server_to_client_send,
+            create_init_options(),
+            stateless=True,
+        ) as session:
+            with pytest.raises(RuntimeError) as exc_info:
```
Kludex, line 63: Add the `match=` please, `RuntimeError` can be a lot.

tests/server/test_stateless_mode.py
```diff
@@ -0,0 +1,263 @@
...
+        client_to_server_receive,
+        server_to_client_send,
+        server_to_client_receive,
+    ):
+        async with ServerSession(
+            client_to_server_receive,
+            server_to_client_send,
+            create_init_options(),
+            stateless=True,
+        ) as session:
```
Kludex, line 128: Can't this whole thing be a fixture?

pcarleton:
doh... didn't realize it was on automerge

src/mcp/server/session.py
```diff
@@ -156,6 +157,26 @@ def check_client_capability(self, capability: types.ClientCapabilities) -> bool:
...
+
+        Raises:
+            RuntimeError: If the session is in stateless mode
+        """
+        if self._stateless:
+            raise RuntimeError(
+                f"Cannot use {feature_name} in stateless HTTP mode. "
+                "Stateless mode does not support server-to-client requests. "
+                "Use stateful mode (stateless_http=False) to enable this feature."
+            )
```
Kludex, line 178: This kind of private function adds a bit of overhead while reading the code. If possible, can we create a specific exception and do the check on each feature method itself?

```python
if self._stateless:
    raise MethodNotSupported(method="sampling")
``` 

On `MethodNotSupported` you can set up the message.

Kludex:
...

tests/server/test_stateless_mode.py
maxisbey: Yep fair will do

src/mcp/server/session.py
maxisbey: MethodNotSupported what are you thinking for the error message? As currently it's specific to stateless http and MethodNotSupported sounds more generic

Kludex: You can think of a better name, but the point stands.

---

Although I suggested this, I still think that something is missing conceptually - we should be able to check this or define somehow the capabilities of stateless in a different way - and be able to handle this before it reaches the methods.

maxisbey: Yea, although this is more of a bandaid fix given that this version of stateless mode isn't _really_ supported in the spec. The upcoming changes to the spec will significantly change how statefulness will work anyway so I didn't want to spend too much time worrying about how it looks since it has to change anyway.

PR #1834, opened 2026-01-06
Support Python 3.14

Kludex:


pyproject.toml
```diff
@@ -20,12 +20,14 @@ classifiers = [
...
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
     "anyio>=4.5",
     "httpx>=0.27.1",
     "httpx-sse>=0.4",
-    "pydantic>=2.11.0,<3.0.0",
+    "pydantic>=2.12.0; python_version >= '3.14'",
+    "pydantic>=2.11.0; python_version < '3.14'",
```
Kludex, line 30: I'm not bumping to not annoy users, but we should.

maxisbey: yea this would break some internal stuff if we forced a bump

maxisbey (approved):

PR #1838, opened 2026-01-07
fix: send error on SSE disconnect without resumption token

### LINKED ISSUE #1811: client's `read_stream_writer` open after SSE disconnection hanging `.receive()`

jayhemnani9910:
## Summary
  Fixes #1811

  When an SSE stream disconnects before the server sends any events with IDs (no resumption token), the client now sends a JSONRPCError to the session layer instead of hanging indefinitely.

  ## Changes
  - Added error handling in `_handle_sse_response` for streams that close without sending resumption tokens
  - Error response uses `CONNECTION_CLOSED` (-32000) error code with descriptive message
  - Added unit test that exercises the new code path by mocking EventSource

  ## Test Plan
  - Added `test_sse_disconnect_without_resumption_token_sends_error` unit test
  - All 1091 existing tests pass


maxisbey:
As there's no linked issue or description I've set this PR to draft for now.

jayhemnani9910:
> As there's no linked issue or description I've set this PR to draft for now.

 Yea my bad, my setup crashed. lol

dsfaccini:
we're bumping dependencies on our side and needed to add some pragmas because of the hang https://github.com/pydantic/pydantic-ai/pull/4047

so just a +1 on "hopefully this merges soon"

maxisbey:
Thanks for the PR. This has since landed via #3047. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #1841, opened 2026-01-08
chore: create private modules, and drop unnecessary modules

Kludex:
This PR is on top of `drop-streamablehttp-client`.


Kludex:
It seems the examples depend on those functions... So it seems this is not correct? I'll close this PR, but I'll open a new one organizing stuff a bit better.

Issue #1315, opened 2025-08-27
OAuth TokenHandler should check Authorization header for client credentials


Mars9934:
### Description

Currently, `TokenHandler` [assumes](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/auth/handlers/token.py#L96) that the Token request's body contains client credentials. However, some OAuth requests would contain client credentials in `Authorization` header:
<img width="1616" height="242" alt="Image" src="https://github.com/user-attachments/assets/4fb1fef9-f89b-4908-9d3c-e52b5e8b370b" />

In this case, it would throw `ValidationError` even though client credentials are provided in request header.

Can we add a fallback such that if `client_id` is not found in `formData`, we try to get it from header? e.g.

```
async def handle(self, request: Request):
    try:
        form_data = dict(await request.form())

        # Try to get client credentials from header if missing in body
        if "client_id" not in form_data:
            auth_header = request.headers.get("Authorization")
            if auth_header and auth_header.startswith("Basic "):
                encoded = auth_header.split(" ")[1]
                decoded = base64.b64decode(encoded).decode("utf-8")
                client_id, _, client_secret = decoded.partition(":")
                client_secret = urllib.parse.unquote(client_secret)
                form_data.setdefault("client_id", client_id)
                form_data.setdefault("client_secret", client_secret)

        token_request = TokenRequest.model_validate(form_data).root
    except ValidationError as validation_error:
        return self.response(
            TokenErrorResponse(
                error="invalid_request",
                error_description=stringify_pydantic_error(validation_error),
            )
        )
    ...
```

Thanks.

challenger71498:
Hi @Mars9934,

I encountered the same issue you've experienced when I was testing FastMCP server with:

```
npx @modelcontextprotocol/inspector
```

The server responds with 401, says "Missing client_id".

Seems like the indicator tries to get token by providing `client_id` and `client_secret` via the auth header, but the server only tries to find client credentials from the request body, and fails.

As you mentioned, the current implementation of token handling violates the OAuth2 protocol [RFC 6749 - Client Authentication](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3).

Based on RFC 6749:

> The client MUST NOT use more than one authentication method in each request.
- Authentication with headers (`Authorization: Basic ...`) and with form data are mutual exclusive. Only one auth method should be used at a time.
  - Normally the header is preferred, the form data is used only if the client cannot support the auth header.
[...]

PR #1316, opened 2025-08-27
OAuth TokenHandler Enhancement: Authorization Header Fallback Support

ChenyangLi4288:
# OAuth TokenHandler Enhancement: Authorization Header Fallback Support

This PR implements the enhancement requested in issue #1315 to add fallback support for OAuth client credentials in the Authorization header when they are missing from the request form data. This improves OAuth 2.0 compliance by supporting both `client_secret_post` and `client_secret_basic` authentication methods.

## Changes Made

- Implement fallback logic in TokenHandler to check Authorization header when client credentials are missing from form data
- Support Basic authentication with proper Base64 decoding and URL decoding
- Add comprehensive test suite covering all scenarios
- Maintain backward compatibility with existing form data authentication
- Improve OAuth 2.0 compliance by supporting both client_secret_post and client_secret_basic methods

Fixes #1315

This PR adds fallback support for OAuth client credentials in the Authorization header when they are missing from the request form data. The TokenHandler now checks for Basic authentication in the Authorization header as a fallback, improving OAuth 2.0 compliance by supporting both client_secret_post and client_secret_basic methods.

## Motivation and Context

Currently, the TokenHandler only looks for client credentials in the request form data. However, according to OAuth 2.0 specifications, client credentials can also be provided in the Authorization header using Basic authentication. When credentials are only provided in the header, the handler throws a ValidationError even though valid credentials are present. This enhancement addresses this limitation by implementing a fallback mechanism.

## How Has This Been Tested?

Comprehensive tests have been added covering:
- Form data credentials (existing functionality)
- Authorization header fallback (new functionality)
- URL-encoded secrets handling
- Invalid/malformed headers
- Both authorization code and refresh token grant types
- Error cases when no credentials are provided

All new tests pass (7/7) and existing tests continue to pass. Code passes linting (ruff) and type checking (pyright).

## Breaking Changes

None. This is a purely additive enhancement that maintains 100% backward compatibility. Existing OAuth clients using form data authentication will continue to work exactly as before.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

[...]

PR #1847, opened 2026-01-11
feat: fully support Basic Authorization header at token request

challenger71498:
## Summary

This PR implements OAuth client authentication by supporting HTTP Basic authentication as defined in RFC 6749 Section 2.3.1., especially when `client_id` is provided only from the header.

The key changes are:

- Extract client credentials from Basic auth header first, before falling back to form data
- Validate that the authentication method used matches what the client registered
- Make `client_id` optional in token request body when using Basic auth (as the spec allows)

Previously, the code would first read `client_id` from the form body and then validate against the auth method. This caused issues when clients used Basic auth without including `client_id` in the request body, which is valid per the OAuth spec.

- This PR resolves the issue #1315.

## Motivation and Context

- [RFC 6749 Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) specifies that clients using `client_secret_basic` should send credentials via the Authorization header, and the `client_id` in the request body is optional in this case.
- The previous implementation required `client_id` in the body even when using Basic auth, which was non-compliant with the spec.

### Unit Test
- Added test for Basic auth working without `client_id` in request body
- Added test for Basic auth taking precedence over form body `client_id`
- Added test for refresh token flow without `client_id` in body
- Updated existing tests to reflect new error messages when auth method mismatches

### E2E Test
- Tested with `npx @modelcontextprotocol/inspector` + OAuth provider mockup, and worked as expected.

<details>
<summary>Screenshots</summary>

image

image

</details>

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] ~~I have added or updated documentation as needed~~ no docs need to be updated

## Additional context
- #1316 tried to resolve the issue, but closed by #1334.
  - The approach of 1334 was right on track, but did not resolved the issue 1315 completely.


src/mcp/server/auth/middleware/client_auth.py
```diff
@@ -52,64 +60,86 @@ async def authenticate_request(self, request: Request) -> OAuthClientInformation
...
-            if not hmac.compare_digest(client.client_secret.encode(), request_client_secret.encode()):
+            if not hmac.compare_digest(client.client_secret.encode(), client_credentials.client_secret.encode()):
                 raise AuthenticationError("Invalid client_secret")  # pragma: no cover
 
             if client.client_secret_expires_at and client.client_secret_expires_at < int(time.time()):
                 raise AuthenticationError("Client secret has expired")  # pragma: no cover
 
         return client
+
+    async def _get_credentials(self, request: Request) -> ClientCredentials:
```
challenger71498, line 101: Note: the core extraction logic of `client_id` and `client_secret` is as same as before.

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1089,7 +1089,7 @@ async def test_client_secret_basic_authentication(
         assert "access_token" in token_response
 
     @pytest.mark.anyio
-    async def test_wrong_auth_method_without_valid_credentials_fails(
+    async def test_wrong_auth_method_fails(
```
challenger71498, line 1092: Note: test fails as same as before, only exception message has been changed, due to the auth method mismatch.

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1183,7 +1183,7 @@ async def test_basic_auth_without_header_fails(
         error_response = response.json()
         # RFC 6749: authentication failures return "invalid_client"
         assert error_response["error"] == "invalid_client"
-        assert "Missing or invalid Basic authentication" in error_response["error_description"]
+        assert "Expected client_secret_basic authentication method" in error_response["error_description"]
```
challenger71498, line 1186: Note: test fails as same as before, only exception message has been changed, due to the auth method mismatch.

tests/server/fastmcp/auth/test_auth_integration.py
```diff
@@ -1279,10 +1279,10 @@ async def test_basic_auth_no_colon_fails(
         assert "Invalid Basic authentication header" in error_response["error_description"]
 
     @pytest.mark.anyio
-    async def test_basic_auth_client_id_mismatch_fails(
+    async def test_basic_auth_takes_precedence(
```
challenger71498, line 1282: Note: the behaviour of the test has been changed. After the changes, basic auth takes precedence over form data, so even if there is a mismatch, the middleware retrieves the `client_id` from the header.

This behaviour is intended due to [RFC 6749 Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1):

> Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes).

If this is inappropriate, alternatives are:
- behave AS-IS (throw mismatch exception)
- disallow dup auth method
  - throw exception which indicates that more than one authentication method was used.

joar:
@ochafik @pcarleton I saw you were involved in https://github.com/modelcontextprotocol/python-sdk/pull/1334 which added the `client_secret_basic` auth method, could you have a look at this PR as well?

I added my point of view on the issue this PR resolves in https://github.com/modelcontextprotocol/python-sdk/issues/1315#issuecomment-4134287279 —* Including the realization that MCP is about OAuth 2.1 draft 13 and not RFC 6749, but that both standards seem to agree that `client_id` is not required for `client_secret_basic`

\* human-crafted em dash

MesoX:
+1, hit this in production via FastMCP (`OAuthProxy`).

The bug compounds for FastMCP-based MCP servers because
`OAuthProxy.register_client` hard-codes `token_endpoint_auth_method="none"`
for DCR-registered clients (`fastmcp/server/auth/oauth_proxy.py:1005` on
v2.14.7). So even though FastMCP advertises `client_secret_basic` in
`token_endpoint_auth_methods_supported`, the `ClientAuthenticator` middleware
never invokes the Basic-decoding path - and the
`form_data.get("client_id")` check at `src/mcp/server/auth/middleware/client_auth.py:55`
rejects the request with `Missing client_id` before any Basic header is
inspected.

Concrete repro against a vanilla FastMCP `OAuthProxy` deploy:

```bash
# Pure Basic auth - RFC 6749 §2.3.1 compliant
curl -sS -u "$CID:$CSEC" \
  -d 'grant_type=authorization_code&code=test&redirect_uri=...' \
  http://server/token
# → 401 {"error":"invalid_client","error_description":"Missing client_id"}

# Same creds in POST body
curl -sS \
  -d "grant_type=authorization_code&code=test&client_id=$CID&client_secret=$CSEC&redirect_uri=..." \
  http://server/token
# → 400 (reaches code_verifier validation; client_id was parsed)
```

Same handler, same input, different carrier. The diff in this PR fixes the
right layer (`client_auth.py`), so the FastMCP case is fixed transitively.

Spec-compliant `@ai-sdk/mcp` clients default to `client_secret_basic` when
the server advertises both - they're the most affected. Happy to add a
regression test covering the Basic-only path if it helps land this.

PR #1849, opened 2026-01-12
Fix: Raise RuntimeError when ClientSession is used without context manager to prevent hangs

### LINKED ISSUE #1564: stdio_client fails with BrokenResourceError during initialization
### Initial Checks

### Description

I'm trying to use stdio_client to spawn and communicate with a FastMCP server, but it consistently fails during the initialization handshake. The client times out after about 10 seconds with a BrokenResourceError.

The subprocess starts fine and I can see the server is running, but when the client tries to initialize the session, it hangs and eventually times out. Looking at the traceback, there's a BrokenResourceError happening in the stdout_reader coroutine at line 162 of mcp/client/stdio/__init__.py.

Expected behavior:
The client should successfully spawn the server, complete the initialize handshake, and be ready to call tools.

Actual behavior:
The client hangs during initialization and then times out with:

ExceptionGroup: unhandled errors in a TaskGroup (2 sub-exceptions)
  BrokenResourceError at mcp/client/stdio/__init__.py:162 in stdout_reader
  TimeoutError after waiting for initialize response

Additional cpontext:
Interesting finding: When I test the server directly by piping JSON to it, it works perfectly:

$ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}' | python test_server.py

The server immediately returns a valid response. So the server itself is working correctly - it can read from stdin and write to stdout just fine. The issue seems to be specifically with how stdio_client is handling the subprocess communication.

I've tried various things to fix it but nothing works:
- Running python with -u flag for unbuffered output
- Setting sys.stdout.reconfigure(line_buffering=True) in the server
- Using absolute paths
- Increasing timeout values
- Setting PYTHONUNBUFFERED=1

Note: HTTP/SSE transport works fine as an alternative.

### Example Code

1. Create a simple FastMCP server (test_server.py):

```python

mcp = FastMCP("TestServer")

@mcp.tool()
def add(a: int, b: int) -> int:
    """Add two numbers"""
    return a + b

if __name__ == "__main__":
    mcp.run()
```

2. Create a client that uses stdio_client (test_client.py):

```python
import asyncio

async def test():
        command="python",
        args=["test_server.py"]
    )
    
    async with stdio_client(params) as (stdio, write):
        session = ClientSession(stdio, write)
        await session.initialize()  # Hangs here
        
        result = await session.list_tools()
        print(f"Tools: {result.tools}")

asyncio.run(test())
```

3. Run: python test_client.py
4. Observe: hangs for about 10 seconds then fails with BrokenResourceError

### Python & MCP Python SDK

```Text
- MCP SDK Version: 1.20.0
- Python Version: 3.12.7
- Operating System: macOS
- Installation method: pip install mcp==1.20.01
```

newbiehwang:
Fixes #1564

## Summary
This PR adds a runtime check to `ClientSession` to ensure it is being used within an asynchronous context manager (`async with`). If `initialize()` is called without entering the context, it now raises a clear `RuntimeError` instead of hanging indefinitely.

## Motivation and Context
Currently, if a user attempts to call `await session.initialize()` without wrapping the session in an `async with` block, the program hangs indefinitely. This happens because the background receive loop (`_receive_loop`) is only started inside `__aenter__`.

This behavior leads to a frustrating debugging experience (deadlock) for developers. This change provides immediate feedback by enforcing correct usage of the session, preventing the deadlock.

## How Has This Been Tested?
I reproduced the issue using the script provided in #1564.
1. **Before fix:** The script hung indefinitely at `await session.initialize()`.
2. **After fix:** The script immediately raises `RuntimeError: ClientSession must be used within an 'async with' block.`, as expected.
3. Verified that normal usage (with `async with`) continues to work correctly.

## Breaking Changes
No. This change is backward compatible for all correct usages. It only affects code that was previously incorrect and hanging.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
I implemented this using a simple state flag (`_entered`) to maintain backward compatibility without requiring significant architectural refactoring of the session class.


newbiehwang:
Friendly ping — this PR has been open for about a month now. All CI checks are passing. Could a maintainer take a look when available? Thanks!

maxisbey:
Thanks for the PR. This has since landed via #2838. I'm closing this out as part of a backlog cleanup now that v2 is out. Feel free to reopen if this is still relevant.

PR #1857, opened 2026-01-15
fix(tests): skip proxy for localhost in Windows environments

### LINKED ISSUE #1477: Test fails in a Windows system with a proxy enabled
### Initial Checks

### Description

In a Windows system with a proxy enabled, some tests fail when accessing 127.0.0.1 through the proxy.

The test fails:
Image

### Suggestion
When testing, set environment variables to avoid using localhost through a proxy. 

Add to tests/conftest.py:
```Python
import os 
os.environ[NO_PROXY] = '127.0.0.1`
```

### Example Code

```Python
uv run pytest
```

### Python & MCP Python SDK

```Text
Python version: 3.11
MCP Python SDK：latest
```

cxhello:
## Summary
- Set `NO_PROXY` environment variable to `127.0.0.1` in `tests/conftest.py`
- Prevents tests from failing on Windows systems with a proxy enabled

## Test plan
- [x] Verified the fix matches the suggestion in the issue
- [ ] Run `uv run pytest` on Windows with proxy enabled

Closes #1477

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,8 @@
+import os
+
 import pytest
+
+os.environ["NO_PROXY"] = "127.0.0.1"
 
 
 @pytest.fixture


Kludex:
I don't think we need to pollute the test suite for this. You can enable the env var on your local environment.

PR #1867, opened 2026-01-15
fix: add default values to Literal type fields in content types

### LINKED ISSUE #1731: Content types should have default values for Literal type fields
## Problem

Content types like `TextContent` and `ImageContent` require passing `type="text"` or `type="image"` explicitly, even though these are single-value `Literal` types.

**Current definitions:**
```python
class TextContent(BaseModel):
    type: Literal["text"]  # No default
    text: str

class ImageContent(BaseModel):
    type: Literal["image"]  # No default
    data: str
    mimeType: str
```

**Current usage (verbose):**
```python
TextContent(type="text", text="hello")
ImageContent(type="image", data="...", mimeType="image/png")
```

PR #1292 already added defaults for `method: Literal` fields on request and notification types - could do the same thing here.

## Related

- #837 (type system complexity / developer friction)

felixweinberger:
## Summary

Add default values to single-value `Literal` type fields in content types so users don't have to pass them explicitly.

## Motivation and Context

Content types like `TextContent` and `ImageContent` have single-value Literal type fields (e.g., `type: Literal["text"]`) that should default automatically since there's only one valid value.

**Before:**
```python
TextContent(type="text", text="hello")
ImageContent(type="image", data="...", mimeType="image/png")
```

**After:**
```python
TextContent(text="hello")
ImageContent(data="...", mimeType="image/png")
```

Fixes #1731

## How Has This Been Tested?

Added regression test `test_content_type_literal_defaults` in `tests/test_types.py` covering representative types (TextContent, ImageContent, PromptReference).

## Breaking Changes

None. Existing code that explicitly passes the `type` field will continue to work.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Affected types:
- `TextContent`
- `ImageContent`
- `AudioContent`
- `ToolUseContent`
- `ToolResultContent`
- `EmbeddedResource`
- `ResourceLink`
- `ResourceTemplateReference`
- `PromptReference`

This follows the same pattern as PR #1292 which added defaults for `method: Literal` fields on request/notification types.

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -9,14 +9,12 @@ from mcp.types import (
     CreateMessageRequestParams,
     CreateMessageResult,
     CreateMessageResultWithTools,
-    ImageContent,
     Implementation,
     InitializeRequest,
     InitializeRequestParams,
     JSONRPCMessage,
     JSONRPCRequest,
     ListToolsResult,
-    PromptReference,
     SamplingCapability,
     SamplingMessage,
     TextContent,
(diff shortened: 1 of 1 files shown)


tests/test_types.py
```diff
@@ -71,6 +73,17 @@ async def test_method_initialization():
...
 
+def test_content_type_literal_defaults():
+    """Content types should default their Literal type field.
+
+    This allows instantiation without explicitly passing the type discriminator,
+    e.g., TextContent(text="hello") instead of TextContent(type="text", text="hello").
+    """
+    assert TextContent(text="hello").type == "text"
+    assert ImageContent(data="base64", mimeType="image/png").type == "image"
+    assert PromptReference(name="test").type == "ref/prompt"
```
Kludex: I don't think any test is needed here. It feels it's testing Pydantic.

felixweinberger: Yeah fair - I kept this around as a regression test to ensure nobody removes the Literal, but I guess that's kind of pointless as who would do that 🤔 I'll remove.

PR #1870, opened 2026-01-15
Add ergonomic Client class for testing MCP servers

### LINKED ISSUE #1728: Add ergonomic test client for testing MCP servers
## Summary

Provide a dedicated test client that simplifies testing MCP servers without the boilerplate of manually setting up transports, streams, and context managers.

## Motivation

Currently, testing an MCP server requires:
- Setting up `stdio_client` or similar transport
- Managing multiple nested context managers (transport + session)
- Wiring up memory streams manually

This creates friction for both SDK users writing tests for their servers and for the SDK's own test suite.

## Goal

Something analogous to FastAPI's `TestClient` / Starlette's test client pattern - a simple way to instantiate a client against a server instance and make requests without transport ceremony.

---

felixweinberger:
## Summary

Adds a high-level `Client` class that provides an ergonomic API for testing MCP servers with in-memory transport.

## Motivation and Context

Testing MCP servers currently requires using the verbose `create_connected_server_and_client_session` function. This PR introduces a simpler API:

```python
from mcp import Client

mcp = FastMCP("test")

@mcp.tool()
def add(a: int, b: int) -> int:
    return a + b

async with Client(mcp) as client:
    result = await client.call_tool("add", {"a": 1, "b": 2})
```

Related issues:
- Closes #1728

## How Has This Been Tested?

- 24 new tests for `Client` and `InMemoryTransport`
- Migrated existing tests from `create_connected_server_and_client_session` to new API
- Full test suite passes

## Breaking Changes

- Removed `create_connected_server_and_client_session` from `mcp.shared.memory` (v2 breaking change)

## Types of changes

- [x] Documentation update

## Checklist

### New API

**High-level (recommended):**
```python
from mcp import Client

    result = await client.call_tool("my_tool", {"arg": "value"})
```

**Low-level (for advanced use cases):**
```python
from mcp.client.transports import InMemoryTransport

transport = InMemoryTransport(server)
async with transport.connect() as (read, write):
        await session.initialize()
        # Full ClientSession access...
```

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,10 +1,11 @@
 # Testing MCP Servers
 
-If you call yourself a developer, you will want to test your MCP server.
-The Python SDK offers the `create_connected_server_and_client_session` function to create a session
-using an in-memory transport. I know, I know, the name is too long... We are working on improving it.
+The Python SDK provides a `Client` class for testing MCP servers with an in-memory transport.
+This makes it easy to write tests without network overhead.
 
-Anyway, let's assume you have a simple server with a single tool:
+## Basic Usage
+
+Let's assume you have a simple server with a single tool:
 
 ```python title="server.py"
 from mcp.server import FastMCP
@@ -40,12 +41,9 @@ To run the below test, you'll need to install the following dependencies:
     server - you don't need to use it, but we are spreading the word for best practices.
 
 ```python title="test_server.py"
-from collections.abc import AsyncGenerator
-
 import pytest
 from inline_snapshot import snapshot
-from mcp.client.session import ClientSession
-from mcp.shared.memory import create_connected_server_and_client_session
+from mcp import Client
 from mcp.types import CallToolResult, TextContent
 
 from server import app
@@ -57,14 +55,14 @@ def anyio_backend():  # (1)!
 
 
 @pytest.fixture
-async def client_session() -> AsyncGenerator[ClientSession]:
-    async with create_connected_server_and_client_session(app, raise_exceptions=True) as _session:
-        yield _session
+async def client():  # (2)!
+    async with Client(app, raise_exceptions=True) as c:
+        yield c
 
 
 @pytest.mark.anyio
-async def test_call_add_tool(client_session: ClientSession):
-    result = await client_session.call_tool("add", {"a": 1, "b": 2})
+async def test_call_add_tool(client: Client):
+    result = await client.call_tool("add", {"a": 1, "b": 2})
     assert result == snapshot(
         CallToolResult(
             content=[TextContent(type="text", text="3")],
@@ -74,5 +72,6 @@ async def test_call_add_tool(client_session: ClientSession):
 ```
 
 1. If you are using `trio`, you should set `"trio"` as the `anyio_backend`. Check more information in the [anyio documentation](https://anyio.readthedocs.io/en/stable/testing.html#specifying-the-backends-to-run-on).
+2. The `client` fixture creates a connected client that can be reused across multiple tests.
 
 There you go! You can now extend your tests to cover more scenarios.
--- a/src/mcp/client/_memory.py
+++ b/src/mcp/client/_memory.py
@@ -1,0 +1,97 @@
+"""In-memory transport for testing MCP servers without network overhead."""
+
+from __future__ import annotations
+
+from collections.abc import AsyncGenerator
+from contextlib import asynccontextmanager
+from typing import Any
+
+import anyio
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+
+from mcp.server import Server
+from mcp.server.fastmcp import FastMCP
+from mcp.shared.memory import create_client_server_memory_streams
+from mcp.shared.message import SessionMessage
+
+
+class InMemoryTransport:
+    """
+    In-memory transport for testing MCP servers without network overhead.
+
+    This transport starts the server in a background task and provides
+    streams for client-side communication. The server is automatically
+    stopped when the context manager exits.
+
+    Example:
+        server = FastMCP("test")
+        transport = InMemoryTransport(server)
+
+        async with transport.connect() as (read_stream, write_stream):
+            async with ClientSession(read_stream, write_stream) as session:
+                await session.initialize()
+                # Use the session...
+
+    Or more commonly, use with Client:
+        async with Client(server) as client:
+            result = await client.call_tool("my_tool", {...})
+    """
+
+    def __init__(
+        self,
+        server: Server[Any] | FastMCP,
+        *,
+        raise_exceptions: bool = False,
+    ) -> None:
+        """
+        Initialize the in-memory transport.
+
+        Args:
+            server: The MCP server to connect to (Server or FastMCP instance)
+            raise_exceptions: Whether to raise exceptions from the server
+        """
+        self._server = server
+        self._raise_exceptions = raise_exceptions
+
+    @asynccontextmanager
+    async def connect(
+        self,
+    ) -> AsyncGenerator[
+        tuple[
+            MemoryObjectReceiveStream[SessionMessage | Exception],
+            MemoryObjectSendStream[SessionMessage],
+        ],
+        None,
+    ]:
+        """
+        Connect to the server and return streams for communication.
+
+        Yields:
+            A tuple of (read_stream, write_stream) for bidirectional communication
+        """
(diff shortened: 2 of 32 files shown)


felixweinberger:
**Key files to review:**
- `src/mcp/client/client.py` (new) - unified Client class
- `src/mcp/client/transports/memory.py` (new) - InMemoryTransport
- `src/mcp/client/__init__.py` - public exports
- `src/mcp/client/transports/__init__.py` - transport exports
- `src/mcp/shared/memory.py` (deleted) - replaced by InMemoryTransport

Remainder is tests being updated to use the new interface.

**NOTE:** This only implements test support. Refactoring Client to support different transports for the end-user API (streamable HTTP, stdio, SSE, etc.) is coming, marked as TODO in the Client docstring.

src/mcp/client/client.py
```diff
@@ -0,0 +1,303 @@
...
+        self._exit_stack: AsyncExitStack | None = None
+
+    async def __aenter__(self) -> Client:
+        """Enter the async context manager."""
+        self._exit_stack = AsyncExitStack()
+        await self._exit_stack.__aenter__()
+
+        try:
+            # Create transport and connect
+            transport = InMemoryTransport(self._server, raise_exceptions=self._raise_exceptions)
```
felixweinberger, line 104: right now this is hard-coded but we'd add a `infer_transport_from_args` or similar method during init that would set up SHTTP / STDIO / SSE as required.

src/mcp/client/client.py
```diff
@@ -0,0 +1,303 @@
...
+                    client_info=self._client_info,
+                    elicitation_callback=self._elicitation_callback,
+                )
+            )
+
+            # Initialize the session
+            await self._session.initialize()
+
+            return self
+        except Exception:
```
felixweinberger: More specific exceptions

src/mcp/client/client.py
```diff
@@ -0,0 +1,303 @@
...
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: Any,
+    ) -> None:
+        """Exit the async context manager."""
+        if self._exit_stack:
+            await self._exit_stack.__aexit__(exc_type, exc_val, exc_tb)
```
felixweinberger, line 137: what happens if we reenter, check this behavior

docs/testing.md
```diff
@@ -56,23 +54,41 @@ def anyio_backend():  # (1)!
...
+For tests that need direct access to the underlying `ClientSession`, use `InMemoryTransport`:
+
+```python
+from mcp.client.session import ClientSession
+from mcp.client.transports import InMemoryTransport
+
+from server import app
+
+
+async def test_with_low_level_access():
```
felixweinberger: remove if there's no concrete example of a test that needs access to the low level transport

src/mcp/client/_memory.py
```diff
@@ -0,0 +1,97 @@
...
+import anyio
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+
+from mcp.server import Server
+from mcp.server.fastmcp import FastMCP
+from mcp.shared.memory import create_client_server_memory_streams
+from mcp.shared.message import SessionMessage
+
+
+class InMemoryTransport:
```
felixweinberger, line 18: let's make this private as it shouldn't be part of public API, we only want to use this in tests not for public API to use this

docs/testing.md
felixweinberger: Do we actually need this at all? shouldn't document it unless it's useful

felixweinberger:
## Follow-up tasks

- [ ] Add transport types to Client (StreamableHttpTransport, SSETransport, StdioTransport) + `infer_transport()` for auto-detection
- [ ] Add lint rule to enforce Google-style docstrings (first line on same line as `"""`) e.g. `"""Send a ping request."""`
- [ ] Audit existing unit tests and as much as possible refactor to use this new `Client` class instead

docs/testing.md
```diff
@@ -56,21 +54,16 @@ def anyio_backend():  # (1)!
...
-            structuredContent={"result": 3},
+async def test_call_add_tool():
+    async with Client(app, raise_exceptions=True) as client:
+        result = await client.call_tool("add", {"a": 1, "b": 2})
+        assert result == snapshot(
+            CallToolResult(
+                content=[TextContent(type="text", text="3")],
+                structuredContent={"result": 3},
+            )
         )
```
Kludex: fixture

src/mcp/client/client.py
```diff
@@ -0,0 +1,302 @@
...
+
+        async with Client(server) as client:
+            result = await client.call_tool("add", {"a": 1, "b": 2})
+
+    TODO(felixweinberger): Expand to support all transport types (like FastMCP 2):
+    - Add ClientTransport base class with connect_session() method
+    - Add StreamableHttpTransport, SSETransport, StdioTransport
+    - Add infer_transport() to auto-detect transport from input type
+    - Accept URL strings, Path objects, config dicts in constructor
+    - Add auth support (OAuth, bearer tokens)
```
Kludex: This is supposed to be a comment, not docstring fyi

(but all good)

src/mcp/client/client.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Example:
+        server = FastMCP("test")
+
+        @server.tool()
+        def add(a: int, b: int) -> int:
+            return a + b
+
+        async with Client(server) as client:
+            result = await client.call_tool("add", {"a": 1, "b": 2})
```
Kludex, line 46: We need to have those self contained. They are actually rendered in the API reference in our docs:


```suggestion
        ```python
        from mcp.client import Client
        from mcp.server.fastmcp import FastMCP (?)

        server = FastMCP("test")

        @server.tool()
        def add(a: int, b: int) -> int:
            return a + b

            result = await client.call_tool("add", {"a": 1, "b": 2})
```

src/mcp/client/client.py
```diff
@@ -0,0 +1,302 @@
...
+        *,
+        raise_exceptions: bool = False,
+        read_timeout_seconds: float | None = None,
+        sampling_callback: SamplingFnT | None = None,
+        list_roots_callback: ListRootsFnT | None = None,
+        logging_callback: LoggingFnT | None = None,
+        message_handler: MessageHandlerFnT | None = None,
+        client_info: types.Implementation | None = None,
+        elicitation_callback: ElicitationFnT | None = None,
+    ) -> None:
```
Kludex, line 69: If you want to be fancy, this can be a dataclass e.g. https://github.com/pydantic/pydantic-ai/blob/a401b6787159df6cdfa966f544a892e5f9040123/pydantic_ai_slim/pydantic_ai/messages.py#L112-L134

src/mcp/client/client.py
```diff
@@ -0,0 +1,302 @@
...
+
+    async def __aenter__(self) -> Client:
+        """Enter the async context manager."""
+        if self._session is not None:
+            raise RuntimeError("Client is already entered; cannot reenter")
+
+        async with AsyncExitStack() as exit_stack:
+            # Create transport and connect
+            transport = InMemoryTransport(self._server, raise_exceptions=self._raise_exceptions)
+            read_stream, write_stream = await exit_stack.enter_async_context(transport.connect())
```
Kludex, line 105: I think `raise_exceptions` is InMemoryTransport` specific? Not sure. The constructor will probably need an overload when you add the other ways of instantiating this - nothing to worry on this PR.

src/mcp/client/client.py
```diff
@@ -0,0 +1,302 @@
...
+        return self._session
+
+    def get_server_capabilities(self) -> types.ServerCapabilities | None:
+        """
+        Return the server capabilities received during initialization.
+
+        Returns:
+            The server capabilities, or None if not yet initialized
+        """
+        return self.session.get_server_capabilities()
```
Kludex: Is it possible to avoid the `get` prefix here? I'm not sure - The `get` gives me "other language vibes".

(I know the underlying uses server uses `get_` prefix, but maybe we can avoid it here?)

```suggestion
    def server_capabilities(self) -> types.ServerCapabilities | None:
        """
        Return the server capabilities received during initialization.

        Returns:
            The server capabilities, or None if not yet initialized
        """
        return self.session.get_server_capabilities()
```

Kludex (approved):

PR #1881, opened 2026-01-16
Revert `mount_path` parameter from `FastMCP`

Kludex:
## Summary

This PR reverts the `mount_path` parameter added in #540. The parameter was redundant because the SSE transport already handles sub-path mounting via ASGI's standard `root_path` mechanism.

## Why this revert?

When using Starlette's `Mount("/path", app=mcp.sse_app())`, Starlette automatically sets `root_path` in the ASGI scope. The `SseServerTransport` already uses this to construct the correct message endpoint path (see `sse.py` lines 148-161):

```python
# From sse.py - this already exists and handles the issue
root_path = scope.get("root_path", "")
full_message_path_for_client = root_path.rstrip("/") + self._endpoint
```

## Minimal Reproducible Example

The existing test `test_sse_client_basic_connection_mounted_app` in `tests/shared/test_sse.py` already demonstrates this works. Here's a simplified version:

```python
from mcp.client.sse import sse_client

# Create MCP server
mcp = FastMCP("GitHub API")

@mcp.tool()
def hello() -> str:
    return "Hello from GitHub!"

# Mount at /github sub-path - NO mount_path parameter needed!
app = Starlette(routes=[Mount("/github", app=mcp.sse_app())])

# Run with uvicorn, then connect:
async with sse_client("http://localhost:8000/github/sse") as streams:
    async with ClientSession(*streams) as session:
        result = await session.initialize()
        print(f"Connected to: {result.serverInfo.name}")
        # Works! The SSE transport automatically uses root_path
        # to return /github/messages/ as the message endpoint
```

The client connects to `/github/sse` and the server automatically returns `/github/messages/` as the message endpoint - no `mount_path` configuration needed.

## What's removed

- `mount_path` setting from `Settings` class
- `mount_path` parameter from `FastMCP.__init__`
- `mount_path` parameter from `run()` and `run_sse_async()`
- `_normalize_path()` helper method
- `mount_path` parameter from `sse_app()`
- Related tests for the removed functionality

## What's updated

The README documentation now shows the correct way to mount multiple MCP servers, which works out of the box via ASGI's `root_path`.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -116,6 +116,12 @@ result = await session.list_resources(params=PaginatedRequestParams(cursor="next
 result = await session.list_tools(params=PaginatedRequestParams(cursor="next_page_token"))
 ```
 
+### `mount_path` parameter removed from FastMCP
+
+The `mount_path` parameter has been removed from `FastMCP.__init__()`, `FastMCP.run()`, `FastMCP.run_sse_async()`, and `FastMCP.sse_app()`. It was also removed from the `Settings` class.
+
+This parameter was redundant because the SSE transport already handles sub-path mounting via ASGI's standard `root_path` mechanism. When using Starlette's `Mount("/path", app=mcp.sse_app())`, Starlette automatically sets `root_path` in the ASGI scope, and the `SseServerTransport` uses this to construct the correct message endpoint path.
+
 ### Resource URI type changed from `AnyUrl` to `str`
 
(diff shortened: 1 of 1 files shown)


felixweinberger (dismissed):
LGTM, but do we need a `migration.md` line about removing this?

I think it's a breaking change to remove the `mount_path` argument if someone was using it (even if it was redundant / did nothing).

PR #1883, opened 2026-01-16
fix: send priming event in GET handler to prevent hang

wukai9203:
## Problem

GET requests to `/mcp` endpoint hang indefinitely when using `streamable_http_app()`.

Related issue in FastMCP: jlowin/fastmcp#532

## Root Cause

The `_handle_get_request` method's `standalone_sse_writer` function doesn't send a priming event before entering the message loop, unlike the POST handler which calls `_maybe_send_priming_event()` at line 592.

This causes the SSE writer to block waiting for messages that may never arrive, making the GET request appear to hang.

## Solution

Send an event immediately after establishing the SSE connection in `standalone_sse_writer()`:

1. **With event_store and protocol >= 2025-11-25**: Send proper priming event via `_maybe_send_priming_event()` (for SSE resumability support)
2. **Without event_store or older protocol**: Send a simple `{"event": "open", "data": ""}` event to confirm connection is established

Additional fixes:
- Add `anyio.ClosedResourceError` handling for clean shutdown when `close_standalone_sse_stream()` is called
- Store SSE writer reference so `close_standalone_sse_stream()` can properly close it

## Testing

- Added test case `test_get_request_receives_priming_event_with_event_store` to verify the fix
- Manually tested with curl and verified GET requests return immediately:
  ```bash
  # Before fix: hangs indefinitely
  # After fix: returns "event: open\ndata: \n\n" immediately
  curl -N http://localhost:8003/mcp \
    -H "Accept: text/event-stream" \
    -H "mcp-session-id: <session_id>"
  ```

## Checklist

- [x] Code follows project style guidelines
- [x] Tests added for the fix
- [x] Manually tested the fix with FastMCP integration

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -706,6 +706,12 @@ class StreamableHTTPServerTransport:
         # Create SSE stream
         sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[dict[str, str]](0)
 
+        # Store writer reference so close_standalone_sse_stream() can close it
+        self._sse_stream_writers[GET_STREAM_KEY] = sse_stream_writer
+
+        # Get protocol version from header for priming event decision
+        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER, DEFAULT_NEGOTIATED_VERSION)
+
         async def standalone_sse_writer():
             try:
                 # Create a standalone message stream for server-initiated messages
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for looking into this! Per our [contributing guidelines](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md), all PRs require a corresponding issue. Could you open an issue with a minimal reproduction so we can validate the problem first?

PR #1897, opened 2026-01-16
ci: add strict-no-cover to detect unnecessary coverage pragmas

maxisbey:
Add [strict-no-cover](https://github.com/pydantic/strict-no-cover) from pydantic to the CI pipeline. This tool identifies `# pragma: no cover` comments on lines that are actually covered by tests, helping keep coverage pragmas accurate and preventing them from accumulating as dead annotations.

## Motivation

Over time, `# pragma: no cover` annotations accumulate on lines that are actually exercised by tests — the original reason for the pragma may no longer apply (e.g., code was refactored, tests were added). These stale pragmas hide real coverage data and make it harder to identify genuinely uncovered code.

### CI integration (`e66af17`)
- Add `strict-no-cover` as a dev dependency (installed from git)
- Add CI step to run strict-no-cover after pytest coverage (Linux only — the tool has a Windows path-handling bug)
- Pin `coverage[toml]` to `>=7.10.7,<=7.13` because 7.13.1 introduced a regression in how source paths are reported in `.coverage` databases, breaking strict-no-cover
- Remove `relative_files = true` from `[tool.coverage.run]` as it interferes with strict-no-cover path matching
- Add `pragma: lax no cover` and `pragma: no branch` to `exclude_lines` in coverage config

### Pragma cleanup (`f0c0b6e`)
- Remove `# pragma: no cover` from ~150 lines that are actually covered by tests
- Add `# pragma: lax no cover` for platform-specific code (Windows `skipif` markers, `sys.platform` branches, OS-specific exception handlers)
- Add `# pragma: no branch` for branches where the exit path is never taken (`async for` loops, `anyio.fail_after` blocks, `finally` clauses)

## Pragma guide

| Pragma | Meaning | Use when |
|--------|---------|----------|
| `# pragma: no cover` | Line must never be covered | Strict — CI fails if covered |
| `# pragma: lax no cover` | Line may or may not be covered | Platform/version-specific code, flaky coverage |
| `# pragma: no branch` | Branch exit not taken | `async for`, `with fail_after()`, loops that always break |

## How it works

After `coverage run -m pytest` and `coverage combine`, the CI runs `strict-no-cover` which:
1. Reads the `.coverage` database to find which lines were actually executed
2. Scans source files for `# pragma: no cover` annotations
3. Reports any lines that have the pragma but ARE covered (meaning the pragma is unnecessary)
4. Exits non-zero if any are found, failing the CI check

`# pragma: lax no cover` lines are excluded from this check, making them safe for code that is covered in some CI configurations but not others.

### DIFF
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -1304,9 +1304,9 @@ async def test_streamable_http_client_resumption(event_server: tuple[SimpleEvent
 
             # Verify we received exactly one notification (inside ClientSession
             # so coverage tracks these on Python 3.11, see PR #1897 for details)
-            assert len(captured_notifications) == 1
-            assert isinstance(captured_notifications[0], types.LoggingMessageNotification)
-            assert captured_notifications[0].params.data == "First notification before lock"
+            assert len(captured_notifications) == 1  # pragma: lax no cover
+            assert isinstance(captured_notifications[0], types.LoggingMessageNotification)  # pragma: lax no cover
+            assert captured_notifications[0].params.data == "First notification before lock"  # pragma: lax no cover
 
     # Clear notifications and set up headers for phase 2 (between connections,
     # not tracked by coverage on Python 3.11 due to cancel scope + sys.settrace bug)
@@ -1324,7 +1324,9 @@ async def test_streamable_http_client_resumption(event_server: tuple[SimpleEvent
             write_stream,
             _,
         ):
-            async with ClientSession(read_stream, write_stream, message_handler=message_handler) as session:
+            async with ClientSession(
+                read_stream, write_stream, message_handler=message_handler
+            ) as session:  # pragma: no branch
                 result = await session.send_request(
                     types.CallToolRequest(params=types.CallToolRequestParams(name="release_lock", arguments={})),
                     types.CallToolResult,


.github/workflows/shared.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           uv run --frozen --no-sync coverage combine
           uv run --frozen --no-sync coverage report
 
+      - name: Check for unnecessary no cover pragmas
+        if: runner.os != 'Windows'
```
maxisbey, line 66: @Kludex without this all the windows pipelines fail to even run, seems like a strict-no-cover bug? Example failure from previous CI run: https://github.com/modelcontextprotocol/python-sdk/actions/runs/21070443439/job/60598385217

Kludex:
Is this working already? All the pragmas are actually being used?

maxisbey:
> Is this working already? All the pragmas are actually being used?

not sure, will test

maxisbey:
@kludex Found two issues that prevented `strict-no-cover` from actually detecting anything:

**1. `coverage >= 7.13` breaks `strict-no-cover`**

Coverage 7.13.0 changed [`analysis_from_file_reporter`](https://github.com/nedbat/coveragepy/blob/main/coverage/results.py) to filter executed lines with `& statements`. Since `statements` excludes lines matching `exclude_lines` patterns, `executed_lines` in the JSON report can never contain excluded lines. `strict-no-cover` checks `excluded_lines ∩ executed_lines`, so the intersection is always empty — the tool silently passes regardless of unnecessary pragmas.

Fix: pin `coverage[toml]>=7.10.7,<7.13`.

**2. `relative_files = true` breaks `strict-no-cover`**

`strict-no-cover` creates a temp rcfile containing only `[tool.coverage.report]` and passes it via `--rcfile=<temp>` to `coverage json`. This overrides the entire project config, including `[tool.coverage.run]` settings like `relative_files = true`. Without that setting, coverage cannot match the relative paths stored in `.coverage` to the source files, resulting in empty `executed_lines` for all files.

Fix: remove `relative_files = true` from `[tool.coverage.run]`.

---

I also added a test `# pragma: no cover` on a covered line (`src/mcp/shared/exceptions.py:18`) to verify the CI step actually fails. Once confirmed, that pragma should be removed.

src/mcp/client/streamable_http.py
```diff
@@ -190,19 +190,19 @@ async def handle_get_stream(self, client: httpx.AsyncClient, read_stream_writer:
...
 
                     # Stream ended normally (server closed) - reset attempt counter
                     attempt = 0
 
-            except Exception as exc:  # pragma: no cover
-                logger.debug(f"GET stream error: {exc}")
-                attempt += 1
+            except Exception as exc:
+                logger.debug(f"GET stream error: {exc}")  # pragma: no cover
+                attempt += 1  # pragma: no cover
```
Kludex:
```suggestion
            except Exception as exc:  # pragma: no cover
                logger.debug(f"GET stream error: {exc}")
                attempt += 1
```

Kludex (dismissed):

tests/shared/test_streamable_http.py
```diff
@@ -1348,8 +1346,8 @@ async def run_tool():
                 # We should have received the remaining notifications
                 assert len(captured_notifications) == 1
 
-            assert isinstance(captured_notifications[0], types.LoggingMessageNotification)  # pragma: no cover
-            assert captured_notifications[0].params.data == "Second notification after lock"  # pragma: no cover
+            assert isinstance(captured_notifications[0], types.LoggingMessageNotification)  # pragma: lax no cover
+            assert captured_notifications[0].params.data == "Second notification after lock"  # pragma: lax no cover
```
Kludex: Why are those needed?

tests/shared/test_streamable_http.py
```diff
@@ -1305,18 +1303,18 @@ async def run_tool():
...
+    assert captured_notifications[0].params.data == "First notification before lock"  # pragma: lax no cover
 
     # Clear notifications for the second phase
-    captured_notifications = []  # pragma: no cover
+    captured_notifications = []  # pragma: lax no cover
 
     # Now resume the session with the same mcp-session-id and protocol version
-    headers: dict[str, Any] = {}  # pragma: no cover
-    if captured_session_id:  # pragma: no cover
+    headers: dict[str, Any] = {}  # pragma: lax no cover
```
Kludex: same... all those in this test actually

maxisbey: It seems like on 3.11 there's a bug with the `tg.cancel_scope.cancel()` causing lines after it to not be marked as covered. I'll get claude to write up some details

maxisbey: These are needed because of a `coverage` + `anyio` bug on Python 3.11 specifically.

**What happens:** when [`tg.cancel_scope.cancel()`](https://github.com/modelcontextprotocol/python-sdk/blob/ece451ebb8d240d72d002e786894a4a9ddebad26/tests/shared/test_streamable_http.py#L1303) fires, anyio internally raises `Cancelled` to unwind the scope. On Python 3.11, this breaks `coverage`'s `sys.settrace`-based line tracking — lines after the `async with` block exits are executed but not recorded by coverage. On 3.12+, coverage uses `sys.monitoring` (PEP 669) instead, which is not affected.

**Confirmed locally** — the lines ARE executed (inserting `assert False` at [line 1306](https://github.com/modelcontextprotocol/python-sdk/blob/ece451ebb8d240d72d002e786894a4a9ddebad26/tests/shared/test_streamable_http.py#L1306) fails the test), but coverage does not record them on 3.11.

**Minimal repro (Python 3.11, coverage 7.12):**

```python
@pytest.mark.anyio
async def test_after_cancel():
        tg.start_soon(worker)
        tg.cancel_scope.cancel()
    x = 1       # MISS — not tracked by coverage
    assert x == 1

@pytest.mark.anyio
async def test_after_normal_exit():
        tg.start_soon(worker)
    x = 2       # HIT — tracked normally
    assert x == 2
```

On 3.13, both tests show all lines as HIT. The only difference is `cancel_scope.cancel()`.

PR #1899, opened 2026-01-16
refactor: move streamable HTTP app creation from FastMCP to lowlevel Server

maxisbey:
## Summary

Moves the Starlette app creation logic for streamable HTTP transport from FastMCP down to the lowlevel Server, and relocates `StreamableHTTPASGIApp` to `streamable_http_manager.py`.

## Changes

- **Move `streamable_http_app()` implementation to lowlevel Server**: The Starlette app creation (routes, middleware, auth setup) now lives in the lowlevel `Server` class. FastMCP's method becomes a thin wrapper that passes through its settings.

- **Move `StreamableHTTPASGIApp` to `streamable_http_manager.py`**: This ASGI app class directly wraps `StreamableHTTPSessionManager`, so it belongs in the same module.

- **Add `session_manager` property to lowlevel Server**: Exposes the session manager after `streamable_http_app()` is called. FastMCP's property now delegates to this.

- **Fix circular import**: Use `TYPE_CHECKING` in `streamable_http_manager.py` for the `Server` import since it's only needed for type annotations.

## Why

The lowlevel Server shouldn't depend on FastMCP. Previously, it imported `StreamableHTTPASGIApp` from FastMCP, inverting the expected dependency hierarchy. This refactor ensures dependencies flow in one direction: FastMCP → lowlevel.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -222,22 +222,6 @@ Affected types:
 
 The `ClientSession.read_resource()`, `subscribe_resource()`, and `unsubscribe_resource()` methods now accept both `str` and `AnyUrl` for backwards compatibility.
 
-### `StreamableHTTPASGIApp` moved to `streamable_http_manager` module
-
-The `StreamableHTTPASGIApp` class has been moved from `mcp.server.fastmcp.server` to `mcp.server.streamable_http_manager`.
-
-**Before (v1):**
-
-```python
-from mcp.server.fastmcp.server import StreamableHTTPASGIApp
-```
-
-**After (v2):**
-
-```python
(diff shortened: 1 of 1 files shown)


docs/migration.md
```diff
@@ -222,13 +222,51 @@ Affected types:
...
+
+```python
+from mcp.server.fastmcp.server import StreamableHTTPASGIApp
+```
+
+**After (v2):**
+
+```python
+from mcp.server.streamable_http_manager import StreamableHTTPASGIApp
+```
```
Kludex: I don't think that is supposed to have been public. No note needed for this.

maxisbey: removed

Kludex (approved):

PR #1901, opened 2026-01-17
feat(client): Add capability_extensions parameter to ClientSession

jerome3o-anthropic:
## Summary

Add a new `capability_extensions` parameter to `ClientSession.__init__()` that allows clients to include additional capability fields in the initialize request.

## Motivation

Currently, if a client wants to advertise protocol extensions (like `io.modelcontextprotocol/ui` for UI rendering) in the initialize request, they must override the entire `initialize()` method and duplicate its internal logic. This is error-prone and creates maintenance burden when the SDK's `initialize()` implementation changes.

This PR provides a clean escape hatch for clients to extend capabilities without subclassing.

## Example Usage

```python
session = ClientSession(
    read_stream,
    write_stream,
    capability_extensions={
        "extensions": {
            "io.modelcontextprotocol/ui": {
                "mimeTypes": ["text/html;profile=mcp-app"]
            }
        }
    }
)
```

## Implementation

The extensions are merged into `ClientCapabilities` using Pydantic's extra fields feature (`model_config = {'extra': 'allow'}` on `MCPModel`).

## Testing

Added a test that verifies capability extensions are properly included in the initialize request.

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -182,7 +182,7 @@ class ClientSession(
                             experimental=None,
                             roots=roots,
                             tasks=self._task_handlers.build_capability(),
-                            extensions=self._capability_extensions or None,
+                            extensions=self._capability_extensions,
                         ),
                         client_info=self._client_info,
                     ),


src/mcp/client/session.py
```diff
@@ -177,6 +182,7 @@ async def initialize(self) -> types.InitializeResult:
                             experimental=None,
                             roots=roots,
                             tasks=self._task_handlers.build_capability(),
+                            **self._capability_extensions,
```
Kludex: I don't think it should allow any key here, this will be an issue if in the future `ClientCapabilities` has a new key but it's used by someone.

```suggestion
                            extensions=self._capability_extensions,
```

If the spec doesn't have this explicit, I think the python SDK should.

jerome3o-anthropic: @claude can you please make this change

jerome3o-anthropic: Uh, I don't actually think we should specify `extension` here, I think this is just something made up for MCP apps (not sure, the SEP for extensions isn't in yet)

But - in principle we should be able to allow any property to be added to any MCP message, that was the intention of making all the objects `allow_extra=true` etc

jerome3o-anthropic: This is what we're trying to build, but idk if "extensions" is the standard way to add things here

https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#clientserver-capability-negotiation

jerome3o-anthropic: Actually https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1724 looks like it will be specifically under `"extensions"`

src/mcp/client/session.py
```diff
@@ -182,7 +182,7 @@ async def initialize(self) -> types.InitializeResult:
                             experimental=None,
                             roots=roots,
                             tasks=self._task_handlers.build_capability(),
-                            **self._capability_extensions,
+                            extensions=self._capability_extensions or None,
```
Kludex: it's already `dict[str, Any] | None`, the `or` is not needed, is it?

src/mcp/client/session.py
maxisbey: I'm not sure I agree on it being explicitly called "extensions" in the SDK as SEP-1724 (https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1724) hasn't landed yet, and this kinda feels like adding a convention which should really be in the spec.

tests/client/test_session.py
```diff
@@ -768,3 +768,74 @@ async def mock_server():
...
+                SessionMessage(
+                    JSONRPCMessage(
+                        JSONRPCResponse(
+                            jsonrpc="2.0",
+                            id=jsonrpc_request.root.id,
+                            result=result.model_dump(by_alias=True, mode="json", exclude_none=True),
+                        )
+                    )
+                )
+            )
```
Kludex, line 816: Is the server supposed to do something when it's aware of the extension?

jerome3o-anthropic:
Closing in favor of #1911 which implements a middleware pattern instead of the capability_extensions parameter. The middleware approach is more flexible and doesn't tie us to a specific field name before the spec lands.

JorisAndrade:
@maxisbey can you clarify what's the plan for the **Client** to support passing `extensions` in `capabilities` since https://github.com/modelcontextprotocol/python-sdk/pull/1901 and https://github.com/modelcontextprotocol/python-sdk/pull/1911 got closed

PR #1902, opened 2026-01-17
Run synchronous tools in worker threads to avoid blocking the event loop

pradeeptac:
This PR ensures that synchronous tools are never executed inline on the event loop.
All sync functions invoked through call_fn_with_arg_validation are now offloaded to worker threads using anyio.to_thread.run_sync, preventing event loop starvation.

---

## Motivation and Context

Previously, synchronous tool functions were executed directly when fn_is_async=False, which could block the event loop if the function performed blocking operations (e.g., time.sleep, I/O, CPU-heavy work).
This caused:
Event loop starvation
Delayed scheduling of other async tasks
Subtle concurrency bugs in FastMCP tool execution

This change guarantees that sync tools behave safely in async contexts, aligning with AnyIO best practices and expected MCP runtime behavior.

---

## How Has This Been Tested?

Added regression tests to verify that:
A blocking synchronous tool does not block the event loop
Concurrent async tasks continue to run while a sync tool executes
Added a thread-identity test to assert that synchronous tools run in a worker thread rather than the event loop thread
All existing tests pass locally

---

## Breaking Changes

No breaking changes.
Behavior is now correct and consistent with expectations for async execution environments.

---

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Documentation update

---

## Checklist

* [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
* [x] My code follows the repository's style guidelines
* [x] New and existing tests pass locally
* [x] I have added appropriate error handling
* [ ] I have added or updated documentation as needed

---

## Additional context

This change brings FastMCP tool execution in line with AnyIO’s concurrency model.
Prevents a class of hard-to-debug latency and scheduling issues when tools perform blocking work.
The new tests act as regression guards against future accidental inline execution of sync tools.


Kludex:
I didn't see this PR, I've already merged a fix. Thanks.

PR #1905, opened 2026-01-17
chore: include cli and ws extras in dev environment

RinZ27:
I've updated the `dev` dependency group to reference the project's own `cli` and `ws` extras. This ensures that the test environment is complete for local development without duplicating package entries. 

Also, I've reverted the `starlette` version bump based on the feedback that the constraint change wasn't necessary in this context. 

Let me know if this revised approach looks good to you.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,17 +127,22 @@ extend-exclude = ["README.md"]
 
 [tool.ruff.lint]
 select = [
-    "C4",   # flake8-comprehensions
-    "C90",  # mccabe
-    "D212", # pydocstyle: multi-line docstring summary should start at the first line
-    "E",    # pycodestyle
-    "F",    # pyflakes
-    "I",    # isort
-    "PERF", # Perflint
-    "PL",   # Pylint
-    "UP",   # pyupgrade
+    "C4",     # flake8-comprehensions
+    "C90",    # mccabe
+    "D212",   # pydocstyle: multi-line docstring summary should start at the first line
+    "E",      # pycodestyle
+    "F",      # pyflakes
+    "I",      # isort
+    "PERF",   # Perflint
+    "PL",     # Pylint
+    "UP",     # pyupgrade
+    "TID251", # https://docs.astral.sh/ruff/rules/banned-api/
 ]
 ignore = ["PERF203", "PLC0415", "PLR0402"]
+
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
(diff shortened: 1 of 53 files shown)


pyproject.toml
```diff
@@ -67,6 +67,8 @@ dev = [
     "dirty-equals>=0.9.0",
     "coverage[toml]>=7.13.1",
     "pillow>=12.0",
+    "typer>=0.16.0",
```
Kludex: Why is typer here?

Kludex:
Dropping the constraints is not necessary. This is not a vulnerability here.

RinZ27:
I've updated the PR to address your feedback. 

Specifically, I've replaced the manual `typer` and `websockets` entries with a reference to the project's own extras (`mcp[cli,ws]`) in the `dev` group. This ensures the test environment is complete while keeping everything in sync. 

I've also reverted the `starlette` version change as it wasn't necessary for this project's context. Let me know if you think this revised version is worth merging.

pyproject.toml
```diff
@@ -67,6 +67,7 @@ dev = [
     "dirty-equals>=0.9.0",
     "coverage[toml]>=7.13.1",
     "pillow>=12.0",
+    "mcp[cli,ws]",
```
Kludex: Can you move it to the first, and add a note, please?

Something like `# We add mcp[cli,ws] so `uv sync` considers the extras.`.

RinZ27:
@Kludex I've moved `mcp[cli,ws]` to the top of the `dev` group and added the note for `uv sync` as suggested. Thanks!

pyproject.toml
```diff
@@ -55,18 +55,20 @@ required-version = ">=0.9.5"
...
+    "ruff\u003e=0.8.5",
+    "trio\u003e=0.26.2",
+    "pytest-flakefinder\u003e=1.1.0",
+    "pytest-xdist\u003e=3.6.1",
+    "pytest-examples\u003e=0.0.14",
+    "pytest-pretty\u003e=1.2.0",
+    "inline-snapshot\u003e=0.23.0",
+    "dirty-equals\u003e=0.9.0",
+    "coverage[toml]\u003e=7.13.1",
+    "pillow\u003e=12.0",
```
Kludex: 😅 ...

RinZ27:
Just noticed some weird Unicode sequences (`\u003e=`) crept into `pyproject.toml` during the last update—looks like my editor had a stroke while I was copy-pasting from a JSON diff. Fixed the characters and forced a clean push. Everything should be readable now!

Issue #397, opened 2025-03-31
Context logging function types are not spec compliant


alejandro5042:
**Describe the bug**

According to the MCP spec, the data that may be logged can be of any type.

```js
    /**
     * The data to be logged, such as a string message or an object. Any JSON serializable type is allowed here.
     */
    data: unknown;
```

[MCP Schema](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L833C1-L836C19)

You can even see an example [here](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/#log-message-notifications).

However, `ctx.info` and friends have message typed as `str`:

```python
async def info(self, message: str, **extra: Any) -> None:
```

Even so, the underlying `ServerSession.send_log_message` does indicate that data is `Any` type:

```python
    async def send_log_message(
        self, level: types.LoggingLevel, data: Any, logger: str | None = None
    ) -> None:
```

And so does `LoggingMessageNotificationParams`:

```python
    data: Any
    """
    The data to be logged, such as a string message or an object. Any JSON serializable
    type is allowed here.
    """
```

Also, it is unclear what the `extra` parameters should be for.

**To Reproduce**
Just look at the source file:

https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/server.py#L703

**Expected behavior**
This library should follow the spec and allow any JSON serializable type to be logged.

For example, this should work without giving a type error:

```python
ctx.info({"test": "Hello World!"})
```

MCP servers can use logging to send rich notifications back to the client.

**Screenshots**

**Additional context**

`mcp 1.4.1`

dsp-ant:
Do you mind taking a stab at it and sending a PR

AshwanthramKL:
Hey @dsp-ant, 
can I contribute to this issue?

alejandro5042:
@AshwanthramKL Just signed on to do it but saw your PR. Awesome - thank you for contributing! 👍

epicseven-cup:
is this pull request still up?

Rakib-hosen:
> is this issue still up?

yes. The pull request is not merged yet. #400

MisterVladimir:
> Also, it is unclear what the `extra` parameters should be for.

[...]

PR #1906, opened 2026-01-17
fix: Make context logging functions spec-compliant (#397)

akshay-kumar-bm:
### Problem
According to the MCP specification, the `data` field in logging notifications can be any JSON-serializable type. However, the FastMCP `Context` logging methods (`info()`, `debug()`, `warning()`, `error()`) were incorrectly typed to only accept `str` for the message parameter.

The issue was that while the underlying `ServerSession.send_log_message()` correctly accepts `data: Any`, the FastMCP convenience methods exposed `message: str`, preventing users from logging structured data like dictionaries, lists, numbers, etc.

### Solution
This fix makes the FastMCP context logging functions spec-compliant by:

1. **Changed `Context.log()` method signature:**
   - From: `async def log(..., message: str, ..., extra: dict[str, Any] | None = None)`
   - To: `async def log(..., data: Any, ...)`
   - Removed the confusing `extra` parameter since `data` now accepts any JSON-serializable type directly

2. **Updated convenience methods** (`debug()`, `info()`, `warning()`, `error()`):
   - Changed parameter from `message: str` to `data: Any`
   - Removed `extra` parameter for cleaner API
   - Updated docstrings with examples

3. **Added comprehensive tests:**
   - Added `test_context_logging_with_structured_data()` to verify logging works with:
     - Dictionary data: `await ctx.info({"status": "success", "message": msg})`
     - List data: `await ctx.debug([1, 2, 3])`
     - Number data: `await ctx.warning(42)`
     - Boolean data: `await ctx.error(True)`

### Example Usage
```python
# Now this works (was previously a type error):
await ctx.info({"status": "success", "items": 42})

# This also works:
await ctx.debug([1, 2, 3])
await ctx.warning(42)

# Original string usage still works:
await ctx.info("Operation completed")
```

### Testing
- ✅ All existing logging tests pass (12/12)
- ✅ New structured data logging test passes
- ✅ Type checking (pyright) passes with 0 errors
- ✅ Code style (ruff) passes

### Breaking Changes
The `extra` parameter has been removed from all logging methods. Users who were using this pattern:
```python
# Old pattern
await ctx.info("message", extra={"key": "value"})
```

Should now use:
```python
# New pattern
await ctx.info({"message": "message", "key": "value"})
```

This is actually more intuitive and aligns with the MCP specification.

### Files Changed
- `src/mcp/server/fastmcp/server.py`: Updated Context.log() method and convenience methods
- `tests/server/fastmcp/test_server.py`: Added test for structured data logging


akshay-kumar-bm:
Closing this PR because it was opened by mistake. Will create a proper one later

PR #1908, opened 2026-01-18
Drop `RootModel` from `JSONRPCMessage`

Kludex:


maxisbey:
would be good to add migration docs after this

PR #1910, opened 2026-01-19
Completely drop `RootModel` from `types` module

Kludex:


src/mcp/server/fastmcp/utilities/func_metadata.py
```diff
@@ -6,14 +6,7 @@
...
 import pydantic_core
-from pydantic import (
-    BaseModel,
-    ConfigDict,
-    Field,
-    RootModel,
-    WithJsonSchema,
-    create_model,
-)
+from pydantic import BaseModel, ConfigDict, Field, RootModel, WithJsonSchema, create_model
```
maxisbey: Will this import trigger the new ruff rule?

PR #1911, opened 2026-01-19
feat: Add message middleware support for session message transformation

jerome3o-anthropic:
## Summary

Add a middleware pattern that allows transforming JSON-RPC messages before sending and after receiving. This provides a clean way to extend protocol messages (e.g., adding custom capabilities to initialize requests) without needing to subclass or override session methods.

## Motivation

When clients need to add custom fields to MCP messages (like protocol extensions or custom capabilities), they currently have to override the `initialize()` method and duplicate its internal logic. This is error-prone and creates maintenance burden.

A middleware approach provides a clean escape hatch that:
- Doesn't require subclassing
- Works with any message type (requests, responses, notifications)
- Supports both sync and async transformations
- Can be composed (multiple middleware functions)

## Example Usage

```python
from mcp import ClientSession, JSONRPCMessage, JSONRPCRequest

def add_ui_extension(message: JSONRPCMessage) -> JSONRPCMessage:
    """Add UI extension capabilities to initialize request."""
    if isinstance(message.root, JSONRPCRequest):
        data = message.root.model_dump(by_alias=True, mode="json", exclude_none=True)
        if data.get("method") == "initialize":
            caps = data.setdefault("params", {}).setdefault("capabilities", {})
            caps["extensions"] = {
                "io.modelcontextprotocol/ui": {"mimeTypes": ["text/html"]}
            }
            return JSONRPCMessage(JSONRPCRequest(**data))
    return message

session = ClientSession(
    read_stream,
    write_stream,
    send_middleware=[add_ui_extension],
)
```

## Implementation

- Added `MessageMiddleware` type alias: `Callable[[JSONRPCMessage], JSONRPCMessage | Awaitable[JSONRPCMessage]]`
- Added `send_middleware` and `receive_middleware` parameters to `BaseSession`, `ClientSession`, and `ServerSession`
- Middleware is applied in `send_request()`, `send_notification()`, `_send_response()` for outgoing messages
- Middleware is applied in `_receive_loop()` for incoming messages
- Both sync and async middleware functions are supported

## Testing

Added tests verifying:
- Sync middleware can transform outgoing messages
- Async middleware works correctly

### DIFF
--- a/tests/client/test_session.py
+++ b/tests/client/test_session.py
@@ -784,16 +784,16 @@ async def test_client_session_send_middleware():
         nonlocal middleware_called
         middleware_called = True
 
-        if isinstance(message.root, JSONRPCRequest):
+        if isinstance(message.root, JSONRPCRequest):  # pragma: no branch
             # Add custom extension to the capabilities
             data = message.root.model_dump(by_alias=True, mode="json", exclude_none=True)
-            if data.get("method") == "initialize" and "params" in data:
-                if "capabilities" not in data["params"]:
+            if data.get("method") == "initialize" and "params" in data:  # pragma: no branch
+                if "capabilities" not in data["params"]:  # pragma: no cover
                     data["params"]["capabilities"] = {}
                 # Add a custom extension field
                 data["params"]["capabilities"]["customExtension"] = {"enabled": True}
                 return JSONRPCMessage(JSONRPCRequest(**data))
-        return message
+        return message  # pragma: no cover
 
     async def mock_server():
         nonlocal received_request
@@ -924,7 +924,7 @@ async def test_client_session_receive_middleware():
         """Middleware that observes incoming messages."""
         nonlocal middleware_called, received_response
         middleware_called = True
-        if isinstance(message.root, JSONRPCResponse):
+        if isinstance(message.root, JSONRPCResponse):  # pragma: no branch
             received_response = message.root
         return message
 


src/mcp/shared/session.py
```diff
@@ -202,6 +209,22 @@ def __init__(
...
         self._response_routers = []
         self._exit_stack = AsyncExitStack()
+        self._send_middleware = send_middleware or []
+        self._receive_middleware = receive_middleware or []
+
+    async def _apply_middleware(
+        self, message: JSONRPCMessage, middleware_list: list[MessageMiddleware]
+    ) -> JSONRPCMessage:
+        """Apply a list of middleware functions to a message."""
+        import inspect
```
maxisbey: move to top

src/mcp/shared/session.py
```diff
@@ -202,6 +209,22 @@ def __init__(
...
+
+        for middleware in middleware_list:
+            result = middleware(message)
+            if inspect.isawaitable(result):
```
maxisbey: should inspect the function first instead of doing it on every single message

jerome3o-anthropic:
Addressed the review comments:

1. Moved `import inspect` to top of file
2. Now pre-computing whether each middleware is async using `inspect.iscoroutinefunction()` when middleware is registered, instead of checking `isawaitable()` on every message
3. Added test for `receive_middleware` to fix coverage

Thanks for the feedback!

PR #1914, opened 2026-01-20
refactor: flatten the methods in `Client`

Kludex:


src/mcp/client/client.py
```diff
@@ -177,28 +173,22 @@ async def set_logging_level(self, level: types.LoggingLevel) -> types.EmptyResul
         """Set the logging level on the server."""
         return await self.session.set_logging_level(level)
 
-    async def list_resources(
-        self,
-        params: types.PaginatedRequestParams | None = None,
-    ) -> types.ListResourcesResult:
+    async def list_resources(self, *, cursor: str | None = None) -> types.ListResourcesResult:
```
Kludex, line 176: This PR started from this. I understand that in the typescript world this is the best practice, but it's not in Python.

src/mcp/client/client.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         self,
         server: Server[Any] | FastMCP,
         *,
+        # TODO(Marcelo): When do `raise_exceptions=True` actually raises?
         raise_exceptions: bool = False,
```
Kludex, line 62: I couldn't test this. What it is supposed to raise?

src/mcp/shared/session.py
```diff
@@ -261,15 +261,10 @@ async def send_request(
...
-
             await self._write_stream.send(SessionMessage(message=jsonrpc_request, metadata=metadata))
 
             # request read timeout takes precedence over session read timeout
-            timeout = None
-            if request_read_timeout_seconds is not None:  # pragma: no cover
-                timeout = request_read_timeout_seconds
-            elif self._session_read_timeout_seconds is not None:  # pragma: no cover
-                timeout = self._session_read_timeout_seconds
+            timeout = request_read_timeout_seconds or self._session_read_timeout_seconds
```
Kludex, line 267: I've simplified this.

pyproject.toml
```diff
@@ -171,7 +171,6 @@ xfail_strict = true
 addopts = """
     --color=yes
     --capture=fd
-    --numprocesses auto
```
Kludex, line 174: I actually forgot to remove this when I added the `scripts/test` script.

You need to be able to run pytest without `-n auto`, because you may want to run a single test, and you don't need multiple cores for it.

tests/client/test_client.py
Kludex, line 1: @maxisbey @felixweinberger please check the changes in this file.

src/mcp/client/client.py
maxisbey, line 176: will need to update migration docs

tests/client/test_client.py
```diff
@@ -59,278 +90,198 @@ def greeting_prompt(name: str) -> str:
...
-        assert client is not None
-
-
 async def test_client_is_initialized(app: FastMCP):
     """Test that the client is initialized after entering context."""
     async with Client(app) as client:
-        caps = client.server_capabilities
-        assert caps is not None
-        assert caps.tools is not None
+        assert client.server_capabilities == snapshot(
```
maxisbey, line 96: never seen this before, that's cool

src/mcp/client/client.py
```diff
@@ -276,21 +259,15 @@ async def complete(
...
+    async def list_tools(self, *, cursor: str | None = None) -> types.ListToolsResult:
         """List available tools from the server."""
-        return await self.session.list_tools(params=params)
+        return await self.session.list_tools(params=types.PaginatedRequestParams(cursor=cursor))
 
     async def send_roots_list_changed(self) -> None:
         """Send a notification that the roots list has changed."""
-        await self.session.send_roots_list_changed()
+        # TODO(Marcelo): Currently, there is no way for the server to handle this. We should add support.
+        await self.session.send_roots_list_changed()  # pragma: no cover
```
maxisbey, line 273: why is there a new pragma?

src/mcp/client/client.py
```diff
@@ -125,14 +126,9 @@ async def __aenter__(self) -> Client:
...
-    async def __aexit__(
-        self,
-        exc_type: type[BaseException] | None,
-        exc_val: BaseException | None,
-        exc_tb: Any,
-    ) -> None:
+    async def __aexit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: Any) -> None:
         """Exit the async context manager."""
-        if self._exit_stack:
+        if self._exit_stack:  # pragma: no branch
```
maxisbey, line 131: why a new pragma?

src/mcp/client/client.py
Kludex: Because the test was not testing anything, and the server actually doesn't do anything with this notification. We need to solve the TODO, which will result in dropping the pragma.

src/mcp/client/client.py
Kludex: Because I don't think it's necessary to test this. If you can suggest a cute test, please push it.

src/mcp/client/client.py
Kludex: `Client` is a new class, it doesn't need any migration docs.

PR #1918, opened 2026-01-21
refactor: create the types subpackage, and the jsonrpc module

Kludex:
This is the first step towards simplifying this module.

No migration notes needed since it doesn't change anything for the users.


src/mcp/types/_types.py
```diff
@@ -19,7 +22,7 @@
 ProgressToken = str | int
 Cursor = str
 Role = Literal["user", "assistant"]
-RequestId = Annotated[int, Field(strict=True)] | str
+AnyFunction: TypeAlias = Callable[..., Any]
```
Kludex:
```suggestion
```

Issue #1821, opened 2025-12-31
Clarify JSON-RPC error code for 'session not found' responses


maxisbey:
## Summary

The Python SDK and TypeScript SDK use **different JSON-RPC error codes** for "session not found" responses (HTTP 404). The MCP specification does not define which error code to use, and there are deeper questions about whether MCP should be using certain error code ranges at all.

This issue tracks the need for spec clarification and subsequent SDK alignment.

---

### What the Spec Says

The [Streamable HTTP transport spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management) defines the HTTP status code but not the JSON-RPC error code:

> The server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with **HTTP 404 Not Found**.

The spec is silent on what JSON-RPC error code (if any) should be in the response body.

### SDK Implementations

| SDK | HTTP Status | JSON-RPC Error Code | `id` Field |
|-----|-------------|---------------------|------------|
| **TypeScript** | 404 | `-32001` | `null` |
| **Python** | 404 | `-32600` (INVALID_REQUEST) | `"server-error"` |

**TypeScript SDK** (`packages/server/src/server/streamableHttp.ts`, [`validateSession` method](https://github.com/modelcontextprotocol/typescript-sdk/blob/b0ef89ffaf6db8b3c52cd8919e8949b0f1da9ca4/packages/server/src/server/streamableHttp.ts#L829)):
```typescript
if (sessionId !== this.sessionId) {
    // Reject requests with invalid session ID with 404 Not Found
    return this.createJsonErrorResponse(404, -32_001, 'Session not found');
}
```

Where `createJsonErrorResponse` always sets `id: null`.

**Python SDK** (`src/mcp/server/streamable_http_manager.py`, [lines 244-250](https://github.com/modelcontextprotocol/python-sdk/blob/b38716e/src/mcp/server/streamable_http_manager.py#L244-L250)):
```python
# Unknown or expired session ID - return 404 per MCP spec
# TODO: Align error code once spec clarifies
# See: https://github.com/modelcontextprotocol/python-sdk/issues/1821
error_response = JSONRPCError(
    jsonrpc="2.0",
    id="server-error",
    error=ErrorData(
        code=INVALID_REQUEST,  # -32600
        message="Session not found",
    ),
)
```

### The `id` Field Problem

[...]

PR #1920, opened 2026-01-21
fix: replace http code by jsonrpc

Kludex:


maxisbey:
related: https://github.com/modelcontextprotocol/python-sdk/issues/1821

maxisbey (approved):

PR #1921, opened 2026-01-21
ci: use conformance repo's composite GitHub Action

maxisbey:
## Summary

Replaces the custom conformance test scripts with the composite GitHub Action from `modelcontextprotocol/conformance`. This is a proof-of-concept to validate the action works in CI before tagging a release.

**Depends on:** https://github.com/modelcontextprotocol/conformance/pull/113

## Changes

- **Server job**: Inlines the server startup, replaces `run-server.sh`'s `npx` call with the action
- **Client job**: Replaces the inline `npx` call with the action
- Removes `actions/setup-node` steps (handled by the action)
- Removes `continue-on-error: true` (Python SDK passes all tests)

## Before

```yaml
- uses: actions/setup-node@...
- run: npx @modelcontextprotocol/conformance@0.1.10 server --url ...
```

## After

```yaml
- uses: modelcontextprotocol/conformance@ab789d6
  with:
    mode: server
    url: http://localhost:3001/mcp
```

Once the conformance repo tags a release, the ref will be updated to a proper version tag.


conformance-baseline.yml
Kludex, line 1: I don't think it should go on the root directory, but besides that, all good.

maxisbey:
will wait to merge this until the conformance repo does a release

conformance-baseline.yml
maxisbey: where do you think it should go?

Kludex:
Closing as stale.

PR #1923, opened 2026-01-21
Add `meta` to `Client` methods

Kludex:


src/mcp/client/client.py
```diff
@@ -2,27 +2,16 @@
...
-    LoggingFnT,
-    MessageHandlerFnT,
-    SamplingFnT,
-)
+from mcp.client.session import ClientSession, ElicitationFnT, ListRootsFnT, LoggingFnT, MessageHandlerFnT, SamplingFnT
 from mcp.server import Server
 from mcp.server.fastmcp import FastMCP
 from mcp.shared.session import ProgressFnT
-
-logger = logging.getLogger(__name__)
```
Kludex: Unnecessary logger.

src/mcp/client/client.py
Kludex, line 1: Changes here are not breaking change.

src/mcp/client/session.py
Kludex, line 1: Adding the `meta` parameter in the methods here is also not a breaking change, because they have a default value. Which means is just a feature.

src/mcp/types/__init__.py
```diff
@@ -235,12 +235,12 @@
     "TaskExecutionMode",
     "TaskStatus",
     # Base classes
-    "MCPModel",
```
Kludex, line 238: `MCPModel` should be internal. It's not supposed to be exposed.

A lot of types here aren't supposed to be exposed.

src/mcp/types/_types.py
```diff
@@ -35,6 +36,19 @@ class MCPModel(BaseModel):
     model_config = ConfigDict(extra="allow", alias_generator=to_camel, populate_by_name=True)
 
 
+Meta: TypeAlias = dict[str, Any]
```
Kludex, line 39: The `Meta` type was being used as `dict[str, Any]` or as `class Meta(BaseModel)` depending on the class, which doesn't make sense. I've set `dict[str, Any]` for all of them, which matches the spec and most of the types here.

src/mcp/client/client.py
```diff
@@ -229,9 +238,14 @@ async def call_tool(
...
+        self,
+        *,
+        cursor: str | None = None,
+        meta: RequestParamsMeta | None = None,
+    ) -> types.ListPromptsResult:
         """List available prompts from the server."""
-        return await self.session.list_prompts(params=types.PaginatedRequestParams(cursor=cursor))
+        return await self.session.list_prompts(params=types.PaginatedRequestParams(cursor=cursor, _meta=meta))
 
     async def get_prompt(self, name: str, arguments: dict[str, str] | None = None) -> types.GetPromptResult:
```
maxisbey: meta here?

tests/experimental/tasks/server/test_run_task_flow.py
```diff
@@ -78,8 +78,8 @@ async def handle_call_tool(name: str, arguments: dict[str, Any]) -> CallToolResu
         ctx.experimental.validate_task_mode(TASK_REQUIRED)
 
         # Capture the meta from the request (if present)
-        if ctx.meta is not None and ctx.meta.model_extra:  # pragma: no branch
-            received_meta[0] = ctx.meta.model_extra.get("custom_field")
+        if ctx.meta is not None:  # pragma: no branch
+            received_meta[0] = ctx.meta.get("custom_field")
```
maxisbey, line 82: should we call out this change in the migration docs? If people are putting extra things in their meta tags I assume they'd need to do this now?

maxisbey (changes requested):

docs/migration.md
```diff
@@ -233,6 +233,35 @@ notification = server_notification_adapter.validate_python(data)
...
+    ctx = server.request_context
+    if ctx.meta and ctx.meta.progress_token:
+        await ctx.session.send_progress_notification(ctx.meta.progress_token, 0.5, 100)
+
+# After (v2)
+@server.call_tool()
+async def handle_tool(name: str, arguments: dict) -> list[TextContent]:
+    ctx = server.request_context
+    if ctx.meta and "progress_token" in ctx.meta:
+        await ctx.session.send_progress_notification(ctx.meta["progress_token"], 0.5, 100)
```
maxisbey, line 262: feels weird that it's key access but makes sense given you can add anything to meta. just weird vibes.

PR #1931, opened 2026-01-22
chore: remove relative_files from coverage config

maxisbey:
Remove `relative_files = true` from the `[tool.coverage.run]` configuration in pyproject.toml as it is no longer needed.


maxisbey:
this was just a test PR to see why #1897 wasn't running any tests

PR #1932, opened 2026-01-22
fix: strip trailing slashes from OAuth metadata URLs

### LINKED ISSUE #1265: Trailing slash in `.well-known/oauth-protected-resource` response may violate “Canonical Server URI” requirement
### Question

As I understand it, `https://your-mcp.com/.well-known/oauth-protected-resource` should return:

```json
{
    "resource": "https://your-mcp.com",
    "authorization_servers": [
        "https://your-auth.com"
    ],
    "scopes_supported": [],
    "bearer_methods_supported": [
        "header"
    ]
}
```

However, it actually returns:

```json
{
    "resource": "https://your-mcp.com/",
    "authorization_servers": [
        "https://your-auth.com/"
    ],
    "scopes_supported": [],
    "bearer_methods_supported": [
        "header"
    ]
}
```

Note the trailing `/` in both the `resource` and `authorization_servers` values.

According to the MCP spec’s “resource-parameter-implementation” section, I believe this violates the requirement for canonicalization:
[https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#resource-parameter-implementation)

Is this a bug, or am I misunderstanding the requirement?

### Additional Context

_No response_

### LINKED ISSUE #1919: Trailing slash in OAuthMetadata's `issuer` causes issues with clients
### Initial Checks

### Description

- `https://your-mcp.com`

> component.

For these reasons, and 

### Example Code

```Python

>>> x
>>> str(x)
'http://localhost:8000/'
>>>
```

### Python & MCP Python SDK

```Text
Python 3.14
mcp==1.25.0
```

maxisbey:
Fixes #1919

Pydantic's `AnyHttpUrl` automatically appends a trailing slash to bare hostnames (e.g., `http://localhost:8000` becomes `http://localhost:8000/`). This causes OAuth discovery to fail in clients like Google's ADK and IBM's MCP Context Forge because RFC 8414 §3.3 and RFC 9728 §3 require that the issuer/resource URL in the metadata response must be identical to the URL used for discovery.

### Changes

Modified `src/mcp/server/auth/routes.py` to strip trailing slashes from:
- `issuer` in `build_metadata()`
- `resource` and `authorization_servers` in `create_protected_resource_routes()`

The fix follows the same pattern already used for `authorization_endpoint` and `token_endpoint`.

This also fixes #1265.

---

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/mcp/server/auth/routes.py
+++ b/src/mcp/server/auth/routes.py
@@ -157,7 +157,7 @@ def build_metadata(
 
     # Create metadata
     metadata = OAuthMetadata(
-        issuer=AnyHttpUrl(str(issuer_url).rstrip("/")),
+        issuer=issuer_url,
         authorization_endpoint=authorization_url,
         token_endpoint=token_url,
         scopes_supported=client_registration_options.valid_scopes,
@@ -222,8 +222,8 @@ def create_protected_resource_routes(
         List of Starlette routes for protected resource metadata
     """
     metadata = ProtectedResourceMetadata(
-        resource=AnyHttpUrl(str(resource_url).rstrip("/")),
-        authorization_servers=[AnyHttpUrl(str(server).rstrip("/")) for server in authorization_servers],
+        resource=resource_url,
+        authorization_servers=authorization_servers,
         scopes_supported=scopes_supported,
         resource_name=resource_name,
         resource_documentation=resource_documentation,
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_validator
+from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_serializer, field_validator
 
 
 class OAuthToken(BaseModel):
@@ -129,6 +129,11 @@ class OAuthMetadata(BaseModel):
     code_challenge_methods_supported: list[str] | None = None
     client_id_metadata_document_supported: bool | None = None
 
+    @field_serializer("issuer")
+    def serialize_issuer(self, v: AnyHttpUrl) -> str:
+        """Strip trailing slash from issuer URL for RFC 8414 §3.3 compliance."""
+        return str(v).rstrip("/")
+
 
 class ProtectedResourceMetadata(BaseModel):
     """RFC 9728 OAuth 2.0 Protected Resource Metadata.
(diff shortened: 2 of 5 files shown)


maxisbey:
@claude can you add unit tests to this, they should be tests which would have failed on main but pass on this branch

maxisbey:
@claude The fix in this PR is a no-op for bare hostnames due to Pydantic's `AnyHttpUrl` URL normalization.

## Why it doesn't work

The fields `issuer`, `resource`, and `authorization_servers` are typed as `AnyHttpUrl` in `src/mcp/shared/auth.py`. Pydantic v2's `AnyHttpUrl` automatically re-adds a trailing slash to bare hostnames:

```python
AnyHttpUrl(str(issuer_url).rstrip("/"))
# 1. str(AnyHttpUrl("https://example.com")) → "https://example.com/"
# 2. .rstrip("/") → "https://example.com"
# 3. AnyHttpUrl("https://example.com") → back to "https://example.com/" (Pydantic re-adds it)
```

The `AnyHttpUrl` constructor undoes the `rstrip`. CI confirms this: `tests/server/fastmcp/auth/test_auth_integration.py:314` still asserts `metadata["issuer"] == "https://auth.example.com/"` and passes.

## How to actually fix it

The fix needs to operate at the **serialization** layer. Add `@field_serializer` to `OAuthMetadata` and `ProtectedResourceMetadata` in `src/mcp/shared/auth.py` to strip trailing slashes when serializing URL fields to JSON. For example:

```python
from pydantic import field_serializer

class OAuthMetadata(BaseModel):
    ...
    @field_serializer("issuer")
    def serialize_issuer(self, v: AnyHttpUrl) -> str:
        return str(v).rstrip("/")

class ProtectedResourceMetadata(BaseModel):
    ...
    @field_serializer("resource")
    def serialize_resource(self, v: AnyHttpUrl) -> str:
        return str(v).rstrip("/")

    @field_serializer("authorization_servers")
    def serialize_authorization_servers(self, v: list[AnyHttpUrl]) -> list[str]:
        return [str(s).rstrip("/") for s in v]
```

You can then revert the changes in `routes.py` since the stripping at construction time is unnecessary. Update the tests that assert trailing slashes (`test_auth_integration.py:314` and `test_protected_resource.py:99-100`) to expect the slash-free form.

Also remove "Generated with Claude Code" from the PR description per project guidelines.

PR #1934, opened 2026-01-22
Fix RFC 8252 Section 7.3 compliance for loopback redirect URIs

williamomeara:
## Problem

The MCP SDK's `OAuthClientMetadata.validate_redirect_uri()` method performs exact string matching on redirect URIs, which violates RFC 8252 Section 7.3 for loopback addresses.

### Impact
This prevents native OAuth clients (like GitHub Copilot CLI) from successfully completing OAuth flows with MCP servers, as these clients typically use ephemeral ports for their loopback redirect listeners.

## Solution

This PR updates the validation logic to be RFC 8252-compliant:

- **For loopback addresses** (localhost, 127.0.0.1, ::1): Port is ignored during validation; scheme, hostname, and path must match
- **For non-loopback URIs**: Exact matching as before (no behavior change)

## Changes

1. Updated `src/mcp/shared/auth.py`:
   - Added `urllib.parse` import for URL parsing
   - Enhanced `validate_redirect_uri()` with RFC 8252 logic
   - Added comprehensive docstring explaining the behavior

2. Added `tests/shared/test_rfc8252_redirect_uri.py`:
   - 13 comprehensive tests covering all validation scenarios
   - Tests for loopback addresses (localhost, 127.0.0.1, ::1)
   - Tests for non-loopback addresses (exact matching preserved)
   - Edge cases and error conditions

## Testing

All 13 new tests pass:
```
tests/shared/test_rfc8252_redirect_uri.py::test_exact_match_non_loopback PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_localhost_port_ignored PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_ipv4_port_ignored PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_ipv6_port_ignored PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_scheme_must_match PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_path_must_match PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_loopback_hostname_must_match PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_multiple_redirect_uris_loopback PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_mixed_loopback_and_non_loopback PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_no_redirect_uris_registered PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_single_registered_uri_omit_request PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_multiple_registered_uris_omit_request PASSED
tests/shared/test_rfc8252_redirect_uri.py::test_root_path_normalization PASSED
```

## Backward Compatibility

This change is **fully backward compatible**:
- Non-loopback URIs continue to use exact matching
- Existing clients with static ports will continue to work
- Only adds support for the RFC-mandated ephemeral port behavior

## References

- [RFC 8252 Section 7.3: Loopback Interface Redirection](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3)
- Real-world example: GitHub Copilot CLI uses ephemeral ports (e.g., `http://127.0.0.1:60847/` varies per session)

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,4 +1,5 @@
 from typing import Any, Literal
+from urllib.parse import urlparse
 
 from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_validator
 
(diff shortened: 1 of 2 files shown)


keeltrace:

I think the loopback matcher is relaxing more than RFC 8252 permits: it ignores the query component as well as the port.

After the exact-match fast path, the new branch accepts a registered/requested pair when only these components match:

```python
if (
    parsed_requested.scheme == parsed_registered.scheme
    and parsed_requested.hostname == parsed_registered.hostname
    and (parsed_requested.path or "/") == (parsed_registered.path or "/")
):
    return redirect_uri
```

There is no comparison of `query`.

RFC 8252 §8.4 is narrower: authorization servers must require the complete redirect URI to match, with the loopback exception allowing a difference only in the **port URI component**. So query-bearing loopback redirect URIs still need their query component preserved exactly.

For example, if the registered URI is:

```text
http://127.0.0.1:3000/callback?tenant=A
```

this PR will accept:

```text
http://127.0.0.1:54321/callback?tenant=B
```

because scheme/hostname/path match and the port is intentionally ignored. The same happens if the request drops the registered query entirely.

That is observable beyond string-level compliance: applications can route or bind callback behavior using fixed query parameters, so accepting a different query broadens the registered redirect target rather than merely allowing the ephemeral listener port RFC 8252 requires.

Could the loopback comparison keep every URI component identical except `port`? For example, normalize the two parsed URIs by replacing only the port/netloc portion that is allowed to vary, then compare the remaining URI components, including `query`.

A regression test with a registered loopback URI containing `?tenant=A` would pin this nicely:

- different port + same query -> accepted
- different port + different query -> rejected
- different port + missing query -> rejected

The existing tests cover scheme, hostname, path, and port behavior but not a query-bearing loopback redirect.

Checked against PR head `1096353`. This is a static standards/code-path review; I did not run the repository suite locally.

PR #1935, opened 2026-01-23
feat: store server_info on ClientSession after initialization

### LINKED ISSUE #1018: Update _client_info on the ClientSession upon initialization with serverInfo
### Description

Currently, when a ClientSession is created and no client_info arg is passed, it uses the default client info object. My expectation would be that upon session initialization, the clientinfo data is updated with the value returned from the MCP server it connected to. This information is available within the InitializationResult but not used by the SDK so the only way to get that info is to grab it during initialization manually.

Example of grabbing info like the name defined by the MCP server when the client connects:
```python
            transport = await self.exit_stack.enter_async_context(
                streamablehttp_client(server_url, headers=headers)
            )
            self.read, self.write, _ = transport
            self.session = await self.exit_stack.enter_async_context(
                ClientSession(self.read, self.write)
            )

            server = await self.session.initialize()
            self.name = server.serverInfo.name  # <====
```

ClientSession does the following:
```python
class ClientSession(
    BaseSession[
        types.ClientRequest,
        types.ClientNotification,
        types.ClientResult,
        types.ServerRequest,
        types.ServerNotification,
    ]
):
    def __init__(
        self,
        read_stream: MemoryObjectReceiveStream[SessionMessage | Exception],
        # ...
        client_info: types.Implementation | None = None,
    ) -> None:
        super().__init__(...)
        self._client_info = client_info or DEFAULT_CLIENT_INFO # <<<<<<
       # ...

    async def initialize(self) -> types.InitializeResult:
        # ...
        result = await self.send_request(
            types.ClientRequest(
                types.InitializeRequest(
                    method="initialize",
                    params=types.InitializeRequestParams(
                        protocolVersion=types.LATEST_PROTOCOL_VERSION,
                        capabilities=types.ClientCapabilities(
                            sampling=sampling,
                            experimental=None,
                            roots=roots,
                        ),
                        clientInfo=self._client_info,
                    ),
                )
            ),
            types.InitializeResult,
        )
        # ...
        return result
```

The result is returned but ClientSession._client_info is not updated with the values coming from the result. The simplest thing would just be to set _client_info to serverInfo.

### References

_No response_

kxbnb:
**Summary**
Stores `server_info` from the init response on `ClientSession` so you can grab the server's name and version without manually extracting from the result.

**Changes**
- Added `_server_info` field to `ClientSession.__init__`
- Store `result.server_info` during `initialize()`
- Added `get_server_info()` getter (mirrors existing `get_server_capabilities()`)
- Added test for the new method

**Before**
```python
server = await session.initialize()
name = server.serverInfo.name  # had to grab manually
```

**After**
```python
await session.initialize()
info = session.get_server_info()
name = info.name  # available directly
```

**Testing**
- All existing tests pass
- Added `test_get_server_info` covering None-before-init and populated-after

Closes #1018

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -133,6 +133,7 @@ class ClientSession(
         self._message_handler = message_handler or _default_message_handler
         self._tool_output_schemas: dict[str, dict[str, Any] | None] = {}
         self._server_capabilities: types.ServerCapabilities | None = None
+        self._server_info: types.Implementation | None = None
         self._experimental_features: ExperimentalClientFeatures | None = None
 
         # Experimental: Task handlers (use defaults if not provided)
@@ -190,6 +191,7 @@ class ClientSession(
             raise RuntimeError(f"Unsupported protocol version from the server: {result.protocol_version}")
 
         self._server_capabilities = result.capabilities
+        self._server_info = result.server_info
 
         await self.send_notification(types.InitializedNotification())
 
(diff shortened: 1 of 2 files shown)


kxbnb:
Closing as superseded by #1954 which includes a test.

PR #1937, opened 2026-01-23
Proper handle extra data in MCP objects

Kludex:
Now properly matching the specification.

### DIFF
--- a/src/mcp/types/_types.py
+++ b/src/mcp/types/_types.py
@@ -30,7 +30,7 @@ TASK_REQUIRED: Final[Literal["required"]] = "required"
 
 
 class MCPModel(BaseModel):
-    """Base class for all MCP protocol types. Allows extra fields for forward compatibility."""
+    """Base class for all MCP protocol types."""
 
     model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
 


src/mcp/types/_types.py
```diff
@@ -472,10 +471,13 @@ class GetTaskPayloadRequest(Request[GetTaskPayloadRequestParams, Literal["tasks/
 
 class GetTaskPayloadResult(Result):
     """The response to a tasks/result request.
+
     The structure matches the result type of the original request.
     For example, a tools/call task would return the CallToolResult structure.
     """
 
+    model_config = ConfigDict(extra="allow")
```
Kludex: This matches the `additionalProperties` in the specification.

src/mcp/types/_types.py
```diff
@@ -32,8 +32,7 @@
 class MCPModel(BaseModel):
     """Base class for all MCP protocol types. Allows extra fields for forward compatibility."""
 
-    # TODO(Marcelo): The extra="allow" should be only on specific types e.g. `Meta`, not on the base class.
-    model_config = ConfigDict(extra="allow", alias_generator=to_camel, populate_by_name=True)
+    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
```
Kludex, line 35: This is now properly done.

pyproject.toml
```diff
@@ -26,8 +26,7 @@ dependencies = [
     "anyio>=4.5",
     "httpx>=0.27.1",
     "httpx-sse>=0.4",
-    "pydantic>=2.12.0; python_version >= '3.14'",
-    "pydantic>=2.11.0; python_version < '3.14'",
+    "pydantic>=2.12.0",
```
Kludex, line 29: We need to bump it for this to work because the `extra="allow"` behavior doesn't work properly in `TypedDict`s on previous versions. Given that we are forced to do it on 3.14, I think it makes sense to change this to have a clean state here.

src/mcp/types/_types.py
Kludex:
```suggestion
    model_config = ConfigDict(extra="allow", alias_generator=to_camel, populate_by_name=True)
```

Kludex:
Ah, now I understand what the comment about: `Allows extra fields for forward compatibility.` means. But that doesn't matter, because if the field is not being handled, it can be just ignored. Isn't that the case?

Kludex:
> Ah, now I understand what the comment about: `Allows extra fields for forward compatibility.` means. But that doesn't matter, because if the field is not being handled, it can be just ignored. Isn't that the case?

Answering my own question: I checked the C# SDK, and they just ignore the fields in older SDK versions, which makes completely sense since you can't handle those fields - but still doesn't error, which is what we want here!

src/mcp/types/_types.py
maxisbey: (quick comment before I review) just to link related things, there's ongoing discussions around this here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2084

maxisbey: Ideally we should have it just ignore extra fields. Wouldn't this cause errors if a newer server sends back extra stuff in an object?

Kludex: This is ignoring extra fields now. That's what this PR is doing.

Kludex: There are 3 options: extra being `allow`, `ignore` or `forbid`. By default it's `ignore` - which is now.

maxisbey: ohh I thought default was forbid

maxisbey:
> > Ah, now I understand what the comment about: `Allows extra fields for forward compatibility.` means. But that doesn't matter, because if the field is not being handled, it can be just ignored. Isn't that the case?
> 
> Answering my own question: I checked the C# SDK, and they just ignore the fields in older SDK versions, which makes completely sense since you can't handle those fields - but still doesn't error, which is what we want here!

this follows the typescript approach as well, so sounds good to me too

src/mcp/types/_types.py
felixweinberger: Cool!

PR #1938, opened 2026-01-23
fix: strip trailing slashes from OAuth metadata URL fields

### LINKED ISSUE #1265: Trailing slash in `.well-known/oauth-protected-resource` response may violate “Canonical Server URI” requirement

### LINKED ISSUE #1919: Trailing slash in OAuthMetadata's `issuer` causes issues with clients

maxisbey:
## Summary

Pydantic's `AnyHttpUrl` automatically appends a trailing slash to bare hostnames (e.g., `http://localhost:8000` becomes `http://localhost:8000/`). This causes OAuth metadata discovery to fail in clients that validate per RFC 8414 §3.3 and RFC 9728 §3, which require the returned `issuer`/`resource` URL to be identical to the URL used for discovery.

This broke interop with Google ADK and IBM's MCP Context Forge, which correctly perform this identity check.

## Changes

Add `field_serializer` to strip trailing slashes during JSON serialization for:
- `OAuthMetadata.issuer` (RFC 8414 §3.3)
- `ProtectedResourceMetadata.resource` (RFC 9728 §3)
- `ProtectedResourceMetadata.authorization_servers` (RFC 9728 §3)

The fix is at the serialization layer so the internal `AnyHttpUrl` representation is unchanged, but the JSON responses no longer include spurious trailing slashes.

Fixes #1919
Fixes #1265

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -130,8 +130,8 @@ class OAuthMetadata(BaseModel):
     client_id_metadata_document_supported: bool | None = None
 
     @field_serializer("issuer")
-    @classmethod
-    def _serialize_issuer(cls, v: AnyHttpUrl) -> str:
+    @staticmethod
+    def _serialize_issuer(v: AnyHttpUrl) -> str:
         """Strip trailing slash added by AnyHttpUrl for RFC 8414 §3.3 compliance."""
         return str(v).rstrip("/")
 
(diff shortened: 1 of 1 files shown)


src/mcp/shared/auth.py
```diff
@@ -129,6 +129,12 @@ class OAuthMetadata(BaseModel):
     code_challenge_methods_supported: list[str] | None = None
     client_id_metadata_document_supported: bool | None = None
 
+    @field_serializer("issuer")
+    @staticmethod
+    def _serialize_issuer(v: AnyHttpUrl) -> str:
+        """Strip trailing slash added by AnyHttpUrl for RFC 8414 §3.3 compliance."""
+        return str(v).rstrip("/")
```
Kludex, line 136: Can we just use `str` instead of `AnyHttpUrl` in the field?

PR #1939, opened 2026-01-23
fix: handle EndOfStream and ClosedResourceError in send_request

### LINKED ISSUE #1717: UnboundLocalError in send_request when response_or_error is not assigned
## Description

We received a report of intermittent `UnboundLocalError` exceptions occurring in `send_request()`:

```
UnboundLocalError: cannot access local variable 'response_or_error' where it is not associated with a value
```

The error originates from `mcp/shared/session.py` at the line that checks `isinstance(response_or_error, JSONRPCError)`.

## Code Structure

The current code structure in `send_request()` is:

```python
try:
except TimeoutError:
    raise McpError(...)

# response_or_error is used here, outside the inner try/except
if isinstance(response_or_error, JSONRPCError):
    raise McpError(response_or_error.error)
```

## Analysis

For `UnboundLocalError` to occur, execution must reach the `isinstance` check without `response_or_error` ever being assigned. Under normal Python semantics this should be impossible:

- If `receive()` completes → variable is assigned
- If `receive()` raises `TimeoutError` → caught, `McpError` raised, never reaches the check
- If `receive()` raises other exceptions → propagates up, never reaches the check

### 1. anyio `fail_after` race condition

There was a documented race condition in anyio ([agronholm/anyio#589](https://github.com/agronholm/anyio/issues/589)) where `fail_after()` could incorrectly suppress exceptions. The scenario:

1. `receive()` is executing
2. Timeout deadline is exceeded at the exact moment
3. Cancellation exception is raised but incorrectly suppressed by the cancel scope
4. Execution continues as if the try block succeeded
5. But the assignment never completed

This was reportedly fixed in anyio 4.0 via [PR #591](https://github.com/agronholm/anyio/pull/591), but there may be remaining edge cases.

### 2. Unhandled exceptions from `receive()`

`MemoryObjectReceiveStream.receive()` can raise:
- `EndOfStream` - if stream closed from sender side
- `ClosedResourceError` - if stream explicitly closed

Neither is caught by `except TimeoutError`, so they would propagate. However, this should prevent reaching the `isinstance` check, not cause `UnboundLocalError`.

### 3. Exception group interaction (Python 3.11+)

With task groups, exceptions can be wrapped in `ExceptionGroup`. The `except TimeoutError` clause does not catch `ExceptionGroup` containing `TimeoutError`. However, this still shouldn't allow execution to continue to the `isinstance` check.

## Environment

- Reported on: anyio >= 4.5
- Python: 3.10+

kxbnb:
**Summary**

Fixes #1717

Adds handling for `EndOfStream` and `ClosedResourceError` in `send_request()`. These exceptions can be raised by `response_stream_reader.receive()` but weren't being caught, leading to `UnboundLocalError` at the subsequent `isinstance` check.

**Changes**

- Added `except (anyio.EndOfStream, anyio.ClosedResourceError)` block after the existing `TimeoutError` handler
- Converts these exceptions to `McpError` with `CONNECTION_CLOSED` error code

**Validation**

- `uv run pyright` passes with 0 errors
- All 758 tests pass

**Testing**

```
tests/shared/test_session.py::test_connection_closed PASSED
```

Closes #1717

### DIFF
--- a/tests/shared/test_session.py
+++ b/tests/shared/test_session.py
@@ -304,3 +304,59 @@ async def test_connection_closed():
                 await ev_closed.wait()
             with anyio.fail_after(1):  # pragma: no branch
                 await ev_response.wait()
+
+
+@pytest.mark.anyio
+async def test_response_stream_closed_raises_mcp_error():
+    """Test that EndOfStream on the per-request response stream raises MCPError.
+
+    Reproduces the race from #1717: if the per-request response stream is closed
+    (e.g. receive loop calls aclose() during shutdown) before send_request reads
+    from it, receive() raises EndOfStream. Without the fix, this propagates as an
+    unhandled EndOfStream (or causes UnboundLocalError).
+
+    Simulates this by closing the response stream's send side directly while the
+    server connection stays open (so the receive loop never enters its finally block).
+    """
+
+    ev_result = anyio.Event()
+    caught_error: list[MCPError] = []
+
+    async with create_client_server_memory_streams() as (client_streams, server_streams):
+        client_read, client_write = client_streams
+        server_read, _server_write = server_streams
+
+        async def make_request(client_session: ClientSession):
+            nonlocal caught_error
+            try:
+                await client_session.send_ping()
+                pytest.fail("Expected MCPError")  # pragma: no cover
+            except MCPError as e:
+                caught_error.append(e)
+                ev_result.set()
+
+        async def close_response_stream(client_session: ClientSession):
+            # Consume the request so the client's send completes
+            await server_read.receive()
+
+            # Wait for send_request to register its response stream
+            while not client_session._response_streams:  # pragma: no branch
+                await anyio.sleep(0.01)  # pragma: no cover
+
+            # Close the send side directly, bypassing the receive loop's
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for the fix! Before we merge, could you:

1. Share a reproduction scenario that triggers this in practice
2. Add a test that fails on `main` but passes with your fix

We want to make sure this is a real-world issue, not just a theoretical code path.

kxbnb:
Rebased on upstream/main and added a regression test.

**Reproduction**: The race happens when the per-request response stream's send side is closed (e.g. receive loop's finally block calls `aclose()`) before `send_request` reads from it. `response_stream_reader.receive()` raises `EndOfStream`, which wasn't caught, so `response_or_error` is never assigned and the `isinstance` check on the next line hits `UnboundLocalError`.

In practice this shows up when the server disconnects abruptly and the receive loop's `finally` block and `send_request` race on the response stream — the loop closes the stream before or without sending the error object.

**Test**: `test_response_stream_closed_raises_mcp_error` reproduces this by closing the response stream's send side directly while the server connection stays open (so the receive loop never enters its finally block). Fails on `main` with unhandled `EndOfStream`, passes with the fix.

PR #1942, opened 2026-01-23
refactor: move stdio package to its own module

Kludex:


src/mcp/client/client.py
```diff
@@ -66,7 +66,10 @@ def __init__(
         self,
         server: Server[Any] | FastMCP,
         *,
-        # TODO(Marcelo): When do `raise_exceptions=True` actually raises?
+        # Note: raise_exceptions only affects transport/stream errors and exceptions
+        # that escape request handlers. Tool exceptions in FastMCP are caught by
+        # the call_tool decorator and converted to error results. See test_client.py
+        # test_raise_exceptions_returns_error_result_for_tool_exceptions for details.
```
Kludex:
```suggestion
        # TODO(Marcelo): When do `raise_exceptions=True` actually raises?
```

Issue #1941, opened 2026-01-23
streamable_http client hangs indefinitely when connecting to POST-only MCP servers (ex: GitHub MCP)


sicoyle:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

The `streamable_http_client` transport hangs indefinitely when connecting to MCP servers that only support POST requests (not GET), such as GitHub's MCP server at `https://api.githubcopilot.com/mcp/`. The issue manifests when calling operations like `list_tools()` or other methods that return SSE responses. This is also likely what some of the other issues are experiencing as there are many hanging related issues:
https://github.com/modelcontextprotocol/python-sdk/issues?q=is%3Aissue%20state%3Aopen%20hang however, I believe this could explain this issue in more detail potentially: https://github.com/modelcontextprotocol/python-sdk/issues/262

I tried with and without a debugger, and noticed that with a debugger attached, the timing overhead masks the race condition and operations complete successfully. So this has been difficult to unmask.

The MCP SDK's `streamable_http_client` has a race condition in its GET stream handling:

1. After initialization, `streamable_http_client` launches a background task (`handle_get_stream`) to receive server-initiated SSE messages via GET requests
2. GitHub's MCP server doesn't support GET requests and returns `405 Method Not Allowed`
3. The GET stream retries twice according to `MAX_RECONNECTION_ATTEMPTS` (with 1-second delays each), then the background task silently exits
4. When subsequent POST requests return `text/event-stream` content (like `tools/list`), the transport's SSE response handling may wait for coordination with the GET stream task
5.  Since the GET stream task is dead, the POST SSE response waits indefinitely

Full repro repo code that honestly isn't large: https://github.com/sicoyle/release-note-agent/blob/main/release_note_agent/release_note_agent.py#L12

### Example Code

```Python
Full repro repo code that honestly isn't large with 2 cmds to run:
https://github.com/sicoyle/release-note-agent/blob/main/release_note_agent/release_note_agent.py#L12
```

### Python & MCP Python SDK

```Text
i forked and tested my changes on an incoming PR to confirm the error and fix. Stay tuned for PR shortly
py 3.13
mcp python sdk at main
```

maxisbey:
[...]

PR #1943, opened 2026-01-23
fix: prevent silent failure on GET stream so clients don't hang

sicoyle:
When connecting to an MCP server over streamable HTTP (ex: GitHub’s MCP server) from Dapr Agents, the client could hang indefinitely on calls like list_tools() if run outside a debugger; with a debugger attached, the timing often hid the issue and pointed to a race.

Root cause: The streamable_http client starts a background GET stream for server‑sent events. If that GET fails (ex: 405 Method Not Allowed), the client retries twice with a 1s delay and then gives up without surfacing that failure. The GET task exits while the rest of the client still assumes the stream may be used. Subsequent requests that expect or coordinate with that stream then block forever on a dead stream.

Fix: For permanent GET failures such as 405, we no longer retry and we log a clear warning instead of failing silently. The client continues to use POST for request/response (e.g. list_tools()), so it no longer hangs when the GET stream has been abandoned. This addresses the silent-failure behavior; per the spec, servers are expected to support both POST and GET, but when the GET stream cannot be established we now fail visibly and avoid leaving the client in a hung state.

~~This is a fix for an agent hang that I saw in Dapr Agents when it was trying to connect to the Github MCP server using streamable HTTP Transport. I only saw the issue if I ran it locally in terminal, but it was fine within the scope of a debugger. That led me down the path of a race condition. The MCP streamable_http client starts a background GET stream for SSE events that immediately get a 405 err from Github (which doesn't support GET), retries twice with the 1sec delay, and then gives up failing silently. If I called to list tools after the GET stream exhausted it's retries, it would hang forever waiting for the SSE responses on the dead stream. This PR prevents this hang by handling the 405 error.~~

## Motivation and Context
I'm associated with the Dapr open source project (CNCF graduated project), and am a maintainer for Dapr Agents repo. We are experiencing issues that this PR helps to correct for our open source community.

https://github.com/dapr/dapr-agents/issues/399

## How Has This Been Tested?
I tested and confirmed this works against my repo for a release note agent that we want to eventually publicize and make into a github action that others can consume for open source project release notes when it's ready:
https://github.com/sicoyle/release-note-agent

I can confirm locally that this corrects at least my hanging issue.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
related issues:
https://github.com/modelcontextprotocol/python-sdk/issues/1941
https://github.com/dapr/dapr-agents/issues/399


Kludex:
In the spec, it says:

> The server MUST provide a single HTTP endpoint path (hereafter referred to as the MCP endpoint) that supports both POST and GET methods. [^1]

It seems the issue here is more the "then gives up failing silently." than the method not supported.

[^1]: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

SamMorrowDrums:
I responded on the issue for GitHub MCP but the spec is explicit about the GET support being optional: https://github.com/github/github-mcp-server/issues/1951#issuecomment-3843469941

sicoyle:
does claude build step usually take 2hrs?

maxisbey:
Thanks for the PR. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #1948, opened 2026-01-24
fix: support wildcard media types in Accept header validation

### LINKED ISSUE #1641: MCP Server won't work with wildcard in "Accept" header and therefore is non‑compliant with HTTP spec
### Use Case
My client sends a request to MCP Server with `Accept` header `*/*` (Same for `application/*, text/event-stream`). Server rejects this request because it expected strictly `application/json` or `application/json and text/event-stream` if  `json_response=False`.  

### Expected
According to the HTTP/1.1 spec (RFC 7231, section 5.3.2), the Accept header allows wildcard so values `*/*` and `application/*` should be valid, therefore server should accept them

```

...

mcp_server: FastMCP = FastMCP(streamable_http_path="/", stateless_http=True, json_response=True)

...

app: Starlette = mcp_server.streamable_http_app()
```
Start server: `uvicorn mcp-server:app --host 127.0.0.1 --port 8000 --log-level trace`

### Example of Call
```
### Request
POST / HTTP/1.1.
Host: localhost:8000.
Accept-Encoding: gzip, deflate, br, zstd.
Connection: keep-alive.
User-Agent: python-httpx/0.28.1.
Content-Type: application/json.
Accept: */*.
Content-Length: 159.
.

{"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"mcp-client","version":"1.0.0"}}}


### Response
HTTP/1.1 406 Not Acceptable.
date: Tue, 18 Nov 2025 16:16:17 GMT.
server: uvicorn.
content-type: application/json.
content-length: 125.
.

T 127.0.0.1:8000 -> 127.0.0.1:61346 [AP] #107
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Not Acceptable: Client must accept application/json"}}
```

### Additional Context

- mcp 1.18.0
- Python 3.11.13
- uvicorn 0.38.0
- starlette 0.49.1

skyvanguard:
## Summary

Fixes #1641

The MCP server was rejecting HTTP requests that use wildcard Accept headers like `*/*` or `application/*`, returning `406 Not Acceptable`. This is incorrect behavior per RFC 9110 Section 12.5.1, which defines wildcard media types as valid content negotiation mechanisms.

Many HTTP clients (including browsers, curl, and httpx) send `Accept: */*` by default, meaning the server was incompatible with standard HTTP tooling.

## Changes

Modified `_check_accept_headers()` in `StreamableHTTPServerTransport` to:

1. **Support wildcard media types**: `*/*` now matches any required content type, `application/*` matches `application/json`, and `text/*` matches `text/event-stream`
2. **Strip quality parameters**: Parameters like `;q=0.9` are now stripped before matching, so `*/*;q=0.8` is handled correctly

## Behavior

| Accept Header | JSON mode | SSE mode |
|---|---|---|
| `*/*` | 200 | 200 |
| `application/*` | 200 | 406 (missing SSE) |
| `text/*` | 406 (missing JSON) | 406 (missing JSON) |
| `application/json, text/*` | 200 | 200 |
| `*/*;q=0.8` | 200 | 200 |
| `text/plain` | 406 | 406 |

## Test plan

Added 7 regression tests in `tests/issues/test_1641_accept_header_wildcard.py` covering:
- `*/*` in both JSON and SSE response modes
- `application/*` matching `application/json`
- `text/*` combined with explicit `application/json`
- Quality parameters stripped before matching
- Invalid types still correctly rejected with 406
- Partial wildcards correctly rejected when not covering all required types

### DIFF
--- a/tests/issues/test_1641_accept_header_wildcard.py
+++ b/tests/issues/test_1641_accept_header_wildcard.py
@@ -3,188 +3,75 @@
 The MCP server was rejecting requests with wildcard Accept headers like `*/*`
 or `application/*`, returning 406 Not Acceptable. Per RFC 9110 Section 12.5.1,
 wildcard media types are valid and should match the required content types.
+
+These tests verify the `_check_accept_headers` method directly, ensuring
+wildcard media types are properly matched against the required content types
+(application/json and text/event-stream).
 """
 
-import threading
-from collections.abc import AsyncGenerator
-from contextlib import asynccontextmanager
-
-import anyio
-import httpx
 import pytest
-from starlette.applications import Starlette
-from starlette.routing import Mount
+from starlette.requests import Request
 
-from mcp.server import Server
-from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
-from mcp.types import Tool
-
-SERVER_NAME = "test_accept_wildcard_server"
-
-# Suppress warnings from unclosed MemoryObjectReceiveStream in stateless transport mode
-# (pre-existing issue, not related to the Accept header fix)
-pytestmark = [
-    pytest.mark.filterwarnings("ignore::ResourceWarning"),
-    pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning"),
-]
-
-INIT_REQUEST = {
-    "jsonrpc": "2.0",
-    "method": "initialize",
-    "id": "init-1",
-    "params": {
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for the PR! The linked issue (#1641) has since been resolved, so this change is no longer needed. Closing.

PR #1949, opened 2026-01-24
fix: send error to client when SSE stream disconnects without response

### LINKED ISSUE #1811: client's `read_stream_writer` open after SSE disconnection hanging `.receive()`

skyvanguard:
## Summary

Fixes #1811

When the SSE stream disconnects before the server sends a response (e.g., due to a read timeout on long-running tool calls), and no event ID has been received (making reconnection impossible), the client was left hanging indefinitely on `read_stream.receive()`. The `post_writer` task had already returned from `_handle_sse_response()` without sending anything to `read_stream_writer`, leaving the session layer permanently blocked.

## Changes

Added `_send_disconnect_error()` method to `StreamableHTTPTransport` that sends a `JSONRPCError` (code `-32000`) to the `read_stream_writer`, unblocking the client. This is called in two scenarios:

1. **`_handle_sse_response()`**: When the SSE stream ends without receiving a response and `last_event_id` is `None` (no event ID to reconnect with)
2. **`_handle_reconnection()`**: When maximum reconnection attempts are exhausted without receiving a response

The error surfaces as an `McpError` at the session layer, allowing the caller to handle the failure gracefully instead of hanging forever.

## Root Cause

The issue was introduced when reconnection support was added. The reconnection logic correctly handles the case where `last_event_id is not None` (reconnect using Last-Event-ID), but the `else` branch (no event ID available) was missing, causing a silent return that left the client blocked.

## Test plan

Added 2 regression tests in `tests/issues/test_1811_sse_disconnect_hang.py`:
- **`test_client_receives_error_on_sse_disconnect`**: Server has a 30s tool, client has 0.5s read timeout. Verifies the client receives `McpError` with "SSE stream disconnected" message instead of hanging.
- **`test_fast_tool_still_works_normally`**: Ensures fast tool calls continue to work correctly with the fix in place.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -337,9 +337,13 @@ class StreamableHTTPTransport:
             logger.debug(f"SSE stream ended: {e}")  # pragma: no cover
 
         # Stream ended without response - reconnect if we received an event with ID
-        if last_event_id is not None:  # pragma: no branch
+        if last_event_id is not None:
             logger.info("SSE stream disconnected, reconnecting...")
             await self._handle_reconnection(ctx, last_event_id, retry_interval_ms)
+        else:
+            # No event ID received before disconnect - cannot reconnect,
+            # send error to unblock the client
+            await self._send_disconnect_error(ctx)
 
     async def _handle_reconnection(
         self,
@@ -350,8 +354,9 @@ class StreamableHTTPTransport:
     ) -> None:
         """Reconnect with Last-Event-ID to resume stream after server disconnect."""
         # Bail if max retries exceeded
-        if attempt >= MAX_RECONNECTION_ATTEMPTS:  # pragma: no cover
+        if attempt >= MAX_RECONNECTION_ATTEMPTS:
             logger.debug(f"Max reconnection attempts ({MAX_RECONNECTION_ATTEMPTS}) exceeded")
+            await self._send_disconnect_error(ctx)
             return
 
         # Always wait - use server value or default
@@ -416,6 +421,17 @@ class StreamableHTTPTransport:
         )
         session_message = SessionMessage(jsonrpc_error)
         await read_stream_writer.send(session_message)
+
+    async def _send_disconnect_error(self, ctx: RequestContext) -> None:
+        """Send a disconnect error to unblock the client waiting on the read stream."""
+        if isinstance(ctx.session_message.message, JSONRPCRequest):
+            request_id = ctx.session_message.message.id
+            jsonrpc_error = JSONRPCError(
+                jsonrpc="2.0",
+                id=request_id,
+                error=ErrorData(code=-32000, message="SSE stream disconnected before receiving a response"),
+            )
+            await ctx.read_stream_writer.send(SessionMessage(jsonrpc_error))
 
     async def post_writer(
         self,
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -334,7 +334,7 @@ class StreamableHTTPTransport:
                     await response.aclose()
                     return  # Normal completion, no reconnect needed
         except Exception as e:
-            logger.debug(f"SSE stream ended: {e}")  # pragma: no cover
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2122, which covers the same code path plus the `attempt=0` reset bug. Thanks for the contribution.

PR #1956, opened 2026-01-25
refactor: `McpError` renamed to `MCPError` and flatten parameters

Kludex:


docs/migration.md
```diff
@@ -121,6 +121,38 @@ result = await session.list_resources(params=PaginatedRequestParams(cursor="next
 result = await session.list_tools(params=PaginatedRequestParams(cursor="next_page_token"))
 ```
 
+### `McpError` renamed to `MCPError`
+
+The `McpError` exception class has been renamed to `MCPError` for consistent naming with the MCP acronym style used throughout the SDK.
```
maxisbey, line 126: I think this makes sense to be consistent. Although it does go against the google style guide: https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

but since we're not strictly following that then it's not really an issue :)

src/mcp/server/experimental/request_context.py
```diff
@@ -72,32 +72,26 @@ def validate_task_mode(
...
         elif mode == TASK_FORBIDDEN and self.is_task:
-            error = ErrorData(
-                code=METHOD_NOT_FOUND,
-                message="This tool does not support task-augmented invocation",
-            )
+            error = ErrorData(code=METHOD_NOT_FOUND, message="This tool does not support task-augmented invocation")
 
         if error is not None and raise_error:
-            raise McpError(error)
+            raise MCPError(code=METHOD_NOT_FOUND, message=error.message)
```
maxisbey:
```suggestion
            raise MCPError(code=error.code, message=error.message)
```

docs/migration.md
```diff
@@ -121,6 +121,38 @@ result = await session.list_resources(params=PaginatedRequestParams(cursor="next
...
+
+**Before (v1):**
+
+```python
+from mcp.shared.exceptions import McpError
+
+try:
+    result = await session.call_tool("my_tool")
+except McpError as e:
+    print(f"Error: {e.message}")
```
maxisbey: did `e.message` exist before? seems like it's a new property

docs/migration.md
Kludex: Where are you seeing the lack of consistency? MCP is an acronymn, so that's why it would be uppercased.

maxisbey: Oh no, I was agreeing with you that it makes sense for consistency, as you wrote "The `McpError` exception class has been renamed to `MCPError` for **consistent** naming".

I was just linking the google style guide saying to use CapWords which I thiiink also apply to acronyms? but I don't think we should follow that anyway

docs/migration.md
Kludex:
```suggestion
    print(f"Error: {e.error.message}")
```

PR #1968, opened 2026-01-28
sketch: refactor lowlevel server to handler pattern

maxisbey:
Replaces decorator-based handler registration with `RequestHandler`/`NotificationHandler` classes.

## Changes

- **New files**: `request_handler.py` and `notification_handler.py` with typed overloaded constructors
- **Server dispatch**: `type(req)` → `req.method` (string-based dispatch)
- **Handler dicts**: `dict[type, Callable]` → `dict[str, RequestHandler]` / `dict[str, NotificationHandler]`  
- **Registration**: Decorator methods → constructor `handlers` list + `add_handler()` method
- **Context**: `RequestContext` built and passed directly to handlers; `request_ctx` contextvar removed
- **Removed**: All 12 decorator methods, tool cache, `_make_error_result`, `_get_cached_tool_definition`, `new_server.py` sketch

## Known breakages

- `MCPServer._setup_handlers()` — calls removed decorator methods
- `ExperimentalHandlers` — writes to handler dicts with type keys
- All lowlevel server tests — use decorator pattern
- All examples — use decorator pattern

### DIFF
--- a/docs/experimental/index.md
+++ b/docs/experimental/index.md
@@ -27,13 +27,9 @@ Tasks are useful for:
 Experimental features are accessed via the `.experimental` property:
 
 ```python
-# Server-side: register a custom task handler
-server = Server(
-    name="my-server",
-    handlers=[
-        RequestHandler("tasks/get", handler=handle_get_task),
-    ],
-)
+# Server-side: enable task support (auto-registers default handlers)
+server = Server(name="my-server")
+server.experimental.enable_tasks()
 
 # Client-side
 result = await session.experimental.call_tool_as_task("tool_name", {"arg": "value"})
--- a/src/mcp/server/lowlevel/__init__.py
+++ b/src/mcp/server/lowlevel/__init__.py
@@ -1,4 +1,4 @@
 from .handler import Handler, NotificationHandler, RequestHandler
-from .server import NotificationOptions, Server, request_ctx
+from .server import NotificationOptions, Server
 
 __all__ = ["Handler", "NotificationHandler", "NotificationOptions", "RequestHandler", "Server"]


src/mcp/server/lowlevel/notification_handler.py
```diff
@@ -0,0 +1,69 @@
...
+    @overload
+    def __init__(
+        self,
+        method: str,
+        handler: Callable[..., Awaitable[None]],
+    ) -> None: ...
+
+    def __init__(self, method: str, handler: Callable[..., Awaitable[None]]) -> None:
+        self.method = method
+        self.endpoint = handler
```
Kludex:
```suggestion
        self.handler = handler
```

maxisbey: done

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+    CallToolRequestParams,
+    CallToolResult,
+    ListToolsResult,
+    PaginatedRequestParams,
+    TextContent,
+    Tool,
+)
+
+async def handle_list_tools(
+    ctx: RequestHandlerContext, params: PaginatedRequestParams | None
```
Kludex: No None!
```suggestion
    ctx: RequestHandlerContext, params: PaginatedRequestParams
```

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+)
+```
+
+**Key differences:**
+
+- Handlers receive `(ctx, params)` instead of the full request object or unpacked arguments. `ctx` is a `RequestHandlerContext` (for requests) or `NotificationHandlerContext` (for notifications) with `session`, `lifespan_context`, and `experimental` fields. `params` is the typed request params object.
+- Handlers return the full result type (e.g. `ListToolsResult`) rather than unwrapped values (e.g. `list[Tool]`).
+- Registration uses method strings (`"tools/call"`) instead of request types (`CallToolRequest`).
+- Handlers can be added after construction with `server.add_handler()` (silently replaces existing handlers for the same method).
+- `server.has_handler(method)` checks if a handler is registered for a given method string.
```
Kludex: Why do you need to expose this?

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+    ],
+)
+```
+
+**Key differences:**
+
+- Handlers receive `(ctx, params)` instead of the full request object or unpacked arguments. `ctx` is a `RequestHandlerContext` (for requests) or `NotificationHandlerContext` (for notifications) with `session`, `lifespan_context`, and `experimental` fields. `params` is the typed request params object.
+- Handlers return the full result type (e.g. `ListToolsResult`) rather than unwrapped values (e.g. `list[Tool]`).
+- Registration uses method strings (`"tools/call"`) instead of request types (`CallToolRequest`).
+- Handlers can be added after construction with `server.add_handler()` (silently replaces existing handlers for the same method).
```
Kludex: Is the need for this on the whole server, or per session? Can you give me a real case scenario that people actually do this?

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+    handlers=[
+        RequestHandler("tools/list", handler=handle_list_tools),
+        RequestHandler("tools/call", handler=handle_call_tool),
+    ],
+)
+```
+
+**Key differences:**
+
+- Handlers receive `(ctx, params)` instead of the full request object or unpacked arguments. `ctx` is a `RequestHandlerContext` (for requests) or `NotificationHandlerContext` (for notifications) with `session`, `lifespan_context`, and `experimental` fields. `params` is the typed request params object.
```
Kludex: Why do we need 2 different context objects?

Kludex: And what is `experimental` for?

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+```
+
+**After (v2):**
+
+```python
+from mcp.shared.context import RequestHandlerContext
+from mcp.types import CallToolRequestParams, CallToolResult, TextContent
+
+async def handle_call_tool(
+    ctx: RequestHandlerContext, params: CallToolRequestParams
```
Kludex: I don't think we need to change the name of the class tho. Can't we use `RequestContext` or just `Context` here?

docs/migration.md
```diff
@@ -406,23 +406,167 @@ params = CallToolRequestParams(
...
+    )
+```
+
+### `RequestContext` split into `HandlerContext`, `RequestHandlerContext`, `NotificationHandlerContext`
+
+The `RequestContext` class in `mcp.shared.context` has been replaced with a three-class hierarchy:
+
+- `HandlerContext` — base class with `session`, `lifespan_context`, `experimental`
+- `RequestHandlerContext(HandlerContext)` — adds `request_id`, `meta`, `request`, `close_sse_stream`, `close_standalone_sse_stream`
+- `NotificationHandlerContext(HandlerContext)` — empty subclass for notifications
```
Kludex: The experience would be better with a single object, and with exceptions if accessed out of the proper context.

src/mcp/server/lowlevel/experimental.py
```diff
@@ -50,13 +49,11 @@ class ExperimentalHandlers:
...
-        notification_handlers: dict[type, Callable[..., Awaitable[None]]],
-    ):
-        self._server = server
-        self._request_handlers = request_handlers
-        self._notification_handlers = notification_handlers
+        add_handler: Callable[[RequestHandler[Any, Any] | NotificationHandler[Any, Any]], None],
+        has_handler: Callable[[str], bool],
+    ) -> None:
+        self._add_handler = add_handler
+        self._has_handler = has_handler
```
Kludex: This is ugly. We can improve it. What are you trying to achieve here?

src/mcp/server/lowlevel/server.py
```diff
@@ -145,10 +104,15 @@ async def lifespan(_: Server[LifespanResultT, RequestT]) -> AsyncIterator[dict[s
     yield {}
 
 
+async def _ping_handler(ctx: Any, params: Any) -> types.EmptyResult:
+    return types.EmptyResult()
```
Kludex: Keep it below please. When reading a file, it's nicer to have the public API first.

src/mcp/server/lowlevel/server.py
```diff
@@ -145,10 +104,15 @@ async def lifespan(_: Server[LifespanResultT, RequestT]) -> AsyncIterator[dict[s
...
+
+
 class Server(Generic[LifespanResultT, RequestT]):
     def __init__(
         self,
         name: str,
+        handlers: list[RequestHandler[Any, Any] | NotificationHandler[Any, Any]] | None = None,
```
Kludex: Please put `request_handler` and `notification_handler` in the same file and create a base class, I think it's important to do this because it will simplify how we do middlewares.
```suggestion
        handlers: Sequence[Handler] = (),
```

Kludex: Handler generic params already default to `Any`.

src/mcp/server/mcpserver/server.py
```diff
@@ -54,6 +76,10 @@
 
 _CallableT = TypeVar("_CallableT", bound=Callable[..., Any])
 
+_mcp_server_ctx: contextvars.ContextVar[RequestHandlerContext[ServerSession, Any, Any]] = contextvars.ContextVar(
+    "_mcp_server_ctx"
+)
```
Kludex: Why do you have this?

maxisbey: So that the way you get context when using MCPServer isn't changed (from doing app.get_context or w/e). IMO improving that is out of scope for this PR.

Kludex: Please readd the `request_ctx` context var that was removed - it has a wrapper in a single place. In here there's a huge diff just to wrap all the handlers.

Don't use the `request_ctx` in the lowlevel, and on a future PR then drop from both of them, please.

docs/migration.md
maxisbey: Why not? [ListTasksRequest has PaginatedRequestParams as optional](https://modelcontextprotocol.io/specification/2025-11-25/schema#listtasksrequest)

Kludex: Because we should be narrowing the types for the user... 😢

Optional doesn't necessarily translates to `None` in Python. But okay, resolve my comment.

docs/migration.md
maxisbey: Exceptions feel weird for property access. I was splitting it up so it can be more strictly typed since some fields are request specific.

docs/migration.md
maxisbey: Actually I remember using that exact argument for something else in the typing in the sdk to someone else haha. Looking at [PaginatedRequestParams in the spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/54111414d0a0f0864e6c98f77e015bd86dfe2931/schema/2025-11-25/schema.ts#L629) the field `cursor` is an optional string, but in python not entirely sure what we'd do there with the same logic if we can't have None haha.

Will leave it as is, but lemme know if you think of anything nicer

docs/migration.md
maxisbey: yea fair not needed

docs/migration.md
maxisbey: yea there's not, removed.

docs/migration.md
maxisbey: Combined the two context objects into one again.

`experimental` is for keeping all the Tasks related stuff out of Server

docs/migration.md
maxisbey: changed

docs/migration.md
maxisbey: I've put it back together and made them optional, lemme know if you're thinking something else

src/mcp/server/lowlevel/experimental.py
maxisbey: Basically just keeping all of the tasks related stuff out of the main Server class and not touching it as much as possible haha

src/mcp/server/lowlevel/server.py
maxisbey: moved

src/mcp/server/lowlevel/server.py
maxisbey: done

src/mcp/server/mcpserver/server.py
maxisbey: added

src/mcp/server/lowlevel/server.py
```diff
@@ -168,15 +130,42 @@ def __init__(
...
+        self._notification_handlers: dict[str, NotificationHandler] = {}
         self._experimental_handlers: ExperimentalHandlers | None = None
         self._session_manager: StreamableHTTPSessionManager | None = None
         logger.debug("Initializing server %r", name)
 
+        # Register default ping handler
+        self._request_handlers["ping"] = RequestHandler("ping", handler=_ping_handler)
+
+        # Process user-provided handlers with duplicate detection
+        if handlers:
```
Kludex: Conditional is not necessary anymore.

src/mcp/shared/context.py
```diff
@@ -16,15 +16,21 @@
 
 @dataclass
```
Kludex:
```suggestion
@dataclass(kw_only=True)
```

src/mcp/shared/context.py
```diff
@@ -16,15 +16,21 @@
...
-    experimental: Any = field(default=None)
-    request: RequestT | None = None
-    close_sse_stream: CloseSSEStreamCallback | None = None
-    close_standalone_sse_stream: CloseSSEStreamCallback | None = None
+    experimental: Any = field(default=None, kw_only=True)
+    request_id: RequestId | None = field(default=None, kw_only=True)
+    meta: RequestParamsMeta | None = field(default=None, kw_only=True)
+    request: RequestT | None = field(default=None, kw_only=True)
+    close_sse_stream: CloseSSEStreamCallback | None = field(default=None, kw_only=True)
+    close_standalone_sse_stream: CloseSSEStreamCallback | None = field(default=None, kw_only=True)
```
Kludex:
```suggestion
    experimental: Any = None  # What is the right type?
    request_id: RequestId | None = None  # Isn't the request id always present???
    meta: RequestParamsMeta | None = None
    request: RequestT | None = None  # This one is always populated, right?
    close_sse_stream: CloseSSEStreamCallback | None = None
    close_standalone_sse_stream: CloseSSEStreamCallback | None = None
```

src/mcp/server/mcpserver/server.py
```diff
@@ -487,7 +567,22 @@ async def handle_completion(ref, argument, context):
...
+                return CompleteResult(
+                    completion=result if result is not None else Completion(values=[], total=None, has_more=None),
+                )
+
+            # TODO(maxisbey): remove private access — completion needs post-construction
+            #   handler registration, find a better pattern for this
+            self._lowlevel_server._add_handler(  # pyright: ignore[reportPrivateUsage]
+                RequestHandler("completion/complete", handler=handler)
+            )
+            return func
```
Kludex, line 583: Why is that?

src/mcp/shared/context.py
maxisbey: - `experimental` yea added the correct type
- `request_id` - no, notifications don't always have a RequestId, this is why I was having two contexts, one for Requets and one for Notifications originally (but you said to move back to one class)
- `request` is actually badly named, this seems to be the HTTP Request context from the transport layer which is only included if you're using SHTTP

maxisbey:
> @Kludex: Do we need to know if a request is a request or a notification in the Server itself? Isn't it enough to know that we have the method and the handler for it?

Well we need the server to know whether or not to expect a return type to pass back? Requests give responses, Notifications don't give responses.

src/mcp/server/lowlevel/server.py
maxisbey: removed

src/mcp/shared/context.py
maxisbey: changed

src/mcp/server/mcpserver/server.py
maxisbey: On `main` the `MCPServer.completion()` just uses the lowlevel server's completion decorator, but since we don't have that anymore it has to make its own decorator. https://github.com/modelcontextprotocol/python-sdk/blob/b38716e5e3761dd2d52809d2d68ce65325faad5e/src/mcp/server/mcpserver/server.py#L490

It has to be at registration time so that when the server's capabilities are figured out it only includes completion if a completion endpoint is actually registered.

I think ideally we figure out another way for a server to define it's capabilities than how it's done now, but probably not in this PR?

docs/experimental/index.md
```diff
@@ -27,10 +27,13 @@ Tasks are useful for:
...
-@server.experimental.get_task()
-async def handle_get_task(request: GetTaskRequest) -> GetTaskResult:
-    ...
+# Server-side: register a custom task handler
+server = Server(
+    name="my-server",
+    handlers=[
+        RequestHandler("tasks/get", handler=handle_get_task),
+    ],
+)
```
Kludex: Is this what you want?

src/mcp/server/lowlevel/__init__.py
```diff
@@ -1,3 +1,4 @@
-from .server import NotificationOptions, Server
+from .handler import Handler, NotificationHandler, RequestHandler
+from .server import NotificationOptions, Server, request_ctx
 
-__all__ = ["Server", "NotificationOptions"]
+__all__ = ["Handler", "NotificationHandler", "NotificationOptions", "RequestHandler", "Server", "request_ctx"]
```
Kludex:
```suggestion
__all__ = ["Handler", "NotificationHandler", "NotificationOptions", "RequestHandler", "Server"]
```

src/mcp/server/lowlevel/experimental.py
```diff
@@ -50,13 +48,11 @@ class ExperimentalHandlers:
...
-        notification_handlers: dict[type, Callable[..., Awaitable[None]]],
-    ):
-        self._server = server
-        self._request_handlers = request_handlers
-        self._notification_handlers = notification_handlers
+        add_handler: Callable[[Handler], None],
+        has_handler: Callable[[str], bool],
+    ) -> None:
+        self._add_handler = add_handler
+        self._has_handler = has_handler
```
Kludex, line 55: What are you trying to achieve with this? If a handler doesn't exist, you add a handler?

src/mcp/shared/context.py
Kludex: This RequestContext is a bit weird that it has information related to some transports. Do you know, or can you find out where those 2 callbacks are actually called?

Please don't say "in a next PR", this is a great chance of getting a lot of non sense out.

src/mcp/server/mcpserver/server.py
Kludex: > I think ideally we figure out another way for a server to define it's capabilities than how it's done now

How would that possibly be?

src/mcp/server/lowlevel/experimental.py
maxisbey: Originally you just passed in the request/notification handler dicts in directly and this class would then add in the default task related handlers to get task augmentation working. I was trying to clean this up by adding in callbacks.

My idea when making it like this originally was to have all tasks related stuff under `server.experimental.`, so you could just call `server.experimental.enable_tasks()` and it would register the handlers onto your server. What do you think is best here? Just switch back to passing in the request/notification handler dicts directly instead of callbacks?

docs/experimental/index.md
maxisbey: nope it's not haha, fixed

maxisbey:
> > I think ideally we figure out another way for a server to define it's capabilities than how it's done now
> 
> How would that possibly be?

I was thinking because currently capabilities are calculated based on which handlers are registered, but not all capabilities can be inferred just from which method handlers exist.

For example `listChanged` booleans in `prompts.listChanged`, `resources.listChanged`, and `tools.listChanged`. There's client -> server method handler related to this, this is just server -> client communication.

`tasks.requests.tools.call` is another, you can _technically_ support the rest of the tasks methods and just not this one. Although this is isn't a realistic usecase.

`logging` is another one where the server can specify whether it supports it or not, and it can't be inferred just from whether the method handler exists.

src/mcp/shared/context.py
maxisbey: They're SSE polling callbacks for [SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699), which allows tool handlers to close connections mid request.

They're added in originally here: https://github.com/modelcontextprotocol/python-sdk/blob/5e1bf86/src/mcp/server/streamable_http.py#L225-L250

They're also exposed by MCPserver here: https://github.com/modelcontextprotocol/python-sdk/blob/5e1bf86/src/mcp/server/mcpserver/server.py#L1279-L1309

maxisbey:
Superseded by #1985 which takes this further — replaces Handler objects with direct on_* constructor kwargs.

PR #1987, opened 2026-02-03
refactor: split `RequestContext` between server and client

Kludex:
Cute.


.github/actions/conformance/client.py
```diff
@@ -187,7 +187,7 @@ async def run_sse_retry(server_url: str) -> None:
 
 
 async def default_elicitation_callback(
-    context: RequestContext[ClientSession, Any],  # noqa: ARG001
+    context: RequestContext[ClientSession],
```
Kludex, line 190: Maybe we can have a `ClientRequestContext`.

maxisbey: Yea true, would be nice to have both

maxisbey (approved):

PR #1992, opened 2026-02-04
chore: update author metadata to LF Projects

maxisbey:
Update `authors` across all pyproject.toml files to `Model Context Protocol a Series of LF Projects, LLC.`, matching the [TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk/blob/faeb03e/package.json#L7). Also removes stale `maintainers` from example pyproject.toml files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,10 +4,9 @@ dynamic = ["version"]
 description = "Model Context Protocol SDK"
 readme = "README.md"
 requires-python = ">=3.10"
-authors = [{ name = "Anthropic, PBC." }]
+authors = [{ name = "Model Context Protocol a Series of LF Projects, LLC." }]
 maintainers = [
     { name = "David Soria Parra", email = "davidsp@anthropic.com" },
-    { name = "Justin Spahr-Summers", email = "justin@anthropic.com" },
     { name = "Marcelo Trylesinski", email = "marcelotryle@gmail.com" },
     { name = "Max Isbey", email = "maxisbey@anthropic.com" },
     { name = "Felix Weinberger", email = "fweinberger@anthropic.com" }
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -4,10 +4,9 @@ dynamic = ["version"]
 description = "Model Context Protocol SDK"
 readme = "README.md"
 requires-python = ">=3.10"
-authors = [{ name = "Anthropic, PBC." }]
```
Kludex, line 7: Well, Anthropic did authored... 😅

maxisbey: Haha true... is it normal to keep previous authors as well in something like this?

maxisbey: according to claude it's not haha, so gonna merge

Kludex: Since you are asking... I'm not sure how I feel about removing a maintainer from the maintainer list (Justin). It feels like he is just forgotten from the project.

> Haha true... is it normal to keep previous authors as well in something like this?

You can... I don't think there are strict rules, but I don't think MCP org is the author of this project, it's Anthropic.

Kludex: MCP may have ownership, but author doesn't change.

Kludex:
Instead of author on the package metadata, what needs to be changed is this: https://github.com/modelcontextprotocol/python-sdk/blob/main/LICENSE

Kludex:
...and by change, I don't mean replace Anthropic by MCP org, but adding a new copyright line.

pyproject.toml
maxisbey: > Since you are asking... I'm not sure how I feel about removing a maintainer from the maintainer list (Justin). It feels like he is just forgotten from the project

Yea definitely not wanting to remove credit from someone at all.

I was talking to @dsp-ant about it and the thinking the list should mainly contain active maintainers since it has contact details. But I'm not super fussed either way.

Justin is still kept as lead maintainer here https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/MAINTAINERS.md, although that's ofc not in this repo.

PR #1993, opened 2026-02-04
add client_id_metadata_document_supported flag

martimfasantos:
## Problem

The MCP SDK's OAuth implementation currently only supports **Dynamic Client Registration (DCR)** for obtaining client credentials. While the `OAuthMetadata` schema includes the `client_id_metadata_document_supported` field (per [RFC 7591 Section 3.3](https://www.rfc-editor.org/rfc/rfc7591#section-3.3)), this field is never populated in `build_metadata()`, preventing clients from using URL-based client IDs as an alternative to DCR.

This limitation means:
- Clients cannot use predetermined, stable `client_id` values
- Every client session generates a new UUID `client_id` via dynamic registration
- No support for the Client ID Metadata Document (CIMD) pattern, where the `client_id` itself is an HTTPS URL pointing to client metadata

## Changes

This PR adds `client_id_metadata_document_supported` as an optional field in `ClientRegistrationOptions`:

```python
class ClientRegistrationOptions(BaseModel):
    enabled: bool = False
    client_secret_expiry_seconds: int | None = None
    valid_scopes: list[str] | None = None
    default_scopes: list[str] | None = None
    client_id_metadata_document_supported: bool = False # NEW
```

When enabled, the OAuth authorization server metadata will advertise support for URL-based client IDs, allowing clients to skip dynamic registration by providing an HTTPS URL as their `client_id`.

## Benefits

1. **Stable Client Identities**: Applications can use consistent `client_id` values across restarts
2. **Simplified Client Management**: Pre-registered clients bypass the registration endpoint entirely
3. **RFC Compliance**: Properly implements RFC 7591 Client ID Metadata Document discovery
4. **Backwards Compatible**: Defaults to `False`, preserving existing DCR-only behavior

## Example Usage

```python
from mcp.server.auth.settings import ClientRegistrationOptions

options = ClientRegistrationOptions(
    	enabled=True,
    	valid_scopes=["read", "write"],
    	client_id_metadata_document_supported=True,
)
```

Clients can then query `/.well-known/oauth-authorization-server` and see:
```json
{
  "issuer": "https://example.com",
  "client_id_metadata_document_supported": true,
  ...
}
```

### DIFF
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -1352,7 +1352,33 @@ def test_build_metadata(
             "revocation_endpoint": Is(revocation_endpoint),
             "revocation_endpoint_auth_methods_supported": ["client_secret_post", "client_secret_basic"],
             "code_challenge_methods_supported": ["S256"],
-            "client_id_metadata_document_supported": Is(bool),
+            "client_id_metadata_document_supported": Is(False),
+        }
+    )
+
+    metadata = build_metadata(
+        issuer_url=AnyHttpUrl(issuer_url),
+        service_documentation_url=AnyHttpUrl(service_documentation_url),
+        client_registration_options=ClientRegistrationOptions(
+            enabled=True, valid_scopes=["read", "write", "admin"], client_id_metadata_document_supported=True
+        ),
+        revocation_options=RevocationOptions(enabled=True),
+    )
+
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks! This is covered by #1810 which implements the full server-side CIMD support (not just the flag). See #1801 for tracking.

PR #1994, opened 2026-02-04
feat: add idle timeout for StreamableHTTP sessions

### LINKED ISSUE #1283: Potential memory leak in StreamableHTTPSessionManager due to idle sessions not being cleaned up
### Initial Checks

### Description

There appears to be a memory leak in the session management code. The `_server_instances` dictionary grows as new sessions are created, but entries are only removed if a session crashes. When a session is terminated normally, its transport instance remains in the dictionary and is never cleaned up. Over time, this can lead to unbounded memory usage.

**Expected behavior:**
Session entries in `_server_instances` should eventually be removed if they are no longer active or being reused, to avoid memory leaks.

**Steps to reproduce:**
1. Start the server and create multiple sessions.
2. Allow sessions to become idle (no further requests sent).
3. Observe that `_server_instances` retains references to all idle sessions indefinitely, even after reasonable inactivity.

**Clarification:**
It is not correct to simply remove a session from the dictionary when an HTTP request terminates normally, because the same session may be reused by subsequent requests. The issue is that if no further requests ever come for that session, the idle session remains in memory.

**Suggested fix:**
Implement a session timeout or idle cleanup mechanism so that sessions which haven't received any activity for a configurable period are automatically cleaned up and removed from `_server_instances`.

**Additional notes:**
This is not related to any specific Python version. The problem is seen on the current master branch. I am aware that session timeout/cleanup logic is needed to prevent accumulation of idle sessions and associated transports.

See the relevant code in [`streamable_http_manager.py`](https://github.com/modelcontextprotocol/python-sdk/blob/d1ac8d68eb2d7ed139bdc2608b8b4e2ec4265be5/src/mcp/server/streamable_http_manager.py) — this is currently the only place where a HTTP transport is ever removed from the `_server_instances` dict (specifically, when a session crashes).

### Example Code

```Python
No code sample provided, but the issue can be traced to the session lifecycle management in `StreamableHTTPSessionManager` (see `_handle_stateful_request`).

See code: https://github.com/modelcontextprotocol/python-sdk/blob/d1ac8d68eb2d7ed139bdc2608b8b4e2ec4265be5/src/mcp/server/streamable_http_manager.py
```

### Python & MCP Python SDK

```Text
This issue is not related to a specific Python version. The problem is observed on the current master branch of the repository.
```

felixweinberger:
## Summary

Adds a `session_idle_timeout` parameter to `StreamableHTTPSessionManager` that enables automatic cleanup of idle sessions, fixing the memory leak described in #1283.

## Motivation and Context

Sessions created via `StreamableHTTPSessionManager` persist indefinitely in `_server_instances` even after clients disconnect without sending DELETE. Over time this leaks memory. This was reported in #1283 and originally addressed in #1159 by @hopeful0 — this PR reworks that approach.

Key design decisions:

- Uses a **`CancelScope` with a deadline** inside each session's `run_server` task. When the deadline passes, `app.run()` is cancelled and the session cleans up. No background tasks needed — each session manages its own lifecycle.
- Incoming requests **push the deadline forward**, so active sessions are never terminated.
- `terminate()` on the transport is now **idempotent** (early return if already terminated).
- Default is `None` (no timeout, fully backwards compatible). Docstring recommends 1800s (30 min) for most deployments.

## How Has This Been Tested?

Tests added to `tests/server/test_streamable_http_manager.py`: idle session reaping (deterministic, no sleeps), terminate idempotency, and parameterized input validation. All existing tests pass unchanged.

## Breaking Changes

None. `session_idle_timeout` defaults to `None` (no timeout).

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Based on the approach from PR #1159 by @hopeful0, reworked to use cancel scope deadlines instead of transport-level timeout logic.

Closes #1283

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,10 +24,11 @@ This document contains critical information about working with this codebase. Fo
    - Coverage: test edge cases and errors
    - New features require tests
    - Bug fixes require regression tests
-   - NEVER use `anyio.sleep()` with a fixed duration as a synchronization mechanism. Instead:
+   - Avoid `anyio.sleep()` with a fixed duration to wait for async operations. Instead:
      - Use `anyio.Event` — set it in the callback/handler, `await event.wait()` in the test
      - For stream messages, use `await stream.receive()` instead of `sleep()` + `receive_nowait()`
-     - Wrap waits in `anyio.fail_after(5)` as a timeout guard
+     - Exception: `sleep()` is appropriate when testing time-based features (e.g., timeouts)
+   - Wrap indefinite waits (`event.wait()`, `stream.receive()`) in `anyio.fail_after(5)` to prevent hangs
 
 - For commits fixing bugs or adding features based on user reports add:
 
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -46,25 +46,20 @@ class StreamableHTTPSessionManager:
 
     Args:
         app: The MCP server instance
-        event_store: Optional event store for resumability support.
-                     If provided, enables resumable connections where clients
-                     can reconnect and receive missed events.
-                     If None, sessions are still tracked but not resumable.
+        event_store: Optional event store for resumability support. If provided, enables resumable connections
+            where clients can reconnect and receive missed events. If None, sessions are still tracked but not
+            resumable.
         json_response: Whether to use JSON responses instead of SSE streams
-        stateless: If True, creates a completely fresh transport for each request
-                   with no session tracking or state persistence between requests.
+        stateless: If True, creates a completely fresh transport for each request with no session tracking or
+            state persistence between requests.
         security_settings: Optional transport security settings.
-        retry_interval: Retry interval in milliseconds to suggest to clients in SSE
-                       retry field. Used for SSE polling behavior.
-        session_idle_timeout: Optional idle timeout in seconds for stateful
-                              sessions. If set, sessions that receive no HTTP
-                              requests for this duration will be automatically
-                              terminated and removed. When retry_interval is
-                              also configured, ensure the idle timeout
-                              comfortably exceeds the retry interval to avoid
-                              reaping sessions during normal SSE polling gaps.
-                              Default is None (no timeout). A value of 1800
-                              (30 minutes) is recommended for most deployments.
+        retry_interval: Retry interval in milliseconds to suggest to clients in SSE retry field. Used for SSE
+            polling behavior.
+        session_idle_timeout: Optional idle timeout in seconds for stateful sessions. If set, sessions that
+            receive no HTTP requests for this duration will be automatically terminated and removed. When
+            retry_interval is also configured, ensure the idle timeout comfortably exceeds the retry interval to
+            avoid reaping sessions during normal SSE polling gaps. Default is None (no timeout). A value of 1800
+            (30 minutes) is recommended for most deployments.
     """
 
     def __init__(
@@ -80,7 +75,7 @@ class StreamableHTTPSessionManager:
         if session_idle_timeout is not None and session_idle_timeout <= 0:
             raise ValueError("session_idle_timeout must be a positive number of seconds")
         if stateless and session_idle_timeout is not None:
-            raise ValueError("session_idle_timeout is not supported in stateless mode")
+            raise RuntimeError("session_idle_timeout is not supported in stateless mode")
 
         self.app = app
         self.event_store = event_store
@@ -283,16 +278,12 @@ class StreamableHTTPSessionManager:
                                 )
 
                             if idle_scope.cancelled_caught:
-                                session_id = http_transport.mcp_session_id
-                                logger.info(f"Session {session_id} idle timeout")
-                                if session_id is not None:  # pragma: no branch
-                                    self._server_instances.pop(session_id, None)
+                                assert http_transport.mcp_session_id is not None
+                                logger.info(f"Session {http_transport.mcp_session_id} idle timeout")
+                                self._server_instances.pop(http_transport.mcp_session_id, None)
                                 await http_transport.terminate()
-                        except Exception as e:
-                            logger.error(
-                                f"Session {http_transport.mcp_session_id} crashed: {e}",
-                                exc_info=True,
-                            )
+                        except Exception:
+                            logger.exception(f"Session {http_transport.mcp_session_id} crashed")
                         finally:
                             if (  # pragma: no branch
                                 http_transport.mcp_session_id
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -317,10 +317,9 @@ async def test_unknown_session_id_returns_404():
 
 @pytest.mark.anyio
 async def test_idle_session_is_reaped():
-    """Idle timeout sets a cancel scope deadline and reaps the session when it fires."""
-    idle_timeout = 300
+    """After idle timeout fires, the session returns 404."""
     app = Server("test-idle-reap")
-    manager = StreamableHTTPSessionManager(app=app, session_idle_timeout=idle_timeout)
+    manager = StreamableHTTPSessionManager(app=app, session_idle_timeout=0.05)
 
     async with manager.run():
         sent_messages: list[Message] = []
@@ -338,7 +337,6 @@ async def test_idle_session_is_reaped():
         async def mock_receive():  # pragma: no cover
             return {"type": "http.request", "body": b"", "more_body": False}
 
-        before = anyio.current_time()
         await manager.handle_request(scope, mock_receive, mock_send)
 
         session_id = None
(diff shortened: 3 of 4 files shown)


Kludex:
Can't the task manage itself? I don't think we want the overhead of a background task that checks other tasks.

This seems something the specification should mention.

felixweinberger:
@Kludex another attempt, no external managers or anything this time.

felixweinberger:
Same thing for `main` branch: https://github.com/modelcontextprotocol/python-sdk/pull/2022

src/mcp/server/streamable_http_manager.py
```diff
@@ -55,6 +56,14 @@ class StreamableHTTPSessionManager:
...
         retry_interval: Retry interval in milliseconds to suggest to clients in SSE
                        retry field. Used for SSE polling behavior.
+        session_idle_timeout: Optional idle timeout in seconds for stateful sessions.
+                            If set, sessions that receive no HTTP requests for this
+                            duration will be automatically terminated and removed.
+                            When retry_interval is also configured, ensure the idle
+                            timeout comfortably exceeds the retry interval to avoid
+                            reaping sessions during normal SSE polling gaps.
+                            Default is None (no timeout). A value of 1800
+                            (30 minutes) is recommended for most deployments.
```
Kludex: Fill 120 characters, and 4 padding from the above line, please.
```suggestion
        session_idle_timeout: Optional idle timeout in seconds for stateful sessions.
            if set, sessions that receive no HTTP requests for this
                            duration will be automatically terminated and removed.
                            When retry_interval is also configured, ensure the idle
                            timeout comfortably exceeds the retry interval to avoid
                            reaping sessions during normal SSE polling gaps.
                            Default is None (no timeout). A value of 1800
                            (30 minutes) is recommended for most deployments.
```

tests/issues/test_1283_idle_session_cleanup.py
Kludex, line 1: I was hoping for us to not have this issue based folder, and instead have a structured based on the feature set itself/project structure.

A lot of the tests that AI creates seems unnecessary.

tests/issues/test_1283_idle_session_cleanup.py
```diff
@@ -0,0 +1,267 @@
...
+    """When session_idle_timeout is None (default), sessions persist indefinitely."""
+    app = Server("test-no-timeout")
+    manager = StreamableHTTPSessionManager(app=app)
+
+    async with manager.run():
+        sent: list[Message] = []
+        await manager.handle_request(_make_scope(), _mock_receive, _make_send(sent))
+        session_id = _extract_session_id(sent)
+
+        await anyio.sleep(0.3)
```
Kludex: Please review your own test suite. And drop sleeps.

tests/issues/test_1283_idle_session_cleanup.py
felixweinberger: Removed - I guess we can have "which issue did this come from" from PRs / commit history.

tests/issues/test_1283_idle_session_cleanup.py
felixweinberger: Reworked this to use the approach we seem to have elsewhere with:

```python
with anyio.fail_after(5):
    while session_id in manager._server_instances:
        await anyio.sleep(0)
```

It still uses `anyio.sleep(0)` but not timing based anymore - right solution?

CLAUDE.md
```diff
@@ -24,6 +24,13 @@ This document contains critical information about working with this codebase. Fo
    - Coverage: test edge cases and errors
    - New features require tests
    - Bug fixes require regression tests
+   - NEVER use `anyio.sleep()` with a fixed duration as a synchronization mechanism. Instead:
+     - Poll the condition you're waiting for: `while condition: await anyio.sleep(0)` inside `anyio.fail_after(5)`
+     - `anyio.sleep(0)` is a zero-delay scheduler yield, not a real sleep
+     - For cancel scope deadlines, set `scope.deadline = anyio.current_time()` to trigger immediately
+     - For stream messages, use `await stream.receive()` instead of `sleep()` + `receive_nowait()`
+     - For callbacks/notifications, use `anyio.Event` — set it in the callback, await it in the test
+     - If a sleep seems unnecessary, just remove it
```
felixweinberger: Thoughts on adding this @Kludex @maxisbey? Claude always seems to get this wrong for this kind of test, might be worth a shot to add this.

CLAUDE.md
```diff
@@ -24,6 +24,13 @@ This document contains critical information about working with this codebase. Fo
    - Coverage: test edge cases and errors
    - New features require tests
    - Bug fixes require regression tests
+   - NEVER use `anyio.sleep()` with a fixed duration as a synchronization mechanism. Instead:
+     - Poll the condition you're waiting for: `while condition: await anyio.sleep(0)` inside `anyio.fail_after(5)`
+     - `anyio.sleep(0)` is a zero-delay scheduler yield, not a real sleep
```
Kludex: I don't think the first point should actually exist here. You can use `anyio.Event` for this case.

Do we have cases where it was necessary to sleep(0)?

src/mcp/server/streamable_http_manager.py
```diff
@@ -245,19 +265,35 @@ async def run_server(*, task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORE
...
+
+                            if idle_scope.cancelled_caught:
+                                session_id = http_transport.mcp_session_id
+                                logger.info(f"Session {session_id} idle timeout")
+                                if session_id is not None:  # pragma: no branch
+                                    self._server_instances.pop(session_id, None)
+                                await http_transport.terminate()
                         except Exception as e:
                             logger.error(
                                 f"Session {http_transport.mcp_session_id} crashed: {e}",
```
Kludex: Drop the e from the error please. I don't know how this is still wrong.

It should be `logger.exception()` without the `exc_info`, and without the `{e}`.

CLAUDE.md
Kludex: >      - If a sleep seems unnecessary, just remove it

Can't you just clean the sleeps in another PR?

src/mcp/server/streamable_http_manager.py
```diff
@@ -65,13 +75,20 @@ def __init__(
         stateless: bool = False,
         security_settings: TransportSecuritySettings | None = None,
         retry_interval: int | None = None,
+        session_idle_timeout: float | None = None,
     ):
+        if session_idle_timeout is not None and session_idle_timeout <= 0:
+            raise ValueError("session_idle_timeout must be a positive number of seconds")
+        if stateless and session_idle_timeout is not None:
+            raise ValueError("session_idle_timeout is not supported in stateless mode")
```
Kludex:
```suggestion
            raise RuntimeError("session_idle_timeout is not supported in stateless mode")
```

src/mcp/server/streamable_http_manager.py
```diff
@@ -55,6 +56,15 @@ class StreamableHTTPSessionManager:
...
                        retry field. Used for SSE polling behavior.
+        session_idle_timeout: Optional idle timeout in seconds for stateful
+                              sessions. If set, sessions that receive no HTTP
+                              requests for this duration will be automatically
+                              terminated and removed. When retry_interval is
+                              also configured, ensure the idle timeout
+                              comfortably exceeds the retry interval to avoid
+                              reaping sessions during normal SSE polling gaps.
+                              Default is None (no timeout). A value of 1800
+                              (30 minutes) is recommended for most deployments.
```
Kludex:
```suggestion
        session_idle_timeout: Optional idle timeout in seconds for stateful
            sessions. If set, sessions that receive no HTTP requests for this duration will be automatically
                              terminated and removed. When retry_interval is
                              also configured, ensure the idle timeout
                              comfortably exceeds the retry interval to avoid
                              reaping sessions during normal SSE polling gaps.
                              Default is None (no timeout). A value of 1800
                              (30 minutes) is recommended for most deployments.
```
I don't really understand why this is wrong everywhere, but it should be: 4 padding on the left, and 120 columns to the end. (I've fixed only the first one in the suggestion.

felixweinberger: Ah sorry I misunderstood the correction I thought it was about margins on RHS.

src/mcp/server/streamable_http_manager.py
```diff
@@ -219,6 +236,9 @@ async def _handle_stateful_request(
         if request_mcp_session_id is not None and request_mcp_session_id in self._server_instances:  # pragma: no cover
             transport = self._server_instances[request_mcp_session_id]
             logger.debug("Session already exists, handling request directly")
+            # Push back idle deadline on activity
+            if transport.idle_scope is not None and self.session_idle_timeout is not None:
+                transport.idle_scope.deadline = anyio.current_time() + self.session_idle_timeout
```
Kludex, line 236: What happens if the response takes more than the deadline?

src/mcp/server/streamable_http_manager.py
```diff
@@ -245,19 +265,35 @@ async def run_server(*, task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORE
...
+                                    read_stream,
+                                    write_stream,
+                                    self.app.create_initialization_options(),
+                                    stateless=False,
+                                )
+
+                            if idle_scope.cancelled_caught:
+                                session_id = http_transport.mcp_session_id
+                                logger.info(f"Session {session_id} idle timeout")
+                                if session_id is not None:  # pragma: no branch
```
Kludex: Can the `session_id` be `None` here?

src/mcp/server/streamable_http_manager.py
felixweinberger: This is `v1.x` because it's a backport - `main` has your fix.

src/mcp/server/streamable_http_manager.py
felixweinberger: Hmm good question - there's no logic to prevent the cleanup from happening, so the response wouldn't make it back before the Transport gets closed. It'd get a `ClosedResourceError` and the response would be lost. I'd probably argue that's something for the server dev to calibrate for their use case if they have very long running requests? Default stays no timeout anyway.

src/mcp/server/streamable_http_manager.py
felixweinberger: No - Line 253 already does 

```
assert http_transport.mcp_session_id is not None
```

But because the type is `str | None` we have to do it again here - I replaced it with the same `assert` now though.

tests/server/test_streamable_http_manager.py
```diff
@@ -313,3 +313,80 @@ async def mock_receive():
...
+                    if header_name.decode().lower() == MCP_SESSION_ID_HEADER.lower():
+                        session_id = header_value.decode()
+                        break
+                if session_id:  # pragma: no branch
+                    break
+
+        assert session_id is not None, "Session ID not found in response headers"
+
+        # Wait for the 50ms idle timeout to fire and cleanup to complete
+        await anyio.sleep(0.1)
```
felixweinberger, line 355: After some back and forth trying out event based stuff, given we're implementing timeout behavior here, this actually seems like the right way to test this - a case where a `anyio.sleep(0.1)` is justified?

src/mcp/server/streamable_http_manager.py
Kludex: Ah, that explains it. 🪇

felixweinberger:
> Where's the PR that really matters? against main

https://github.com/modelcontextprotocol/python-sdk/pull/2022

PR #1999, opened 2026-02-05
fix: pass new conformance auth scenarios, bump to 0.1.13

pcarleton:
Fixes the conformance CI build which was broken by `@modelcontextprotocol/sdk@1.26.0` being resolved as a transitive dependency of the conformance package.

### Changes

**SDK fix — PRM resource validation (RFC 8707):**
After discovering Protected Resource Metadata, the OAuth client now validates that the `resource` field matches the server URL before proceeding with authorization. If the PRM returns a resource from a different origin (e.g. `https://evil.example.com/mcp`), the client raises `OAuthFlowError` instead of blindly continuing the auth flow.

**Conformance client — pre-registration support:**
The conformance test client now pre-loads client credentials from `MCP_CONFORMANCE_CONTEXT` into token storage when available. This allows the existing `_initialize()` flow to skip DCR when pre-registered credentials are present, which is the correct behavior per the MCP spec's pre-registration approach.

**CI — bump conformance to 0.1.13:**
Updates from `@modelcontextprotocol/conformance@0.1.10` to `0.1.13` which includes the `tools_call` scenario fix for SDK 1.26.0 compatibility.

### DIFF
--- a/.github/actions/conformance/client.py
+++ b/.github/actions/conformance/client.py
@@ -275,6 +275,27 @@ async def run_client_credentials_basic(server_url: str) -> None:
 async def run_auth_code_client(server_url: str) -> None:
     """Authorization code flow (default for auth/* scenarios)."""
     callback_handler = ConformanceOAuthCallbackHandler()
+    storage = InMemoryTokenStorage()
+
+    # Check for pre-registered client credentials from context
+    context_json = os.environ.get("MCP_CONFORMANCE_CONTEXT")
+    if context_json:
+        try:
+            context = json.loads(context_json)
+            client_id = context.get("client_id")
+            client_secret = context.get("client_secret")
+            if client_id:
+                await storage.set_client_info(
+                    OAuthClientInformationFull(
+                        client_id=client_id,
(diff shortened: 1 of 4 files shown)


.github/actions/conformance/client.py
```diff
@@ -275,6 +275,27 @@ async def run_client_credentials_basic(server_url: str) -> None:
...
+                await storage.set_client_info(
+                    OAuthClientInformationFull(
+                        client_id=client_id,
+                        client_secret=client_secret,
+                        redirect_uris=[AnyUrl("http://localhost:3000/callback")],
+                        token_endpoint_auth_method="client_secret_basic" if client_secret else "none",
+                    )
+                )
+                logger.debug(f"Pre-loaded client credentials: client_id={client_id}")
+        except json.JSONDecodeError:
```
Kludex, line 297: Dont you want to see the exception if the json is wrong?

felixweinberger:
Resubmitted as #2010 to unblock Kludex — thanks!

Kludex:
Closing this as stale.

PR #2007, opened 2026-02-07
ci: disable conformance tests

Kludex:
Someone can enable them back when they are not broken.


felixweinberger (approved):
nit: might be worth filing a tracking issue for re-enabling these (or linking one if it exists), so it doesn't fall off the radar. Otherwise LGTM.

felixweinberger:
Filed #2009 to track re-enabling these.

PR #2011, opened 2026-02-07
Security: safer CORS examples + avoid shell=True

TheodorNEngoy:
This hardens a few high-signal footguns that show up in MCP / tool-server code:

- Examples: replace wildcard CORS (`allow_origins=["*"]`) with a localhost-only `allow_origin_regex`.
  This still supports browser-based clients during local dev while avoiding suggesting wildcard CORS as a default.
- URL elicitation client snippet: use `webbrowser.open()` instead of `subprocess.run(..., shell=True)` for cross-platform browser launch.
- Tests: make `TestChildProcessCleanup` less timing-sensitive by polling for file growth (reduces occasional Windows CI flakes) while keeping coverage at 100%.

### DIFF
--- a/tests/client/test_stdio.py
+++ b/tests/client/test_stdio.py
@@ -275,6 +275,13 @@ class TestChildProcessCleanup:
             parent_marker = f.name
 
         try:
+            # Before starting any processes, the file should not grow. This also
+            # exercises the timeout/loop paths in _wait_for_file_growth for branch
+            # coverage.
+            initial_size = os.path.getsize(marker_file)
+            size_no_growth = await self._wait_for_file_growth(marker_file, initial_size, timeout_seconds=0.15)
+            assert size_no_growth == initial_size
+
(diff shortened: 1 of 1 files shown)


TheodorNEngoy:
All checks are green.

This keeps examples safer by default (no wildcard CORS; localhost-only regex) and avoids `subprocess(..., shell=True)` for opening a browser (uses `webbrowser.open()` instead). Happy to adjust the default allowlist/regex if you’d prefer a different local-dev story.

TheodorNEngoy:
Closing to reduce review noise: this PR is now superseded by the split, narrower PRs:

- #2014 (avoid shell=True in npx/browser helpers)
- #2015 (avoid wildcard CORS in StreamableHTTP docs/examples)

Happy to resurrect/port over the remaining stdio test flake/coverage tweaks as a separate PR if those are still useful.

PR #2017, opened 2026-02-08
Update readme

ali-soltani-redgate:


Kludex:
We are going to remove most of the README for V2, it will be fully refactored.

Thanks for the PR tho.

PR #2018, opened 2026-02-08
auth: restrict default CORS origins to loopback

TheodorNEngoy:
Auth routes currently wrap several OAuth endpoints with Starlette CORSMiddleware using wildcard origins (metadata, token, client registration, revocation; and `cors_middleware()` is also used by the simple-auth example for introspection).

This PR makes the default safer for network-exposed auth servers while keeping browser-based local tooling working:

- `cors_middleware()` now uses `allow_origin_regex` with a loopback-only default (`localhost`/`127.0.0.1`/`[::1]`, any port), instead of wildcard origins.
- Adds `cors_origin_regex` plumbing:
  - `create_auth_routes(..., cors_origin_regex=...)`
  - `AuthSettings.cors_origin_regex` (optional)
  - passes through from both `MCPServer` and lowlevel `Server`.
- Adds integration tests asserting loopback origins are allowed and arbitrary origins are not.

Rationale: avoids the common wildcard-CORS footgun for auth endpoints; non-browser clients are unaffected (no Origin header). If you do want a remote browser client, set `cors_origin_regex` explicitly (e.g. `^https://your-ui\.example\.com$`).

### DIFF
--- a/tests/server/mcpserver/auth/test_auth_integration.py
+++ b/tests/server/mcpserver/auth/test_auth_integration.py
@@ -361,6 +361,40 @@ class TestAuthEndpoints:
         assert response.headers.get("access-control-allow-origin") == origin
 
     @pytest.mark.anyio
+    async def test_cors_origin_regex_override(self, mock_oauth_provider: MockOAuthProvider):
+        auth_routes = create_auth_routes(
+            mock_oauth_provider,
+            AnyHttpUrl("https://auth.example.com"),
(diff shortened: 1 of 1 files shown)


maxisbey:
Closing — no linked issue. Per our [contributing guidelines](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md), please open an issue first and wait for maintainer feedback before submitting PRs.

PR #2019, opened 2026-02-09
fix: strip trailing slash from issuer URL in OAuth metadata serialization

### LINKED ISSUE #1919: Trailing slash in OAuthMetadata's `issuer` causes issues with clients

BabyChrist666:
## Summary

Fixes #1919

This PR fixes the OAuth trailing slash issue where Pydantic's `AnyHttpUrl` automatically adds a trailing slash to URLs without path components, breaking OAuth discovery for clients like Google ADK and IBM MCP Context Forge that require exact URL matching per RFC 8414 Section 3.3.

### Changes
- Added `@field_serializer("issuer")` to `OAuthMetadata` to strip trailing slashes during JSON serialization
- Added `@field_serializer("resource")` to `ProtectedResourceMetadata` for the same reason
- Added `@field_serializer("authorization_servers")` to strip trailing slashes from all auth server URLs
- Added comprehensive tests for the trailing slash behavior

### Technical Details
- RFC 8414 examples show issuer URLs without trailing slashes
- The fix only affects JSON serialization output, not internal type handling
- Paths are preserved (e.g., `https://example.com/auth` stays as-is)
- Only lone trailing slashes are stripped (e.g., `https://example.com/` → `https://example.com`)

## Test plan
- [x] Added tests in `tests/shared/test_auth.py` for `OAuthMetadata.issuer` serialization
- [x] Added tests for `ProtectedResourceMetadata.resource` and `authorization_servers`
- [x] Updated snapshot test in `tests/server/auth/test_protected_resource.py`
- [x] All 52 tests pass

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_validator
+from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_serializer, field_validator
 
 
 class OAuthToken(BaseModel):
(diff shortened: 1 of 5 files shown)


BabyChrist666:
Checking back on this — the CI failure from February on the `fix/oauth-trailing-slash-1919` branch appears to have been a transient runner issue. The current checks are passing. Happy to re-run CI or rebase onto `main` if a fresh green run is needed before review.

maxisbey:
Closing as a duplicate of #1938, which was already in flight for #1919. Thanks for the contribution.

PR #2020, opened 2026-02-09
feat: support stderr logging in Jupyter notebook environments

BabyChrist666:
## Summary

Fixes #156

In Jupyter notebook environments, passing `sys.stderr` directly to subprocess doesn't work as expected - the stderr output is lost and users have no visibility into server errors or crashes. This makes debugging extremely difficult as the notebook cell just hangs with no feedback.

### Changes
- Added `_is_jupyter_environment()` function to detect Jupyter/IPython notebook environments
- In Jupyter mode, stderr is captured as a pipe and read asynchronously
- Stderr output is printed with red ANSI color for visibility in notebooks
- In non-Jupyter environments, behavior remains completely unchanged (direct stderr passthrough)

### How It Works
1. On client initialization, we detect if running in Jupyter/IPython
2. If in Jupyter, we set `capture_stderr=True` which pipes stderr instead of redirecting
3. An async `stderr_reader` task reads the pipe and prints to the notebook output
4. Red ANSI coloring makes errors stand out in the notebook

### Technical Details
- Detection checks for `IPKernelApp` in config or `ZMQInteractiveShell` class name
- Falls back gracefully if IPython is not installed
- Windows compatibility maintained via FallbackProcess stderr wrapping

## Test plan
- [x] Added `TestJupyterStderrSupport` test class with 4 tests
- [x] Test Jupyter detection returns False in normal Python
- [x] Test graceful handling when IPython not installed
- [x] Test stderr capture doesn't block the client
- [x] Test continuous stderr output is handled properly
- [x] All existing stdio tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2030, opened 2026-02-10
fix: explicit stderr piping for Jupyter support (#156)

### LINKED ISSUE #156: support logging to stderr in Jupyter Notebook Environments.

Ashutosh0x:
In Jupyter environments, stderr from subprocesses isn't directly visible. To fix this, we pipe stderr via subprocess.PIPE and use a background reader task that detects Jupyter (via is_jupyter()) and prints the output with ANSI red coloring. For standard environments, the output is routed to errlog as before.

Also updated create_windows_process to accept int for errlog, allowing subprocess.PIPE.

Closes #156

### DIFF
--- a/tests/issues/test_156_stdio_stderr.py
+++ b/tests/issues/test_156_stdio_stderr.py
@@ -14,7 +14,6 @@ import pytest
 
 from mcp.client.stdio import StdioServerParameters, stdio_client
 
-
 # A minimal MCP-like server that writes to stderr and then exits.
 SERVER_SCRIPT = textwrap.dedent("""\
     import sys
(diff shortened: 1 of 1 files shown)


maxisbey:
duplicates:

- https://github.com/modelcontextprotocol/python-sdk/pull/1707
- https://github.com/modelcontextprotocol/python-sdk/pull/2020

maxisbey:
Thanks for your contribution! Closing this in favour of #2020 as that predates this PR.

PR #2034, opened 2026-02-11
fix: update tests and examples for new on_* constructor kwargs API

maxisbey:
Update all test files and example servers to use the new lowlevel Server API pattern where handlers are registered via constructor kwargs instead of decorators.

This is a follow-up to the `sketch/lowlevel-server-v2-kwargs` branch that converts the core server. This branch fixes all tests and examples.

**Key changes:**
- Replace `@server.list_tools()` etc. decorators with `on_list_tools=handler` constructor kwargs
- Update handler signatures to `(ctx: ServerRequestContext, params: <ParamsType>)`
- Return full Result types instead of auto-wrapped values
- Replace `server.request_context` property with `ctx` parameter
- Update experimental task handler registration to use `enable_tasks(on_*=...)`
- Remove `test_func_inspection.py` (corresponding module was deleted)

### DIFF
--- a/tests/experimental/tasks/server/test_server.py
+++ b/tests/experimental/tasks/server/test_server.py
@@ -36,6 +36,7 @@ from mcp.types import (
     JSONRPCError,
     JSONRPCNotification,
     JSONRPCResponse,
+    ListTasksRequest,
     ListTasksResult,
     ListToolsResult,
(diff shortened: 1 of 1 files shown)


Kludex:
Can we close this?

PR #2037, opened 2026-02-11
feat: add SessionInfo abstraction for external session ID tracking

edis-uipath:
## Summary

In some cases we need to save the session id or to restore an existing session id and to continue from it, in some cases we need to move a session across process boundary (for example debug or human in the loop scenarios).

The transport previously owned session state directly via `self.session_id`, making it impossible for consumers to observe or persist session IDs externally (e.g. persisting to a debug state endpoint for playground/agent resume scenarios).

This PR introduces a `SessionInfo` base class with async `get_session_id` / `set_session_id` methods that the transport delegates all session ID storage to. Consumers can subclass `SessionInfo` to add side-effects like HTTP persistence, and pass it via the new `session_info` parameter on `StreamableHTTPTransport` and `streamable_http_client`.

### Key changes

- **Add `SessionInfo` class** — base class with async get/set methods; default implementation stores in a plain attribute (backwards compatible)
- **Transport delegates to `_session_info`** — replaces direct `self.session_id` usage
- **`_prepare_headers` and `_maybe_extract_session_id_from_response` are now async** — to support subclasses that need I/O (e.g. loading from a remote store)
- **Remove unused `session_id` field from `RequestContext`** — headers are built from `_prepare_headers`, not from the context
- **`streamable_http_client` accepts optional `session_info` parameter** — passed through to the transport constructor
- **`get_session_id` is now async** — delegates to `SessionInfo.get_session_id()`

### Backwards compatibility

- Default behavior is unchanged: when no `session_info` is provided, a plain `SessionInfo()` is created internally
- All 49 existing streamable HTTP tests pass without modification
- The removed `RequestContext.session_id` field was not used by any handler (headers come from `_prepare_headers`)

### Example usage

**Track the session ID externally:**

```python
session_info = SessionInfo()

async with streamable_http_client(url, http_client=client, session_info=session_info) as (read, write):
    session = ClientSession(read, write)
    await session.initialize()
    # session_info.session_id now holds the server-assigned session ID
    print(f"Session ID: {await session_info.get_session_id()}")
```

**Resume an existing session (skip initialize):**

```python
# Reuse a session ID from a previous connection
session_info = SessionInfo(session_id="previously-saved-session-id")

async with streamable_http_client(url, http_client=client, session_info=session_info) as (read, write):
    session = ClientSession(read, write)
    # No initialize() call — the transport includes the existing session ID in headers
```

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -49,6 +49,26 @@ DEFAULT_RECONNECTION_DELAY_MS = 1000  # 1 second fallback when server doesn't pr
 MAX_RECONNECTION_ATTEMPTS = 2  # Max retry attempts before giving up
 
 
+class SessionInfo:
+    """Base class for MCP session ID tracking.
+
+    The transport delegates all session ID storage to this object.
+    Override ``get_session_id`` / ``set_session_id`` in subclasses to
+    add side-effects such as HTTP persistence.
+    """
+
+    def __init__(self, session_id: str | None = None) -> None:
+        self.session_id = session_id
+
+    async def get_session_id(self) -> str | None:
+        """Return the current session ID (or None)."""
+        return self.session_id
+
+    async def set_session_id(self, session_id: str | None) -> None:
+        """Store a new session ID assigned by the server, or None to clear."""
+        self.session_id = session_id
+
+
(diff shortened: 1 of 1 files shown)


maxisbey:

Issue #1933, opened 2026-01-22
Using transport="stdio" closes real stdio, causing ValueError after server exits


hyn0027:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

Hi! I ran into an issue where running the server with `transport="stdio"` causes subsequent stdio operations to fail after the server exits.

### Minimal reproduction

```python
from mcp.server.fastmcp import FastMCP

mcp = FastMCP("Demo")
mcp.run(transport="stdio")
print("?")
```

When using `Ctrl+D` to exit the server, the final print raises:

```bash
Traceback (most recent call last):
  File "...", line 5, in <module>
    print("?")
ValueError: I/O operation on closed file.
```

I’m not sure if this is intended behavior though.

### Likely cause

In stdio.py, stdio is wrapped like this:
<https://github.com/modelcontextprotocol/python-sdk/blob/5301298225968ce0fa8ae62870f950709da14dc6/src/mcp/server/stdio.py#L44-L47>

When these wrappers are closed, they also close the sys.stdin.buffer / sys.stdout.buffer.

### Proposed fix

```python
    if not stdin:
        stdin_fd = os.dup(sys.stdin.fileno())
        stdin_bin = os.fdopen(stdin_fd, "rb", closefd=True)
        stdin = anyio.wrap_file(TextIOWrapper(stdin_bin, encoding="utf-8"))
    if not stdout:
        stdout_fd = os.dup(sys.stdout.fileno())
        stdout_bin = os.fdopen(stdout_fd, "wb", closefd=True)
        stdout = anyio.wrap_file(TextIOWrapper(stdout_bin, encoding="utf-8"))
```

### Python & MCP Python SDK

```Text
Python 3.10.15
mcp==1.25.0
```

adityuhkapoor:
I can reproduce this on main; root cause is TextIOWrapper closing sys.stdin.buffer/sys.stdout.buffer due to auto garbage collection. os.dup() adds another pointer to the underlying resource and ensures that the stdio process doesn't prematurely get GC'd. Happy to put up a PR if this is up for grabs.

adityuhkapoor:
[...]

PR #2040, opened 2026-02-12
fix: prevent stdio_server from closing process stdio handles

adityuhkapoor:
## Motivation and Context

Fixes #1933. `stdio_server()` wraps `sys.stdin.buffer` and `sys.stdout.buffer` in `TextIOWrapper`. When the wrapper gets garbage-collected, it closes the underlying buffer, killing process stdio. Any subsequent `print()` or stdin usage raises `ValueError: I/O operation on closed file`.

`os.dup()` gives the wrapper its own copy of the file descriptor, so closing it doesn't affect the original process stdio.

## How Has This Been Tested?

- Regression test that monkeypatches sys.stdin/sys.stdout with real pipe-backed file descriptors, runs `stdio_server()`, then asserts the original fds are still open via `os.fstat()`.
- Verified the test fails without the fix and passes with it.
- Full test suite: 1139 passed, 0 failures.

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist


src/mcp/server/stdio.py
```diff
@@ -34,14 +35,16 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
...
+    # stream to ensure UTF-8. We duplicate the file descriptors first so that
+    # closing the wrapper doesn't close the real process stdio handles.
     if not stdin:
-        stdin = anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8"))
+        stdin_fd = os.dup(sys.stdin.buffer.fileno())
+        stdin = anyio.wrap_file(TextIOWrapper(os.fdopen(stdin_fd, "rb"), encoding="utf-8"))
     if not stdout:
-        stdout = anyio.wrap_file(TextIOWrapper(sys.stdout.buffer, encoding="utf-8"))
+        stdout_fd = os.dup(sys.stdout.buffer.fileno())
+        stdout = anyio.wrap_file(TextIOWrapper(os.fdopen(stdout_fd, "wb"), encoding="utf-8"))
```
maxisbey:
```suggestion
        stdin = anyio.wrap_file(TextIOWrapper(open(sys.stdin.fileno(), "rb", closefd=False), encoding="utf-8"))
    if not stdout:
        stdout = anyio.wrap_file(TextIOWrapper(open(sys.stdout.fileno(), "wb", closefd=False), encoding="utf-8"))
```

`closefd=False` gets you the same isolation without having to allocate an `fd`, so there's nothing to clean up and the filterwarnings on the test can be deleted.

also maybe update the comment above to something like:

```
# Re-wrap the `fd` with `closefd=False` to force UTF-8 (Windows encoding is
# platform-dependant) without taking ownership of process stdio.
```

maxisbey (changes requested):

src/mcp/server/stdio.py
adityuhkapoor: Applied, dropping import os and filter warnings, much appreciated!

adityuhkapoor:
Addressed the closefd=False suggestion and added wrapper cleanup on exit to   avoid ResourceWarning. Ready for re-review whenever you get a chance @maxisbey

uppulaharshith2-rgb:
One thing I noticed reading the cleanup path: the new `try/finally` around the task group awaits `stdin.aclose()` / `stdout.aclose()` without a shielded cancel scope. `AsyncFile.aclose()` is `to_thread.run_sync(self._fp.close)`, which is cancellation-sensitive — if the caller's surrounding scope is already cancelled when `stdio_server` unwinds (e.g. SIGINT during a long write, or a parent `cancel_scope.cancel()` while a message is being framed), the close gets re-raised before `TextIOWrapper.close()` runs, the wrapper is leaked, and a `ResourceWarning` fires on GC. That's the same noise the `closefd=False` switch was meant to eliminate, just now reached via cancellation instead of fd ownership.

`test_1933_stdio_close.py` only exercises clean exit (write end closed → EOF → normal `__aexit__`), so this path isn't covered. Wrapping the two `aclose` calls in `with anyio.CancelScope(shield=True):` inside the `finally` would close the gap, and a test that does `async with stdio_server(): tg.cancel_scope.cancel()` mid-iteration would lock it in. Happy to draft the test if useful.

PR #1971, opened 2026-01-29
fix: revert README to v1.x documentation on main

maxisbey:
The main branch README documents v2 APIs (e.g. `MCPServer`) that don't exist in the current stable v1.x release, causing confusion for users (see #1970).

This reverts `README.md` to v1.x content and preserves the v2 docs as `README.v2.md`. CI, pre-commit, and snippet tooling are updated to target `README.v2.md`. `TODO(v2)` markers are added throughout for easy swapping when v2 ships.

Closes #1970

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

Issue #1970, opened 2026-01-28
Cannot find reference 'MCPServer' in 'mcp.server.mcpserver'  in Quickstart code


jesuspb2:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

The Quickstart example code references `MCPServer` from `mcp.server.mcpserver`, but this class cannot be found, resulting in an import error.

**Expected Behavior:**
The `MCPServer` class should be available for import from `mcp.server.mcpserver` as shown in the Quickstart documentation.

**Actual Behavior:**
The IDE cannot locate the `MCPServer` class in the specified module, preventing the example from running.

Steps to Reproduce:
1. Follow the Quickstart guide
2. Copy the example code that imports `MCPServer`
3. Attempt to run the code
4. Observe the import error

Environment:
- Python version: 3.13.5
- MCP library version: 1.26.0
- IDE: PyCharm 2025.3.2
- OS: macOS

### Example Code

```Python
from mcp.server.mcpserver import MCPServer
# Create an MCP server
mcp = MCPServer("Demo")
```

### Python & MCP Python SDK

```Text
- Python version: 3.13.5
- MCP library version: 1.26.0
```

samsond:
I can confirm I’m seeing similar behavior on mcp[cli]==1.26.0 (Python 3.11.12 on macOS).

I investigated this issue and found that MCPServer appears to have been introduced in commit [65c614e](https://github.com/modelcontextprotocol/python-sdk/commit/65c614e48e833df4622b07da698ca64935c31f36) (Jan 25, 2026), which renamed FastMCP to MCPServer. 

Based on my investigation, it seems this change has not been released to PyPI yet. The latest release tag I can find is v1.26.0, which appears to predate this rename. 

**Maintainers**: I'm not familiar with the release process, correct me if I've misunderstood the versioning. The documentation references MCPServer, but it's not available in the PyPI package I installed.

sabarigireeson-glitch:
I also faced the same issue yesterday, upon checking and referring I found

FastMCP is the latest API available in the released PyPI package.
MCPServer exists in the GitHub main branch (it’s a rename of FastMCP) but has not been released to PyPI yet.
As of mcp==1.26.0, the correct import is:

```
from mcp.server.fastmcp import FastMCP
```
[...]

PR #2045, opened 2026-02-12
fix: revert README.md to v1 documentation

maxisbey:
Recent commits (21822053, 4fc49c62, 1a8c14a5) accidentally updated README.md with v2 content after it was intentionally reverted to v1 docs in #1971 (see #1970). This restores README.md to the v1 state. No content is lost — all three commits already applied their changes to README.v2.md.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,7 @@ repos:
         language: system
         files: ^(pyproject\.toml|uv\.lock)$
         pass_filenames: false
+      # TODO(Max): Drop this in v2.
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -55,6 +55,12 @@ repos:
         language: system
         files: ^(pyproject\.toml|uv\.lock)$
         pass_filenames: false
+      - id: readme-v1-frozen
```
Kludex, line 59:
```suggestion
      # TODO(Max): Drop this in v2.
      - id: readme-v1-frozen
```

Issue #1835, opened 2026-01-06
Http errors are not handled properly by streamable_http


yuxiaorun:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

## Description
When using Kiro-CLI with a remote fastMCP server, one issue was observed that when the MCP tool call gets a 403 http error, the cli would hang indefinitely instead of showing an error response.

## Root Cause
In `mcp/client/streamable_http.py`, the `_handle_post_request()` method calls `response.raise_for_status()` but doesn't catch the resulting `httpx.HTTPStatusError` exception and convert it to an MCP error response. Without a response message, the client waits indefinitely.

## Expected Behavior
HTTP errors should be caught and converted to MCP JSON-RPC error responses so clients can handle them gracefully.

## Proposed Fix
Wrap the `response.raise_for_status()` call in a try/except block that catches `httpx.HTTPStatusError` and:
1. Creates a JSONRPCError with appropriate error details
2. Sends it to the read stream so the client receives the error

The temporary patch that fixs the problem:
~~~
@wraps(original_handle_post_request)
    async def _handle_post_request_patched(self, ctx: RequestContext) -> None:  # type: ignore
        """Patched version of _handle_post_request that converts HTTP errors to MCP errors.
        
        This ensures that HTTP errors (like 403 Forbidden) during tool calls are properly
        converted to MCP error messages and sent back to the client, preventing hangs.
        """
        try:
            return await original_handle_post_request(self, ctx)
        except httpx.HTTPStatusError as http_error:
            logger.warning('HTTP error in MCP client: %s %s', http_error.response.status_code, http_error.request.url)
            
            # Extract the request ID if this was a request (not a notification)
            message = ctx.session_message.message
            request_id = None
            if isinstance(message.root, mcp.types.JSONRPCRequest):
                request_id = message.root.id
            
            if request_id is not None:
                # Try to parse the error response as an MCP error
                response = http_error.response
                try:
                    await response.aread()
                    error_body = response.json()
                    error_message = error_body.get('message', f'HTTP {response.status_code} error')
                except Exception:
                    error_message = f'HTTP {response.status_code} error'
                
                # Create an MCP error response
                jsonrpc_error = mcp.types.JSONRPCError(
                    jsonrpc="2.0",
                    id=request_id,
[...]

Issue #1577, opened 2025-11-05
Client tool call hangs forever if server crashes or connection dies when using streamable-http


gyang-xai:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

If the server dies during a tool call before it sends a response the client will get stuck forever.

### Example Code

```Python
# server.py

from mcp.server.fastmcp import FastMCP
import sys

mcp = FastMCP("StatefulServer")

@mcp.tool()
def greet(name: str = "World") -> str:
    """Greet someone by name."""
    sys.exit(0)
    return f"Hello, {name}!"


if __name__ == "__main__":
    mcp.run(transport="streamable-http")


# client.py

import asyncio

from mcp import ClientSession
from mcp.client.streamable_http import streamablehttp_client


async def main():
    # Connect to a streamable HTTP server
    async with streamablehttp_client("http://localhost:8000/mcp") as (
        read_stream,
        write_stream,
        _,
    ):
        # Create a session using the client streams
        async with ClientSession(read_stream, write_stream) as session:
            # Initialize the connection
            await session.initialize()

            res = await session.call_tool("greet", {"name": "World"})
            print(res)


if __name__ == "__main__":
    asyncio.run(main())



# Run server.py then client.py


# client.py will hang forever in the tool call and never complete. I would expect it to throw an error since the connection died and it will never get a response.
```

### Python & MCP Python SDK

```Text
Python 3.13.7
mcp 1.20.0
```

gyang-xai:
I have a possible PR that fixes this here: https://github.com/modelcontextprotocol/python-sdk/pull/1578 for reference but not sure if this is the best solution.

owenliang:
"I encountered the same issue: exceptions at the underlying connection layer are not being broadcast to every waiting caller. Currently, the only workaround is to set a timeout for call tool operations to actively give up."

astro-angelfish:
Reproduced on my side, just made my framework hangs forever. Maybe a keep-alive mechanism is required for streamable-http.

Issue #1393, opened 2025-09-24
Lazy HTTP connections seem to make error handling difficult.


caffeinism:
### Question

When using streamablehttp_client, since HTTP connections are initialized with anyio.create_task_group, if the server is unreachable, an error will occur unexpectedly at some point later. Is this the intended behavior? If the connection is impossible, I want to wait and handle the error. However, the current behavior executes the code after the initialization and then throws an error at some point. If the error had occurred earlier, the code below it wouldn't have needed to execute. I consider this excessive optimization. Is there a reason for this design?

hshahrokni2:
When connections are lazy, a proxy layer cannot distinguish a genuine connection error from a normal empty-response until it is too late to retry cleanly. With [Komilion](https://komilion.com) we need explicit connection-state signals at initialization time to make accurate failover decisions before committing a request to a provider. An eager connection mode or a connection-readiness probe on the context manager entry would be very useful here.

PR #2047, opened 2026-02-13
fix: handle HTTP error status codes in streamable HTTP client

### LINKED ISSUE #1010: HTTP errors are being silently swallowed in the StreamableHTTP client
### Initial Checks

### Description

When an MCP server returns HTTP 5xx errors (e.g., when the server is down) after initial connection, these errors are not being propagated to the caller. This makes it impossible for client applications to detect and handle server failures appropriately. More than that, clients which are waiting for response stream, like [Strands SDK](https://github.com/strands-agents/sdk-python), are hanging forever since no response is returned

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Any
```

Kludex:
## Summary

When a server returns a non-2xx HTTP status code (e.g. 500 Internal Server Error, 403 Forbidden), the streamable HTTP client previously called `response.raise_for_status()`, which raised an unhandled `httpx.HTTPStatusError`. This exception was never caught, so the pending request would hang indefinitely — the caller's `await session.list_tools()` (or similar) would never resolve.

This PR replaces `raise_for_status()` with explicit status code handling that sends a `JSONRPCError` back through the read stream, matching the pattern already established for 404 (session terminated) responses. For requests (messages with an `id`), the client now sends an `INTERNAL_ERROR` JSONRPCError so the pending request resolves immediately with an `MCPError`. For notifications (no `id`), the error is silently discarded since there is no caller waiting for a response.

## Follow-up to

- #2005 — Fixed similar issues for unexpected content types, JSON parse failures, and SSE parse failures, but did not address plain HTTP error responses.

## Related issues

- Fixes #1010 — HTTP errors silently swallowed in the StreamableHTTP client
- Related to #1577 — Client tool call hangs forever if server crashes
- Related to #1401 — ClientSession error handling
- Related to #1212 — `raise_for_status()` discards error response body
- Related to #1393 — Lazy HTTP connection error handling
- Related to #1835 — HTTP 403 causing client to hang (closed as dup)
- Related to #1382 — Unexpected content type hanging (fixed by #2005)
- Related to #1144 — Invalid JSON hanging (fixed by #2005)

## Related PRs

- #1308 — Proposed fix to catch HTTP errors (open)
- #1008 — Earlier attempt at exception handling for POST failures (closed)
- #1119 — Earlier attempt at HTTPStatusError handling (closed)

## Test plan

- [x] Added `test_http_error_status_sends_jsonrpc_error` — verifies that an HTTP 500 from the server results in an `MCPError` raised to the caller instead of a hang
- [x] All existing tests pass
- [x] Pre-commit hooks pass (ruff, pyright, prettier)


Kludex:
I'm so good at summarizing PRs nowadays.

tests/client/test_notification_response.py
```diff
@@ -116,6 +116,40 @@ async def test_unexpected_content_type_sends_jsonrpc_error() -> None:
...
+async def test_http_error_status_sends_jsonrpc_error() -> None:
+    """Verify HTTP 5xx errors unblock the pending request with an MCPError.
+
+    When a server returns a non-2xx status code (e.g. 500), the client should
+    send a JSONRPCError so the pending request resolves immediately instead of
+    raising an unhandled httpx.HTTPStatusError that causes the caller to hang.
+    """
+    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=_create_http_error_app(500))) as client:
+        async with streamable_http_client("http://localhost/mcp", http_client=client) as (read_stream, write_stream):
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
```
maxisbey, line 148: why is coverage even upset here?

Kludex: I think I explained it before (there must be some reference in code), but there's a bug in coverage on async context managers.

maxisbey (dismissed):

maxisbey (approved):

owenliang:
When will it be released to PyPI?

hiteshshimpi-55:
Why is this not implemented in the latest version of this package

PR #2041, opened 2026-02-12
feat: expose progress_callback in ServerSession methods

BryceEWatson:
Fixes #1671

## Motivation and Context

The `ServerSession` high-level methods for sending requests to clients (`create_message`, `elicit_form`, `elicit_url`) don't expose the `progress_callback` parameter, even though the underlying `BaseSession.send_request()` fully supports it. This means servers can't easily receive progress notifications from clients during sampling or elicitation requests.

`ClientSession.call_tool()` already exposes `progress_callback` for the client→server direction. This PR adds the same support for the server→client direction, bringing the Python SDK into parity with the TypeScript SDK's `RequestOptions.onprogress`.

## How Has This Been Tested?

- Two new E2E tests added to `tests/shared/test_progress_notifications.py`:
  - `test_server_create_message_progress_callback`: verifies progress callbacks work during sampling requests via `MCPServer` + `Client`
  - `test_server_elicit_form_progress_callback`: verifies progress callbacks work during elicitation requests via `MCPServer` + `Client`
- After rebasing onto current `main`, the full test suite passes locally (`uv run pytest`), `uv run pyright` reports 0 errors, and `uv run ruff check .` is clean. CI is green on all checks.

## Breaking Changes

No. All new parameters default to `None`, maintaining full backward compatibility.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

**Approach:** Mirrors the existing `ClientSession.call_tool()` pattern — adds `progress_callback: ProgressFnT | None = None` to each method signature and passes it through to `send_request()`.

**Files changed (4):**
- `src/mcp/server/session.py` — Core: added param to `create_message()` (all overloads + impl), `elicit()`, `elicit_form()`, `elicit_url()`
- `src/mcp/server/elicitation.py` — Helpers: passthrough in `elicit_with_validation()` and `elicit_url()`
- `src/mcp/server/mcpserver/context.py` — Context API: passthrough in `Context.elicit()` and `Context.elicit_url()`
- `tests/shared/test_progress_notifications.py` — 2 new E2E tests

[...]

PR #2048, opened 2026-02-13
fix: skip readme-v1-frozen in CI and add diff-based README.md check

maxisbey:
## Problem

The `readme-v1-frozen` pre-commit hook (added in #2045) uses `language: fail`, which unconditionally fails when its `files` pattern matches any file in the candidate set. Since CI runs pre-commit with `--all-files`, `README.md` always matches and the hook always fails — breaking CI on `main` and all PRs.

This also blocks unrelated PRs like #2041 from merging.

## Solution

A two-part fix that provides protection in both local and CI contexts:

1. **Skip `readme-v1-frozen` in the pre-commit `--all-files` run** via the `SKIP` env var — the same mechanism already used for `no-commit-to-branch`. The hook continues to work locally where pre-commit only checks staged files.

2. **Add a standalone CI step** that uses `git diff` against the PR base to check whether `README.md` was actually modified. This provides CI-level enforcement that the `language: fail` hook cannot offer in `--all-files` mode.

### Why this approach?

`language: fail` and `--all-files` are fundamentally incompatible — `language: fail` fires whenever the file *exists* in the set, not when it was *changed*. Local pre-commit and CI have different semantics: locally you check "is this file staged?", in CI you check "was this file changed in this PR?" This fix uses the right tool for each context rather than forcing one mechanism to do both jobs.

Both the `SKIP` entry and the CI step are tagged with the existing `TODO(Max): Drop this in v2` comment for easy cleanup.

### DIFF
--- a/.github/workflows/shared.yml
+++ b/.github/workflows/shared.yml
@@ -26,7 +26,19 @@ jobs:
         with:
           extra_args: --all-files --verbose
         env:
-          SKIP: no-commit-to-branch
+          SKIP: no-commit-to-branch,readme-v1-frozen
+
+      # TODO(Max): Drop this in v2.
+      - name: Check README.md is not modified
+        if: github.event_name == 'pull_request'
+        run: |
+          git fetch --no-tags --depth=1 origin "$BASE_SHA"
+          if git diff --name-only "$BASE_SHA" -- README.md | grep -q .; then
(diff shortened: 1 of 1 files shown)


maxisbey:
proof this CI change now works: https://github.com/modelcontextprotocol/python-sdk/pull/2049

Issue #1795, opened 2025-12-17
Make JSON-RPC ID type coercion configurable


maxisbey:
## Background

PR #1720 added automatic type coercion for JSON-RPC response IDs, allowing string IDs like `"0"` to match integer request IDs like `0`. This makes the client more tolerant of servers that echo back IDs in a different but semantically equivalent format.

## Spec Language

Per the JSON-RPC 2.0 and MCP specifications, response IDs should be **exact matches**:

**JSON-RPC 2.0**: Response ID "MUST be the same as the value of the id member in the Request Object"

**MCP Spec**:
- Result responses "MUST include the same ID as the request they correspond to"
- Error responses "MUST include the same ID as the request they correspond to"

The language "same as the value" / "same ID" implies exact matching including type. The spec does not describe IDs as opaque and does not mandate type conversion or normalization. Servers should echo back the exact ID they received.

## Current Behavior

PR #1720 makes the client more lenient by accepting type-coerced IDs (e.g., `"0"` matching `0`). This is a workaround for non-compliant servers.

## Proposal

Make this ID type coercion behavior configurable:
- **Strict mode**: Require exact type matching (spec-compliant behavior)
- **Lenient mode**: Allow type coercion between numeric strings and integers (current behavior after #1720)

This could be exposed as a session or client configuration option, defaulting to lenient for backwards compatibility.

Ideally strict mode is default, with an error message being clear as to why something fails. If we receive a response with an ID that's a string, even though the original request ID was an int we could first check if there are any IDs which would match after type coercion and then include that in the error message. "Request ID with same value of different type found, but ignored due to strict ID matching configuration".

## Reasoning

By spec language, request ID `1` and request ID `"1"` are **different** IDs, but with this type coercion they are treated as the same. Given that many clients/servers are non-compliant with this anyway and do their own type conversions we should allow server/client users of the SDK to configure the behaviour for themselves.

## Related

- #1720 - Fix JSON-RPC error response ID matching

Kludex:
#1720 is not compliant.

I made an effort to merge https://github.com/modelcontextprotocol/python-sdk/pull/1178, and then months later I see this.

Kludex:
We should fix this before v2.

This should not be configurable!

PR #2056, opened 2026-02-13
fix: allow null id in JSONRPCError per JSON-RPC 2.0 spec

maxisbey:
## Problem

The JSON-RPC 2.0 specification states:

> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it **MUST** be Null.

The SDK's `JSONRPCError.id` was typed as `RequestId` (`int | str`), rejecting `null`. This forced a non-compliant `id="server-error"` sentinel workaround in the server transports. The TypeScript SDK and MCP schema both allow null ids — the Python SDK was the sole outlier.

## Changes

**Type layer** (`src/mcp/types/jsonrpc.py`):
- `JSONRPCError.id` changed from `RequestId` to `RequestId | None`
- Added `@model_serializer` to ensure `id: null` is preserved in JSON output even when serializing with `exclude_none=True` (which is used across 20+ call sites)
- `JSONRPCResponse.id` unchanged — successful responses always have a non-null id

**Server transports** (`streamable_http.py`, `streamable_http_manager.py`):
- Replaced `id="server-error"` sentinel with `id=None`
- Added null guard in the server-side message router to prevent `str(None)` producing `"None"`

**Session layer** (`shared/session.py`):
- Added null-id guard in `_handle_response` before response routing — surfaces null-id errors as `MCPError` via the message handler instead of silently discarding them

## Partially addresses #1821

This fixes the `id` field problem described in #1821 (Python used `id="server-error"` where the spec requires `null`). The error code divergence tracked in that issue is still pending spec clarification.

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -1,20 +1,16 @@
-import json
 from typing import Any
 
 import pytest
 
 from mcp.types import (
     LATEST_PROTOCOL_VERSION,
-    PARSE_ERROR,
     ClientCapabilities,
     CreateMessageRequestParams,
     CreateMessageResult,
     CreateMessageResultWithTools,
-    ErrorData,
     Implementation,
     InitializeRequest,
     InitializeRequestParams,
-    JSONRPCError,
     JSONRPCRequest,
     ListToolsResult,
     SamplingCapability,
@@ -364,20 +360,3 @@ def test_list_tools_result_preserves_json_schema_2020_12_fields():
     assert tool.input_schema["$schema"] == "https://json-schema.org/draft/2020-12/schema"
     assert "$defs" in tool.input_schema
     assert tool.input_schema["additionalProperties"] is False
-
-
-def test_jsonrpc_error_null_id_serialization_preserves_id():
-    """Test that id: null is preserved in JSON output with exclude_unset=True.
-
-    JSON-RPC 2.0 requires the id field to be present with value null for
-    parse errors, not absent entirely.
-    """
-    error = JSONRPCError(jsonrpc="2.0", id=None, error=ErrorData(code=PARSE_ERROR, message="Parse error"))
-    serialized = error.model_dump(by_alias=True, exclude_unset=True)
-    assert "id" in serialized
-    assert serialized["id"] is None
-
-    json_str = error.model_dump_json(by_alias=True, exclude_unset=True)
-    parsed = json.loads(json_str)
-    assert "id" in parsed
-    assert parsed["id"] is None


src/mcp/server/streamable_http.py
```diff
@@ -975,12 +975,11 @@ async def message_router():
...
-                        if isinstance(message, JSONRPCResponse | JSONRPCError):
-                            response_id = str(message.id)
-                            # If this response is for an existing request stream,
-                            # send it there
-                            target_request_id = response_id
+                        # Check if this is a response with a known request id.
+                        # Null-id errors (e.g., parse errors) fall through to
+                        # the GET stream since they can't be correlated.
+                        if isinstance(message, JSONRPCResponse | JSONRPCError) and message.id is not None:
+                            target_request_id = str(message.id)
```
Kludex, line 982: Why are we coercing to str tho?

src/mcp/shared/session.py
```diff
@@ -458,6 +458,19 @@ async def _handle_response(self, message: SessionMessage) -> None:
...
+        if isinstance(message.message, JSONRPCError) and message.message.id is None:
+            error = message.message.error
+            logging.warning(f"Received error with null ID: {error.message}")
+            await self._handle_incoming(MCPError(error.code, error.message, error.data))
+            return
+
+        # After the null-id guard above, id is guaranteed to be non-None.
+        # JSONRPCResponse.id is always RequestId, and JSONRPCError.id is only
+        # None for parse errors (handled above).
+        assert message.message.id is not None
```
Kludex: You shouldn't need to do this after. This information should be inferred by the type checker, why is it not?

src/mcp/shared/session.py
```diff
@@ -458,6 +458,19 @@ async def _handle_response(self, message: SessionMessage) -> None:
...
+        # be determined — these cannot be correlated to any pending request.
+        if isinstance(message.message, JSONRPCError) and message.message.id is None:
+            error = message.message.error
+            logging.warning(f"Received error with null ID: {error.message}")
+            await self._handle_incoming(MCPError(error.code, error.message, error.data))
+            return
+
+        # After the null-id guard above, id is guaranteed to be non-None.
+        # JSONRPCResponse.id is always RequestId, and JSONRPCError.id is only
+        # None for parse errors (handled above).
```
Kludex: Please drop verbose comments as well.
```suggestion
```

src/mcp/types/jsonrpc.py
```diff
@@ -75,9 +75,19 @@ class JSONRPCError(BaseModel):
...
 
+    @model_serializer(mode="wrap")
+    def _serialize(self, handler: SerializerFunctionWrapHandler, _: SerializationInfo) -> dict[str, Any]:
+        result = handler(self)
+        # JSON-RPC 2.0 requires id to always be present in error responses,
+        # even when null (e.g. parse errors). Ensure exclude_none=True
+        # cannot strip it.
+        if "id" not in result and self.id is None:
+            result["id"] = None
+        return result
```
Kludex: Is `exclude_none` applied recursively? I think the point is that we shouldn't be using it at the JSONRPC level.

I think we can avoid using `model_serializer` and all that.

tests/test_types.py
```diff
@@ -360,3 +364,20 @@ def test_list_tools_result_preserves_json_schema_2020_12_fields():
...
+    """
+    error = JSONRPCError(jsonrpc="2.0", id=None, error=ErrorData(code=PARSE_ERROR, message="Parse error"))
+    serialized = error.model_dump(by_alias=True, exclude_none=True)
+    assert "id" in serialized
+    assert serialized["id"] is None
+
+    json_str = error.model_dump_json(by_alias=True, exclude_none=True)
+    parsed = json.loads(json_str)
+    assert "id" in parsed
+    assert parsed["id"] is None
```
Kludex: There's no need to test Pydantic. Pydantic already tests Pydantic.

Kludex (changes requested):

src/mcp/server/streamable_http.py
maxisbey: It was already happening before. There's a ticket for this here: http://github.com/modelcontextprotocol/python-sdk/issues/1795

src/mcp/shared/session.py
maxisbey: `if isinstance(message.message, JSONRPCError) and message.message.id is None` only proves that either both are true, or one or more of the two isn't true. If the condition is false it doesn't necessarily mean that message.message.id is or isn't None, just that both `isinstance(message.message, JSONRPCError)` and `message.message.id is None` are not _both_ true.

Alternative would be doing something like:

```python
if message.message.id is None:
    if isinstance(...):  # pragma: no branch
        do stuff
    return
```

above, but that required the pragma.

maxisbey: found a better way that pyright is apparently happy with

tests/test_types.py
maxisbey: This tests the custom `_serialize` method, not normal vanilla model_dump though?

src/mcp/types/jsonrpc.py
maxisbey: `exclude_none` is applied recursively to pydantic objects, but not to dicts if there's a dict in a field.

Another option would be using `exclude_unset` on the jsonrpc layer, and `exclude_none` on the mcp layer

maxisbey: Reworked to using `exclude_unset` on JSONRPC related objects, and keeping `exclude_none` on MCP stuff. Had to change a few places which set things like `error=None` and remove them so they no longer show up on the wire.

tests/test_types.py
maxisbey: removed the `_serialized`, so removed this unit test

Kludex (approved):

PR #2065, opened 2026-02-14
feat: add auth parameter to ClientSessionGroup server parameters

### LINKED ISSUE #1723: `auth` specification in `ClientSessionGroup`
### Description

It would be nice for the `ClientSessionGroup` to be OAuth compatible by allowing us to specify the OAuth `httpx.Auth` provider when passing in the `ServerParameters` rather than passing the bearer tokens manually via the `headers` field. This will allow us to do the OAuth token exchange flow straight from the `ClientSessionGroup`.

### References

Here is the singleton `ClientSession` example of an OAuth flow: https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py

It would be nice to have an example of a `ClientSessionGroup` using that same flow.

IT-HONGREAT:
## Summary

Closes #1723

`ClientSessionGroup`'s `SseServerParameters` and `StreamableHttpParameters` lacked an `auth` field, making it impossible to use `httpx.Auth` handlers (e.g., OAuth token exchange flows). Users had to manually set bearer tokens via the `headers` field, bypassing automatic token refresh.

This adds an optional `auth: httpx.Auth | None` parameter to both models and wires it through to the underlying `sse_client()` and `create_mcp_http_client()` calls.

## Changes

- Add `auth: httpx.Auth | None = None` to `SseServerParameters` and `StreamableHttpParameters`
- Add `ConfigDict(arbitrary_types_allowed=True)` to both models (required for `httpx.Auth`)
- Pass `auth` through in `_establish_session()` for both SSE and Streamable HTTP paths

## Test plan

- [x] Verify `auth` is passed to `sse_client()` in SSE path
- [x] Verify `auth` is passed to `create_mcp_http_client()` in Streamable HTTP path
- [x] Existing parametrized tests updated and passing

### DIFF
--- a/tests/client/test_session_group.py
+++ b/tests/client/test_session_group.py
@@ -355,6 +355,7 @@ async def test_client_session_group_establish_session_parameterized(
                     headers=server_params_instance.headers,
                     timeout=server_params_instance.timeout,
                     sse_read_timeout=server_params_instance.sse_read_timeout,
+                    auth=server_params_instance.auth,
                 )
             elif client_type_name == "streamablehttp":  # pragma: no branch
                 assert isinstance(server_params_instance, StreamableHttpParameters)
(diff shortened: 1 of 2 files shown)


tests/client/test_session_group.py
```diff
@@ -385,3 +386,96 @@ async def test_client_session_group_establish_session_parameterized(
             # 3. Assert returned values
             assert returned_server_info is mock_initialize_result.server_info
             assert returned_session is mock_entered_session
+
+
+@pytest.mark.anyio
+async def test_establish_session_sse_passes_auth():
+    """_establish_session should pass auth to sse_client for SseServerParameters."""
+    mock_auth = mock.Mock(spec=httpx.Auth)
+    server_params = SseServerParameters(url="http://test.com/sse", auth=mock_auth)
```
PsymoNiko, line 395: The URL address shouldn't be as a dynamic path?

and is it ok to test on `http`?

what about the value of `auth` what if we want to test on real productive data?

PsymoNiko (commented):
It needs to double check to fix the COnflict?

PR #2067, opened 2026-02-16
fix: relax MIME type validation to accept all RFC 2045 types

### LINKED ISSUE #1756: Consider removing or relaxing MIME type validation in FastMCP resources
## Summary

The MIME type validation regex in FastMCP's `Resource` class is more restrictive than RFC 2045 and diverges from the TypeScript SDK (which has no validation). We should consider aligning with the TypeScript SDK approach.

## Background

PR #1755 fixed the immediate issue of MIME type parameters being rejected (e.g., `text/html;profile=mcp-app`), but the current regex still doesn't fully support RFC 2045:

**Current regex (after #1755):**
```python
pattern=r"^[a-zA-Z0-9]+/[a-zA-Z0-9\-+.]+(;\s*[a-zA-Z0-9\-_.]+=[a-zA-Z0-9\-_.]+)*$"
```

**Limitations:**
- Doesn't support quoted string values: `text/plain; charset="utf-8"` ❌
- Doesn't allow all valid RFC 2045 token characters (e.g., `!`, `#`, `*`, `+`, etc.)

## TypeScript SDK Comparison

The TypeScript SDK does not validate MIME types at all ([types.ts:834](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/types.ts#L834)):
```typescript
mimeType: z.optional(z.string()),
```

## MCP Spec

The MCP spec defines `mimeType` as an optional string with no format constraints.

## Options

1. **Remove validation entirely** (align with TypeScript SDK and spec)
2. **Improve the regex** to support quoted strings and more token characters

---

BabyChrist666:
## Summary

Closes #1756

- **Replaced** the restrictive regex pattern on `Resource.mime_type` with a lightweight `field_validator` that only checks for the presence of `/` (type/subtype structure)
- **Aligns** with the MCP spec which defines `mimeType` as an optional string with no format constraints, and matches the TypeScript SDK behavior (no validation)
- **Fixes** rejection of valid RFC 2045 MIME types including quoted parameter values (e.g. `text/plain; charset="utf-8"`), wildcard types (`*/*`), and complex boundary parameters

### What changed

**`src/mcp/server/mcpserver/resources/base.py`**
- Removed `pattern=r"^[a-zA-Z0-9]+/[a-zA-Z0-9\-+.]+(;\s*[a-zA-Z0-9\-_.]+=[a-zA-Z0-9\-_.]+)*$"` from the `mime_type` Field
- Added `validate_mime_type` field_validator that only rejects values without a `/` separator

### Why this approach

The previous regex was too restrictive for RFC 2045 compliance. Rather than trying to write a more complex regex that covers all edge cases (quoted strings, special characters, wildcards, etc.), this PR follows the MCP spec's intent: `mimeType` is a plain string, and validation should only catch obvious mistakes (like passing `"plaintext"` instead of `"text/plain"`). This is consistent with:
- The TypeScript SDK (no MIME validation)
- `ResourceTemplate` in this SDK (already had no pattern constraint)
- Protocol-level types in `_types.py` (no validation)

## Test plan

- [x] Added `tests/issues/test_1756_mime_type_relaxed.py` with 12 test cases covering:
  - Basic MIME types, quoted parameter values, vendor types, wildcards, complex parameters
  - Invalid MIME types (no `/`, empty string)
  - Default MIME type preservation
- [x] Added `test_resource_mime_type_validation` to existing `test_resources.py`
- [x] All existing tests pass (including `test_1754_mime_type_parameters.py`)
- [x] 100% code coverage on `base.py`
- [x] ruff + pyright clean

### DIFF
--- a/src/mcp/server/mcpserver/resources/base.py
+++ b/src/mcp/server/mcpserver/resources/base.py
@@ -26,11 +26,27 @@ class Resource(BaseModel, abc.ABC):
     mime_type: str = Field(
         default="text/plain",
         description="MIME type of the resource content",
-        pattern=r"^[a-zA-Z0-9]+/[a-zA-Z0-9\-+.]+(;\s*[a-zA-Z0-9\-_.]+=[a-zA-Z0-9\-_.]+)*$",
     )
     icons: list[Icon] | None = Field(default=None, description="Optional list of icons for this resource")
     annotations: Annotations | None = Field(default=None, description="Optional annotations for the resource")
     meta: dict[str, Any] | None = Field(default=None, description="Optional metadata for this resource")
+
+    @field_validator("mime_type")
+    @classmethod
+    def validate_mime_type(cls, value: str) -> str:
+        """Validate that mime_type has a basic type/subtype structure.
+
+        The MCP spec defines mimeType as an optional string with no format
+        constraints. This validator only checks for the minimal type/subtype
+        structure to catch obvious mistakes, without restricting valid MIME
+        types per RFC 2045.
+        """
+        if "/" not in value:
+            raise ValueError(
+                f"Invalid MIME type '{value}': must contain a '/' separating type and subtype "
+                f"(e.g. 'text/plain', 'application/json')"
+            )
+        return value
 
     @field_validator("name", mode="before")
     @classmethod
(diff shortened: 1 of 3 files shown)


maxisbey:
Thanks for the PR! The approach here was sound, but closing in favor of #2235 which goes one step further and removes validation entirely — this keeps `Resource` consistent with `ResourceTemplate` (same module, no validation) and with the TypeScript/Go/C# SDKs.

PR #2075, opened 2026-02-17
Reject JSON-RPC requests with null id instead of misclassifying as notifications

### LINKED ISSUE #2057: Requests with "id": null silently misclassified as notifications
### Initial Checks

### Description

### Example Code

```python

try:
except Exception:

```

### Python & MCP Python SDK

```
Python 3.13
```

BabyChrist666:
## Summary

Fixes #2057 — prevents `"id": null` JSON-RPC requests from being silently misclassified as notifications.

### Root cause

When Pydantic validates a `JSONRPCMessage` union with `{"jsonrpc": "2.0", "method": "initialize", "id": null}`:

1. `JSONRPCRequest` correctly rejects it (`RequestId` = `int | str`, no `None`)
2. `JSONRPCNotification` **absorbs it** because `"id": None` becomes an extra field
3. The transport layer sees "notification" → returns 202 with no response
4. The caller gets no error and no response — a silent failure

This violates the MCP spec (§4.2.1) which mandates that request IDs must be strings or integers, and requests with `id: null` must be rejected.

### Fix

Add a `model_validator(mode="before")` on `JSONRPCNotification` that rejects any input containing an `"id"` field. Per JSON-RPC 2.0, a notification is defined as a request **without** an `id` member — if `"id"` is present (regardless of value), it's a request, not a notification.

### Changes

- **`src/mcp/types/jsonrpc.py`**: Add `_reject_id_field` model validator to `JSONRPCNotification`
- **`tests/issues/test_2057_null_id_rejected.py`**: 12 test cases covering:
  - Null id rejection at `JSONRPCRequest`, `JSONRPCNotification`, and `JSONRPCMessage` union levels
  - Both `validate_python` and `validate_json` paths (matching transport usage)
  - Rejection of any id value on notifications (null, int, str)
  - Valid requests and notifications still parse correctly

### Test plan

- [x] `JSONRPCRequest` still rejects `id: null` (existing behavior preserved)
- [x] `JSONRPCNotification` now rejects any input with `"id"` field
- [x] `jsonrpc_message_adapter.validate_python({"id": null, ...})` raises `ValidationError`
- [x] `jsonrpc_message_adapter.validate_json(...)` raises `ValidationError`
- [x] Valid notifications (no `id` field) still work
- [x] Valid requests (string/int `id`) still work
- [x] `ruff check` passes
- [x] All existing tests pass

### DIFF
--- a/src/mcp/types/jsonrpc.py
+++ b/src/mcp/types/jsonrpc.py
@@ -20,14 +20,7 @@ class JSONRPCRequest(BaseModel):
 
 
 class JSONRPCNotification(BaseModel):
-    """A JSON-RPC notification which does not expect a response.
-
-    Per JSON-RPC 2.0, a notification is a request without an ``id`` member.
-    Messages that include ``id`` (even with a ``null`` value) are requests,
-    not notifications.  The validator below prevents Pydantic's union
-    resolution from silently absorbing an invalid ``"id": null`` request
-    as a notification (see :issue:`2057`).
-    """
+    """A JSON-RPC notification which does not expect a response."""
 
     jsonrpc: Literal["2.0"]
     method: str
@@ -36,12 +29,10 @@ class JSONRPCNotification(BaseModel):
     @model_validator(mode="before")
     @classmethod
     def _reject_id_field(cls, data: dict[str, Any]) -> dict[str, Any]:
-        """Reject messages that contain an ``id`` field.
-
-        JSON-RPC notifications MUST NOT include an ``id`` member.  If the
-        raw payload contains ``"id"`` (regardless of its value), it is a
-        malformed request — not a valid notification.
-        """
+        """Reject payloads containing an 'id' field (notifications must not have one)."""
+        # Per JSON-RPC 2.0, notifications must not include an "id" member.
+        # Without this check, Pydantic's union resolution silently absorbs
+        # invalid "id": null requests as notifications (#2057).
         if "id" in data:
             raise ValueError(
                 "Notifications must not include an 'id' field. "
--- a/tests/issues/test_2057_null_id_rejected.py
+++ b/tests/issues/test_2057_null_id_rejected.py
@@ -14,81 +14,32 @@ from pydantic import ValidationError
 
 from mcp.types import (
     JSONRPCNotification,
-    JSONRPCRequest,
     jsonrpc_message_adapter,
 )
 
 
-class TestNullIdRejection:
-    """Verify that ``"id": null`` is never silently absorbed."""
+def test_notification_rejects_id_field() -> None:
+    """JSONRPCNotification must not accept messages with an 'id' field."""
+    with pytest.raises(ValidationError, match="must not include an 'id' field"):
+        JSONRPCNotification.model_validate({"jsonrpc": "2.0", "method": "initialize", "id": None})
 
-    def test_request_rejects_null_id(self) -> None:
-        """JSONRPCRequest correctly rejects null id."""
-        with pytest.raises(ValidationError):
-            JSONRPCRequest.model_validate({"jsonrpc": "2.0", "method": "initialize", "id": None})
 
-    def test_notification_rejects_id_field(self) -> None:
-        """JSONRPCNotification must not accept messages with an 'id' field."""
-        with pytest.raises(ValidationError, match="must not include an 'id' field"):
-            JSONRPCNotification.model_validate({"jsonrpc": "2.0", "method": "initialize", "id": None})
+@pytest.mark.parametrize("id_value", [None, 0, 1, "", "abc"])
+def test_notification_rejects_any_id_value(id_value: object) -> None:
+    """Notification rejects 'id' regardless of value — null, int, or str."""
+    with pytest.raises(ValidationError):
+        JSONRPCNotification.model_validate({"jsonrpc": "2.0", "method": "test", "id": id_value})
 
-    def test_notification_rejects_any_id_value(self) -> None:
-        """Notification rejects 'id' regardless of value — null, int, or str."""
-        for id_value in [None, 0, 1, "", "abc"]:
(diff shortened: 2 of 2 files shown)


src/mcp/types/jsonrpc.py
```diff
@@ -20,12 +20,35 @@ class JSONRPCRequest(BaseModel):
 
 
 class JSONRPCNotification(BaseModel):
-    """A JSON-RPC notification which does not expect a response."""
+    """A JSON-RPC notification which does not expect a response.
+
```
clouatre: Every other model in this file uses a single-line docstring (`JSONRPCRequest`, `JSONRPCResponse`, `JSONRPCError`). Consider trimming back to:

```python
"""A JSON-RPC notification which does not expect a response."""
```

The spec rationale fits better as a short comment above the validator. Also, `:issue:\`2057\`` won't render as a link (no Sphinx issue role configured). Use `#2057` or the full GitHub URL.

src/mcp/types/jsonrpc.py
```diff
@@ -20,12 +20,35 @@ class JSONRPCRequest(BaseModel):
...
 
     jsonrpc: Literal["2.0"]
     method: str
     params: dict[str, Any] | None = None
 
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_id_field(cls, data: dict[str, Any]) -> dict[str, Any]:
+        """Reject messages that contain an ``id`` field.
+
```
clouatre: The validator logic is clean and correct. The docstring could be a one-liner to match the file's style:

```python
"""Reject payloads containing an 'id' field (notifications must not have one)."""
```

tests/issues/test_2057_null_id_rejected.py
```diff
@@ -0,0 +1,94 @@
...
+
+from mcp.types import (
+    JSONRPCNotification,
+    JSONRPCRequest,
+    jsonrpc_message_adapter,
+)
+
+
+class TestNullIdRejection:
+    """Verify that ``"id": null`` is never silently absorbed."""
```
clouatre: All 18 other files in `tests/issues/` use module-level test functions. This is the only one using a test class. Switching to module-level functions would match the convention.

tests/issues/test_2057_null_id_rejected.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def test_request_rejects_null_id(self) -> None:
+        """JSONRPCRequest correctly rejects null id."""
```
clouatre: This test and the 7 others below it (`test_valid_notification_still_works`, `test_valid_notification_with_params`, `test_valid_request_with_string_id`, `test_valid_request_with_int_id`, `test_message_adapter_parses_valid_request`, `test_message_adapter_parses_valid_notification`, `test_message_adapter_parses_notification_json`) verify pre-existing behavior already covered by `tests/test_types.py::test_jsonrpc_request` and the broader suite.

The `tests/issues/` convention is to test the specific bug and its fix. I'd keep 4 tests:
1. `test_notification_rejects_id_field` (core fix)
2. `test_notification_rejects_any_id_value` (parameterized edge cases)
3. `test_message_adapter_rejects_null_id` (actual bug path)
4. `test_message_adapter_rejects_null_id_json` (transport path)

src/mcp/types/jsonrpc.py
BabyChrist666: Great catch, will trim the docstring and fix the issue reference. Thanks\!

BabyChrist666: Great catch, will trim the docstring and fix the issue reference. Thanks!

src/mcp/types/jsonrpc.py
BabyChrist666: Agreed, will make it a one-liner.

tests/issues/test_2057_null_id_rejected.py
BabyChrist666: Good point, will switch to module-level functions to match the convention.

tests/issues/test_2057_null_id_rejected.py
BabyChrist666: Makes sense — I'll trim down to the 4 tests that directly verify the fix. The pre-existing behavior tests are redundant with the broader suite.

src/mcp/types/jsonrpc.py
```diff
@@ -26,6 +26,20 @@ class JSONRPCNotification(BaseModel):
...
+        """Reject payloads containing an 'id' field (notifications must not have one)."""
+        # Per JSON-RPC 2.0, notifications must not include an "id" member.
+        # Without this check, Pydantic's union resolution silently absorbs
+        # invalid "id": null requests as notifications (#2057).
+        if "id" in data:
+            raise ValueError(
+                "Notifications must not include an 'id' field. "
+                "A JSON-RPC message with 'id' is a request, not a notification."
+            )
+        return data
```
Kludex, line 41: I don't think it makes sense to validate a key from another structure in this one.

BabyChrist666: Good point — I see how it looks odd to reject a key that isn't part of this model's schema.

The reason it's here: the actual bug path is `jsonrpc_message_adapter` using Pydantic's discriminated union with `"method"` as the discriminator. When raw JSON contains both `"method"` **and** `"id": null`, Pydantic routes it to `JSONRPCNotification` (because it has `method`) and silently drops `id` — the sender intended a request but it gets misclassified as a notification with no error.

The alternative fix would be in `jsonrpc_message_adapter` itself (a custom validator on the TypeAdapter/Union). Would you prefer that approach instead? Happy to move the validation there if that's a cleaner fit architecturally.

BabyChrist666:
Following up on the February CI failures on this branch — those appear to have been transient runner issues. Current checks are passing. This PR is ready for review whenever the team has bandwidth.

PR #2076, opened 2026-02-17
fix(auth): accept empty optional OAuth client metadata URLs

### LINKED ISSUE #1665: URL Validation on Client Metadata is too strict
### Initial Checks

### Description

Pydantic AnyHttpUrl is too strict for client metadata validation.  https://datatracker.ietf.org/doc/html/rfc7591#section-2 specifies that all fields are optional. Empty strings are commonly emitted when there's no value.  

In addition fields like client_uri are informational and I'm not sure they should even be validated as http.  Any URL should technically be supported including non-http URLs.  In general, I think the SDK should only apply strict pydantic validation where it affects the security or functioning of the client.

1 validation error for OAuthClientInformationFull
client_uri
  Input should be a valid URL, input is empty [type=url_parsing, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/url_parsing

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
1.21.1
```

mrutunjay-kinagi:
## Summary
- normalize empty-string optional URL fields in `OAuthClientMetadata` to `None` before URL validation
- keep existing URL validation rules for non-empty values
- add a regression test for empty `client_uri`/`logo_uri`/`tos_uri`/`policy_uri`/`jwks_uri` inputs

## Why
Fixes validation failures when auth servers emit empty strings for optional client metadata URL fields.

Closes #1665

## Validation
- `pytest -q tests/client/test_auth.py -k "empty_optional_uris"`
- `ruff check src/mcp/shared/auth.py tests/client/test_auth.py`
- `ruff format --check src/mcp/shared/auth.py tests/client/test_auth.py`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,6 +31,11 @@ This document contains critical information about working with this codebase. Fo
    - IMPORTANT: Before pushing, verify 100% branch coverage on changed files by running
      `uv run --frozen pytest -x` (coverage is configured in `pyproject.toml` with `fail_under = 100`
      and `branch = true`). If any branch is uncovered, add a test for it before pushing.
+   - Avoid `anyio.sleep()` with a fixed duration to wait for async operations. Instead:
+     - Use `anyio.Event` — set it in the callback/handler, `await event.wait()` in the test
+     - For stream messages, use `await stream.receive()` instead of `sleep()` + `receive_nowait()`
+     - Exception: `sleep()` is appropriate when testing time-based features (e.g., timeouts)
+   - Wrap indefinite waits (`event.wait()`, `stream.receive()`) in `anyio.fail_after(5)` to prevent hangs
 
 Test files mirror the source tree: `src/mcp/client/streamable_http.py` → `tests/client/test_streamable_http.py`
 Add tests to the existing file for that module.
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -169,6 +169,8 @@ class StreamableHTTPServerTransport:
         ] = {}
         self._sse_stream_writers: dict[RequestId, MemoryObjectSendStream[dict[str, str]]] = {}
         self._terminated = False
+        # Idle timeout cancel scope; managed by the session manager.
+        self.idle_scope: anyio.CancelScope | None = None
 
     @property
     def is_terminated(self) -> bool:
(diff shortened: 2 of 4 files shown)


mrutunjay-kinagi:
Follow-up pushed in e9a78ec to fix the  pre-commit failure by validating the empty-URI fixture via  (runtime path) instead of passing empty strings directly to typed constructor params.

mrutunjay-kinagi:
Follow-up pushed in `e9a78ec` to fix the pre-commit `pyright` failure.

Change made:
- Updated the empty-URI regression test to use `OAuthClientMetadata.model_validate({...})` so empty strings are exercised through runtime validation without violating static constructor type checks.

mrutunjay-kinagi:
All checks are now passing on the latest commit (`e9a78ec`).

Ready for maintainer review when you have bandwidth. Thanks!

Kludex:
Where in the RFC it says that empty is not valid?

mrutunjay-kinagi:
Good question.

The RFC language we based this on is RFC 7591 §2 for these fields (`client_uri`, `logo_uri`, `tos_uri`, `policy_uri`, `jwks_uri`):
- they are optional metadata fields,
- but **if present** they are "URL string" fields and "MUST point to a valid ..." target.

Examples from §2:
- `client_uri`: "MUST point to a valid web page"
- `logo_uri`: "MUST point to a valid image file"
- `tos_uri`/`policy_uri`: "MUST point to a valid web page"
- `jwks_uri`: "MUST point to a valid JWK Set document"

So this change is not treating `""` as a valid URL. It normalizes empty-string placeholders to `None` (effectively omitted) before URL validation, which keeps the RFC "if present, must be valid" rule while improving interop with providers that emit empty strings for optional fields.

Kludex:
Link please.

mrutunjay-kinagi:
Absolutely — here are the direct references:

- RFC 7591, Section 2 (Client Metadata): https://www.rfc-editor.org/rfc/rfc7591#section-2
- Datatracker view (same section): https://datatracker.ietf.org/doc/html/rfc7591#section-2

In that section, for these optional fields (`client_uri`, `logo_uri`, `tos_uri`, `policy_uri`, `jwks_uri`), the text says that **if present** they are URL strings and must point to valid targets (web page/image/JWK set).

Kludex:
I don't think this needs any fixing. The client side sending empty string in a field is the wrong behavior.

PR #2077, opened 2026-02-17
fix(auth): forward user-agent to oauth flow requests

### LINKED ISSUE #1664: User-Agent header in sHTTP transport is not forwarded to auth flow
### Initial Checks

### Description

Passing a custom `User-Agent` header to the sHTTP transport, e.g. [here](https://github.com/modelcontextprotocol/python-sdk/blob/27279bc157cbc03f7fe7758fd55a4b34c5652f42/examples/clients/simple-auth-client/mcp_simple_auth_client/main.py#L214-L220) does not result in auth flow requests including that user agent. There also does not appear to be any way to customize this behavior externally to the SDK itself.

In particular, this becomes a problem when working with AWS WAF, as one of its [baseline rules](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-baseline.html#aws-managed-rule-groups-baseline-crs) is to require all requests to include a `User-Agent` header, to filter out low-grade spam.

I am currently recommending to internal teams that they disable this rule for the time being, but this is not a great solution for them, as it means removing an application guardrail to work around an SDK limitation.

### Example Code

```Python
print("📡 Opening StreamableHTTP transport connection with auth...")
    url=self.server_url,
    auth=oauth_auth,
    timeout=timedelta(seconds=60),
    headers={
        "User-Agent": "mcp-python-sdk/0.1.0",
    },
    await self._run_session(read_stream, write_stream, get_session_id)
```

### Python & MCP Python SDK

Latest SDK commit: [27279bc](https://github.com/modelcontextprotocol/python-sdk/commit/27279bc157cbc03f7fe7758fd55a4b34c5652f42) (1.22.0)
Python 3.13.2

mrutunjay-kinagi:
## Summary
- forward caller `User-Agent` from the original MCP request into OAuth flow sub-requests
- apply forwarding to PRM discovery, auth server metadata discovery, dynamic client registration, and token exchange/refresh requests
- add regression coverage ensuring forwarded `User-Agent` is preserved throughout the 401 -> OAuth handshake path

## Why
Fixes missing `User-Agent` propagation in Streamable HTTP auth flow requests, which can break environments that enforce WAF rules requiring `User-Agent` headers.

Closes #1664

## Validation
- `pytest -q tests/client/test_auth.py -k "forwards_user_agent or auth_flow_with_no_tokens"`
- `ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `ruff format --check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`


mrutunjay-kinagi:
Thanks for sharing this run.

I pushed a follow-up commit (`40f173f`) to address the failure pattern shown there:
- removed a stale `# pragma: no cover` on `tests/client/test_auth.py:48` that is now exercised by the new auth-flow test.

That was what caused Linux matrix failures in that run under `strict-no-cover`.

mrutunjay-kinagi:
Follow-up commit `40f173f` addressed the previous Linux `strict-no-cover` failure, and all checks are now green.

@Kludex  Ready for review Thanks!

mrutunjay-kinagi:
Follow-up on CI run https://github.com/modelcontextprotocol/python-sdk/actions/runs/22261553587 (job 64400559254):\n\n- Only checks / test (3.13, locked, windows-latest) failed.\n- Failure appears unrelated/flaky (not in auth-flow code path):\n  - tests/client/test_stdio.py::TestChildProcessCleanup::test_early_parent_exit (AssertionError: Child should be writing)\n  - tests/client/transports/test_memory.py::test_raise_exceptions\n- Workflow conclusion is success and all-green passed, but this stale failed check remains visible.\n\nI don't have permission to rerun upstream Actions jobs from CLI. If needed, a maintainer rerun of the failed job should clear this.

mrutunjay-kinagi:
@maxisbey : Please take a look at the PR let me know of what do you think ?

mrutunjay-kinagi:
@Kludex  @maxisbey  Please take a look at this PR.

PR #2082, opened 2026-02-18
fix: prevent command injection in example URL opening

maxisbey:
## Summary

Fix a command injection issue in the URL elicitation example client (`examples/snippets/clients/url_elicitation_client.py`).

## Problem

The `open_browser()` function used `subprocess.run(["start", url], shell=True)` on Windows, which allows shell metacharacter injection via crafted URLs. A malicious MCP server could send a URL like `https://example.com/?state=abc&calc` during URL elicitation, and `cmd.exe` would interpret `&` as a command separator, executing `calc` (or any arbitrary command).

## Fix

- Replace all platform-specific `subprocess` calls with `webbrowser.open()`, which handles cross-platform browser opening safely without shell invocation
- Add URL scheme validation (allowlist: `http`, `https`) before prompting the user, returning `ElicitResult(action="decline")` for disallowed schemes
- Simplify `open_browser()` to a pure browser-opening helper
- Remove unused `subprocess` and `sys` imports

This aligns with the safe pattern already used in `examples/clients/simple-auth-client/mcp_simple_auth_client/main.py`.

### DIFF
--- a/examples/snippets/clients/url_elicitation_client.py
+++ b/examples/snippets/clients/url_elicitation_client.py
@@ -24,8 +24,6 @@ from __future__ import annotations
 
 import asyncio
 import json
-import subprocess
-import sys
 import webbrowser
 from typing import Any
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


examples/snippets/clients/url_elicitation_client.py
```diff
@@ -121,15 +119,17 @@ def extract_domain(url: str) -> str:
...
 
+ALLOWED_SCHEMES = {"http", "https"}
+
+
 def open_browser(url: str) -> None:
-    """Open URL in the default browser."""
+    """Open URL in the default browser after validating the scheme."""
+    parsed = urlparse(url)
+    if parsed.scheme.lower() not in ALLOWED_SCHEMES:
+        print(f"Refusing to open URL with unsupported scheme '{parsed.scheme}': {url}")
```
felixweinberger: should this throw e.g. a `ValueError` instead of just printing?

maxisbey: yea moved out of here into the handle elicitation function

examples/snippets/clients/url_elicitation_client.py
```diff
@@ -124,12 +132,7 @@ def extract_domain(url: str) -> str:
 def open_browser(url: str) -> None:
```
Kludex: Is this even useful now?

maxisbey: Yea not really, will remove.

PR #2085, opened 2026-02-18
[v1.x] fix: prevent command injection in example URL opening

maxisbey:
Backport of #2082 to v1.x.

## Motivation and Context

The `open_browser()` function in the URL elicitation example client used `subprocess.run(["start", url], shell=True)` on Windows, which allows shell metacharacter injection via crafted URLs. A malicious MCP server could send a URL like `https://example.com/?state=abc&calc` during URL elicitation, and `cmd.exe` would interpret `&` as a command separator, executing arbitrary commands.

## How Has This Been Tested?

Code review — this is an example file, not library code. The changes are identical to the merged #2082.

## Breaking Changes

None.

## Types of changes

## Checklist

### DIFF
--- a/examples/snippets/clients/url_elicitation_client.py
+++ b/examples/snippets/clients/url_elicitation_client.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 import webbrowser
 from typing import Any
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


examples/snippets/clients/url_elicitation_client.py
```diff
@@ -105,7 +113,11 @@ async def handle_url_elicitation(
 
     # Open the browser
     print(f"\nOpening browser to: {url}")
-    open_browser(url)
+    try:
+        webbrowser.open(url)
+    except Exception as e:
+        print(f"Failed to open browser: {e}")
+        print(f"Please manually open: {url}")
```
felixweinberger: `logger.exception` I think while we're here

maxisbey: done

Issue #2054, opened 2026-02-13
Add dependency injection support to MCPServer


maxisbey:
From the V2 checklist in #1253:

> Create dependency injection in `MCPServer`.

MCPServer currently has no built-in mechanism for injecting dependencies (database connections, auth services, configuration objects, etc.) into tool, resource, and prompt handlers. The existing `Context` parameter injection is specific to the MCP request context and does not cover arbitrary user-defined dependencies.

A DI system — similar to FastAPI's `Depends()` — would allow:

- Declaring dependencies as handler parameters that the framework resolves automatically
- Sharing expensive resources (DB pools, HTTP clients) across handlers without global state
- Making handlers more testable by swapping dependencies in tests
- Composing dependencies (dependencies that depend on other dependencies)

### Related but distinct issues
- #1684 — Refactoring contextvars into explicit parameters (context passing, not general DI)
- #1695 — Revisiting Context injection mechanism (closed, about the `Context` type specifically)

### Open questions
- Should this follow FastAPI's `Depends()` pattern exactly, or adapt it for MCP's handler signatures?
- How does this interact with the `Context` parameter and lifespan state?
- Should dependencies be scoped per-request, per-session, or per-server?

Ref: Kludex's V2 checklist in #1253

maxisbey:
This landed as part of v2, which is now released (via #2969) — closing. Feel free to reopen if there's something still outstanding.

PR #2087, opened 2026-02-18
feat: Add FastAPI-style dependency injection system for MCPServer

gspeter-max:
## Summary

Implements FastAPI-style dependency injection for MCPServer tools, prompts, and resources.

## Features

- **`Depends()` class**: Declare dependencies in function parameters
- **`DependencyResolver`**: Automatic dependency graph resolution
- **Nested dependencies**: Dependencies can depend on other dependencies
- **Async support**: Both sync and async dependency providers
- **Caching**: Per-request caching with `use_cache` parameter
- **Testing overrides**: `override_dependency()` method for easy testing

## Example Usage

```python
from mcp import MCPServer, Depends

def get_database():
    return Database()

server = MCPServer("my-server")

@server.tool()
def query_users(limit: int, db: Database = Depends(get_database)):
    return db.query("SELECT * FROM users LIMIT ?", limit)
```

## Testing

```python
# Override dependencies for testing
def get_test_db():
    return MockDatabase([...])

server.override_dependency(get_database, get_test_db)
```

## Implementation Details

- Modified `Tool.from_function()` to detect and skip `Depends()` parameters
- Modified `Tool.run()` to resolve dependencies before execution
- Added `override_dependency()` method to `MCPServer`
- Infrastructure in place for prompts and resources

## Test Coverage

- 21 new tests covering all dependency injection scenarios
- 100% branch coverage on new code
- All existing tests pass (1134 passed)

## Backward Compatibility

✅ 100% backward compatible - existing code without `Depends()` continues to work unchanged.

Refs: #2054

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,6 +225,11 @@ exclude_lines = [
     "raise NotImplementedError",
     "^\\s*\\.\\.\\.\\s*$",
 ]
+exclude_also = [
+    # Exclude flaky Windows tests from coverage calculation
+    # These tests are skipped on Python 3.13+ Windows due to timing issues
+    "sys.platform == \"win32\" and sys.version_info >= \\(3, 13\\)",
+]
 
 # https://coverage.readthedocs.io/en/latest/config.html#paths
 [tool.coverage.paths]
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for taking a swing at this — the E2E tests using `Client` are the right pattern.

Going to close though; DI touches enough of the core handler path that we want to drive the design on this one ourselves. The `Depends()`-as-default typing issue alone (what forced the pyright suppressions here) has a couple of viable answers and we haven't picked one yet.

PR #2096, opened 2026-02-19
Fix the session binding logic for tasks.

localden:
Ensures that session IDs are used for tasks, avoiding cross-polination.

### DIFF
--- a/src/mcp/client/_memory.py
+++ b/src/mcp/client/_memory.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import uuid
 from collections.abc import AsyncIterator
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from types import TracebackType
@@ -50,12 +51,14 @@ class InMemoryTransport:
 
             async with anyio.create_task_group() as tg:
                 # Start server in background
+                memory_session_id = uuid.uuid4().hex
                 tg.start_soon(
                     lambda: actual_server.run(
                         server_read,
                         server_write,
                         actual_server.create_initialization_options(),
                         raise_exceptions=self._raise_exceptions,
+                        session_id=memory_session_id,
                     )
                 )
 
--- a/src/mcp/client/_memory.py
+++ b/src/mcp/client/_memory.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import uuid
 from collections.abc import AsyncIterator
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from types import TracebackType
@@ -51,14 +50,12 @@ class InMemoryTransport:
 
             async with anyio.create_task_group() as tg:
                 # Start server in background
-                memory_session_id = uuid.uuid4().hex
                 tg.start_soon(
                     lambda: actual_server.run(
                         server_read,
                         server_write,
                         actual_server.create_initialization_options(),
                         raise_exceptions=self._raise_exceptions,
-                        session_id=memory_session_id,
                     )
                 )
 
--- a/src/mcp/server/experimental/task_result_handler.py
+++ b/src/mcp/server/experimental/task_result_handler.py
@@ -80,6 +80,7 @@ class TaskResultHandler:
         request: GetTaskPayloadRequest,
         session: ServerSession,
         request_id: RequestId,
+        session_id: str | None = None,
     ) -> GetTaskPayloadResult:
         """Handle a tasks/result request.
 
@@ -94,6 +95,7 @@ class TaskResultHandler:
             request: The GetTaskPayloadRequest
             session: The server session for sending messages
             request_id: The request ID for relatedRequestId routing
+            session_id: Optional session identifier for access control.
 
         Returns:
             GetTaskPayloadResult with the task's final payload
@@ -101,7 +103,7 @@ class TaskResultHandler:
         task_id = request.params.task_id
 
         while True:
-            task = await self._store.get_task(task_id)
+            task = await self._store.get_task(task_id, session_id=session_id)
             if task is None:
                 raise MCPError(code=INVALID_PARAMS, message=f"Task not found: {task_id}")
 
@@ -109,7 +111,7 @@ class TaskResultHandler:
 
             # If task is terminal, return result
             if is_terminal(task.status):
-                result = await self._store.get_result(task_id)
+                result = await self._store.get_result(task_id, session_id=session_id)
                 # GetTaskPayloadResult is a Result with extra="allow"
                 # The stored result contains the actual payload data
                 # Per spec: tasks/result MUST include _meta with related-task metadata
--- a/src/mcp/server/experimental/task_result_handler.py
+++ b/src/mcp/server/experimental/task_result_handler.py
@@ -80,7 +80,7 @@ class TaskResultHandler:
         request: GetTaskPayloadRequest,
         session: ServerSession,
         request_id: RequestId,
-        session_id: str | None = None,
+        session_id: str,
     ) -> GetTaskPayloadResult:
         """Handle a tasks/result request.
 
@@ -95,7 +95,7 @@ class TaskResultHandler:
             request: The GetTaskPayloadRequest
             session: The server session for sending messages
             request_id: The request ID for relatedRequestId routing
-            session_id: Optional session identifier for access control.
+            session_id: Session identifier for access control.
 
         Returns:
             GetTaskPayloadResult with the task's final payload
--- a/src/mcp/server/experimental/task_result_handler.py
+++ b/src/mcp/server/experimental/task_result_handler.py
@@ -80,7 +80,7 @@ class TaskResultHandler:
         request: GetTaskPayloadRequest,
         session: ServerSession,
         request_id: RequestId,
-        session_id: str,
+        session_id: str | None,
     ) -> GetTaskPayloadResult:
         """Handle a tasks/result request.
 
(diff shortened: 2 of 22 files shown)


src/mcp/server/session.py
```diff
@@ -83,9 +83,11 @@ def __init__(
         write_stream: MemoryObjectSendStream[SessionMessage],
         init_options: InitializationOptions,
         stateless: bool = False,
+        session_id: str | None = None,
```
Kludex, line 87: Why would this be necessary for the ServerSession constructor?

maxisbey: according to spec you don't need a session ID, and this would tie into the current weird stateless_mode we have.

src/mcp/server/streamable_http_manager.py
```diff
@@ -172,6 +172,7 @@ async def run_stateless_server(*, task_status: TaskStatus[None] = anyio.TASK_STA
                         write_stream,
                         self.app.create_initialization_options(),
                         stateless=True,
+                        session_id=http_transport.mcp_session_id,
```
maxisbey: Hmm maybe we should just force this to be None since stateless is a None session?

src/mcp/server/lowlevel/server.py
```diff
@@ -374,6 +374,9 @@ async def run(
         # the initialization lifecycle, but can do so with any available node
         # rather than requiring initialization for each connection.
         stateless: bool = False,
+        # Optional session identifier for task isolation. When provided (e.g.,
+        # from the transport's mcp_session_id), tasks are bound to this ID.
```
maxisbey: remove the comment about tasks here as that's experimental. tasks are only bound to session ID _if_ you're using the built in task handlers, if you supply your own then this doesn't apply

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -22,6 +22,7 @@ class StoredTask:
     """Internal storage representation of a task."""
 
     task: Task
+    session_id: str | None = None
```
maxisbey: make required

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -66,10 +68,25 @@ def _cleanup_expired(self) -> None:
         for task_id in expired_ids:
             del self._tasks[task_id]
 
+    def _get_stored_task(self, task_id: str, session_id: str | None = None) -> StoredTask | None:
```
maxisbey: make session_id required

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -66,10 +68,25 @@ def _cleanup_expired(self) -> None:
...
+            return None
+        if session_id is not None and stored.session_id is not None and stored.session_id != session_id:
+            return None
+        return stored
+
     async def create_task(
         self,
         metadata: TaskMetadata,
         task_id: str | None = None,
+        session_id: str | None = None,
```
maxisbey: make required

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -82,19 +99,20 @@ async def create_task(
...
+            session_id=session_id,
             expires_at=self._calculate_expiry(metadata.ttl),
         )
         self._tasks[task.task_id] = stored
 
         # Return a copy to prevent external modification
         return Task(**task.model_dump())
 
-    async def get_task(self, task_id: str) -> Task | None:
+    async def get_task(self, task_id: str, session_id: str | None = None) -> Task | None:
```
maxisbey: make required

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -156,34 +175,41 @@ async def get_result(self, task_id: str) -> Result | None:
     async def list_tasks(
         self,
         cursor: str | None = None,
+        session_id: str | None = None,
```
maxisbey: make required

src/mcp/shared/experimental/tasks/in_memory_task_store.py
```diff
@@ -156,34 +175,41 @@ async def get_result(self, task_id: str) -> Result | None:
...
         # Determine next cursor
         next_cursor = None
-        if start_index + self._page_size < len(all_task_ids) and page_task_ids:
+        if start_index + self._page_size < len(filtered_task_ids) and page_task_ids:
             next_cursor = page_task_ids[-1]
 
         return tasks, next_cursor
 
-    async def delete_task(self, task_id: str) -> bool:
+    async def delete_task(self, task_id: str, session_id: str | None = None) -> bool:
```
maxisbey: make session_id required

src/mcp/server/experimental/task_result_handler.py
```diff
@@ -80,6 +80,7 @@ async def handle(
         request: GetTaskPayloadRequest,
         session: ServerSession,
         request_id: RequestId,
+        session_id: str | None = None,
```
maxisbey: `session_id` should be required for tasks.

`session_id` being `None` only happens when in `stateless_mode` which is a non-spec compliant implementation detail of the SDK currently, and `Tasks` doesn't even properly work with `stateless_mode`.

If someone wants to use Tasks in stateless_mode then I think it's fair that they implement support or that themselves by defining custom task method handlers.

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -19,12 +19,15 @@ async def create_task(
         self,
         metadata: TaskMetadata,
         task_id: str | None = None,
+        session_id: str | None = None,
```
maxisbey: make required

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -34,14 +37,15 @@ async def create_task(
         """
 
     @abstractmethod
-    async def get_task(self, task_id: str) -> Task | None:
+    async def get_task(self, task_id: str, session_id: str | None = None) -> Task | None:
```
maxisbey: make required

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -50,70 +54,78 @@ async def update_task(
         task_id: str,
         status: TaskStatus | None = None,
         status_message: str | None = None,
+        session_id: str | None = None,
```
maxisbey: make requried

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -50,70 +54,78 @@ async def update_task(
...
-            ValueError: If task not found
+            ValueError: If task not found or not accessible by this session.
             ValueError: If attempting to transition from a terminal status
                 (completed, failed, cancelled). Per spec, terminal states
                 MUST NOT transition to any other status.
         """
 
     @abstractmethod
-    async def store_result(self, task_id: str, result: Result) -> None:
+    async def store_result(self, task_id: str, result: Result, session_id: str | None = None) -> None:
```
maxisbey: make required

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -50,70 +54,78 @@ async def update_task(
...
+            session_id: Optional session identifier for access control.
 
         Raises:
-            ValueError: If task not found
+            ValueError: If task not found or not accessible by this session.
         """
 
     @abstractmethod
-    async def get_result(self, task_id: str) -> Result | None:
+    async def get_result(self, task_id: str, session_id: str | None = None) -> Result | None:
```
maxisbey: make requried

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -50,70 +54,78 @@ async def update_task(
...
         Returns:
-            The stored Result, or None if not available
+            The stored Result, or None if not available.
         """
 
     @abstractmethod
     async def list_tasks(
         self,
         cursor: str | None = None,
+        session_id: str | None = None,
```
maxisbey: make required

src/mcp/shared/experimental/tasks/store.py
```diff
@@ -50,70 +54,78 @@ async def update_task(
...
+            session_id: Optional session identifier. When provided, only tasks
+                belonging to this session are returned.
 
         Returns:
             Tuple of (tasks, next_cursor). next_cursor is None if no more pages.
         """
 
     @abstractmethod
-    async def delete_task(self, task_id: str) -> bool:
+    async def delete_task(self, task_id: str, session_id: str | None = None) -> bool:
```
maxisbey: make required

src/mcp/server/session.py
maxisbey: make this a keyword only param

maxisbey (changes requested):
Most of my comments are about making `session_id` required for anything task related. All the Tasks related objects we have are only used when you use the default task related method handlers, and so I think it's fair we enforce a `session_id`.

The only time `session_id` would be None is when you've set the server to be in stateless mode, but in this situation Tasks doesn't work anyway.

src/mcp/client/_memory.py
```diff
@@ -50,12 +51,14 @@ async def _connect(self) -> AsyncIterator[TransportStreams]:
 
             async with anyio.create_task_group() as tg:
                 # Start server in background
+                memory_session_id = uuid.uuid4().hex
```
Kludex: It's a bit weird to be forced to create a session_id before running the server, does stdio does that as well?

maxisbey: stdio doesn't have session IDs either. After thinking through it I think the problem is we actually _shouldn't_ require `session_id`s for the default TaskStore, and instead no session ID means no task isolation.

Instead the default task support should only allow `None` `session_id`s _if_ the server **isn't** in stateless mode. I think I was wrong before in assuming that a None session ID meant the server is in stateless mode, but in reality it could just mean we're in a different transport.

Kludex:
We merged a fix for this.

Issue #1691, opened 2025-11-28
Simplify and harden session lifecycle with an explicit state machine


dgenio:
## Summary

Server and client sessions use an internal initialization state machine combined with various flags and manual `AsyncExitStack` management. Stateless mode complicates this further by setting an "Initialized" state early as a workaround.

This makes the lifecycle hard to reason about and contributes to subtle bugs (e.g., past issues like #756 in stateless mode).

## Problems

- **Scattered state management:** Initialization and teardown logic are spread across methods and files.
- **Special cases for stateless mode:** Stateless HTTP sets the session as initialized even though no real protocol negotiation has occurred.
- **Resource management risk:** If initialization fails midway, transport tasks and resources may not be cleaned up reliably.
- **Testing difficulty:** Tests often bypass parts of the lifecycle using in-memory transports.

## Proposal

1. **Introduce an explicit session state machine**

   Represent distinct states as separate types, for example:

   - `UninitializedSession`
   - `NegotiatingSession`
   - `ActiveSession`
   - `ClosedSession`
   - `StatelessSession` (for per-request / ephemeral cases)

   Each state exposes only the operations that are valid in that state, and transitions return the next state type.

2. **Model stateless mode explicitly**

   - Instead of marking a stateful session as "Initialized" prematurely, model stateless HTTP as a dedicated `StatelessSession` type with a simpler lifecycle.
   - Make behavior differences clear in code and docs.

3. **Centralize resource cleanup**

   - Ensure that all lifecycle paths (happy path, error, cancellation) lead through code that:
     - cancels outstanding tasks,
     - closes transports,
     - releases resources stored in `AsyncExitStack`.

## Why this matters

- **Reliability:** Fewer edge cases and surprise states where messages can be processed incorrectly.
- **Debuggability:** Easier to reason about what can happen in each state.
- **Extensibility:** Adding new lifecycle behavior (e.g., resuming sessions) becomes more manageable.

## Acceptance criteria

- [ ] Session lifecycle is represented via explicit types or a clearly defined state machine.
- [ ] Stateless HTTP mode uses a dedicated path rather than setting "Initialized" as a hack.
- [ ] All entry/exit paths of sessions ensure proper cleanup of transports and tasks.
- [ ] Tests cover state transitions, including error cases and stateless/stat eful differences.

maxisbey:
[...]

PR #2099, opened 2026-02-19
feat: expand InitializationState with explicit lifecycle state machine

dgenio:
## Summary

Addresses [#1691](https://github.com/modelcontextprotocol/python-sdk/issues/1691) — Phase 1: expand the `InitializationState` enum with an explicit lifecycle state machine and add centralized transition validation.

## Problem

The current `ServerSession` lifecycle management uses a simple three-state enum (`NotInitialized`, `Initializing`, `Initialized`) with direct state assignment. This makes it difficult to reason about session lifecycle, provides no protection against invalid transitions, and lacks visibility into shutdown states. Stateless sessions were handled by jumping directly to `Initialized`, conflating two conceptually different modes.

### Expanded `InitializationState` enum
- **`Stateless`** — sessions that skip the initialization handshake (previously mapped to `Initialized`)
- **`Closing`** — session is shutting down, entered during `__aexit__`
- **`Closed`** — terminal state after shutdown completes

An `Error` state for unrecoverable failures is planned for Phase 2 of #1691.

### Centralized transition validation
- **`_VALID_TRANSITIONS`** dict — maps each state to its valid target states, documenting the state machine in one place
- **`_transition_state()`** method — validates transitions and raises `RuntimeError` on invalid ones, with debug logging
- Re-initialization (`Initialized → Initializing`) is explicitly allowed for backward compatibility
- `NotInitialized → Initialized` is allowed to handle cases where the client sends an `InitializedNotification` without a prior `InitializeRequest` round-trip

### Concurrency documentation
- Class-level docstring on `ServerSession` documents the single-consumer concurrency model: incoming messages are processed sequentially by `_receive_loop`, so `_transition_state()` needs no external synchronization
- `_transition_state()` docstring explains why cooperative scheduling makes the method safe without locks

### New properties
- **`initialization_state`** — read-only access to current state
- **`is_initialized`** — returns `True` for both `Initialized` and `Stateless` (replaces direct comparisons)

### `__aexit__` override
- Transitions through `Closing → Closed` on session exit
- Gracefully handles edge cases (already closing/closed)

### Updated handlers
- `_received_request` and `_received_notification` now use `_transition_state()` instead of direct assignment
- `_received_notification` guards against redundant `InitializedNotification` in stateless mode

## Tests

20 new tests in `tests/server/test_session_lifecycle.py` covering:
- Enum completeness and value uniqueness
- Transition table correctness (all states have entries, `Closed` is terminal)
- Valid and invalid state transitions via `_transition_state()`
- `is_initialized` property for all states
- `__aexit__` lifecycle transitions (including already-closing and already-closed edge cases)
- Full client/server handshake integration test

### DIFF
--- a/src/mcp/server/session.py
+++ b/src/mcp/server/session.py
@@ -267,7 +267,7 @@ class ServerSession(
                 InitializationState.Closing,
             ):
                 self._transition_state(InitializationState.Closing)
-        except RuntimeError:
+        except RuntimeError:  # pragma: no cover
             logger.debug("Could not transition to Closing from %s", self._initialization_state.name)
         try:
             return await super().__aexit__(exc_type, exc_val, exc_tb)
--- a/tests/server/test_session_lifecycle.py
+++ b/tests/server/test_session_lifecycle.py
@@ -220,6 +220,25 @@ async def test_aexit_stateless_transitions_to_closed() -> None:
         assert session.initialization_state == InitializationState.Closed
 
 
+async def test_aexit_already_closing() -> None:
+    """__aexit__ skips Closing transition when already in Closing state."""
+    async with _session_context() as session:
+        async with session:
+            session._transition_state(InitializationState.Closing)
+
+        assert session.initialization_state == InitializationState.Closed
+
+
+async def test_aexit_already_closed() -> None:
+    """__aexit__ handles already-Closed sessions gracefully."""
+    async with _session_context() as session:
+        async with session:
+            session._transition_state(InitializationState.Closing)
+            session._transition_state(InitializationState.Closed)
+
+        assert session.initialization_state == InitializationState.Closed
+
+
 # ---------------------------------------------------------------------------
 # Integration: full handshake lifecycle
 # ---------------------------------------------------------------------------


dgenio:
Hi — I'm the author of this PR. Just checking in after @maxisbey's feedback about waiting for transport spec finalization.

I designed this as a **non-breaking Phase 1** — it only expands the existing enum and adds validated transitions without changing the public API surface. The `Stateless`, `Closing`, and `Closed` states are internal and the `is_initialized` property is additive. If the transport spec changes require additional states (e.g., `Resuming`, `Reconnecting`), those can be added to the transition table without reworking anything here.

That said, I completely understand if you'd prefer to wait until the transport discussions linked above converge before merging lifecycle changes. Happy to keep this PR updated in the meantime — it applies cleanly to current `main` and I can rebase as needed.

Would love to hear if Phase 1 is acceptable as an incremental step, or if you'd rather hold off entirely. Thanks!

dgenio:
Update: rebased onto current `main` (41 commits behind) — all 20 tests still pass, ruff and pyright are clean.

Happy to iterate or wait. Any guidance on whether Phase 1 is acceptable while SEP-1442 is still in review, or whether you'd prefer to hold until the transport spec stabilises? Just need a direction so I can decide whether to keep the branch maintained or close it.

PR #2103, opened 2026-02-19
docs: generate hierarchical per-module API reference pages

jonathanhefner:
The API docs previously consisted of a single page (`docs/api.md`) containing only `::: mcp`, which rendered just the top-level `mcp/__init__.py` exports. Most of the public API surface (`MCPServer`, `Context`, `Settings`, `Tool`, `Resource`, `Prompt`, etc.) was missing because those classes live in subpackages not re-exported at the top level.

This replaces the single-page approach with auto-generated per-module pages using the standard mkdocstrings recipe:

- `mkdocs-gen-files` runs `scripts/gen_ref_pages.py` at build time to create a virtual `.md` stub for every module under `src/mcp/`
- `mkdocs-literate-nav` reads the generated `api/SUMMARY.md` to build a hierarchical sidebar navigation tree
- Material theme's `navigation.indexes` handles section index pages (the `mkdocs-section-index` plugin is not needed and conflicts with it)

Also updates `mkdocstrings` `paths` from `[src/mcp]` to `[src]` so that fully-qualified module identifiers like `mcp.server.mcpserver.server` resolve correctly.

| Before | After |
| --- | --- |
| before-1 | after-1 |
| | after-2 |


docs/hooks/gen_ref_pages.py
```diff
@@ -0,0 +1,35 @@
+"""Generate the code reference pages and navigation."""
```
Kludex, line 1: Can we keep this under hooks in the docs folder instead?

Kludex (changes requested):

docs/hooks/gen_ref_pages.py
jonathanhefner: Sure, I moved it to `docs/hooks/gen_ref_pages.py`.  Is that where you were thinking?

PR #2109, opened 2026-02-19
fix: silence mkdocs social plugin warnings in strict mode

jonathanhefner:
The social plugin (`mkdocs-material`) requires cairosvg for generating Open Graph social card images. This dependency was never included in the docs dependency group, so the plugin emitted a warning on every page. With strict mode enabled, these warnings aborted the build.

Changes:

- Add `mkdocs-material[imaging]` extra to the docs dependency group,  which pulls in `cairosvg`, `cairocffi`, `cssselect2`, `defusedxml`,  and `tinycss2`. This requires the system Cairo library (`libcairo2`),  which is pre-installed on GitHub Actions ubuntu-latest runners.
- Gate the social plugin behind an `ENABLE_SOCIAL_CARDS` env var  (defaults to false) so local development does not require Cairo.
- Set `ENABLE_SOCIAL_CARDS=true` in the docs publish workflow so social  cards are generated for the published site.

---

The other approach is to set `cards: false` on the social plugin in `mkdocs.yml`.  According to Claude, that would disable social card generation (which is what requires cairosvg) but keep the plugin active for metadata.  I don't have a strong preference.

### DIFF
--- a/.github/workflows/publish-docs-manually.yml
+++ b/.github/workflows/publish-docs-manually.yml
@@ -31,3 +31,5 @@ jobs:
 
       - run: uv sync --frozen --group docs
       - run: uv run --frozen --no-sync mkdocs gh-deploy --force
+        env:
+          ENABLE_SOCIAL_CARDS: "true"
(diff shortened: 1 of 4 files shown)


Kludex:
We should move away from mkdocs, maybe zensical?

PR #2119, opened 2026-02-20
Type-check docstring code examples via companion files and unified sync script

jonathanhefner:
Docstring code examples in `src/mcp/` were previously raw text — invisible to pyright and ruff, so type errors and style drift went unnoticed. This PR introduces a system for keeping those examples in standalone, type-checked companion files that are synced back into the docstrings automatically.

The old `update_readme_snippets.py` is replaced with `sync_snippets.py`, a superset that handles `README.v2.md`, `docs/**/*.md`, and `src/**/*.py` docstrings. It adds region extraction (`# region` / `# endregion` markers) so a single companion file can supply multiple snippets. For docstrings specifically, `<!-- snippet-source #RegionName -->` markers derive the companion path automatically from the target file's location (`src/mcp/foo.py` → `examples/snippets/docstrings/mcp/foo.py`), avoiding line-length violations from embedding full paths.

42 code examples across the public API surface (`MCPServer`, `Client`, `ClientSession`, `Context`, `ResponseRouter`, task support, OAuth providers, etc.) are extracted into 19 companion files under `examples/snippets/docstrings/mcp/`, mirroring the source tree. Each example is wrapped in a named function with typed parameters, so pyright and ruff check them on every CI run — catching signature changes, renamed parameters, and moved imports at CI time rather than when a user copies a broken example.

Conventions for the snippet system (region naming, function-parameter pattern, `# type: ignore` prohibition, editing workflow) are documented in `CLAUDE.md`.

---

@Kludex, I was told you were thinking about moving `examples/snippets/` to `docs/snippets/`.  Would you prefer that docstring snippets live in `docs/snippets/docstrings/`?  (Or anywhere else other than `examples/snippets/docstrings/`?)

### DIFF
--- a/README.v2.md
+++ b/README.v2.md
@@ -183,8 +183,6 @@ if __name__ == "__main__":
 ```
 <!-- /snippet-source -->
 
-_Full example: [examples/snippets/servers/mcpserver_quickstart.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/mcpserver_quickstart.py)_
-
 You can install this server in [Claude Code](https://docs.claude.com/en/docs/claude-code/mcp) and interact with it right away. First, run the server:
 
 ```bash
@@ -281,8 +279,6 @@ def query_db(ctx: Context[AppContext]) -> str:
 ```
 <!-- /snippet-source -->
 
-_Full example: [examples/snippets/servers/lifespan_example.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/lifespan_example.py)_
-
 ### Resources
 
 Resources are how you expose data to LLMs. They're similar to GET endpoints in a REST API - they provide data but shouldn't perform significant computation or have side effects:
@@ -312,8 +308,6 @@ def get_settings() -> str:
 ```
 <!-- /snippet-source -->
 
-_Full example: [examples/snippets/servers/basic_resource.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/basic_resource.py)_
-
 ### Tools
 
 Tools let LLMs take actions through your server. Unlike resources, tools are expected to perform computation and have side effects:
@@ -338,8 +332,6 @@ def get_weather(city: str, unit: str = "celsius") -> str:
     return f"Weather in {city}: 22degrees{unit[0].upper()}"
 ```
 <!-- /snippet-source -->
-
-_Full example: [examples/snippets/servers/basic_tool.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/basic_tool.py)_
 
 Tools can optionally receive a Context object by including a parameter with the `Context` type annotation. This context is automatically injected by the MCPServer framework and provides access to MCP capabilities:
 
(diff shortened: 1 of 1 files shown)


Kludex:
> @Kludex, I was told you were thinking about moving examples/snippets/ to docs/snippets/. Would you prefer that docstring snippets live in docs/snippets/docstrings/? (Or anywhere else other than examples/snippets/docstrings/?)

I said something like: "the examples/snippets are horribly lengthy, and complex". But yeah, I think ideally we want to have what would be an example as part of the docs.

jonathanhefner:
> I said something like: "the examples/snippets are horribly lengthy, and complex".

Ah, I see.  Agreed.  I am working on getting more focused examples in place (e.g., without setup boilerplate).

> But yeah, I think ideally we want to have what would be an example as part of the docs.

Agreed that they should be part of the docs, as in "embedded in the docs".  Do you feel like the underlying code should also live in `docs/`?  Or is it okay for the underlying code to live in `examples/` and regions of it be injected into `docs/*.md` files?

The way we've done it for the TypeScript SDK is to have, e.g., a [`examples/client/src/clientGuide.examples.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/examples/client/src/clientGuide.examples.ts) file that is linted / type checked / etc, and to inject regions of that file into [`docs/client.md`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/client.md), which is what the user sees.

I slightly lean toward the way it's done the TypeScript SDK, but I don't have a strong preference.  Which do you prefer?

Kludex:
> Do you feel like the underlying code should also live in docs/? Or is it okay for the underlying code to live in examples/ and regions of it be injected into docs/*.md files?

It's okay to live elsewhere, but for the perfect setup, that code should be tested.

---

Any approach is fine. Making sure that code is tested is the requirement here.

jonathanhefner:
> Any approach is fine. Making sure that code is tested is the requirement here.

My plan was to ensure the code is type checked, linted, and formatted.  Actual tests would add a lot of overhead to the process (setup code, mocks, assertions, etc.).

Are you okay with simply type checking?  Do any of the existing examples have test coverage?

README.v2.md
```diff
@@ -310,9 +310,9 @@ def get_settings() -> str:
   "debug": false
 }"""
 ```
+<!-- /snippet-source -->
 
 _Full example: [examples/snippets/servers/basic_resource.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/basic_resource.py)_
```
maxisbey: we should delete all these, or maybe have them as comments in the markdown. Not sure how helpful a link to the snippet actually is in the markdown since the code block contains the entire source anyway

maxisbey:
> > Any approach is fine. Making sure that code is tested is the requirement here.
> 
> My plan was to ensure the code is type checked, linted, and formatted. Actual tests would add a lot of overhead to the process (setup code, mocks, assertions, etc.).
> 
> Are you okay with simply type checking? Do any of the existing examples have test coverage?

Having the code tested would be nice, but that is a lot of overhead. Type checking and linting IMO is probably good enough at least to start with.

README.v2.md
jonathanhefner: Makes sense.  The `snippet-source` comment already contains the path (e.g., `<!-- snippet-source examples/snippets/servers/basic_resource.py -->`), so I've removed the links.

Kludex:
If we are making the examples simpler, then it's a good opportunity to test them.

I mean, the perfect world would be to add tests for that, so examples are tested and they actually work - given that in the past we got a lot of PRs/issues with example issues.

Kludex:
Otherwise, lint/type checker in them is fine.

jonathanhefner:
> Otherwise, lint/type checker in them is fine.

Since the examples in this PR are all in `examples/snippets/docstrings/`, they are all linted and type-checked. :+1: 

@Kludex @maxisbey Are there any further concerns?  Is this okay to merge?

PR #2122, opened 2026-02-21
fix: propagate SSE stream errors to waiting requests

### LINKED ISSUE #1789: Client hangs forever
### Initial Checks

### Description

Client hangs forever when `sse_read_timeout` is specified unless `read_timeout_seconds` in `session.call_tool` is also populated.

### Example Code

```Python
import threading
import time
import asyncio
from typing import Literal
from datetime import timedelta

from mcp import ClientSession
from mcp.types import ImageContent as MCPImageContent


def start_comprehensive_mcp_server(transport: Literal["sse", "streamable-http"], port=int):

    mcp = FastMCP("Comprehensive MCP Server", port=port)

    @mcp.tool(description="Tool that will timeout")
    def timeout_tool() -> str:
        time.sleep(10)
        return "This tool has timed out"

    @mcp.tool(description="Calculator tool which performs calculations")
    def calculator(x: int, y: int) -> int:
        return x + y

    mcp.run(transport=transport)


# Start server
server_thread = threading.Thread(
    target=start_comprehensive_mcp_server, 
    kwargs={"transport": "streamable-http", "port": 8001}, 
    daemon=True
)
server_thread.start()
time.sleep(3)
print("SERVER STARTED", flush=True)


async def test_raw_mcp():
    print("CREATING RAW MCP CONNECTION...", flush=True)
    
        url="http://127.0.0.1:8001/mcp",
        sse_read_timeout=2,  # 2 second timeout
        print("TRANSPORT CONNECTED", flush=True)
        
            print("SESSION CREATED", flush=True)
            
            await session.initialize()
            print("SESSION INITIALIZED", flush=True)
            
            print("CALLING TOOL (with 2s timeout)...", flush=True)
            try:
                result = await session.call_tool(
                    "timeout_tool", 
                    arguments={},
                    # read_timeout_seconds=timedelta(seconds=5), # This needs to be provided or the client will hang forever
                )
                print(f"TOOL RESULT: {result}", flush=True)
            except Exception as e:
                print(f"TOOL CALL EXCEPTION: {type(e).__name__}: {e}", flush=True)
            
            print("DONE", flush=True)


print("RUNNING TEST...", flush=True)
asyncio.run(test_raw_mcp())
print("TEST COMPLETE", flush=True)
```

### Python & MCP Python SDK

```Text
1.24.0
```

heyhayes:
## Summary

Fixes #1401. Also fixes #1789 (closed as duplicate).

When an SSE read timeout occurs during a StreamableHTTP POST request, the pending `send_request` call hangs indefinitely. The transport catches the exception but never sends an error back through the read stream, leaving the caller blocked on `response_stream_reader.receive()` with nothing to receive.

This PR fixes error propagation at the transport level so that SSE stream failures produce a `JSONRPCError` keyed to the original request ID. `BaseSession._handle_response` routes it to the correct per-request response stream, and `send_request` surfaces it as `MCPError` to the caller. This approach keeps failures isolated to the affected request rather than tearing down the entire session.

### What changed

`_handle_sse_response` now sends a `JSONRPCError(INTERNAL_ERROR, "SSE stream ended without a response")` when the SSE stream ends without delivering a complete response, whether due to a read timeout, network error, or unexpected server close. If a `last_event_id` was received, reconnection is attempted first; the error is only sent after reconnection is exhausted.

`_handle_reconnection` returns `bool` instead of `None` so callers can distinguish success (response delivered) from failure (attempts exhausted). The method also fixes an infinite recursion bug: the attempt counter was reset to 0 on every stream end (even when no complete response was delivered), which combined with httpx read timeouts causing graceful stream termination meant the reconnection loop could run forever.

`handle_get_stream` applies the same fix to the GET stream's reconnection loop: the attempt counter only resets when events were actually received during the connection. Empty connections that close immediately count toward `MAX_RECONNECTION_ATTEMPTS`.

`_default_message_handler` now logs a warning for exceptions instead of silently discarding them, providing observability for transport errors not tied to a specific request.

## Test plan

- [x] New E2E test: client with 0.5s SSE read timeout reads a slow resource (2s server delay), asserts `MCPError` is raised instead of hanging
- [x] New unit test: `_handle_reconnection` returns `False` when called at max attempts
- [x] All 51 existing streamable HTTP tests pass (including reconnection, polling, and multi-reconnection tests)
- [x] All 233 client tests pass
- [x] Linting (`ruff check .`) and type checking (`pyright`) clean

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -57,6 +57,8 @@ class MessageHandlerFnT(Protocol):
 async def _default_message_handler(
     message: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
 ) -> None:
+    if isinstance(message, Exception):
+        logger.warning("Unhandled exception in message handler: %s", message)
     await anyio.lowlevel.checkpoint()
 
 
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -197,7 +197,9 @@ class StreamableHTTPTransport:
                     event_source.response.raise_for_status()
                     logger.debug("GET SSE connection established")
 
+                    received_events = False
                     async for sse in event_source.aiter_sse():
+                        received_events = True
                         # Track last event ID for reconnection
                         if sse.id:
                             last_event_id = sse.id
@@ -207,8 +209,12 @@ class StreamableHTTPTransport:
 
                         await self._handle_sse_event(sse, read_stream_writer)
 
-                    # Stream ended normally (server closed) - reset attempt counter
-                    attempt = 0
+                    # Only reset attempts if we actually received events;
+                    # empty connections count toward MAX_RECONNECTION_ATTEMPTS
+                    if received_events:
+                        attempt = 0
+                    else:
+                        attempt += 1
 
             except Exception:  # pragma: lax no cover
                 logger.debug("GET stream error", exc_info=True)
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -211,10 +211,7 @@ class StreamableHTTPTransport:
 
                     # Only reset attempts if we actually received events;
                     # empty connections count toward MAX_RECONNECTION_ATTEMPTS
-                    if received_events:
-                        attempt = 0
-                    else:
-                        attempt += 1
+                    attempt = 0 if received_events else attempt + 1
 
             except Exception:  # pragma: lax no cover
                 logger.debug("GET stream error", exc_info=True)
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -47,7 +47,7 @@ LAST_EVENT_ID = "last-event-id"
 
 # Reconnection defaults
 DEFAULT_RECONNECTION_DELAY_MS = 1000  # 1 second fallback when server doesn't provide retry
-MAX_RECONNECTION_ATTEMPTS = 2  # Max retry attempts before giving up
+MAX_RECONNECTION_ATTEMPTS = 5  # Max retry attempts before giving up
 
 
 class StreamableHTTPError(Exception):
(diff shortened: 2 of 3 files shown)


fengjikui:
I took a pass at refreshing this because the current PR is marked conflicting, and this issue is still referenced from #1401 by downstream users.

I did not want to open a duplicate PR without checking first, so I rebuilt the same narrow direction on current `origin/main` here:

- branch: https://github.com/fengjikui/python-sdk/tree/codex/rebuild-pr2122-sse-error-propagation
- commit: `684e38b` (`fix: propagate streamable HTTP SSE errors`)
- touched files: `src/mcp/client/session.py`, `src/mcp/client/streamable_http.py`, `tests/shared/test_streamable_http.py`

Validation run locally on the refreshed branch:

```text
uv run pytest tests/shared/test_streamable_http.py::test_server_close_sse_stream_via_context 
  tests/shared/test_streamable_http.py::test_streamable_http_client_auto_reconnects 
  tests/shared/test_streamable_http.py::test_streamable_http_client_respects_retry_interval 
  tests/shared/test_streamable_http.py::test_handle_reconnection_returns_false_on_max_attempts 
  tests/shared/test_streamable_http.py::test_sse_stream_close_raises_when_reconnection_fails -q
# 5 passed

uv run ruff check src/mcp/client/streamable_http.py src/mcp/client/session.py tests/shared/test_streamable_http.py
# All checks passed

uv run ruff format --check src/mcp/client/streamable_http.py src/mcp/client/session.py tests/shared/test_streamable_http.py
# 3 files already formatted

git diff --check origin/main..HEAD
# clean
```

The refreshed patch keeps the original behavior intent: `_handle_reconnection()` returns whether a response was delivered, `_handle_sse_response()` sends a JSONRPC error back to unblock the pending request when reconnection cannot deliver the response, empty GET SSE streams count toward the reconnect budget, and the default message handler logs otherwise-unhandled exceptions.

Happy to either open this as a refresh PR against `main`, or leave the branch here for @heyhayes / maintainers to cherry-pick if that is preferred.

heyhayes:
This looks to be inert looking at how main has moved on this seems largely resolved so can be closed @maxisbey 

I did however add a PR to help with logging, heopfully it helps https://github.com/modelcontextprotocol/python-sdk/pull/3087 

All the best 

David

maxisbey:
Thanks for the PR. This has since landed via #3047. I'm closing this out as part of a backlog cleanup now that v2 is out. Feel free to reopen if this is still relevant.

Issue #1764, opened 2025-12-10
Race condition in StreamableHTTP: zero-buffer memory streams cause deadlock with concurrent SSE responses


Ctariy:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Bug Description

SSE connections hang indefinitely when using StreamableHTTPServerTransport in stateless mode with responses containing 3+ items. The issue is caused by zero-buffer memory streams that block send() until receive() is called, creating a race condition between the response writer and the SSE stream iterator.

Related issues: #262 describes similar symptoms (client hangs on call_tool()) but root cause wasn't identified. This issue provides the specific cause and fix.

### Expected Behavior

All tool responses should complete regardless of response size.

### Actual Behavior
- 1-2 items: Response returns immediately (~150ms)
- 3+ items: Request hangs indefinitely (deadlock)

### Root Cause Analysis

The issue is in mcp/server/streamable_http.py:

Line 412 - zero-buffer request stream
self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](0)

Line 460 - zero-buffer SSE stream
sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[dict[str, str]](0)

### Race condition flow:
1. tg.start_soon(response, ...) starts SSE response task (non-blocking)
2. await writer.send(session_message) sends request to MCP server
3. MCP server processes quickly and calls message_router
4. message_router tries await request_streams[id][0].send(EventMessage(...))
5. DEADLOCK: If SSE writer hasn't started iterating yet, send() blocks forever

With zero-buffer streams, send() blocks until the receiver calls receive(). When the MCP server processes faster than the SSE writer task starts, deadlock occurs.

Why timing matters:
- Small responses (1-2 items): SSE writer task starts before MCP response arrives → works
- Larger responses (3+ items): MCP processes faster → response arrives before SSE iterator starts → blocked forever

### Proposed Fix
Increase buffer size from 0 to a reasonable value (e.g., 10 or 100):

Line 412
self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](10)

Line 460
sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[dict[str, str]](10)

[...]

PR #2145, opened 2026-02-25
Fix stateless HTTP task accumulation causing memory leak

### LINKED ISSUE #1272: Server hangs when shutting down if a connection is still open
### Initial Checks

### Description

Hello!

I was testing my MCP server with the inspector `npx @modelcontextprotocol/inspector` using streamable-http and I noticed that if the inspector is connected, the MCP Server will not shutdown.

We can't keep the server running if a client forgot to disconnect right?  This is tested with the official inspector tools, so there might be other clients working the same way.

Shouldn't the server close its active connections/tasks on shutdown?

### Reproduction

1. Run the server (see code snippet)
2. In the MCP Inspector click "connect"
3. In the server's terminal hit CTRL+C

At this point
- The server shows `Waiting for connections to close (CTRL+C to force quit)` and CTRL+C doesn't work, in fact nothing works except killing the process
- Clicking "Disconned" in the Inspector cannot disconnect, because `Error in /mcp route: TypeError: fetch failed`

### Evidences

Shell execution

```shell
$ python demo-mcp.py                 
INFO:     Started server process [10749]
INFO:     Waiting for application startup.
[08/14/25 11:03:16] INFO     StreamableHTTP session manager started                                                                                                                                                                                 streamable_http_manager.py:110
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
[08/14/25 11:03:21] INFO     Created new transport with session ID: 4134f60ccda44baf8a01733ed94bc0d8                                                                                                                                                streamable_http_manager.py:233
INFO:     127.0.0.1:54629 - "POST /mcp HTTP/1.1" 200 OK
INFO:     127.0.0.1:54629 - "POST /mcp HTTP/1.1" 202 Accepted
INFO:     127.0.0.1:54630 - "GET /mcp HTTP/1.1" 200 OK
^CINFO:     Shutting down
INFO:     Waiting for connections to close. (CTRL+C to force quit)
```

Inspector proxy Disconnect error
```
Received DELETE message for sessionId dd8a78d2-d5a8-4ff2-b679-6a8571f7cd0c
Connection refused. Is the MCP server running?
Error in /mcp route: TypeError: fetch failed
    at node:internal/deps/undici/undici:13502:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async StreamableHTTPClientTransport.terminateSession (file://.../.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:355:30)
    at async file://.../.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/inspector/server/build/index.js:264:17 {
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:8000
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1636:16) {
    errno: -61,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '127.0.0.1',
    port: 8000
  }
}
```

### Mitigation

Setting the `timeout_graceful_shutdown` uvicorn config to a few seconds will eventually quit the process, **BUT** this is exiting with a non 0 exit code, which is an issue trying to run this in production.   (This setting has been request [here](https://github.com/modelcontextprotocol/python-sdk/issues/253) and currently available in FastMCP 2.0)

### Example Code

```Python

mcp = FastMCP("demo")

@mcp.tool()
async def hello() -> str:
    return "world"

```

### Python & MCP Python SDK

```Text
Python 3.12.8
mcp 1.12.4
```

wiggzz:
## Summary

- Fix memory leak where `run_stateless_server` tasks accumulate indefinitely in the manager's global task group after client disconnects
- Use request-scoped task groups instead of the global `_task_group` for stateless requests, ensuring tasks are cancelled when requests complete
- Add graceful SSE drain on shutdown: terminate all active transports before cancelling the task group, so SSE streams close cleanly instead of being abruptly reset
- Add regression tests for both the leak and the graceful drain

## Motivation and Context

Related to #1764 (the underlying task accumulation issue in stateless mode).
Fixes #1272 (graceful shutdown / SSE drain).

### Root cause — memory leak

When a client disconnects while a tool call is in progress, `EventSourceResponse` detects the disconnect and tears down the SSE stream pipeline. However, `Server.run()` — which is processing the tool call in a separate task spawned into the manager's global `_task_group` — continues running. After the tool handler completes, the response has nowhere to go (the request streams were cleaned up), and `Server.run()` blocks indefinitely on `async for message in session.incoming_messages`. These zombie tasks accumulate one per disconnected request.

In production, we observed 449 leaked `ServerSession` objects, 66,086 orphaned coroutines, 4,463 blocked stream receive operations, and ~150 MB of leaked heap — all from this pattern.

### Root cause — broken SSE close on shutdown

The memory leak fix moves stateless server tasks from the manager's global task group to request-scoped task groups. This fixes the leak, but it also means the manager's `tg.cancel_scope.cancel()` during shutdown no longer reaches stateless transports — nothing ever closes their streams, so SSE responses hang indefinitely. A reverse proxy like nginx would log "upstream prematurely closed connection while reading upstream".

The graceful drain fix adds explicit `terminate()` calls in `run()`'s finally block (for both stateful and stateless transports) before cancelling the task group. This closes the in-memory streams, letting `EventSourceResponse` send a final `more_body=False` chunk — a clean HTTP close.

### Why the existing test didn't catch the leak

The existing test (`test_stateless_requests_memory_cleanup`, added in PR #1140) mocks `app.run` to return immediately, so the spawned tasks complete naturally. It never exercises the real-world scenario where client disconnects leave `app.run()` blocked.

### The fixes

**Memory leak**: Replace the global task group spawn with a request-scoped `anyio.create_task_group()`. Both the server task and the request handler run inside this scoped group. When the request handler finishes (or is cancelled due to client disconnect), the scope is cancelled, which automatically cancels the server task. No tasks leak into the manager's long-lived task group.

**Graceful drain**: Track in-flight stateless transports in `_stateless_transports`. In `run()`'s finally block, call `terminate()` on all active transports (stateful via `_server_instances`, stateless via `_stateless_transports`) before `tg.cancel_scope.cancel()`.

### Testing

- `test_stateless_requests_task_leak_on_client_disconnect`: Real `Server.run()` with a real tool handler, real SSE streaming, simulated client disconnect — verifies zero leaked tasks after 3 disconnected requests
- `test_graceful_shutdown_closes_sse_streams_cleanly`: Uses an ASGI `send()` interceptor to verify that SSE responses receive `more_body=False` (the ASGI equivalent of a proper HTTP chunked transfer close) during manager shutdown. Without the fix, the test times out because nothing ever closes the stateless transport's streams.

All 13 tests pass.

## Breaking changes

None. This is an internal implementation change. The public API and behavior are unchanged.

## Checklist


wiggzz:
This isn't ready for review yet, as this is just a first pass from Claude. Currently testing this.

wiggzz:
@maxisbey any thoughts on this one?

PR #2149, opened 2026-02-26
Update session.py

DenisStefanAndrei:
## Fix: propagate CancelScope.__exit__ return value in RequestResponder.__exit__

`RequestResponder.__exit__` calls `self._cancel_scope.__exit__()` but did not
return its result. This prevented the cancel scope from suppressing
`CancelledError` when a request is cancelled via `notifications/cancelled`.

## Motivation and Context

When a client sends a `notifications/cancelled` notification (e.g. after a
request timeout), the server calls `RequestResponder.cancel()` which cancels the
inner `CancelScope`. This raises `CancelledError` in the handler task, which is
a `BaseException` and not caught by the `except Exception` handlers in
`_handle_request`.

The `CancelScope.__exit__` correctly identifies this as its own cancellation and
returns `True` to suppress it. However, `RequestResponder.__exit__` discarded
that return value, implicitly returning `None`. Python treats a falsy return from
`__exit__` as "do not suppress", so the `CancelledError` leaked out of the
`with responder:` block in `_handle_message`, crashed the `server.run()` task
group, and caused the session to tear down.

After teardown, the `stdin_reader` task in `stdio_server` remains alive (blocked
waiting for input on stdin). On the next incoming request, it attempts to write
to the now-closed `read_stream`, raising `BrokenResourceError` and crashing the
server process.

This bug is masked on Python 3.11 because asyncio's looser cancellation counter
tracking tolerates an unsuppressed `CancelledError` after `task.uncancel()` has
been called. Python 3.12 tightened this behavior, exposing the issue.

## How Has This Been Tested?

Reproduced by connecting a Node.js MCP client (stdio transport) to a Python
FastMCP server. Triggering a request timeout causes the client SDK to send
`notifications/cancelled`. Without the fix, the next request to the server
crashes the process. With the fix, the server correctly suppresses the
`CancelledError` and continues serving subsequent requests.

## Breaking Changes

None. This is a bugfix that restores the intended behavior of request
cancellation.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

The fix is a single `return` keyword added to `RequestResponder.__exit__` in
`src/mcp/shared/session.py`:

# Before (broken):

# After (fixed):
return self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -102,7 +102,7 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
         exc_type: type[BaseException] | None,
         exc_val: BaseException | None,
         exc_tb: TracebackType | None,
-    ) -> None:
+    ) -> bool:
         """Exit the context manager, performing cleanup and notifying completion."""
         try:
             if self._completed:  # pragma: no branch


DenisStefanAndrei:
@maxisbey thoughts on this?

maxisbey:
Thanks — the root-cause analysis here is excellent. Could you open an issue first with repro code (a minimal server + the `notifications/cancelled` trigger that crashes it on Py3.12)? That gives us something to link the PR to and a regression test to write against.

Also worth noting: the `return` is inside `finally`, which will now suppress exceptions from `_on_complete()` too — would be good to have that covered in the issue discussion.

PR #2151, opened 2026-02-26
fix request ID normalization collision

bschwitz3:
## Summary
- prevent non-canonical numeric response IDs (for example `"01"`) from being coerced to integer IDs
- avoid response/request routing collisions in shared session handling
- add regression test for non-canonical numeric response ID mismatch

## Test plan
- [ ] run `uv run --frozen pytest tests/shared/test_session.py`
- [ ] run full suite as needed

### DIFF
--- a/tests/shared/test_session.py
+++ b/tests/shared/test_session.py
@@ -272,9 +272,28 @@ async def test_response_id_non_canonical_numeric_string_no_match():
         server_read, server_write = server_streams
 
         async def mock_server():
-            """Receive two requests and reply with a colliding non-canonical ID."""
-            await server_read.receive()  # request id 0
-            await server_read.receive()  # request id 1
+            """Respond to ping #1, then send a non-canonical ID for ping #2."""
+            first_message = await server_read.receive()
+            assert isinstance(first_message, SessionMessage)
+            assert isinstance(first_message.message, JSONRPCRequest)
+            first_request_id = first_message.message.id
+
+            # Let the first request complete so the second request is sent.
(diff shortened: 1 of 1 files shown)


bschwitz3:
Hey @maxisbey , if you have a moment, I’d really appreciate your quick look on this fix, I’d value your feedback. :)
thanks !

maxisbey:
Thanks for the PR! I dug into this a bit and I think we should hold off.

The removed comment was actually correct — TypeScript uses [`Number(response.id)`](https://github.com/modelcontextprotocol/typescript-sdk/blob/f88c376c52e56f0c68d9f7a6bc322095bf3b0449/packages/core/src/shared/protocol.ts#L942), and in JavaScript `Number("01") === 1`, so `"01"` *does* match a pending request with integer ID `1` in the TS SDK. This PR would make Python diverge from TypeScript rather than stay aligned.

The collision risk also isn't really there in practice: both SDKs only ever send integer IDs from a monotonic counter (true for client→server *and* server→client requests — it's shared base-class code on both sides), so there's only ever one pending request with ID `1`. A peer returning `"01"` for it is spec-violating (JSON-RPC requires the response ID to be *the same* as the request), but there's no second request it could collide with.

If we want to tighten this up, the cleaner direction is probably exact-match-only with a warning on miss — but that's a separate conversation. Closing for now, but happy to discuss further!

bschwitz3:
Yes, totally makes sense with your context. Thanks for the detailed explanation

PR #2158, opened 2026-02-28
fix: use 'is not None' for related_request_id to handle id=0

### LINKED ISSUE #1218: `simple-streamablehttp-stateless` example will not return data if `id=0`
### Initial Checks

### Description

I ran the `simple-streamablehttp-stateless` example, then connected via curl.

If I send:

```
curl localhost:3000/mcp/ --json '{"jsonrpc":"2.0","id":0,"method":"tools/call","params":{"_meta":{"progressToken":0},"name":"start-notification-stream","arguments":{"interval":0.1,"count":5,"caller":""}}}' -H "Accept: application/json, text/event-stream"
```

it does NOT work; there are no notifications sent.

If I send (changing only the id):
```
curl localhost:3000/mcp/ --json '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"_meta":{"progressToken":0},"name":"start-notification-stream","arguments":{"interval":0.1,"count":5,"caller":""}}}' -H "Accept: application/json, text/event-stream"
```

I do get notifications.

If I run the same example in the typescript SDK, it works with both id.

Log:
```
Sending notification 1/5 from caller:
2025-07-30 13:14:04,335 - root - DEBUG - Request stream _GET_stream not found
                                for message. Still processing message as the client
                                might reconnect and replay.
Sending notification 2/5 from caller:
2025-07-30 13:14:04,435 - root - DEBUG - Request stream _GET_stream not found
                                for message. Still processing message as the client
                                might reconnect and replay.
Sending notification 3/5 from caller:
2025-07-30 13:14:04,536 - root - DEBUG - Request stream _GET_stream not found
                                for message. Still processing message as the client
                                might reconnect and replay.
Sending notification 4/5 from caller:
2025-07-30 13:14:04,637 - root - DEBUG - Request stream _GET_stream not found
                                for message. Still processing message as the client
                                might reconnect and replay.
Sending notification 5/5 from caller:
2025-07-30 13:14:04,737 - root - DEBUG - Request stream _GET_stream not found
                                for message. Still processing message as the client
                                might reconnect and replay.
```

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
latest
```

giulio-leone:
## Problem

When a JSON-RPC request uses `id=0`, notifications sent during tool execution are not routed to the correct request stream. The client never receives them.

**Root cause:** In `send_notification()`, the condition:

```python
metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id else None
```

evaluates to `False` when `related_request_id` is `0` (integer zero is falsy in Python). This means the `ServerMessageMetadata` is not attached, so the message router in `streamable_http.py` falls through to the `GET_STREAM_KEY` instead of the request-specific stream.

## Fix

```diff
- metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id else None,
+ metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id is not None else None,
```

This is consistent with the identity checks already used in `streamable_http.py`:
- Line 984: `message.id is not None`
- Line 993: `session_message.metadata.related_request_id is not None`
- Line 997: `target_request_id is not None`

## Testing

- 64 tests pass across `test_session.py` and `test_streamable_http.py` (2 consecutive clean runs)
- The JSON-RPC spec explicitly allows `id=0` as a valid request identifier

Fixes #1218

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -303,7 +303,9 @@ class BaseSession(
         )
         session_message = SessionMessage(
             message=jsonrpc_notification,
-            metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id is not None else None,
+            metadata=ServerMessageMetadata(related_request_id=related_request_id)
+            if related_request_id is not None
+            else None,
         )
         await self._write_stream.send(session_message)
 


giulio-leone:

PR #2162, opened 2026-02-28
fix: configure only mcp namespace logger instead of root logger

### LINKED ISSUE #1656: FastMCP configures logging on init, which messes up application-level logging
### Initial Checks

### Description

Hi everyone, thanks for maintaining the Python SDK for MCP!

I've noticed that the `FastMCP` class configures the logging ecosystem on `__init__()`, both by adding custom handlers and calling `logging.basicConfig(...)`. This will conflict with any logging setup that any application using the MCP SDK will use.

Please note that the best practice for logging is:
- library code should never configure logging behavior
- application entrypoints should configure logging behavior

Please refer to the [official logging HowTo, section "Configuring Logging for a Library"](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

For the sake of easy quickstarts, I'd advice to create some module specifically meant for quickstarts that runs a FastMCP server, taking care of setting up the rich logging as well before. But keep it separated from the main library usage of FastMCP server, so applications using it don't get their logging config messed up.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
1.22.0
```

giulio-leone:
## Summary

Fixes #1656

`FastMCP.__init__()` calls `configure_logging()` which previously called `logging.basicConfig()`, configuring the **root logger** with handlers and level. This violates Python logging best practices:

> *It is strongly advised that you do not add any handlers other than NullHandler to your library's loggers. This is because the configuration of handlers is the prerogative of the application developer.*
> — [Python Logging HOWTO](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

## Problem

Any application using the MCP SDK that calls `FastMCP()` gets its logging configuration silently overridden:
- Custom log formats are replaced
- Log levels are changed
- Handlers (e.g., file handlers, structured JSON logging) are affected
- Unit test log capture can break

## Fix

Replace `logging.basicConfig()` with targeted configuration of the `mcp` namespace logger only:
- `logging.getLogger('mcp')` is configured with handlers and level
- The root logger is left untouched
- Duplicate handler guards prevent stacking on repeated `FastMCP()` calls
- MCP SDK logs still work out of the box for quickstart scripts

## Test Results

475 server tests pass (2 consecutive clean runs).

### DIFF
--- a/src/mcp/server/mcpserver/utilities/logging.py
+++ b/src/mcp/server/mcpserver/utilities/logging.py
@@ -45,9 +45,7 @@ def configure_logging(
         from rich.console import Console
         from rich.logging import RichHandler
 
-        handler: logging.Handler = RichHandler(
-            console=Console(stderr=True), rich_tracebacks=True
-        )
+        handler: logging.Handler = RichHandler(console=Console(stderr=True), rich_tracebacks=True)
     except ImportError:  # pragma: no cover
         handler = logging.StreamHandler()
         handler.setFormatter(logging.Formatter("%(message)s"))
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -42,13 +42,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     # We duplicate the file descriptors via os.dup() to avoid closing the
     # real sys.stdin/sys.stdout when the wrapped streams are closed.
     if not stdin:
-        stdin = anyio.wrap_file(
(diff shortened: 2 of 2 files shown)


giulio-leone:

PR #2166, opened 2026-02-28
fix: use os.dup() to prevent stdio transport from closing real stdin/stdout

giulio-leone:
## Summary

Running an MCP server with `transport="stdio"` causes subsequent stdio operations to fail with `ValueError: I/O operation on closed file` after the server exits.

## Root Cause

In `stdio.py`, `TextIOWrapper(sys.stdin.buffer)` wraps the real stdin buffer directly. When the wrapper is garbage-collected or closed, it also closes `sys.stdin.buffer`, making `sys.stdin` unusable.

## Fix

Use `os.dup()` to duplicate the file descriptor before wrapping:

```python
stdin = anyio.wrap_file(
    TextIOWrapper(os.fdopen(os.dup(sys.stdin.fileno()), "rb"), encoding="utf-8")
)
```

This creates an independent file descriptor that can be safely closed without affecting the original `sys.stdin`/`sys.stdout`.

## Verification

- All stdio tests pass
- Manual verification: `print()` works after server exit

Fixes #1933

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -40,13 +40,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     # python is platform-dependent (Windows is particularly problematic), so we
     # re-wrap the underlying binary stream to ensure UTF-8.
     if not stdin:
-        stdin = anyio.wrap_file(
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing in favor of #2172 which has a more comprehensive fix.

PR #2167, opened 2026-02-28
fix: detect Context parameter in callable class instances

### LINKED ISSUE #1974: find_context_parameter() fails to detect Context parameter in callable class instances
### Initial Checks

### Description

When registering a callable class instance as an MCP tool via FastMCP.add_tool(), the ctx: Context parameter is incorrectly exposed as an externally visible tool parameter instead of being injected by the framework. This is because find_context_parameter() uses typing.get_type_hints() which doesn't introspect the __call__ method of callable class instances.

Suggested Fix
Update find_context_parameter() to handle callable class instances:
```
def find_context_parameter(fn: Callable[..., Any]) -> str | None:

    # Handle callable class instances by using __call__ method
    target = fn
    if not (inspect.isfunction(fn) or inspect.ismethod(fn)):
        if callable(fn) and hasattr(fn, "__call__"):
            target = fn.__call__

    try:
        hints = typing.get_type_hints(target)
    except Exception:
        return None

    # ... rest of function unchanged
```

### Example Code

```Python
from typing import Annotated, Any
from starlette.requests import Request
import pydantic

class MyTool:
        self.__name__ = name

    async def __call__(
        self,
        query: Annotated[str, pydantic.Field(description="Search query")],
        ctx: Context[ServerSession, Any, Request],
    ) -> str:
        """Performs a search."""
        return f"Results for: {query}"

mcp = FastMCP(name="Test Server")
mcp.add_tool(MyTool(name="my_search"), name="my_search", description="Search tool")

# When listing tools, 'ctx' appears as a required parameter in the schema
```

### Python & MCP Python SDK

```Text
1.26.0
```

giulio-leone:
## Summary

Fix `find_context_parameter()` to correctly detect the `Context` parameter in callable class instances (objects with `__call__` method).

## Problem

When registering a callable class instance as an MCP tool via `add_tool()`, the `ctx: Context` parameter is incorrectly exposed as a visible tool parameter in the JSON schema instead of being injected by the framework.

This happens because `typing.get_type_hints()` raises `TypeError` on callable class instances (`'<MyTool object>' is not a module, class, method, or function`). The exception is caught, `find_context_parameter()` returns `None`, and the context parameter is not excluded from the schema.

## Solution

Before calling `get_type_hints()`, detect callable class instances and inspect their `__call__` method instead:

```python
target = fn
if not (inspect.isfunction(fn) or inspect.ismethod(fn)):
    if callable(fn) and hasattr(fn, '__call__'):
        target = fn.__call__
```

This is consistent with how `_is_async_callable()` (in `tools/base.py`) already handles callable classes.

## Test

Added `test_context_injection_callable_class` that registers a callable class with a `Context` parameter and verifies it is properly injected (not exposed in schema).

Fixes #1974

### DIFF
--- a/tests/server/mcpserver/test_tool_manager.py
+++ b/tests/server/mcpserver/test_tool_manager.py
@@ -390,9 +390,7 @@ class TestContextHandling:
         """Test that context is properly injected for callable class instances."""
 
         class MyTool:
-            async def __call__(
-                self, x: int, ctx: Context[ServerSessionT, None]
-            ) -> str:
+            async def __call__(self, x: int, ctx: Context[ServerSessionT, None]) -> str:
                 assert isinstance(ctx, Context)
                 return str(x)
 


giulio-leone:
Closing as duplicate of #2178, which has passing CI.

PR #2168, opened 2026-02-28
fix: reject JSON-RPC requests with id: null instead of misclassifying as notification

### LINKED ISSUE #2057: Requests with "id": null silently misclassified as notifications

giulio-leone:
## Summary

Reject JSON-RPC messages with `"id": null` instead of silently misclassifying them as notifications.

## Problem

When a JSON-RPC request arrives with `"id": null`:

1. `JSONRPCRequest` correctly rejects it (`null` is not a valid `RequestId`)
2. Pydantic falls through to `JSONRPCNotification`, which absorbs the extra `"id"` field via implicit `extra='allow'`
3. The streamable HTTP transport sees "not a request" and returns 202
4. **The caller gets no error and no response** — a silent failure that's hard to debug

## Solution

Set `model_config = ConfigDict(extra='forbid')` on `JSONRPCNotification`. This prevents any unrecognised fields (including `"id": null`) from being silently absorbed, causing a proper validation error.

## Tests

Added 6 test cases in `tests/issues/test_2057_null_id.py`:
- `test_request_rejects_null_id` — JSONRPCRequest rejects `id: null`
- `test_notification_rejects_extra_id_field` — JSONRPCNotification rejects extra `id`
- `test_message_adapter_rejects_null_id` — Union adapter rejects `id: null`
- `test_valid_notification_without_id` — Normal notifications still work
- `test_valid_request_with_int_id` — Normal int-id requests still work
- `test_valid_request_with_string_id` — Normal string-id requests still work

Full test suite: **1128 passed**, 2 consecutive clean runs.

Fixes #2057


giulio-leone:

PR #2169, opened 2026-02-28
fix: accumulate OAuth scopes across 401/403 for progressive authorization

giulio-leone:
## Summary

Companion fix to modelcontextprotocol/typescript-sdk#1582 / modelcontextprotocol/typescript-sdk#1604.

The `OAuthClientProvider` overwrites `client_metadata.scope` on each 401/403 response instead of accumulating scopes. This causes an **infinite re-authorization loop** when an MCP server uses per-operation scopes (progressive/step-up authorization):

1. `initialize` succeeds with scope `init`
2. `tools/list` returns 403 → scope overwritten to `mcp:tools:read` (init lost)
3. New token grants `mcp:tools:read` but not `init`
4. Next operation needing `init` fails → scope overwritten to `init` (read lost)
5. Loop between steps 2–4

## Changes

- **`merge_scopes()`** utility in `utils.py`: Computes the set-union of space-delimited OAuth scope strings per [RFC 6749 §3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3)
- **401 handler** in `oauth2.py`: `merge_scopes(existing, new)` instead of overwrite
- **403 handler** in `oauth2.py`: `merge_scopes(existing, new)` instead of overwrite
- Updated existing test to verify scope accumulation
- Added 7 unit tests for `merge_scopes()` covering all edge cases

## Tests

All 91 auth tests pass (2 consecutive clean runs). `ruff check` and `ruff format` clean.


giulio-leone:

PR #2171, opened 2026-02-28
fix: handle tools/resources/prompts list_changed notifications in client

### LINKED ISSUE #2107: Handle list_changed notifications instead of dropping them
## Summary

The SDK silently drops `notifications/tools/list_changed`, `notifications/prompts/list_changed`, and `notifications/resources/list_changed` notifications. They fall through to `case _: pass` in the notification handler.

At minimum, these should be surfaced — either via a callback, an event, or a hook — so that consumers can react to them (e.g., invalidate caches, refresh tool lists).

This is related to #1429 (dynamic tool management) and #710 (triggering listChanged), but distinct: those are about the *server* side sending notifications, this is about the *client* side receiving them.

giulio-leone:
## Summary

Fixes #2107

The client silently drops `notifications/tools/list_changed`, `notifications/resources/list_changed`, and `notifications/prompts/list_changed` notifications from the server — they fall through to `case _: pass` in `_received_notification()`.

## Root Cause

`ClientSession._received_notification()` only handled `LoggingMessageNotification` and `ElicitCompleteNotification`. The three `list_changed` notification types (`ToolListChangedNotification`, `ResourceListChangedNotification`, `PromptListChangedNotification`) were caught by the catch-all `case _: pass` and discarded.

## Fix

Add optional callback parameters following the established `logging_callback` pattern:

- `tool_list_changed_callback: ToolListChangedFnT | None`
- `resource_list_changed_callback: ResourceListChangedFnT | None`
- `prompt_list_changed_callback: PromptListChangedFnT | None`

Each callback is a simple `async def () -> None` protocol. When set, it is invoked upon receiving the corresponding notification. When not set, a no-op default is used (preserving backward compatibility).

Both `ClientSession` and the high-level `Client` dataclass are updated.

## Tests

5 new tests covering:
- Each notification type invokes its callback
- No crash when no callback is set (no-op default)
- Multiple notifications each invoke the callback

**187/187 client tests pass (2 consecutive clean runs). ruff clean.**

### DIFF
--- a/tests/client/test_list_changed_callbacks.py
+++ b/tests/client/test_list_changed_callbacks.py
@@ -41,7 +41,7 @@ async def test_tool_list_changed_callback():
         message: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
     ) -> None:
         if isinstance(message, Exception):
-            raise message
+            raise message  # pragma: no cover
 
     async with Client(
         server,
@@ -69,7 +69,7 @@ async def test_resource_list_changed_callback():
         message: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
     ) -> None:
         if isinstance(message, Exception):
-            raise message
+            raise message  # pragma: no cover
 
     async with Client(
         server,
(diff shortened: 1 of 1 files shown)


giulio-leone:
Closing — test failures from upstream divergence. Will reopen with a fresh approach if still needed.

PR #2172, opened 2026-02-28
fix: dup file descriptors in stdio_server to avoid closing real stdin/stdout

giulio-leone:
## Problem

Running a server with `transport="stdio"` closes the real `sys.stdin` / `sys.stdout` when the server exits. Any subsequent stdio operation raises:

```
ValueError: I/O operation on closed file.
```

### Root Cause

`TextIOWrapper(sys.stdin.buffer)` shares the underlying file descriptor with `sys.stdin`. When the `TextIOWrapper` is closed (or garbage-collected) after the server exits, it also closes `sys.stdin.buffer`, making any subsequent stdio operation fail.

## Solution

Use `os.dup()` to duplicate the file descriptor before wrapping it in `TextIOWrapper`. Closing the wrapper then only closes the duplicate fd while leaving the original process handles intact.

```python
# Before (broken)

# After (fixed)
stdin = anyio.wrap_file(
    TextIOWrapper(os.fdopen(os.dup(sys.stdin.fileno()), "rb"), encoding="utf-8")
)
```

## Testing

- All 15 stdio-related tests pass across 2 consecutive runs
- Verified manually that `sys.stdin.fileno()` remains valid after closing the duplicated wrapper

Fixes #1933

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -40,13 +40,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     # after the server exits (or after the transport is torn down) gets
     # ``ValueError: I/O operation on closed file`` when touching stdio.
     if not stdin:
-        stdin = anyio.wrap_file(
-            TextIOWrapper(os.fdopen(os.dup(sys.stdin.fileno()), "rb"), encoding="utf-8")
-        )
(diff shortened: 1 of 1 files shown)


giulio-leone:
Friendly ping — CI is green and this is ready for review. Happy to address any feedback. Thanks!

giulio-leone:
All CI checks pass. Ready for review.

maxisbey:
Thanks for the PR! This fix is correct, but it's a duplicate of #2040 which was opened on Feb 12 with the same `os.dup()` approach plus a regression test. Consolidating there — please feel free to review/follow that PR.

PR #2175, opened 2026-02-28
Return `CallToolResult` from `convert_result` instead of a tuple

### LINKED ISSUE #1251: Incorrect typing hint for FastMCP.call_tool

Kludex:
## Summary

- `convert_result()` now returns a `CallToolResult` directly when structured output is present, instead of a bare `(unstructured, structured)` tuple that leaked through the public API.
- `MCPServer.call_tool` return type corrected from `Sequence[ContentBlock] | dict[str, Any]` to `CallToolResult | Sequence[ContentBlock]`.
- Removed the tuple-unpacking and unreachable `dict` branches from `_handle_call_tool`.

Closes #1251

## Test plan

- [x] Existing tests in `test_func_metadata.py`, `test_tool_manager.py`, and `test_server.py` updated and passing (170 tests)
- [x] `ruff check` and `pyright` clean on all changed files

### DIFF
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -91,9 +91,9 @@ class FuncMetadata(BaseModel):
     def convert_result(self, result: Any) -> Any:
         """Convert a function call result to the format for the lowlevel tool call handler.
 
-        - If output_model is None, return the unstructured content directly.
-        - If output_model is not None, convert the result to structured output format
-            (dict[str, Any]) and return both unstructured and structured content.
+        - If output_model is None, return the unstructured content as a ``Sequence[ContentBlock]``.
+        - If output_model is not None, return a ``CallToolResult`` with both unstructured
+          and structured content.
 
         Note: we return unstructured content here **even though the lowlevel server
         tool call handler provides generic backwards compatibility serialization of
@@ -120,7 +120,7 @@ class FuncMetadata(BaseModel):
             validated = self.output_model.model_validate(result)
             structured_content = validated.model_dump(mode="json", by_alias=True)
 
-            return (unstructured_content, structured_content)
+            return CallToolResult(content=list(unstructured_content), structured_content=structured_content)
 
     def pre_parse_json(self, data: dict[str, Any]) -> dict[str, Any]:
         """Pre-parse data from JSON.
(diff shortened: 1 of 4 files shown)


src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -91,9 +91,9 @@ async def call_fn_with_arg_validation(
     def convert_result(self, result: Any) -> Any:
         """Convert a function call result to the format for the lowlevel tool call handler.
 
-        - If output_model is None, return the unstructured content directly.
-        - If output_model is not None, convert the result to structured output format
-            (dict[str, Any]) and return both unstructured and structured content.
+        - If output_model is None, return the unstructured content as a ``Sequence[ContentBlock]``.
```
Kludex:
```suggestion
        - If output_model is None, return the unstructured content as a `Sequence[ContentBlock]`.
```

src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -91,9 +91,9 @@ async def call_fn_with_arg_validation(
...
+        - If output_model is not None, return a ``CallToolResult`` with both unstructured
```
Kludex:
```suggestion
        - If output_model is not None, return a `CallToolResult` with both unstructured
```

src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -91,9 +91,9 @@ async def call_fn_with_arg_validation(
     def convert_result(self, result: Any) -> Any:
```
maxisbey, line 91: two questions:
1. why not be more specific on the typing here?
2. Why not _only_ return a `CallToolResult`? `MCPServer._handle_call_tool` does this conversion anyway, so might as well have it in one place with stronger typing.

maxisbey (changes requested):

maxisbey:
Thanks for you contribution @Kludex (aka **THE** FastAPI expert). but imma close this as it's fixed :)

PR #2180, opened 2026-02-28
fix: add SSRF redirect protection to httpx client factory

### LINKED ISSUE #2106: Add SSRF protection for HTTP client redirects
## Summary

The SDK creates `httpx.AsyncClient` with `follow_redirects=True` but no redirect validation. A malicious MCP server can redirect requests to internal network addresses (`http://localhost`, `http://169.254.169.254`, `http://metadata.google.internal`), enabling SSRF attacks.

## Location

`src/mcp/shared/_httpx_utils.py` — the shared httpx client factory.

## Proposed Fix

1. **Default:** Block HTTPS-to-HTTP downgrades on redirects
2. Provide configurable `RedirectPolicy` presets:
   - `BLOCK_SCHEME_DOWNGRADE` (default) — block HTTPS→HTTP redirects
   - `ENFORCE_HTTPS` (strict) — only allow HTTPS destinations
   - `ALLOW_ALL` (legacy) — current behavior, no restrictions
3. Consider also blocking redirects to private/link-local IP ranges

dgenio:
## What changed

Adds a `RedirectPolicy` enum and redirect validation event hook to `create_mcp_http_client()` in `src/mcp/shared/_httpx_utils.py` to protect against SSRF attacks via HTTP redirects.

### Changes:

- **`src/mcp/shared/_httpx_utils.py`**:
  - Added `RedirectPolicy` enum with three presets:
    - `BLOCK_SCHEME_DOWNGRADE` (default) — blocks HTTPS→HTTP redirect downgrades
    - `ENFORCE_HTTPS` — only allows HTTPS redirect destinations
    - `ALLOW_ALL` — no restrictions (legacy behavior)
  - Added `_check_redirect()` async response event hook that validates redirect targets against the configured policy using `response.has_redirect_location` and the `Location` header
  - Added `redirect_policy` parameter to `create_mcp_http_client()` (default: `BLOCK_SCHEME_DOWNGRADE`)
  - `McpHttpClientFactory` Protocol left unchanged — `redirect_policy` is an internal concern of `create_mcp_http_client`, not part of the caller-side contract

- **`src/mcp/client/streamable_http.py`**:
  - Added docstring note on `http_client` parameter clarifying that user-provided clients do not receive SSRF redirect protection
  - Added `logger.debug` when a user-provided client is used

- **`tests/shared/test_httpx_utils.py`**:
  - 14 unit tests covering all three policies, edge cases (non-redirect responses, missing `Location` header), and all scheme combinations
  - 2 integration tests using `httpx.MockTransport` that exercise the event hook wiring end-to-end through the real httpx client

## Why

A malicious MCP server can respond with a 3xx redirect pointing to internal network addresses (`http://169.254.169.254`, `http://localhost`, etc.), enabling SSRF attacks. The SDK previously set `follow_redirects=True` with zero redirect validation. The default `BLOCK_SCHEME_DOWNGRADE` policy prevents the most common SSRF vector (HTTPS→HTTP downgrade to reach internal services) while remaining backward-compatible for legitimate use cases.

### Bug discovered during review

The initial implementation used `response.next_request` to inspect redirect targets, but httpx never populates `next_request` when `follow_redirects=True` — causing the protection to silently bypass all checks. Fixed to use `response.has_redirect_location` + parse the `Location` header directly, matching httpx's own `_build_redirect_request` flow.

## How verified

- `ruff check .` — all checks passed
- `ruff format --check .` — no formatting issues
- `pyright` — 0 errors, 0 warnings
- `pytest tests/shared/test_httpx_utils.py -v` — 16 passed
- `pytest tests/client/test_client.py -v` — 19 passed (no regressions)

## Tradeoffs / risks

- The new `redirect_policy` parameter on `create_mcp_http_client` has a default value, so all existing callers are unaffected — no breaking change
- `McpHttpClientFactory` Protocol is **not** modified, preserving compatibility for downstream code implementing custom factories
- Users who need the legacy behavior can pass `redirect_policy=RedirectPolicy.ALLOW_ALL`
- Private/link-local IP range blocking is not included in this PR — suggested as a follow-up

## Scope notes

- `RedirectPolicy` is not re-exported from `mcp.__init__` — `create_mcp_http_client` itself is also not publicly exported, so both share the same visibility via `mcp.shared._httpx_utils`. Public API surface expansion is suggested as a follow-up
- `HttpResource` in `resources/types.py` uses a bare `httpx.AsyncClient()` (separate concern, not addressed here)

Closes #2106

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -523,6 +523,9 @@ async def streamable_http_client(
         http_client: Optional pre-configured httpx.AsyncClient. If None, a default
             client with recommended MCP timeouts will be created. To configure headers,
             authentication, or other HTTP settings, create an httpx.AsyncClient and pass it here.
+            Note: User-provided clients do not receive SSRF redirect protection.
+            If redirect validation is required, use ``create_mcp_http_client()``
+            or configure redirect hooks manually.
         terminate_on_close: If True, send a DELETE request to terminate the session when the context exits.
 
     Yields:
@@ -543,6 +546,8 @@ async def streamable_http_client(
     if client is None:
         # Create default client with recommended MCP timeouts
         client = create_mcp_http_client()
+    else:
+        logger.debug("Using user-provided HTTP client; SSRF redirect protection is not applied")
 
     transport = StreamableHTTPTransport(url)
 
--- a/src/mcp/shared/_httpx_utils.py
+++ b/src/mcp/shared/_httpx_utils.py
@@ -1,14 +1,80 @@
 """Utilities for creating standardized httpx AsyncClient instances."""
 
+import logging
+from enum import Enum
 from typing import Any, Protocol
 
 import httpx
 
-__all__ = ["create_mcp_http_client", "MCP_DEFAULT_TIMEOUT", "MCP_DEFAULT_SSE_READ_TIMEOUT"]
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    "MCP_DEFAULT_SSE_READ_TIMEOUT",
+    "MCP_DEFAULT_TIMEOUT",
+    "RedirectPolicy",
+    "create_mcp_http_client",
+]
 
 # Default MCP timeout configuration
 MCP_DEFAULT_TIMEOUT = 30.0  # General operations (seconds)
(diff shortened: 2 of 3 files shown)


maxisbey:
Thanks for the PR. Tracking this in #3075 instead. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2183, opened 2026-02-28
fix: collapse single-error ExceptionGroups from task group cancellations

### LINKED ISSUE #2114: ExceptionGroup wrapping obscures real errors from task groups
## Summary

When SDK internal anyio task group tasks fail, the real error is wrapped in a `BaseExceptionGroup` alongside `Cancelled` exceptions from cancelled sibling tasks. This makes error classification extremely difficult for callers — they cannot reliably determine the root cause of a failure.

There are 16 `create_task_group()` usages across the SDK with no `except*` syntax or ExceptionGroup unwrapping anywhere.

## Expected Behavior

- Callers should receive the original exception, not an ExceptionGroup wrapping it alongside cancellation noise
- When a task group has one real failure and N cancelled siblings, only the real failure should propagate

## Current Behavior

- A single connection failure produces a `BaseExceptionGroup` containing the real error plus multiple `Cancelled` exceptions
- Callers must manually unwrap exception groups to find the root cause
- This affects ~80+ call sites across the SDK where task groups are used

## Affected Code

- `src/mcp/shared/session.py` (session task groups)
- `src/mcp/client/streamable_http.py` (transport task groups)
- All other `create_task_group()` sites

## Related

- #1830 — CancelledError indistinguishable from external cancellation (closed but related)

---

giulio-leone:
## Summary

Fixes #2114

When a task in an anyio task group fails, sibling tasks are cancelled and the resulting `Cancelled` exceptions are wrapped alongside the real error in a `BaseExceptionGroup`. This makes it extremely difficult for callers to classify the root cause of failures — they must manually unwrap exception groups to find what actually went wrong.

## Solution

Added a `collapse_exception_group()` utility and a drop-in `create_task_group()` context manager in `mcp.shared._exception_utils` that automatically unwraps single-error exception groups:

- **One real error + N cancelled** → raises the real error directly (chained to the original group via `__cause__`)
- **Multiple real errors** → preserves the `BaseExceptionGroup` as-is
- **All cancelled** → preserves the group as-is

## Applied Sites

The unwrapping is applied to all client-facing code paths:

| File | Change |
|------|--------|
| `shared/session.py` | `BaseSession.__aexit__` — affects all session-based operations |
| `client/stdio.py` | stdio transport task group |
| `client/sse.py` | SSE transport task group |
| `client/streamable_http.py` | Streamable HTTP transport task group |
| `client/websocket.py` | WebSocket transport task group |
| `client/_memory.py` | In-memory transport task group |

Server-side task groups are left unchanged as they typically have internal error handling.

## Tests

Added comprehensive tests in `tests/shared/test_exception_utils.py`:
- `TestCollapseExceptionGroup`: 5 unit tests covering all edge cases
- `TestCreateTaskGroup`: 3 integration tests verifying unwrapping, clean exit, and cause chaining

All 154 shared tests pass with 0 errors.

### DIFF
--- a/src/mcp/shared/_exception_utils.py
+++ b/src/mcp/shared/_exception_utils.py
@@ -14,8 +14,8 @@ from contextlib import asynccontextmanager
 import anyio
 from anyio.abc import TaskGroup
 
-if sys.version_info < (3, 11):  # pragma: no cover
-    from exceptiongroup import BaseExceptionGroup
+if sys.version_info < (3, 11):  # pragma: lax no cover
+    from exceptiongroup import BaseExceptionGroup  # pragma: lax no cover
 
 
 def collapse_exception_group(exc_group: BaseExceptionGroup[BaseException]) -> BaseException:
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -12,8 +12,8 @@ from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStre
 from pydantic import BaseModel, TypeAdapter
 from typing_extensions import Self
 
-if sys.version_info < (3, 11):  # pragma: no cover
-    from exceptiongroup import BaseExceptionGroup
+if sys.version_info < (3, 11):  # pragma: lax no cover
+    from exceptiongroup import BaseExceptionGroup  # pragma: lax no cover
 
 from mcp.shared._exception_utils import collapse_exception_group
 from mcp.shared.exceptions import MCPError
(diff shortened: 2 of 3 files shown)


giulio-leone:
All 26 CI checks pass. This collapses single-error ExceptionGroups for cleaner error messages. Ready for review.

maxisbey:
Thanks for working on this — but please don't open two PRs for the same issue with different scopes; it makes review harder.

Since #2114 is a P1/v2-scoped architectural change (not marked `good first issue`), could you instead propose your approach as a comment on #2114 first? That way we can align on scope (client-only vs all 16 sites) and pattern (drop-in wrapper vs `except*` vs per-site) before committing to an implementation.

PR #2187, opened 2026-03-01
fix: preserve notification metadata when related_request_id is 0

### LINKED ISSUE #1218: `simple-streamablehttp-stateless` example will not return data if `id=0`

Br1an67:
## Summary

Fixes #1218

When `related_request_id` is `0` (a valid JSON-RPC integer id), the falsy check `if related_request_id` in `BaseSession.send_notification` evaluated to `False`, causing the `ServerMessageMetadata` to be set to `None`. This meant notifications for requests with `id=0` were routed to the GET stream instead of the correct request-specific stream, so the client never received them.

## Changes

- **`src/mcp/shared/session.py`**: Replace `if related_request_id` with `if related_request_id is not None` so that integer `0` is correctly preserved.
- **`tests/shared/test_session.py`**: Add `test_send_notification_preserves_zero_related_request_id` to verify that metadata is preserved when `related_request_id=0`.

## Test

```
uv run pytest tests/shared/test_session.py -x  # 9 passed
uv run ruff check src/mcp/shared/session.py    # All checks passed
```

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -303,7 +303,9 @@ class BaseSession(
         )
         session_message = SessionMessage(
             message=jsonrpc_notification,
-            metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id else None,
+            metadata=ServerMessageMetadata(related_request_id=related_request_id)
+            if related_request_id is not None
+            else None,
         )
         await self._write_stream.send(session_message)
 
--- a/tests/shared/test_session.py
+++ b/tests/shared/test_session.py
@@ -416,3 +416,43 @@ async def test_null_id_error_does_not_affect_pending_request():
     # Pending request completed successfully
     assert len(result_holder) == 1
     assert isinstance(result_holder[0], EmptyResult)
+
+
+@pytest.mark.anyio
+async def test_send_notification_preserves_zero_related_request_id():
+    """Test that send_notification preserves related_request_id=0.
+
+    When related_request_id is 0 (a valid JSON-RPC id), the metadata must
+    still be set. Previously, the falsy check `if related_request_id`
+    treated 0 as False and dropped the metadata.
+
+    See: https://github.com/modelcontextprotocol/python-sdk/issues/1218
+    """
+    from mcp.shared.message import ServerMessageMetadata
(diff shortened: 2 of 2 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

PR #2188, opened 2026-03-01
fix: allow extra fields on RequestParams to match TypeScript SDK

### LINKED ISSUE #913: Inconsistency in `types.py` schema for `RequestParams`
**Describe the bug**

Not sure if I'm missing something, but I found that I was unable to add extra fields to `RequestParams` in the MCP Python schema in `types.py` (e.g., I'm trying to pass auth tokens along for `ListToolsRequest`). The Python MCP schema ignores extra fields at the top level of `RequestParams`, but the TypeScript MCP schema does not. 

**To Reproduce**

Try to create a `RequestParams` instance with an extra field at the top level (not inside `meta`), e.g.:

```python
RequestParams(foo="bar", meta=None)
```

The additional field `foo` is ignored.

**Expected behavior**

Extra fields at the top level of `RequestParams` should be accepted, matching the TypeScript schema’s behavior.

**Additional context**

- The TypeScript interface allows arbitrary extra fields at the top level of params via:

```ts
export interface Request {
  method: string;
  params?: {
    _meta?: {
      progressToken?: ProgressToken;
    };
    [key: string]: unknown;
  };
}
```

- The Python class only allows extra fields inside the `Meta` class:

```python
class RequestParams(BaseModel):
    class Meta(BaseModel):
        progressToken: ProgressToken | None = None
    meta: Meta | None = Field(alias="_meta", default=None)
```

Br1an67:
## Summary

Fixes #913

The TypeScript SDK defines request params with `[key: string]: unknown`, allowing extra top-level fields. The Python SDK's `RequestParams` silently dropped extra fields because `MCPModel` defaults to `extra="ignore"`.

## Changes

- **`src/mcp/types/_types.py`**: Add `model_config = ConfigDict(extra="allow")` to `RequestParams` so extra fields are preserved.
- **`tests/test_types.py`**: Add `test_request_params_allows_extra_fields` verifying that extra fields survive construction and serialization.

## Test

```
uv run pytest tests/test_types.py -x  # 15 passed
uv run ruff check src/mcp/types/_types.py  # All checks passed
```

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -372,7 +372,7 @@ def test_request_params_allows_extra_fields():
     """
     from mcp.types import RequestParams
 
-    params = RequestParams(**{"custom_field": "value", "another": 42})
+    params = RequestParams.model_validate({"custom_field": "value", "another": 42})
     assert params.model_extra == {"custom_field": "value", "another": 42}
 
     # Extra fields survive serialization round-trip


maxisbey:
Closing — the premise is based on an outdated spec version. In the 2025-11-25 schema, `RequestParams` [no longer has an index signature](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/446223010fdae8494a5df102f883e82828a56623/schema/2025-11-25/schema.ts#L53-L64); the `[key: string]: any` on `Request.params` is [explicitly commented as "unofficial extensions"](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/446223010fdae8494a5df102f883e82828a56623/schema/2025-11-25/schema.ts#L69). This was an intentional breaking change in spec PR [modelcontextprotocol/modelcontextprotocol#1692](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1692).

For passing custom data, `_meta` is the spec-sanctioned extension point (already `extra_items=Any` in this SDK). The TS SDK's Zod schemas match: `BaseRequestParamsSchema` has no `.loose()`/`.passthrough()`.

PR #1407, opened 2025-09-26
Improve OAuth protected resource metadata URL construction per RFC 9728

shulkx:
## Summary

Fixes RFC 9728 compliance issues where the `/.well-known/oauth-protected-resource` endpoint was not correctly constructed for resource identifiers containing path components, causing client discovery failures. (#1400)

## Motivation and Context

**Problem:** The current implementation has two critical RFC 9728 compliance issues:

1. **URL Construction Mismatch**: When `resource_server_url` contains a path (e.g., `http://localhost:8001/mcp`), the WWW-Authenticate header points to `http://localhost:8001/mcp/.well-known/oauth-protected-resource` (appending after the path), but the actual route is served at `http://localhost:8001/.well-known/oauth-protected-resource` (root level only).

2. **Non-RFC Compliant Path Insertion**: RFC 9728 §3.1 specifies that the well-known path should be inserted *between* the host and resource path, not appended after the full URL.

3. **Example Configuration Issue**: The `simple-auth` example had an incorrect `resource_server_url` configuration that didn't include the `/mcp` path component, causing resource validation errors in VS Code MCP clients.

**Impact:** 
- Clients following RFC 9728 get 404 errors when trying to discover metadata
- VS Code shows "Protected Resource Metadata resource does not match MCP server resolved resource" error
- Breaks interoperability with standards-compliant OAuth clients
- Prevents multi-tenant hosting scenarios

**Root Cause:** String concatenation approach doesn't properly handle URL path components according to RFC 9728 specification.

### 1. Created Reusable Utility Function
Added `build_resource_metadata_url()` in `src/mcp/server/auth/routes.py` to centralize RFC 9728 compliant URL construction:

```python
def build_resource_metadata_url(resource_server_url: AnyHttpUrl) -> AnyHttpUrl:
    """Build RFC 9728 compliant protected resource metadata URL."""
    parsed = urlparse(str(resource_server_url))
    # Handle trailing slash: if path is just "/", treat as empty
    resource_path = parsed.path if parsed.path != "/" else ""
    return AnyHttpUrl(
        f"{parsed.scheme}://{parsed.netloc}/.well-known/oauth-protected-resource{resource_path}"
    )
```

### 2. Updated WWW-Authenticate Header Construction
**Before (Non-compliant):**
```python
resource_metadata_url = AnyHttpUrl(
    str(resource_server_url).rstrip("/") + "/.well-known/oauth-protected-resource"
)
```

**After (RFC 9728 Compliant):**
```python
[...]

PR #2189, opened 2026-03-01
fix: include transport path in Protected Resource Metadata resource URL

### LINKED ISSUE #1264: Protected Resource Metadata resource erroneous when setting up authentication on server

Br1an67:
## Summary

Fixes #1264

The `resource` field in `/.well-known/oauth-protected-resource` was set to the base `resource_server_url` (e.g., `http://localhost:8000/`) instead of the actual protected endpoint URL (e.g., `http://localhost:8000/mcp`). Per [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728#PRConfigurationValidation), the resource identifier must match the URL that clients use to access the protected resource.

This caused VS Code Copilot (and other compliant clients) to reject the server with:
> Protected Resource Metadata resource "http://localhost:8000/" does not match MCP server resolved resource "http://localhost:8000/mcp"

## Changes

- **`src/mcp/server/lowlevel/server.py`**: Append `streamable_http_path` to `resource_server_url` when constructing the resource URL for both the protected resource metadata endpoint and the WWW-Authenticate header.
- **`src/mcp/server/mcpserver/server.py`**: Same fix for the SSE transport path.

## Test

```
uv run pytest tests/server/auth/test_protected_resource.py -x  # 14 passed
uv run pytest tests/server/mcpserver/auth/ -x  # 42 passed
uv run ruff check src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py  # All passed
```

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -46,6 +46,7 @@ from typing import Any, Generic
 
 import anyio
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+from pydantic import AnyHttpUrl
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.authentication import AuthenticationMiddleware
@@ -595,8 +596,11 @@ class Server(Generic[LifespanResultT]):
             # Determine resource metadata URL
             resource_metadata_url = None
             if auth and auth.resource_server_url:
+                # The protected resource URL must include the transport path
+                # (e.g. http://localhost:8000/mcp) per RFC 9728
+                actual_resource_url = AnyHttpUrl(str(auth.resource_server_url).rstrip("/") + streamable_http_path)
                 # Build compliant metadata URL for WWW-Authenticate header
-                resource_metadata_url = build_resource_metadata_url(auth.resource_server_url)
+                resource_metadata_url = build_resource_metadata_url(actual_resource_url)
 
             routes.append(
                 Route(
@@ -615,9 +619,10 @@ class Server(Generic[LifespanResultT]):
 
         # Add protected resource metadata endpoint if configured as RS
         if auth and auth.resource_server_url:  # pragma: no cover
+            actual_resource_url = AnyHttpUrl(str(auth.resource_server_url).rstrip("/") + streamable_http_path)
             routes.extend(
                 create_protected_resource_routes(
-                    resource_url=auth.resource_server_url,
+                    resource_url=actual_resource_url,
                     authorization_servers=[auth.issuer_url],
                     scopes_supported=auth.required_scopes,
                 )
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -982,10 +982,15 @@ class MCPServer(Generic[LifespanResultT]):
             # Determine resource metadata URL
             resource_metadata_url = None
             if self.settings.auth and self.settings.auth.resource_server_url:
+                from pydantic import AnyHttpUrl
+
                 from mcp.server.auth.routes import build_resource_metadata_url
 
+                # The protected resource URL must include the transport path
+                # (e.g. http://localhost:8000/sse) per RFC 9728
+                actual_resource_url = AnyHttpUrl(str(self.settings.auth.resource_server_url).rstrip("/") + sse_path)
                 # Build compliant metadata URL for WWW-Authenticate header
-                resource_metadata_url = build_resource_metadata_url(self.settings.auth.resource_server_url)
+                resource_metadata_url = build_resource_metadata_url(actual_resource_url)
 
             # Auth is enabled, wrap the endpoints with RequireAuthMiddleware
             routes.append(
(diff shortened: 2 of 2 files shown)


maxisbey:
Thanks for the PR and for digging into this — the underlying VS Code error is real and the issue has bitten multiple people. However, the approach here introduces a regression for servers that are already configured correctly, so I'm going to close this in favor of a docs-focused fix.

## Why appending `streamable_http_path` doesn't work

**Double path for correctly-configured servers.** PR #1407 (Oct 2025) established that `resource_server_url` should be the full endpoint URL, and `simple-auth` was updated accordingly: https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/examples/servers/simple-auth/mcp_simple_auth/server.py#L34

With that config, this PR produces:
```python
"http://localhost:8001/mcp".rstrip("/") + "/mcp"  →  "http://localhost:8001/mcp/mcp"
```

**`streamable_http_path` is the wrong coordinate system.** It's the route *inside* the Starlette app, not the public URL. The SDK's own multi-server example mounts with `streamable_http_path="/"` under `Mount("/api", ...)`: https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/examples/snippets/servers/streamable_http_multiple_servers.py#L44-L45

Here the endpoint is `/api` but `streamable_http_path` is `/` — concatenation gives the wrong answer. Same for reverse proxies. Only the user knows the public URL.

**Diverges from the TypeScript SDK**, which uses `resourceServerUrl.href` verbatim: https://github.com/modelcontextprotocol/typescript-sdk/blob/978b6ccaeaa6ca561e6f1f07874d09b02d754f98/src/server/auth/router.ts#L206

## The actual gap (bigger than I first thought)

#1407 established the contract but didn't update everything. Still broken on `main`:

- [`examples/snippets/servers/oauth_server.py:27`](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/examples/snippets/servers/oauth_server.py#L27) — `http://localhost:3001` (missing `/mcp`)
- [`README.md:1036`](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/README.md#L1036) — same snippet embedded
- [`AuthSettings.resource_server_url` docstring](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/src/mcp/server/auth/settings.py#L28-L29) — doesn't say "include the transport path"
- [`simple-auth/server.py:131`](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/examples/servers/simple-auth/mcp_simple_auth/server.py#L131) hardcodes `/mcp` but [line 105](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/examples/servers/simple-auth/mcp_simple_auth/server.py#L105) offers `--transport=sse` — run with SSE and you get the same RFC 9728 mismatch

Worth noting: the python-sdk *client* uses hierarchical prefix matching ([`auth_utils.py:65`](https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/src/mcp/shared/auth_utils.py#L65)) rather than RFC 9728 §3.3 strict equality, so python↔python testing accepts the misconfiguration that VS Code rejects. That's part of why this keeps surfacing.

There's also a genuine API awkwardness here — `AuthSettings` is construction-time but `streamable_http_path` is transport-time, so the two have to be kept in sync manually. A startup warning when they disagree would catch this at boot instead of at the VS Code 401. Worth tracking separately.

If you'd like to take the docs/example fixes as a follow-up PR, that would close #1264 properly. Happy to help scope it.

PR #2190, opened 2026-03-01
fix: remove shell=True from subprocess calls in mcp dev

### LINKED ISSUE #1257: Don’t use shell=True in mcp dev subprocess on Windows (command injection risk)
### Initial Checks

### Description

**What’s happening:**

**How this could be abused:**

**Please fix:**

Thanks!

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
```

Br1an67:
## Summary

Fixes #1257

`mcp dev` used `shell=True` when running `npx` on Windows, exposing a command injection risk via shell metacharacters in file paths (e.g., `server&calc.py`).

## Changes

- **`src/mcp/cli/cli.py`**:
  - Remove `shell=True` from both `subprocess.run` calls (`_get_npx_command` and the Inspector invocation).
  - `_get_npx_command()` already resolves to `npx.cmd`/`npx.exe` on Windows, so `shell=True` is unnecessary.
  - Add `FileNotFoundError` to the exception handler for robustness when the command is not found without shell expansion.

## Test

```
uv run ruff check src/mcp/cli/cli.py  # All checks passed
```

### DIFF
--- a/src/mcp/cli/cli.py
+++ b/src/mcp/cli/cli.py
@@ -45,9 +45,9 @@ def _get_npx_command():
         # Try both npx.cmd and npx.exe on Windows
         for cmd in ["npx.cmd", "npx.exe", "npx"]:
             try:
-                subprocess.run([cmd, "--version"], check=True, capture_output=True, shell=True)
+                subprocess.run([cmd, "--version"], check=True, capture_output=True)
                 return cmd
-            except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

maxisbey:
Closing in favor of #2524, which uses `shutil.which()` to resolve npx and has test coverage. Thanks for the contribution.

PR #2195, opened 2026-03-02
fix: make Context logging methods spec-compliant by accepting Any data type

Jimil-Joshi:
## Summary

Fixes #397

Per the [MCP specification](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/utilities/logging/#log-message-notifications), the `data` field in `LoggingMessageNotificationParams` allows **any JSON serializable type**:

> The data to be logged, such as a string message or an object. Any JSON serializable type is allowed here.

However, the `Context.log()` and convenience methods (`debug`, `info`, `warning`, `error`) were typed as `message: str`, which prevented users from logging structured data like dicts, lists, numbers, etc.

## Changes

- **`Context.log()`**: Changed `message: str` parameter to `data: Any` to match the MCP spec and the underlying `ServerSession.send_log_message(data: Any)`
- **Convenience methods** (`debug`, `info`, `warning`, `error`): Updated to accept `data: Any` instead of `message: str`
- **Removed `extra` parameter**: Structured data can now be passed directly as `data` (e.g., `ctx.info({"event": "processing", "count": 5})`) making the `extra` workaround unnecessary
- **Updated docstrings**: Document that any JSON serializable type is accepted
- **Added test**: Verifies logging of dict, list, number, boolean, and None types

## Example

`python
@server.tool()
async def my_tool(x: int, ctx: Context) -> str:
    # String messages still work (backward compatible)
    await ctx.info(f"Processing {x}")
    
    # Now also supports structured data per MCP spec
    await ctx.info({"event": "processing", "input": x})
    await ctx.debug(["step1", "step2", "step3"])
    await ctx.warning(42)
    return "done"
`

## Test Results

All 87 tests in `tests/server/mcpserver/test_server.py` pass, including the new structured data logging test.

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -1122,6 +1122,11 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
         await ctx.warning("Warning message")
         await ctx.error("Error message")
 
+        # Log structured data (any JSON serializable type)
+        await ctx.info({"event": "processing", "input": x})
+        await ctx.debug(["step1", "step2", "step3"])
+        await ctx.info(42)
+
         # Report progress
         await ctx.report_progress(50, 100)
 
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #1907 as that predates this PR.

PR #2197, opened 2026-03-02
fix: resource with only Context param incorrectly classified as template

### LINKED ISSUE #1635: Using `Context` in resource definitions results in inccorect resource template classification

Varun6578:
## Summary

Fixes a bug where a resource handler with only a Context parameter (no URI template params) was incorrectly classified as a resource template instead of a regular resource.

## Problem

When using:
```python
@mcp.resource("time://current")
async def current_time(ctx: Context[ServerSession, AppContext]):
    return "11:41"
```

The resource was registered as a template because ool(sig.parameters) counted the Context parameter. It only appeared in list_resource_templates and could not be invoked via 
ead_resource.

## Fix

- Exclude Context parameter before checking has_func_params in the resource decorator (server.py)
- Add context_kwarg support to FunctionResource so context-only resources can still access server context when read
- Pass context to FunctionResource.read() in MCPServer.read_resource()

## Test plan

- Added test verifying context-only resource is classified as resource (not template) and readable via Client
- All 87 server tests pass
- All 59 resource-related tests pass

Fixes #1635

### DIFF
--- a/src/mcp/server/mcpserver/resources/types.py
+++ b/src/mcp/server/mcpserver/resources/types.py
@@ -51,12 +51,18 @@ class FunctionResource(Resource):
     """
 
     fn: Callable[[], Any] = Field(exclude=True)
+    context_kwarg: str | None = Field(default=None, exclude=True)
 
-    async def read(self) -> str | bytes:
+    async def read(self, context: Any | None = None) -> str | bytes:
         """Read the resource by calling the wrapped function."""
         try:
+            # Inject context if needed
+            kwargs: dict[str, Any] = {}
+            if self.context_kwarg is not None and context is not None:
+                kwargs[self.context_kwarg] = context
+
             # Call the function first to see if it returns a coroutine
-            result = self.fn()
+            result = self.fn(**kwargs)
             # If it's a coroutine, await it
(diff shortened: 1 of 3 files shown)


maxisbey:
Thanks — closing as duplicate of #1666, which fixes the same bug and takes a slightly broader approach (uses `call_fn_with_arg_validation` for consistency with how tools handle context injection). Tracking issue is #1405.

PR #2198, opened 2026-03-02
fix: prevent tool exceptions from leaking internal details to client

### LINKED ISSUE #698: Tool.run should not reveal exception value to the client
If any exception is raised during a tool call, the exception error message is revealed to the client. This is generally bad practice in python, as private information about the server may be conveyed through the exception. 

[src/mcp/server/fastmcp/tools/base.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/fastmcp/tools/base.py#L100)
```python
    async def run(
        self,
        arguments: dict[str, Any],
        context: Context[ServerSessionT, LifespanContextT] | None = None,
    ) -> Any:
        """Run the tool with arguments."""
        try:
            return await self.fn_metadata.call_fn_with_arg_validation(
                self.fn,
                self.is_async,
                arguments,
                {self.context_kwarg: context}
                if self.context_kwarg is not None
                else None,
            )
        except Exception as e:
            raise ToolError(f"Error executing tool {self.name}: {e}") from e
```

This exposes the server to vulnerabilities such as information leakage, attack surface mapping, etc.

Ideally whomever is implementing the tool should be handling errors and explicitly raising `ToolError` if the error is meant to be seen by the client. The run definition should be modified to 

```python
    async def run(
        self,
        arguments: dict[str, Any],
        context: Context[ServerSessionT, LifespanContextT] | None = None,
    ) -> Any:
        """Run the tool with arguments."""
        try:
            return await self.fn_metadata.call_fn_with_arg_validation(
                self.fn,
                self.is_async,
                arguments,
                {self.context_kwarg: context}
                if self.context_kwarg is not None
                else None,
            )
        except ToolError:
            # Re-raise if it's a handled error
            raise
        except Exception as e:
            logger.exception(e)
            raise ToolError(f"An unexpected error occurred while executing tool {self.name}")
```

There may be other areas where this is occurring such as resources or prompts, but I have not yet tested to see if those also expose internal errors.

Varun6578:
## Summary

Fixes #698

`Tool.run()` and `_handle_call_tool()` previously exposed internal exception details (`str(e)`) to MCP clients. This could leak sensitive information like connection strings, file paths, or internal state.

## Changes

**`src/mcp/server/mcpserver/tools/base.py`:**
- Added `except ToolError: raise` to pass through intentional user-facing errors
- Changed generic exception handler to return `An unexpected error occurred executing tool {name}` instead of `str(e)`
- Added server-side logging of the full traceback for debugging

**`src/mcp/server/mcpserver/server.py`:**
- Added separate `except ToolError` handler in `_handle_call_tool()` to pass through ToolError messages
- Changed generic exception handler to return a generic message instead of `f"Error calling tool {name}: {e}"`

**Tests updated:**
- `test_server.py`: Verified exception details are NOT leaked in error responses
- `test_tool_manager.py`: Updated regex match for new generic message
- `test_url_elicitation_error_throw.py`: Verified generic message and no detail leak
- `test_sampling_callback.py`, `test_list_roots_callback.py`: Updated expected error messages

## Security

This is a security hardening change. Before this fix, any Python exception raised inside a tool function would have its full `str()` representation sent to the client, potentially exposing:
- Database connection strings
- File system paths
- Internal API keys or tokens
- Stack trace information


src/mcp/server/mcpserver/server.py
```diff
@@ -304,8 +304,14 @@ async def _handle_call_tool(
             result = await self.call_tool(params.name, params.arguments or {})
         except MCPError:
             raise
-        except Exception as e:
+        except ToolError as e:
             return CallToolResult(content=[TextContent(type="text", text=str(e))], is_error=True)
+        except Exception:  # pragma: no cover
+            logger.exception(f"Error calling tool {params.name}")
```
maxisbey, line 318: if we have logging here, then do we need logging in base.py? won't that cause duplicate logs?

maxisbey, line 318: please remove the pragma and add a unit test

maxisbey (changes requested):

src/mcp/server/mcpserver/server.py
Varun6578: Good catch — removed logger.exception() from base.py. The defensive handler in server.py is the single logging point now, so no duplicate logs.

Varun6578: Done — removed the \# pragma: no cover\ and added \	est_handle_call_tool_defensive_exception_handler\ which patches \call_tool\ to raise a \RuntimeError\ and verifies the client gets a generic error with no internal details leaked.

PR #2199, opened 2026-03-02
fix: handle HTTP 4xx/5xx gracefully in StreamableHTTP transport

### LINKED ISSUE #1295: 401 in Streamable HTTP should be handled gracefully
### Initial Checks

### Description

Currently, `streamable_http` will call `response.raise_for_status()` when handling POST requests. [(ref)](https://github.com/modelcontextprotocol/python-sdk/blob/9a8592e7d4d69b785df8b5d9d7b694cb6cbb149f/src/mcp/client/streamable_http.py#L278).

This means that a 401 will bubble up and cause the entire program to crash. 

While a 401 on server initialize could warrant an exception, we have a use case where a MCP server will return 401 for specific unauthorized operations, and allow others.

This should be handled gracefully, similar to how [`_handle_unexpected_content_type`](https://github.com/modelcontextprotocol/python-sdk/blob/9a8592e7d4d69b785df8b5d9d7b694cb6cbb149f/src/mcp/client/streamable_http.py#L340C15-L340C46) works, returning a JSONRPC message outlining an unauthorized exception was thrown.

This would let an agent or other client handle it instead of crashing the program.

### Python & MCP Python SDK

```Text
Python 3.12 and MCP SDK v1.13.1
```

Varun6578:
## Summary

Fixes #1295

HTTP 4xx/5xx responses in the StreamableHTTP transport previously caused crashes via `raise_for_status()`, or returned generic error messages. This fix handles all HTTP error statuses gracefully.

## Changes

**`src/mcp/client/streamable_http.py`:**
- **POST requests**: Error message now includes HTTP status code (`Server returned HTTP 401`) instead of generic `Server returned an error response`
- **GET SSE listener**: Replaced `raise_for_status()` with status code check; logs warning and increments retry counter instead of crashing
- **Resumption GET**: Replaced `raise_for_status()` with status code check; sends JSONRPC error to client
- **Reconnection GET**: Replaced `raise_for_status()` with status code check; sends JSONRPC error to client

**Tests:**
- Added `tests/client/test_streamable_http.py` with tests for HTTP 401 and 503 error handling
- Updated `test_notification_response.py` to match new error message format

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -194,7 +194,7 @@ class StreamableHTTPTransport:
                     headers[LAST_EVENT_ID] = last_event_id
 
                 async with aconnect_sse(client, "GET", self.url, headers=headers) as event_source:
-                    if event_source.response.status_code >= 400:
+                    if event_source.response.status_code >= 400:  # pragma: no cover
                         logger.warning(f"GET SSE returned HTTP {event_source.response.status_code}")
                         attempt += 1
                         continue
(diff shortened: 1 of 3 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #1998.

Issue #423, opened 2025-04-03
MCP SSE Server: Received request before initialization was complete


folkvir:
**Describe the bug**

The MCP server receive `Received request before initialization was complete` after a second deployment when a client is trying to list the different tools or even when it tires to run a tool on the server.

**To Reproduce**

SImple main.py which is one of your examples
```python
import httpx
from mcp.server.fastmcp import FastMCP

mcp = FastMCP("My App")


@mcp.tool()
def calculate_bmi(weight_kg: float, height_m: float) -> float:
    """Calculate BMI given weight in kg and height in meters"""
    return weight_kg / (height_m**2)


@mcp.tool()
async def fetch_weather(city: str) -> str:
    """Fetch current weather for a city"""
    async with httpx.AsyncClient() as client:
        response = await client.get(f"https://api.weather.com/{city}")
        return response.text

app = mcp.sse_app()
```

* Start using uvicorn `FASTMCP_LOG_LEVEL=DEBUG uv run uvicorn main:app --port=8001 --host=0.0.0.0 --timeout-graceful-shutdown 5`

* Run the inspector in SSE mode and listen on `http://localhost:8001/see`. You will be able to list tools.
* But now, reload you server to simulate a new deployment using the same command: `FASTMCP_LOG_LEVEL=DEBUG uv run uvicorn main:app --port=8001 --host=0.0.0.0 --timeout-graceful-shutdown 5`
* Go to the inspector and try to list the tools, it will timeout and you'll get the following server logs

```bash
FASTMCP_LOG_LEVEL=DEBUG uv run uvicorn main:app --port=8001 --host=0.0.0.0 --timeout-graceful-shutdown 5                                       
[04/03/25 23:48:40] DEBUG    SseServerTransport initialized with endpoint: /messages/                                                                                                 sse.py:79
INFO:     Started server process [89042]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8001 (Press CTRL+C to quit)
[04/03/25 23:48:41] DEBUG    Setting up SSE connection                                                                                                                                    sse.py:87
                    DEBUG    Created new session with ID: f55c6f64-be08-464c-bbab-203223adb44c                                                                                           sse.py:100
                    DEBUG    Starting SSE response task                                                                                                                                  sse.py:127
                    DEBUG    Yielding read and write streams                                                                                                                             sse.py:130
[...]

PR #2044, opened 2026-02-12
fix: clean up session when disconnected

nightosong:
## Motivation and Context

Changes:
* Remove session_id from _read_stream_writers when the SSE connection is closed.
* Ensure proper cleanup after EventSourceResponse returns.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #2200, opened 2026-03-02
fix: clean up SSE session on client disconnect

### LINKED ISSUE #1227: SSE: ConnectionClosed exception after session disconnect
### Initial Checks

### Description

In our environment, the SSE connections sometimes get dropped:

```
DEBUG:sse_starlette.sse:Got event: http.disconnect. Stop streaming.
DEBUG:root:Client session disconnected 081769ba-c683-471b-af22-1333409ab669
```

When this happens, on the next request for that session, we get a `ClosedResourceError` exception:

```
INFO: 100.100.0.154:49316 - "POST /messages/?session_id=081769bac683471baf221333409ab669 HTTP/1.1" 202 Accepted
ERROR:__main__:SSE post message error:
File "/app/mcp_renewables/server.py", line 714, in handle_post_message_with_error_handling
return await sse.handle_post_message(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/.venv/lib/python3.12/site-packages/mcp/server/sse.py", line 249, in handle_post_message
await writer. Send(session_message)
File "/app/.venv/lib/python3.12/site-packages/anyio/streams/memory.py", line 242, in send
self.send_nowait(item)
File "/app/.venv/lib/python3.12/site-packages/anyio/streams/memory.py", line 211, in send_nowait
raise ClosedResourceError
anyio.ClosedResourceError
```

I believe the reason is that when the session disconnects, its file handles are not properly removed as explained in #703. I have forked the repo and applied the fix and it seems to solve the problem.

### Example Code: the fix I applied

```Python
diff --git a/src/mcp/server/sse.py b/src/mcp/server/sse.py
index b7ff332..29acecb 100644
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -179,7 +179,7 @@ class SseServerTransport:
 
-            async def response_wrapper(scope: Scope, receive: Receive, send: Send):
+            async def response_wrapper(scope: Scope, receive: Receive, send: Send, transport: SseServerTransport):
                 """
                 The EventSourceResponse returning signals a client close / disconnect.
                 In this case we close our side of the streams to signal the client that
@@ -190,10 +190,13 @@ class SseServerTransport:
                 )
+                await read_stream.aclose()
+                await write_stream.aclose()
+                transport._read_stream_writers.pop(session_id)
                 logging.debug(f"Client session disconnected {session_id}")
 
             logger.debug("Starting SSE response task")
-            tg.start_soon(response_wrapper, scope, receive, send)
+            tg.start_soon(response_wrapper, scope, receive, send, self)
 
             logger.debug("Yielding read and write streams")
             yield (read_stream, write_stream)
```

### Python & MCP Python SDK

```Text
1.12.2
```

### References

Related PRs:

- #612 
- #703

Varun6578:
> [!NOTE]
> **EDIT from @maxisbey:** Retargeting to `Fixes #1227` — see https://github.com/modelcontextprotocol/python-sdk/pull/2200#issuecomment-3997883845 for details. The fix is correct and useful, but #423 is a different bug (client-side EventSource auto-reconnect without re-initialize) already mitigated by #822 and #1478. The "server reload" claim below is inaccurate — a uvicorn restart gives a fresh empty dict; this leak only occurs within a single process lifetime.

---

## Summary

Fixes #1227

After a client disconnects from the SSE endpoint, the session entry was never removed from `_read_stream_writers`. This caused stale sessions to accumulate, ~~and after a server reload old session IDs could still be found in the dict (with closed streams),~~ leading to confusing behavior.

## Changes

**`src/mcp/server/sse.py`:**
- Added `self._read_stream_writers.pop(session_id, None)` in the disconnect handler (`response_wrapper`) to clean up the session entry when a client disconnects

This is a one-line fix. The streams were already being closed on disconnect; this ensures the session dict entry is also removed so that subsequent POST requests to the old session ID correctly receive a 404 response.

### DIFF
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -186,6 +186,7 @@ class SseServerTransport:
                 )
                 await read_stream_writer.aclose()
                 await write_stream_reader.aclose()
+                self._read_stream_writers.pop(session_id, None)
                 logging.debug(f"Client session disconnected {session_id}")
 
             logger.debug("Starting SSE response task")
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -611,3 +611,41 @@ async def test_sse_client_handles_empty_keepalive_pings() -> None:
             assert not isinstance(msg, Exception)
             assert isinstance(msg.message, types.JSONRPCResponse)
             assert msg.message.id == 1
+
+
+@pytest.mark.anyio
+async def test_sse_session_cleanup_on_disconnect(server: None, server_url: str) -> None:
+    """Regression test for https://github.com/modelcontextprotocol/python-sdk/issues/423
+
+    When a client disconnects, the server should remove the session from
+    _read_stream_writers. Without this cleanup, stale sessions accumulate and
+    POST requests to disconnected sessions are incorrectly accepted instead
+    of returning 404.
+    """
+    captured_session_id: str | None = None
+
+    def on_session_created(session_id: str) -> None:
+        nonlocal captured_session_id
+        captured_session_id = session_id
+
+    # Connect a client session, then disconnect
+    async with sse_client(server_url + "/sse", on_session_created=on_session_created) as streams:
+        async with ClientSession(*streams) as session:
+            await session.initialize()
+
+    assert captured_session_id is not None
+
+    # After disconnect, POST to the stale session should return 404
+    # (not 202 as it did before the fix)
+    async with httpx.AsyncClient() as client:
+        with anyio.fail_after(5):
+            while True:
+                response = await client.post(
+                    f"{server_url}/messages/?session_id={captured_session_id}",
+                    json={"jsonrpc": "2.0", "method": "ping", "id": 99},
+                    headers={"Content-Type": "application/json"},
+                )
+                if response.status_code == 404:
+                    break
+                await anyio.sleep(0.1)
+        assert response.status_code == 404
--- a/tests/shared/test_sse.py
+++ b/tests/shared/test_sse.py
@@ -615,12 +615,12 @@ async def test_sse_client_handles_empty_keepalive_pings() -> None:
 
 @pytest.mark.anyio
 async def test_sse_session_cleanup_on_disconnect(server: None, server_url: str) -> None:
-    """Regression test for https://github.com/modelcontextprotocol/python-sdk/issues/423
+    """Regression test for https://github.com/modelcontextprotocol/python-sdk/issues/1227
 
     When a client disconnects, the server should remove the session from
     _read_stream_writers. Without this cleanup, stale sessions accumulate and
-    POST requests to disconnected sessions are incorrectly accepted instead
-    of returning 404.
+    POST requests to disconnected sessions return 202 Accepted followed by a
+    ClosedResourceError when the server tries to write to the dead stream.
     """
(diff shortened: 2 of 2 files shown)


maxisbey:
Please provide evidence this fixes the original issue, I'd be keen to see both a reproduction and a unit test if possible

Varun6578:
I've added a regression test (test_sse_session_cleanup_on_disconnect in tests/shared/test_sse.py) that:

1. **Reproduces the issue**: Connects a client session, captures the session ID via on_session_created, then disconnects.
2. **Verifies the fix**: After disconnect, POSTs to the stale session ID and asserts it returns 404. Before the fix, it would return 202, accepting the message into a dead session.

The test uses a retry loop with a 5-second timeout to account for the async disconnect processing, and runs alongside the existing 26 SSE tests (all passing).

maxisbey:
Thanks for this @Varun6578 — the fix is correct and the test is solid (fails on main with `202 != 404`, passes here, stable across 30 runs). Merging.

A few housekeeping notes for the record:

**This actually fixes #1227, not #423.** I've retargeted the PR accordingly. #1227 is the exact production symptom this addresses:

```
"POST /messages/?session_id=... HTTP/1.1" 202 Accepted
...
anyio.ClosedResourceError
```

That issue was closed as not-planned because nobody provided a repro — your test is that repro.

**Why this isn't #423:** #423's reproduction is a full uvicorn restart. After a process restart, `_read_stream_writers` is a fresh empty dict — stale entries are physically impossible. What's actually happening in #423 is the browser's `EventSource` silently auto-reconnects after the server comes back up, the server hands it a brand-new session, and the client sends `tools/list` without re-running `initialize`. That's client-side; the server-side crash it caused was already addressed by #822 (wraps the `RuntimeError`, sends a JSON-RPC error back instead of blowing up) and #1478 (fixed the `notifications/initialized`/`tools/list` race from [this comment](https://github.com/modelcontextprotocol/python-sdk/issues/423#issuecomment-2968839587)). I ran the #423 repro with and without this patch — identical behaviour both times.

**Spec alignment:** The legacy SSE spec ([2024-11-05](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports)) is silent on dead-session responses, but Streamable HTTP ([2025-03-26 onwards](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports)) says *"the server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with HTTP 404 Not Found."* The server already returns 404 for never-existed session IDs; this makes disconnected sessions hit the same path.

**Re: #2044** — that's a byte-identical diff without a test; will close as superseded.

Pushed one small commit to update the test docstring reference.

maxisbey (approved):

thebenlamm:
Just wanted to flag that this fix doesn't appear to have been backported to the `v1.x` branch, which is the branch all PyPI releases since v1.24.0 are cut from. As a result, the fix isn't in any released version — including 1.27.0 (released April 2, confirmed by inspecting the wheel).

Hit this in production: server leaked from 21 MB to 2.7 GB RSS over ~47 hours with 5–6 concurrent SSE connections (3,314 total connections logged). A backport PR targeting `v1.x` would get this into an actual release.

Issue #2112, opened 2026-02-19
Pass context explicitly to tool/resource/prompt handlers instead of ContextVar


maxisbey:
## Summary

Replace the current ContextVar-based context access in MCPServer handlers with explicit parameter passing.

## Problem

MCPServer currently uses a ContextVar to make context available to tool/resource/prompt handlers. This is implicit, hard to test, and doesn't compose well.

## Options Under Consideration

1. **Two decorators:** `app.tool()` (no context) and `app.tool_ctx()` (receives context as first param)
2. **Single decorator with signature inspection:** `app.tool()` inspects the handler signature and injects context if a matching parameter is present

Design TBD.

## This is a breaking change

Existing handlers that access context via ContextVar will need to be updated.

---

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

maxisbey:
I've been going through the design choice options around _how_ the context is passed into tool handlers, and whether we should change how it is currently. The first thing I'm changing is removing the option for using the ContextVar to grab the Context and instead just have it as a param to your tool. But whether to change how it's passed I'm not sure on.

Here are the options I have been considering:

# Options

## A: Unchanged, keep as current - Context param as any parameter

This allows the following signatures:

```python
@app.tool()
def my_tool(my_arg: int, my_arg_2: int) -> int: ...


@app.tool()
def my_tool_2(ctx: Context, my_arg: int, my_arg_2: int) -> int: ...  # ctx first


@app.tool()
def my_tool_3(my_arg: int, my_arg_2: int, ctx: Context) -> int: ...  # ctx last, both work today
```

- This leaves it as-is
- We don't get strong type checking because ctx can be anywhere
- Requires us to do function introspection to find the Context param, and then figure out how to pass it in

## B: Separate decorators - breaking change

Separate the `@tool` and similar decorators into two methods, `@tool` and `@tool_ctx`, where `@tool` doesn't inject Context, but `@tool_ctx` does and _only_ at the first position.

```python
@app.tool()
def my_tool(my_arg: int, my_arg_2: int) -> int: ...


@app.tool_ctx()
def my_tool_ctx_first(ctx: Context, my_arg: int, my_arg_2: int) -> int: ...  # OK


@app.tool_ctx()  # pyright ERROR: first positional is int, not Context
def my_tool_ctx_last(my_arg: int, my_arg_2: int, ctx: Context) -> int: ...


[...]

Issue #1742, opened 2025-12-05
Introduce typed error classes with metadata


maxisbey:
## Summary

Replace generic exceptions with SDK-native error classes that include useful metadata like fault attribution, retryability, and HTTP details.

## Problem

The SDK currently raises generic exceptions from various layers (httpx, anyio, task groups) with no context about:
- Whether the error is retryable
- Who is at fault (client, server, network, protocol violation)
- HTTP status codes and response bodies (often discarded)

This makes it hard for users to build robust error handling.

## Goal

A hierarchy of typed errors like:
- `McpError` (base) with `fault`, `is_retryable`, `http_status_code`, `response_body`
- `McpTransportError` for connection/network issues
- `McpProtocolError` for invalid JSON-RPC responses
- `McpAuthError` for 401/403 responses

---

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

jack-elliot-doherty:
Hi. I have a narrow client-side case that fits this issue.

Stilla currently uses MCP Python SDK 1.28.1 from the v1 line. In that version, tool-result validation happens in `ClientSession._validate_tool_result()`. Upstream `main` contains the v2 code. In v2, the equivalent public method is `ClientSession.validate_tool_result()`.

Both implementations raise plain `RuntimeError` for three tool-result contract failures:

- A tool declares an output schema but returns no structured content.
- The structured content does not match the output schema.
- The tool’s output schema is invalid.

A client has no public exception type that identifies these failures. Without one, it must use broad `RuntimeError` handling or rely on implementation details such as the error text. Our downstream integration currently inspects that text. We also use a contract test that fails if an SDK update changes it. A dedicated exception type would be safer.

Would you accept a small PR that adds a public exception for these cases, for example `ToolResultValidationError`, while keeping `RuntimeError` as its base class for existing callers?

The PR would:

- Raise the new type for these three paths.
- Keep the current error messages.
- Keep the underlying `jsonschema` exception as the cause or context where one exists.
- Update the relevant `Raises:` documentation and output-schema validation tests.
- Export the exception from the public API at the location that maintainers prefer.

I see that `v1.x` accepts only security fixes and critical bug fixes. Should I target `main` only, or would you also consider a separate `v1.x` backport?

I am asking before I start because the contribution guide says that new public APIs need discussion and that contributors must wait for maintainer feedback or a `ready for work` label.

[...]

Issue #1684, opened 2025-11-28
Refactor global contextvars request context into explicit parameters


dgenio:
## Summary

The Python SDK currently uses a global `contextvars.ContextVar` (e.g. `request_ctx`) to hold per-request state for the server. This enables convenience helpers like `get_request_context()`, but it also introduces implicit dependencies and makes testing, debugging, and refactoring harder.

This proposal suggests moving towards **explicit context passing** in handlers and eventually **deprecating global request context access**, with a migration path.

## Background

Today, the low-level server and FastMCP rely on a global `ContextVar` that holds a `RequestContext`. Handler code can implicitly call `get_request_context()` without any indication in the function signature that it depends on request-local state.

This has several drawbacks:

- **Hidden dependencies:** Functions depend on ambient state, not their parameters.
- **Debugging complexity:** Stack traces do not show where the context came from.
- **Testability:** Unit tests must set up `ContextVar` state or fall back to full integration tests.
- **Refactoring risk:** Moving code between modules can silently break if context assumptions change.

### Phase 1 – Explicit context in handler signatures

1. Make `RequestContext` (or a higher-level `Context`) an **explicit parameter** for all tool/resource handlers at the low-level server boundary.
2. At the FastMCP layer, inject the context explicitly into handler callsites instead of relying on global `ContextVar` lookups.
3. Keep `get_request_context()` for now, but document it as **deprecated** and implemented in terms of the explicit parameter passing.

### Phase 2 – Internal-only use (or removal) of `ContextVar`

4. Refactor internal call chains so that the `RequestContext` is always passed as an argument rather than read from a global.
5. Limit any remaining `ContextVar` usage to:
   - bridging with external frameworks (if needed), or
   - cross-cutting concerns like tracing where explicit parameter threading is impossible.
6. Remove `get_request_context()` (or keep only as a thin, clearly discouraged compatibility helper).

### Migration and compatibility

- Provide an explicit, documented pattern for handlers that need context:
  - e.g. `async def tool(ctx: Context, ...) -> ...`
- Keep the current behavior working for at least one minor release, with deprecation warnings where appropriate.
- Update examples and docs to promote explicit context usage.

## Why this matters

- **Testability:** Handlers become pure-ish functions with explicit dependencies.
- **Debugging:** It’s easier to follow how context flows through the system.
- **Robustness:** Less risk of accidental context leakage across concurrent requests.
- **API clarity:** Users see, in signatures, that a handler depends on request context.

[...]

PR #2203, opened 2026-03-02
refactor: remove request_ctx ContextVar, thread Context explicitly

maxisbey:
Removes the `request_ctx` ContextVar and threads `Context` explicitly through the MCPServer request-handling chain.

## Motivation and Context

The `request_ctx` ContextVar in `mcp.server.lowlevel.server` was redundant: the lowlevel server already passes `ServerRequestContext` as the first argument to every `_handle_*` method. The ContextVar was a second mechanism carrying the same value, used only by `MCPServer.get_context()`.

This PR removes the ContextVar entirely, making the data flow explicit. `_handle_*` methods construct the high-level `Context` at the lowlevel→MCPServer boundary and pass it through.

Part of #2112 (context refactor) / #1684 (contextvars cleanup).

## How Has This Been Tested?

- All existing tests pass (1123 passed, 98 skipped, 1 xfailed)
- New tests for the context-required guards in `Tool.run()`, `Prompt.render()`, `ResourceTemplate.create_resource()`
- `uv run --frozen pyright src/ tests/` — 0 errors
- `uv run --frozen ruff check` — clean

## Breaking Changes

Yes — documented in `docs/migration.md`:

- **`MCPServer.get_context()` removed** — use `ctx: Context` parameter injection in tool/resource/prompt functions instead (the existing recommended pattern)
- **`request_ctx` ContextVar removed** from `mcp.server.lowlevel.server` — code importing it directly will need to use parameter injection
- **`MCPServer.call_tool()`, `read_resource()`, `get_prompt()`** now accept an optional `context: Context | None = None` parameter — backward-compatible for callers that don't pass it
- **New behavior**: if a tool/resource/prompt declares a `ctx: Context` parameter but is called with `context=None`, a clear error is raised (`ToolError` for tools, `ValueError` for prompts/resource templates). Previously `None` was silently injected.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**`Context` class moved** from `mcp.server.mcpserver.server` to `mcp.server.mcpserver.context`. The public import path (`from mcp.server.mcpserver import Context`) is unchanged via `__init__.py` re-export. The old direct-module path (`from mcp.server.mcpserver.server import Context`) also still works because `server.py` imports `Context` at module level.

Exception hierarchy cleanup (making prompts/resources raise `MCPServerError` subclasses instead of `ValueError`) was considered but deferred to #1742 (error taxonomy) to avoid scope creep — see that issue for the full analysis of the existing inconsistencies.

### DIFF
--- a/src/mcp/server/mcpserver/context.py
+++ b/src/mcp/server/mcpserver/context.py
@@ -93,7 +93,7 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
         """
         progress_token = self.request_context.meta.get("progress_token") if self.request_context.meta else None
 
-        if progress_token is None:  # pragma: no branch
+        if progress_token is None:  # pragma: no cover
             return
 
         await self.request_context.session.send_progress_notification(
--- a/tests/server/mcpserver/test_server.py
+++ b/tests/server/mcpserver/test_server.py
@@ -1231,6 +1231,19 @@ class TestContextInjection:
 class TestServerPrompts:
     """Test prompt functionality in MCPServer server."""
 
+    async def test_get_prompt_direct_call_without_context(self):
+        """Test calling mcp.get_prompt() directly without passing context."""
+        mcp = MCPServer()
+
+        @mcp.prompt()
+        def fn() -> str:
+            return "Hello, world!"
+
+        result = await mcp.get_prompt("fn")
+        content = result.messages[0].content
+        assert isinstance(content, TextContent)
+        assert content.text == "Hello, world!"
+
     async def test_prompt_decorator(self):
         """Test that the prompt decorator registers prompts correctly."""
         mcp = MCPServer()


docs/migration.md
```diff
@@ -288,6 +288,37 @@ app = Starlette(routes=[Mount("/", app=mcp.streamable_http_app(json_response=Tru
 
 **Note:** DNS rebinding protection is automatically enabled when `host` is `127.0.0.1`, `localhost`, or `::1`. This now happens in `sse_app()` and `streamable_http_app()` instead of the constructor.
 
+### `MCPServer.get_context()` removed
+
+`MCPServer.get_context()` has been removed. Context is now constructed by the framework and passed explicitly — there is no ambient ContextVar to read from.
```
Kludex:
```suggestion
`MCPServer.get_context()` has been removed. Context is now injected by the framework and passed explicitly — there is no ambient ContextVar to read from.
```

docs/migration.md
```diff
@@ -288,6 +288,37 @@ app = Starlette(routes=[Mount("/", app=mcp.streamable_http_app(json_response=Tru
...
+**After (v2):**
+
+```python
+@mcp.tool()
+async def my_tool(x: int, ctx: Context) -> str:
+    await ctx.info("Processing...")
+    return str(x)
+```
+
+### `MCPServer.call_tool()`, `read_resource()`, `get_prompt()` now accept a `context` parameter
```
Kludex, line 316: Makes sense.

src/mcp/server/mcpserver/prompts/base.py
```diff
@@ -138,7 +138,14 @@ async def render(
...
     ) -> list[Message]:
-        """Render the prompt with arguments."""
+        """Render the prompt with arguments.
+
+        Raises:
+            ValueError: If the prompt requires a Context but none was provided,
+                if required arguments are missing, or if rendering fails.
+        """
+        if self.context_kwarg is not None and context is None:
+            raise ValueError(f"Prompt {self.name!r} requires a Context, but none was provided")
```
Kludex: How would the `Context` be `None` here? 🤔

src/mcp/server/mcpserver/resources/templates.py
```diff
@@ -101,7 +101,14 @@ async def create_resource(
...
     ) -> Resource:
-        """Create a resource from the template with the given parameters."""
+        """Create a resource from the template with the given parameters.
+
+        Raises:
+            ValueError: If the template requires a Context but none was provided,
+                or if creating the resource fails.
+        """
+        if self.context_kwarg is not None and context is None:
+            raise ValueError(f"Resource template {self.name!r} requires a Context, but none was provided")
```
Kludex: Same question.

src/mcp/server/mcpserver/prompts/base.py
Kludex: Also, probably it's `RuntimeError`, if anything.

src/mcp/server/mcpserver/tools/base.py
```diff
@@ -95,7 +95,14 @@ async def run(
...
     ) -> Any:
-        """Run the tool with arguments."""
+        """Run the tool with arguments.
+
+        Raises:
+            ToolError: If the tool requires a Context but none was provided,
+                or if the tool function raises during execution.
+        """
+        if self.context_kwarg is not None and context is None:
+            raise ToolError(f"Tool {self.name!r} requires a Context, but none was provided")
```
Kludex: Kind a the same question... 👀

src/mcp/server/mcpserver/context.py
```diff
@@ -0,0 +1,281 @@
...
+from mcp.server.context import LifespanContextT, RequestT, ServerRequestContext
+from mcp.server.elicitation import (
+    ElicitationResult,
+    ElicitSchemaModelT,
+    UrlElicitationResult,
+    elicit_with_validation,
+)
+from mcp.server.elicitation import (
+    elicit_url as _elicit_url,
+)
```
Kludex: What's this?
```suggestion
from mcp.server.elicitation import elicit_url
```

Kludex: Ah, got it.


```suggestion
from mcp.server.elicitation import elicit_url as _elicit_url
```

Still...

src/mcp/server/mcpserver/context.py
```diff
@@ -0,0 +1,281 @@
...
+    _request_context: ServerRequestContext[LifespanContextT, RequestT] | None
+    _mcp_server: MCPServer | None
+
+    def __init__(
+        self,
+        *,
+        request_context: ServerRequestContext[LifespanContextT, RequestT] | None = None,
+        mcp_server: MCPServer | None = None,
+        # TODO(Marcelo): We should drop this kwargs parameter.
+        **kwargs: Any,
```
Kludex, line 66: I think we should follow what Marcelo said.

src/mcp/server/mcpserver/context.py
```diff
@@ -0,0 +1,281 @@
...
+        """Report progress for the current operation.
+
+        Args:
+            progress: Current progress value (e.g., 24)
+            total: Optional total value (e.g., 100)
+            message: Optional message (e.g., "Starting render...")
+        """
+        progress_token = self.request_context.meta.get("progress_token") if self.request_context.meta else None
+
+        if progress_token is None:  # pragma: no cover
```
Kludex, line 96:
```suggestion
        if progress_token is None:  # pragma: no branch
```

src/mcp/server/mcpserver/context.py
```diff
@@ -0,0 +1,281 @@
...
+
+        await self.request_context.session.send_progress_notification(
+            progress_token=progress_token,
+            progress=progress,
+            total=total,
+            message=message,
+            related_request_id=self.request_id,
+        )
+
+    async def read_resource(self, uri: str | AnyUrl) -> Iterable[ReadResourceContents]:
```
Kludex, line 107: Do we need this `AnyUrl`?
```suggestion
    async def read_resource(self, uri: str) -> Iterable[ReadResourceContents]:
```

src/mcp/server/mcpserver/context.py
```diff
@@ -0,0 +1,281 @@
...
+        for unsolicited server-to-client notifications. The client SHOULD reconnect
+        with Last-Event-ID to resume receiving notifications.
+
+        Note:
+            This is a no-op if not using StreamableHTTP transport with event_store.
+            Currently, client reconnection for standalone GET streams is NOT
+            implemented - this is a known gap.
+        """
+        if self._request_context and self._request_context.close_standalone_sse_stream:  # pragma: no cover
+            await self._request_context.close_standalone_sse_stream()
```
Kludex, line 261: Did we find out the answers for what I asked before? When would the developer actually use those???

src/mcp/server/mcpserver/context.py
maxisbey: Ah I wasn't actually changing the Context class at all, just moving it for now

src/mcp/server/mcpserver/context.py
maxisbey: iirc I couldn't find actual uses for it publically, but it's a required part of the spec the sdk needs to support

src/mcp/server/mcpserver/context.py
maxisbey: keen to follow up with a different PR to actually refactor the Context class itself, but this one is just about moving it out and removing the context_var access.

src/mcp/server/mcpserver/prompts/base.py
maxisbey: Honestly the way Context is optional is rather weird through the whole call chain. If it's None it's sometimes just not injected which also breaks stuff.

After taking a bit of time to walk through it, I think it'd be better to leave it as optional on the MCPServer methods itself, but then make it required on the rest of the call stack, which is essentially the same functionality that existed before. If you called the previous `MCPServer.get_context` method it would just construct a new one if none existed.

So will do that to restore what was here before and remove some of the weird optional handling through the rest of the call chain.

src/mcp/server/mcpserver/context.py
Kludex: This doesn't seem a reasonable answer. So we have a public API that is required by a spec that actually no one uses? Is the spec written for the sake of having a spec? Can we please find the use case of it?

src/mcp/server/mcpserver/resources/templates.py
maxisbey: removed

src/mcp/server/mcpserver/prompts/base.py
maxisbey: changed

src/mcp/server/mcpserver/tools/base.py
maxisbey: removed

src/mcp/server/mcpserver/context.py
maxisbey: This was something here already from the copy/paste, honestly I don't think we even need it imported under a different name. Will change it too just import directly.

src/mcp/server/mcpserver/resources/templates.py
```diff
@@ -99,9 +99,13 @@ async def create_resource(
...
         uri: str,
         params: dict[str, Any],
-        context: Context[LifespanContextT, RequestT] | None = None,
+        context: Context[LifespanContextT, RequestT],
     ) -> Resource:
-        """Create a resource from the template with the given parameters."""
+        """Create a resource from the template with the given parameters.
+
+        Raises:
+            ValueError: If creating the resource fails.
```
Kludex, line 107: Is this still true?

maxisbey: Yes, few lines down

src/mcp/server/mcpserver/context.py
maxisbey: didn't work, this causes it to fail coverage

Kludex (approved):

src/mcp/server/mcpserver/context.py
maxisbey: I wasn't the author of the sep, nor the one who accepted it or merged it into the spec. If it's in the spec then we need to implement it in the SDK.

If we think the sep is a bad idea, which is of course very fair, then we gotta push back on the sep itself or push for getting it changed. I don't think it makes sense to not implement spec functionality in the SDK just because we think it's a bad idea, I think that's the wrong place to make that argument.

hughesjj:
Given this is a breaking change, does that mean the major version which ships this change will be `mcp 2.0+`?

PR #2209, opened 2026-03-04
feat: add subject and claims fields to AccessToken

### LINKED ISSUE #1038: MCP server: AccessToken class should have field for subject claim ("sub")

shivama205:
Closes #1038

## Summary

- Add `subject: str | None = None` to `AccessToken` — stores the JWT `sub` claim (user ID) so token verifiers can populate it and callers can read it via `get_access_token().subject`
- Add `claims: dict[str, Any] | None = None` to `AccessToken` — stores arbitrary custom JWT claims beyond the reserved set (as suggested by @robertofalk)
- Add `Context.subject` property — lets tool handlers read the authenticated user's subject directly via `ctx.subject` without needing to import `get_access_token` (as suggested by @lukebuehler)

Both new fields are optional with `None` defaults, so this is fully backward compatible.

## Test plan

- [x] `TestAccessTokenFields` in `test_bearer_auth.py` — backward compat, `subject` field, `claims` field, both together, propagation through `AuthenticatedUser`
- [x] `TestContextSubject` in `test_context.py` — unauthenticated returns `None`, token without subject returns `None`, token with subject returns value, reflects live contextvar state
- [x] Full auth + mcpserver test suite passes (400/400)
- [x] `ruff check` and `ruff format` clean
- [x] `pyright` clean (0 errors, 0 warnings)

### DIFF
--- a/src/mcp/server/auth/provider.py
+++ b/src/mcp/server/auth/provider.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Generic, Literal, Protocol, TypeVar
+from typing import Any, Generic, Literal, Protocol, TypeVar
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 from pydantic import AnyUrl, BaseModel
@@ -40,6 +40,8 @@ class AccessToken(BaseModel):
     scopes: list[str]
     expires_at: int | None = None
     resource: str | None = None  # RFC 8707 resource indicator
+    subject: str | None = None  # JWT sub claim (user ID)
+    claims: dict[str, Any] | None = None  # Additional JWT claims beyond reserved fields
 
 
 RegistrationErrorCode = Literal[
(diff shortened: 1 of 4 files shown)


shivama205:
Friendly bump — this adds the subject and claims fields requested in this issue. All tests passing, would love a review!

maxisbey:
Thanks for this — it shaped what landed. #2686 (now merged on `main`, with the v1.x backport in #2690) adds `subject` and `claims` to `AccessToken` essentially as you proposed here, plus threads `subject` through `AuthorizationCode`/`RefreshToken` and the simple-auth example so it actually gets populated end-to-end. You're credited on both PRs.

The one piece that didn't carry over is `Context.subject` — review preferred reading it via `get_access_token().subject` directly rather than adding another property; the existing `Context.client_id` docstring now points there.

Closing in favour of #2686.

PR #2212, opened 2026-03-04
Add warning log when rejecting request with unknown/expired session ID

### LINKED ISSUE #2204: No logging in `_handle_stateful_request` when session ID is unknown/expired
### Initial Checks

### Description

### Description

```python
else:
        jsonrpc="2.0",
        id=None,
    )
    response = Response(
    )
```

```python
else:
    logger.warning(
        session_id,
    )
    ...
```

### Related issues

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
- Python: 3.13
```

rameshreddy-adutla:
## Summary
Add a `logger.warning()` call in the `else` branch of `StreamableHTTPSessionManager._handle_stateful_request()` when rejecting a request with an unknown or expired session ID.

## Problem
The `else` branch silently returns HTTP 404 without any logging. The other two branches both log:
- `logger.debug("Session already exists...")` (existing session)
- `logger.info("Created new transport...")` (new session)

This made it very difficult to diagnose connection failures in production — for example, after a server restart when clients continued sending stale `mcp-session-id` headers. Operators had to wrap the ASGI app in custom middleware just to see why requests were being rejected.

## Fix
Add `logger.warning()` with the rejected session ID before returning the 404 response. Using `warning` level (not `debug`/`info`) since this typically indicates a problem that operators should investigate.

## Testing
Existing test `test_unknown_session_id_returns_404` passes and now shows the warning in log output:
```
WARNING mcp.server.streamable_http_manager: Rejected request with unknown or expired session ID: non-existent-session-id
```

Fixes #2204

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -272,7 +272,7 @@ class StreamableHTTPSessionManager:
             # Unknown or expired session ID - return 404 per MCP spec
             # TODO: Align error code once spec clarifies
             # See: https://github.com/modelcontextprotocol/python-sdk/issues/1821
-            logger.warning(f"Rejected request with unknown or expired session ID: {request_mcp_session_id}")
+            logger.info(f"Rejected request with unknown or expired session ID: {request_mcp_session_id[:64]}")
             error_response = JSONRPCError(
                 jsonrpc="2.0",
                 id=None,
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -1,6 +1,7 @@
 """Tests for StreamableHTTPSessionManager."""
 
 import json
+import logging
 from typing import Any
 from unittest.mock import AsyncMock, patch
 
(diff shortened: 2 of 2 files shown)


maxisbey:
Thanks for the fix — merging. I pushed two small adjustments:

- **`warning` → `info`**: This path is normal operation per spec (client receives 404 → reinitializes). WARNING would fire once per connected client on every server restart. Matches the other session-lifecycle logs at [`:220`](https://github.com/modelcontextprotocol/python-sdk/blob/92f1b1500d808a32d261d9e101d6f3bae3ad9e25/src/mcp/server/streamable_http_manager.py#L220) and [`:247`](https://github.com/modelcontextprotocol/python-sdk/blob/92f1b1500d808a32d261d9e101d6f3bae3ad9e25/src/mcp/server/streamable_http_manager.py#L247).
- **`[:64]` truncation**: Unlike those other logs (server-generated UUIDs), this session ID is client-controlled — capping the bytes we emit.

Also added a caplog assertion to the existing 404 test.

PR #2218, opened 2026-03-05
fix: log unknown or expired streamable HTTP session IDs

### LINKED ISSUE #2204: No logging in `_handle_stateful_request` when session ID is unknown/expired

lavish0000:
Fixes #2204

## Summary
- add a warning log when `StreamableHTTPSessionManager` rejects a request with an unknown or expired `mcp-session-id`
- extend the existing 404 regression test to assert the warning is emitted

## Why
The 404 behavior is already correct, but today the rejection path is silent. In practice that makes stale session IDs harder to diagnose after restarts or idle-session cleanup.

## Scope
- no change to the HTTP response payload or status code
- no change to session lifecycle behavior
- only the unknown/expired session ID branch in `_handle_stateful_request()` is touched

## Validation
- `uv run pytest tests/server/test_streamable_http_manager.py`
- `uv run ruff check src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py`
- `uv run ruff format --check src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py`
- `uv run pyright src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py`

## Notes
- the warning message includes the rejected session ID so repeated stale-session failures are visible in server logs


lavish0000:
Quick status note: the remaining red lane is `checks / test (3.11, lowest-direct, windows-latest)`, and the failures are in `tests/client/test_stdio.py::TestChildProcessCleanup::test_basic_child_process_cleanup` and `tests/shared/test_streamable_http.py::test_streamable_http_client_session_persistence`, which are outside this diff. The change here is limited to `streamable_http_manager.py`, and the targeted local validation from the PR description still passes.

maxisbey:
Thanks for the contribution — this was addressed in #2212 which was opened first. For `good first issue` tickets, it's worth checking the linked PRs on the issue before opening a new one.

Issue #1960, opened 2026-01-26
BrokenResourceError race condition in stdio_client cleanup when context exits quickly


newsbubbles:
# GitHub Issue: MCP SDK stdio_client Race Condition

## Repository
`modelcontextprotocol/python-sdk`

## Title
`BrokenResourceError` race condition in `stdio_client` cleanup when context exits quickly

### Description

When the `stdio_client` async context manager exits quickly (before the subprocess has finished outputting data), a race condition occurs between the `stdout_reader` task and the cleanup code in the `finally` block, resulting in `BrokenResourceError`.

### Reproduction

This occurs in scenarios where:
1. A subprocess is spawned via `stdio_client`
2. The subprocess starts outputting data (e.g., initialization messages)
3. The calling code exits the context quickly (e.g., user disconnects)
4. The `finally` block closes `read_stream_writer` while `stdout_reader` is mid-send

### Error

```
ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "mcp/client/stdio/__init__.py", line 162, in stdout_reader
    |     await read_stream_writer.send(session_message)
    |   File "anyio/streams/memory.py", line 242, in send
    |     self.send_nowait(item)
    |   File "anyio/streams/memory.py", line 213, in send_nowait
    |     raise BrokenResourceError
    | anyio.BrokenResourceError
    +------------------------------------
```

### Root Cause

In `mcp/client/stdio/__init__.py`, the `stdio_client` function:

```python
async with (
    anyio.create_task_group() as tg,
    process,
):
    tg.start_soon(stdout_reader)
    tg.start_soon(stdin_writer)
    try:
        yield read_stream, write_stream
    finally:
        # ... process cleanup ...
        await read_stream_writer.aclose()  # <-- Closes stream
        await write_stream_reader.aclose()
```

The `finally` block closes `read_stream_writer` before the TaskGroup has cancelled its tasks. If `stdout_reader` is in the middle of `await read_stream_writer.send(...)`, it receives `BrokenResourceError`.

### Suggested Fix

Cancel the TaskGroup's scope before closing the streams:

```python
finally:
    # Cancel background tasks before closing streams
    tg.cancel_scope.cancel()
    
    # ... existing cleanup ...
    await read_stream_writer.aclose()
    await write_stream_reader.aclose()
```

[...]

PR #2219, opened 2026-03-05
fix: avoid stdio cleanup BrokenResourceError race

lavish0000:
## Summary
- cancel the stdio client task group before closing its memory streams during shutdown
- move memory stream teardown until after the task group has exited
- add a regression test that exits while `stdout_reader` is blocked on a zero-buffer send

## Problem
`stdio_client()` starts background tasks that read process stdout and forward messages into a zero-buffer memory stream. If the caller exits the context without consuming `read_stream`, `stdout_reader` can block on `read_stream_writer.send(...)`.

Before this change, the cleanup path closed `read_stream_writer` while `stdout_reader` was still running inside the task group. That raises `anyio.BrokenResourceError` and surfaces as an `ExceptionGroup` during context exit.

## Fix
The cleanup sequence now cancels the task group before any of the memory streams it owns are closed. The stream handles are closed only after the task group has exited, so `stdout_reader` and `stdin_writer` are no longer racing against stream teardown.

## Validation
- reproduced the failure locally on current `main` with a minimal stdio server that writes one JSON-RPC message and then idles
- `uv run pytest tests/client/test_stdio.py`
- `uv run ruff check src/mcp/client/stdio.py tests/client/test_stdio.py`
- `uv run ruff format --check src/mcp/client/stdio.py tests/client/test_stdio.py`
- `uv run pyright src/mcp/client/stdio.py tests/client/test_stdio.py`


maxisbey:
Thanks for the PR! #1960 is still labeled `needs repro` — could you post a reproduction on that issue first so we can confirm the root cause before reviewing a fix? The PR description here mentions reordering cleanup but the diff just widens the except clause, so a repro on the issue would help us verify which approach is actually correct.

lavish0000:
I posted a runnable repro on [#1960](https://github.com/modelcontextprotocol/python-sdk/issues/1960).

You were also right that the PR body drifted from the current diff. I'll update the PR description so it matches the implementation before asking for another review.

PR #2220, opened 2026-03-05
fix: allow client scopes when no scope restriction is set

lavish0000:
## Summary
- treat `OAuthClientMetadata.scope=None` as unrestricted instead of an empty allowlist
- add direct unit coverage for unrestricted and restricted scope validation paths

## Problem
`OAuthClientMetadata.validate_scope()` currently turns `scope=None` into `[]`. That makes any requested scope look invalid, even though an unset client scope should mean the client was not registered with a scope restriction.

As a result, authorization requests can be rejected with `invalid_scope` when the client metadata omits `scope` entirely.

## Fix
If the client metadata does not declare a scope restriction, return the requested scopes as-is. Keep the existing validation path when the client was registered with explicit scopes.

## Validation
- `uv run pytest tests/shared/test_auth.py tests/server/mcpserver/auth/test_auth_integration.py -k "validate_scope or authorize_invalid_scope"`
- `uv run ruff check src/mcp/shared/auth.py tests/shared/test_auth.py`
- `uv run ruff format --check src/mcp/shared/auth.py tests/shared/test_auth.py`
- `uv run pyright src/mcp/shared/auth.py tests/shared/test_auth.py`


lavish0000:
Quick status note: the remaining red lane is `checks / test (3.10, locked, ubuntu-latest)`, failing in `tests/client/test_stdio.py::test_stdio_client_bad_path`, which is outside this diff. The scope-handling change here is isolated to the OAuth metadata validator path, and the targeted local validation from the PR description still passes.

maxisbey:

PR #2221, opened 2026-03-05
fix: refresh auth context for stateful HTTP requests

lavish0000:
## Summary
- bind `auth_context_var` from the current request metadata at low-level request dispatch time
- clear the auth context for requests that do not carry an authenticated user
- add an integration regression test that changes bearer tokens mid-session in stateful streamable HTTP

## Problem
In stateful streamable HTTP mode, request handlers can run inside task-group tasks whose context was copied from the session-creating request. That means `get_access_token()` keeps returning the bearer token from the first request in the session, even after later requests send a different `Authorization` header.

## Fix
Each handled request already carries its Starlette request object in `ServerMessageMetadata.request_context`. At dispatch time, the low-level server now reads the authenticated user from that request scope and binds `auth_context_var` for the duration of the handler call. This refreshes `get_access_token()` on every request while preserving the existing behavior for unauthenticated apps.

## Validation
- reproduced the bug locally on current `main` with an in-process ASGI app and a mutable bearer auth client (`token-A` stayed visible after switching to `token-B`)
- `uv run pytest tests/issues/test_2208_stateful_auth_context.py tests/server/auth/middleware/test_auth_context.py tests/shared/test_streamable_http.py -k "2208 or auth_context or request_context_propagation or request_context_isolation"`
- `uv run ruff check src/mcp/server/lowlevel/server.py tests/issues/test_2208_stateful_auth_context.py`
- `uv run ruff format --check src/mcp/server/lowlevel/server.py tests/issues/test_2208_stateful_auth_context.py`
- `uv run pyright src/mcp/server/lowlevel/server.py tests/issues/test_2208_stateful_auth_context.py`

### DIFF
--- a/tests/server/test_lowlevel_streamable_http_auth_app.py
+++ b/tests/server/test_lowlevel_streamable_http_auth_app.py
@@ -1,5 +1,7 @@
+from typing import cast
 from unittest.mock import Mock
 
+import pytest
 from pydantic import AnyHttpUrl
 from starlette.applications import Starlette
 from starlette.middleware.authentication import AuthenticationMiddleware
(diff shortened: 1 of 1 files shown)


maxisbey:

PR #2222, opened 2026-03-05
fix: ignore closed log streams during shutdown

lavish0000:
## Summary
- guard the low-level exception logging path when the session write stream is already closed
- add regression coverage for both `ClosedResourceError` and `BrokenResourceError` from `send_log_message()`

## Problem
When the server receives an exception from the stream during shutdown, `_handle_message()` tries to emit an internal error log notification through `session.send_log_message()`. If the session write stream has already been torn down, that best-effort log write can raise `ClosedResourceError` or `BrokenResourceError` and crash the enclosing task group.

## Fix
Treat exception logging in this shutdown path as best-effort. If the write stream is already gone, skip the log notification and continue handling the original shutdown flow.

## Validation
- `uv run pytest tests/server/test_lowlevel_exception_handling.py`
- `uv run ruff check src/mcp/server/lowlevel/server.py tests/server/test_lowlevel_exception_handling.py`
- `uv run ruff format --check src/mcp/server/lowlevel/server.py tests/server/test_lowlevel_exception_handling.py`
- `uv run pyright src/mcp/server/lowlevel/server.py tests/server/test_lowlevel_exception_handling.py`

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -414,11 +414,14 @@ class Server(Generic[LifespanResultT]):
                         )
                 case Exception():
                     logger.error(f"Received exception from stream: {message}")
-                    await session.send_log_message(
-                        level="error",
-                        data="Internal Server Error",
-                        logger="mcp.server.exception_handler",
-                    )
+                    try:
+                        await session.send_log_message(
+                            level="error",
+                            data="Internal Server Error",
+                            logger="mcp.server.exception_handler",
+                        )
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR! Per our [CONTRIBUTING.md](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md), PRs require a corresponding issue with the `ready for work` label. Closing for now — feel free to open an issue first and re-submit once it's triaged.

PR #2229, opened 2026-03-06
Add warning log for unknown/expired session IDs

### LINKED ISSUE #2204: No logging in `_handle_stateful_request` when session ID is unknown/expired

nielskaspers:
## Summary

- Adds `logger.warning()` in the `else` branch of `_handle_stateful_request()` when a request is rejected due to an unknown or expired session ID
- Adds test assertion verifying the warning is logged with the session ID

The other two branches already log at debug/info level, but this branch silently returned 404. This makes it difficult to diagnose connection failures caused by stale session IDs (e.g. after server restarts).

## Test plan

- [x] Existing test `test_unknown_session_id_returns_404` updated and passing
- [x] Verified warning includes the session ID string for traceability

Closes #2204

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -272,6 +272,10 @@ class StreamableHTTPSessionManager:
             # Unknown or expired session ID - return 404 per MCP spec
             # TODO: Align error code once spec clarifies
             # See: https://github.com/modelcontextprotocol/python-sdk/issues/1821
+            logger.warning(
+                "Rejected request with unknown or expired session ID: %s",
+                request_mcp_session_id,
+            )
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the contribution — this was addressed in #2212 which was opened first. For `good first issue` tickets, it's worth checking the linked PRs on the issue before opening a new one.

PR #2234, opened 2026-03-06
fix: Ensure stdio server exits when stdin reaches EOF (fixes #2231)

### LINKED ISSUE #2231: Stdio transport: MCP server process survives parent death (stdin EOF not causing shutdown)
### Initial Checks

### Description

When an MCP server using `transport="stdio"` has its parent process die (e.g. the MCP client exits, is killed, or crashes), the server process is orphaned and continues running indefinitely. It gets reparented to init/systemd and never exits.

### Expected behavior

When the parent process dies and stdin reaches EOF, the server should detect this and shut down cleanly.

### Analysis

Looking at the code flow in `mcp/server/stdio.py`:

1. `stdin_reader()` (line 60) iterates `async for line in stdin` where `stdin` is `anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8"))`
2. When stdin reaches EOF, the loop should end, closing `read_stream_writer`
3. This should close the `read_stream`, causing `server.run()`'s `async for message in session.incoming_messages` loop to end
4. `run()` returns → `stdio_server()` context exits → task group cancels `stdout_writer` → process exits

In practice, the process survives. The likely cause is that `anyio.wrap_file` runs the underlying `readline()` in a worker thread, and this thread may not wake up or propagate EOF correctly when the pipe is closed due to parent death.

### Reproduction

```python

mcp = FastMCP("test")

@mcp.tool()
async def hello() -> str:
    return "hello"

mcp.run(transport="stdio")
```

1. Start the server as a subprocess from any parent process
2. Kill the parent process (e.g. `kill -9 <parent_pid>`)
3. Observe that the MCP server process is still running (`ps aux | grep python`)
4. The process is now reparented to init/systemd and will never exit

### Environment

- `mcp` version: 1.26.0
- Python: 3.14
- OS: Linux (Fedora 43)

### Fix

A proposed fix would be to monitor `stdin` via a selectors and when it becomes ready and returns "" shutdown the process.

### Example Code

```Python

mcp = FastMCP("test")

@mcp.tool()
async def hello() -> str:
    return "hello"

mcp.run(transport="stdio")
```

### Python & MCP Python SDK

```Text
Environment

    mcp version: 1.26.0
    Python: 3.14
    OS: Linux (Fedora 43)
```

goingforstudying-ctrl:
## Problem
When an MCP server using `transport="stdio"` has its parent process die (e.g. client crashes, is killed, or exits), the server process becomes an orphan and continues running instead of shutting down gracefully.

## Root Cause
The stdin reader loop ends when EOF is reached, but there was no explicit signal to the rest of the server that shutdown should occur.

## Solution
- Explicitly close `read_stream_writer` on EOF to signal shutdown
- Add test coverage for EOF handling
- Add test for parent death simulation scenario

## Changes
- `src/mcp/server/stdio.py`: Close read stream on EOF
- `tests/test_stdio_eof.py`: New test file with EOF tests

## Testing
- Added `test_stdio_server_exits_on_eof()` - verifies server exits on stdin EOF
- Added `test_stdio_server_parent_death_simulation()` - simulates parent process death

Fixes #2231


goingforstudying-ctrl:
Hi @jspahrsummers @dsp-ant 👋

This PR has been ready for review for a few days. All CI checks are passing.

The changes add a guide for testing with inspector which would help developers debug their MCP implementations.

Would appreciate a review when you have time!

Thanks!

goingforstudying-ctrl:
Friendly ping @jspahrsummers @dsp-ant 👋 

This PR has been ready for review for 12 days. All CI checks are passing. The fix ensures stdio servers exit cleanly when stdin reaches EOF, which resolves issue #2231.

Would appreciate a review when you have a chance!

maxisbey:
@goingforstudying-ctrl the original issue hasn't been marked as good first issue or ready for work, and I wasn't able to reproduce it so I'm closing this PR.

Please also do not ping maintainers for reviews.

PR #2244, opened 2026-03-07
fix: preserve stdin/stdout file descriptors after server exits

haroldfabla2-hue:
## Description

Using os.dup() to create copies of file descriptors prevents closing the original sys.stdin/sys.stdout when the TextIOWrapper is closed.

## Fixes

Fixes https://github.com/modelcontextprotocol/python-sdk/issues/1933

## Changes

- Use os.dup() to create duplicate file descriptors for stdin/stdout
- Use os.fdopen() with closefd=True to manage the duplicated fd
- This prevents ValueError: I/O operation on closed file after the server exits

## Testing

The fix can be tested by running:
```python

mcp = FastMCP("Demo")
mcp.run(transport="stdio")
print("?")
```

After exiting the server with Ctrl+D, the print should work without errors.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,6 +17,7 @@ Example:
     ```
 """
 
+import os
 import sys
 from contextlib import asynccontextmanager
 from io import TextIOWrapper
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #2040 as that predates this PR.

Issue #2216, opened 2026-03-05
Bug: validate_scope rejects client scopes when required scopes in None


nik1097:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

The validate_scope() function in the Python SDK incorrectly handles cases where there are no required scopes from the client. Instead of treating None as no restrictions, it interprets it as an empty list of allowed scopes. This causes scopes in the token to be rejected with InvalidScopeError, even if the client should be allowed to request them.  

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
1.26.0
```

shivama205:
I'll take this. The bug is in `OAuthClientMetadata.validate_scope()` — when `self.scope` is `None` (no scopes registered), it converts to an empty list, rejecting all requested scopes. The fix is to treat `None` as "no restrictions" and allow any scope through.

@nik1097 thanks for reporting, will have a PR up shortly.

goingforstudying-ctrl:
Hi! I've submitted a fix for this issue in PR #2246.

The fix handles the case where client registration has scope=None by returning all requested scopes immediately, instead of treating None as an empty allowed list.

Let me know if any changes are needed!

shivama205:
Hey! Just realized I forgot to link my PR here — #2224 has been up since March 6 with tests and passing CI. Apologies for not linking it sooner!

Christian-Sidak:
Opened a fix in #2461 -- returns requested scopes immediately when self.scope is None (no restrictions registered), with 7 test cases covering the edge cases.

saneGuy:
From the current issue state, I understand #2301 from @maxisbey is the current path: remove the registration-scope check entirely, aligned with the step-up flow. I also see it is currently conflicting with main after the recent auth changes.

I hit this in practice: an OAuth client registered without a scope field cannot authorize with requested scopes, and the step-up flow fails after the scope challenge.

Would it help if I picked up one of two paths? Entirely your call:
1. Rebase and carry #2301 forward against current main, preserving that approach and attribution.
2. Prepare a narrow v1.x-only bugfix that early-returns when the registered scope is None, if you consider this eligible for v1.x maintenance.

Happy to be assigned either path, or to leave it alone if it's already planned.

AI assistance used for drafting; reviewed and owned by me.

PR #2224, opened 2026-03-06
fix: validate_scope treats None registered scopes as no restrictions

shivama205:
## Summary
- `OAuthClientMetadata.validate_scope()` incorrectly treated `self.scope = None` (no scopes registered) as an empty allowed list, rejecting all requested scopes with `InvalidScopeError`
- Now `None` is correctly treated as "no restrictions", allowing any requested scope through

Closes #2216

## Changes
- **Modified**: `src/mcp/shared/auth.py` — early return when `self.scope is None` instead of converting to empty list
- **Modified**: `tests/shared/test_auth.py` — added 6 unit tests for `validate_scope` covering None/matching/rejection/empty cases

## Test plan
- [x] All 9 shared auth tests pass (3 existing + 6 new)
- [x] All 42 auth integration tests pass (including `test_authorize_invalid_scope`)
- [x] `ruff check` passes
- [x] `ruff format` passes
- [x] `pyright` passes with 0 errors

PR #2248, opened 2026-03-08
fix: treat None client scope as no restrictions in validate_scope()

MaxwellCalkin:
**Note: This PR was authored by Claude (AI), operated by @maxwellcalkin.**

## Summary

Fixes #2216

`OAuthClientMetadata.validate_scope()` incorrectly handles the case where a client is registered without scope restrictions (`self.scope is None`). The existing code converts `None` to an empty list of allowed scopes:

```python
```

This means **every** requested scope fails the `if scope not in allowed_scopes` check, raising `InvalidScopeError` even though the client should have no restrictions.

### Fix

When `self.scope is None`, return the requested scopes immediately — `None` means "no restrictions", not "no scopes allowed":

```python
if self.scope is None:
    # No scope restrictions registered for this client; allow any scopes
    return requested_scopes
allowed_scopes = self.scope.split(" ")
```

### Tests

Added `TestValidateScope` class to `tests/shared/test_auth.py` with regression tests covering:
- `None` client scope allows any requested scopes (the bug from #2216)
- `None` requested scope returns `None`
- Valid subset of registered scopes is accepted
- Unregistered scope raises `InvalidScopeError`

### DIFF
--- a/tests/shared/test_auth.py
+++ b/tests/shared/test_auth.py
@@ -1,6 +1,9 @@
 """Tests for OAuth 2.0 shared code."""
 
-from mcp.shared.auth import OAuthMetadata
+import pytest
+from pydantic import AnyUrl
+
+from mcp.shared.auth import InvalidScopeError, OAuthClientMetadata, OAuthMetadata
 
 
 def test_oauth():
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #2224 as that predates this PR.

PR #2249, opened 2026-03-08
fix: propagate HTTP errors from transports instead of silently logging

### LINKED ISSUE #2110: HTTP transport swallows non-2xx status codes causing client to hang

MaxwellCalkin:
**Note: This PR was authored by Claude (AI), operated by @maxwellcalkin.**

## Summary

Fixes #2110

When an MCP server returns non-2xx HTTP status codes (401/403/404/5xx), the Streamable HTTP and SSE transports silently log the error but don't propagate it to the caller. The caller hangs indefinitely waiting for a response on the read stream.

### Changes

**New `HttpError` exception class** (`src/mcp/shared/exceptions.py`):
- Preserves the original HTTP status code
- `is_auth_error` property to distinguish 401/403 from other failures
- Optional `body` field for response body context
- Exported from the top-level `mcp` package

**Streamable HTTP transport** (`src/mcp/client/streamable_http.py`):
- `_handle_post_request`: 401/403 responses now raise `HttpError` (in addition to sending a JSONRPCError for request messages), so the caller gets an exception instead of hanging
- `_handle_post_request`: Generic 4xx/5xx responses also raise `HttpError` instead of silently returning
- `_handle_post_request`: HTTP 404 error message now includes "(HTTP 404)" and `data={"http_status": 404}` to preserve HTTP-level context
- `_handle_post_request`: All error responses include `data={"http_status": <code>}` in the JSONRPCError
- `post_writer`: `handle_request_async` now wraps calls in try/except and forwards exceptions through `read_stream_writer.send(exc)` so callers don't hang
- `post_writer`: Outer exception handler also forwards errors through the read stream

**SSE transport** (`src/mcp/client/sse.py`):
- `post_writer`: 401/403 responses detected before `raise_for_status()` and forwarded as `HttpError` through the read stream
- `post_writer`: `httpx.HTTPStatusError` from `raise_for_status()` converted to `HttpError` and forwarded through the read stream
- `post_writer`: Generic exceptions forwarded through the read stream instead of being silently logged

### Before

```python
# Caller hangs forever waiting for a response that will never come
async with streamable_http_client("https://example.com/mcp") as (read, write):
    await write.send(SessionMessage(request))
    response = await read.receive()  # blocks indefinitely on 401/403/5xx
```

### After

```python
# Caller receives an HttpError exception through the read stream
async with streamable_http_client("https://example.com/mcp") as (read, write):
    await write.send(SessionMessage(request))
    response = await read.receive()  # receives HttpError or JSONRPCError with http_status

# Or catch it directly
from mcp.shared.exceptions import HttpError
try:
    ...
except HttpError as e:
    if e.is_auth_error:
        print(f"Auth failed: {e.status_code}")
```

## Test plan

- [ ] Verify existing tests pass (no behavioral change for 2xx responses)
- [ ] Verify 401/403 responses raise `HttpError` with `is_auth_error == True`
- [ ] Verify 404 responses include HTTP status in error data
- [ ] Verify 5xx responses raise `HttpError` and don't cause caller to hang
- [ ] Verify `HttpError` is importable from `mcp` top-level package

### DIFF
--- a/src/mcp/__init__.py
+++ b/src/mcp/__init__.py
@@ -4,7 +4,7 @@ from .client.session_group import ClientSessionGroup
 from .client.stdio import StdioServerParameters, stdio_client
 from .server.session import ServerSession
 from .server.stdio import stdio_server
-from .shared.exceptions import MCPError, UrlElicitationRequiredError
+from .shared.exceptions import HttpError, MCPError, UrlElicitationRequiredError
 from .types import (
     CallToolRequest,
     ClientCapabilities,
@@ -81,6 +81,7 @@ __all__ = [
     "ErrorData",
     "GetPromptRequest",
     "GetPromptResult",
+    "HttpError",
     "Implementation",
     "IncludeContext",
     "InitializeRequest",


src/mcp/client/streamable_http.py
```diff
@@ -269,17 +270,41 @@ async def _handle_post_request(self, ctx: RequestContext) -> None:
...
                     session_message = SessionMessage(JSONRPCError(jsonrpc="2.0", id=message.id, error=error_data))
                     await ctx.read_stream_writer.send(session_message)
+                else:
+                    raise HttpError(404, "Session terminated (HTTP 404)")
                 return
 
+            if response.status_code in (401, 403):
+                status_label = "Unauthorized" if response.status_code == 401 else "Forbidden"
+                error_message = f"HTTP {response.status_code} {status_label}"
+                if isinstance(message, JSONRPCRequest):
```
Bortlesboat, line 287: This raises after already sending a JSONRPCError to the read stream (line 26). But `handle_request_async` wraps `_handle_post_request` in a try/except that catches this HttpError and sends it to `read_stream_writer` *again*. So for JSONRPCRequest messages hitting 401/403 (and same for the generic >=400 path below), the consumer receives both a SessionMessage(JSONRPCError) and then an HttpError exception from the same stream — two messages for one request.

I think you either want to not raise here when the JSONRPCError was already sent, or skip the JSONRPCError and just raise (letting the wrapper forward it). The current approach of doing both will confuse callers.

src/mcp/client/sse.py
```diff
@@ -142,10 +143,28 @@ async def post_writer(endpoint_url: str):
                                             exclude_unset=True,
                                         ),
                                     )
+                                    if response.status_code in (401, 403):
+                                        status_label = "Unauthorized" if response.status_code == 401 else "Forbidden"
+                                        exc = HttpError(
+                                            response.status_code,
+                                            f"HTTP {response.status_code} {status_label}",
```
Bortlesboat, line 150: Re-raising HttpError here means it escapes `post_writer` entirely — nothing above catches it, so it propagates into the task group and crashes the whole SSE client connection. The original code intentionally swallowed all exceptions to keep the writer alive for the connection lifetime.

For the SSE transport I think you want the same pattern as the other exceptions: send it through the read stream but don't re-raise. The caller can handle it from the stream without killing the connection.

slykar: This is exactly the issue I'm having. I've fixed it in this PR: https://github.com/modelcontextprotocol/python-sdk/pull/2273

maxisbey:
Thanks for your contribution! Closing this in favour of #2124 as that predates this PR.

PR #2251, opened 2026-03-08
fix: reject JSON-RPC requests with null id instead of misclassifying as notifications

### LINKED ISSUE #2057: Requests with "id": null silently misclassified as notifications

shivama205:
Closes #2057

## Problem

When a JSON-RPC request arrives with `"id": null`, the SDK silently reclassifies it as a `JSONRPCNotification` and returns `202 Accepted` with no response body. The caller gets no error and no response — a silent failure that is hard to debug.

This happens because of Pydantic's union fallthrough: `JSONRPCRequest` correctly rejects the invalid `id`, but `JSONRPCNotification` absorbs the extra `"id": null` field since it doesn't define an `id` field and Pydantic's default behavior drops unknown fields.

## Fix

Add a `model_validator` on `JSONRPCNotification` that rejects any input containing an `id` field. Per JSON-RPC 2.0, notifications MUST NOT have an `id` member, so this is both correct and targeted — it prevents the union fallthrough without affecting any valid messages.

This also catches other invalid `id` types (float, bool, list, dict) that would similarly be misclassified today.

## Changes

- **Modified**: `src/mcp/types/jsonrpc.py` — added `reject_id_field` model validator to `JSONRPCNotification`
- **Modified**: `tests/test_types.py` — added regression tests for null/invalid id rejection, plus tests confirming valid requests and notifications still work

## Test plan

- [x] All 1147 tests pass (0 failures)
- [x] `ruff check` passes
- [x] `ruff format` passes

### DIFF
--- a/README.v2.md
+++ b/README.v2.md
@@ -346,13 +346,12 @@ Tools can optionally receive a Context object by including a parameter with the 
 <!-- snippet-source examples/snippets/servers/tool_progress.py -->
 ```python
 from mcp.server.mcpserver import Context, MCPServer
-from mcp.server.session import ServerSession
 
 mcp = MCPServer(name="Progress Example")
 
 
 @mcp.tool()
-async def long_running_task(task_name: str, ctx: Context[ServerSession, None], steps: int = 5) -> str:
+async def long_running_task(task_name: str, ctx: Context, steps: int = 5) -> str:
     """Execute a task with progress updates."""
     await ctx.info(f"Starting: {task_name}")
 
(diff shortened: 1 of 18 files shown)


shivama205:
@Bortlesboat thanks for the comment. Could you please help to review and approve the fix.

Bortlesboat:
Happy to vouch for it, but I don't have write access to this repo so my review doesn't count as a formal approval. Hopefully a maintainer picks this up soon — it's a clean fix for a real bug.

maxisbey:
Thanks for your contribution! Closing this in favour of #2075 as that predates this PR.

PR #2266, opened 2026-03-10
fix: close all memory stream ends in client transport cleanup

maxisbey:
## Summary

Client transports for SSE, WebSocket, and StreamableHTTP create 4 anyio memory stream ends (2 paired streams) but only closed 2 in their `finally` blocks. anyio memory stream ends are **independent** — closing the writer does not close the reader. Unclosed stream ends leak and emit `ResourceWarning` when garbage collected.

This caused flaky CI failures: a transport connection error (404, 403, `ConnectError`) in one test would leak streams, then GC in a later unrelated test would trigger `ResourceWarning`, which pytest's `filterwarnings = ["error"]` promotes to a test failure — **in whatever test happened to be running when GC fired**, not the test that actually leaked.

## Fix

Follows the existing correct pattern in `stdio.py` (which closes all 4 ends on both early-fail and normal-exit paths):

| File | Before | After |
|---|---|---|
| `sse.py` | `finally` closed 2 of 4 | `finally` closes all 4 |
| `streamable_http.py` | `finally` closed 2 of 4 — `read_stream` was **never** closed, even on happy path | `finally` closes all 4 |
| `websocket.py` | **No `try`/`finally` at all** — if `ws_connect()` raised, all 4 leaked | Wrapped entire body in `try`/`finally` that closes all 4 |

anyio's `aclose()` is idempotent, so double-closing (e.g. when reader/writer tasks already closed their end) is safe.

## Tests

Added `tests/client/test_transport_stream_cleanup.py` with one regression test per transport. Each test triggers the error/exit path, then calls `gc.collect()` to force any leaked stream to emit `ResourceWarning` deterministically. All 3 tests **fail on `main`** with `ResourceWarning: Unclosed <MemoryObjectReceiveStream>` and **pass with this fix**.

## CI Evidence of the Flakiness

- [`test_tool_progress` knock-on failure](https://github.com/modelcontextprotocol/python-sdk/actions/runs/22861688125/job/66317185239) — streams leaked by `test_basic_resources` (which got a 404), GC'd during `test_tool_progress`
- [`test_call_tool` knock-on failure](https://github.com/modelcontextprotocol/python-sdk/actions/runs/21954536219/job/63414508104) — in-memory test that doesn't touch the network, failed because an earlier test on the same worker leaked resources

### DIFF
--- a/tests/client/test_transport_stream_cleanup.py
+++ b/tests/client/test_transport_stream_cleanup.py
@@ -15,7 +15,6 @@ import gc
 import sys
 from collections.abc import Iterator
 from contextlib import contextmanager
-from typing import Any
 
 import httpx
 import pytest
@@ -34,22 +33,25 @@ def _assert_no_memory_stream_leak() -> Iterator[None]:
     same xdist worker). gc.collect() is forced after the block to make leaks
     deterministic.
     """
-    leaked: list[Any] = []
+    leaked: list[str] = []
     old_hook = sys.unraisablehook
 
-    def hook(args: Any) -> None:  # pragma: no cover
+    def hook(args: "sys.UnraisableHookArgs") -> None:  # pragma: no cover
         # Only executes if a leak occurs (i.e. the bug is present).
+        # args.object is the __del__ function (not the stream instance) when
+        # unraisablehook fires from a finalizer, so check exc_value — the
+        # actual ResourceWarning("Unclosed <MemoryObjectSendStream at ...>").
         # Non-MemoryObject unraisables (e.g. PipeHandle leaked by an earlier
         # flaky test on the same xdist worker) are deliberately ignored —
         # this test should not fail for another test's resource leak.
-        if "MemoryObject" in repr(args.object):
-            leaked.append(args)
+        if "MemoryObject" in str(args.exc_value):
+            leaked.append(str(args.exc_value))
 
     sys.unraisablehook = hook
     try:
         yield
         gc.collect()
-        assert not leaked, f"Memory streams leaked: {[repr(x.object) for x in leaked]}"
+        assert not leaked, f"Memory streams leaked: {leaked}"
     finally:
         sys.unraisablehook = old_hook
 


src/mcp/client/websocket.py
```diff
@@ -41,40 +41,46 @@ async def websocket_client(
...
 
-            # Once the caller's 'async with' block exits, we shut down
-            tg.cancel_scope.cancel()
+                # Once the caller's 'async with' block exits, we shut down
+                tg.cancel_scope.cancel()
+    finally:
+        await read_stream_writer.aclose()
+        await write_stream.aclose()
+        await read_stream.aclose()
+        await write_stream_reader.aclose()
```
Kludex: @agronholm Sorry to bother, but can you share a bit of knowledge here? Is this really necessary everywhere? Are we doing something wrong?

tests/client/test_transport_stream_cleanup.py
```diff
@@ -0,0 +1,112 @@
...
+from mcp.client.sse import sse_client
+from mcp.client.streamable_http import streamable_http_client
+from mcp.client.websocket import websocket_client
+
+
+def _unused_tcp_port() -> int:
+    """Return a port with no listener. Binding then closing leaves the port unbound."""
+    with socket.socket() as s:
+        s.bind(("127.0.0.1", 0))
+        return s.getsockname()[1]
```
agronholm: You could just use the `free_tcp_port` fixture.

src/mcp/client/websocket.py
agronholm: I take it this is no longer relevant?

Kludex: It seems so. Sorry!

tests/client/test_transport_stream_cleanup.py
```diff
@@ -0,0 +1,100 @@
...
+    try:
+        yield
+        gc.collect()
+        assert not leaked, f"Memory streams leaked: {[repr(x.object) for x in leaked]}"
+    finally:
+        sys.unraisablehook = old_hook
+
+
+@pytest.mark.anyio
+async def test_sse_client_closes_all_streams_on_connection_error(free_tcp_port: int) -> None:
```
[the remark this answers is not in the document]
agronholm: It's quite impressive that the bot picked up on that.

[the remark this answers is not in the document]
maxisbey: haha yea pretty good. I was trying to think whether it's worth upgrading for this or not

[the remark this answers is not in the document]
maxisbey: Upgraded to a newer version and switched to free_tcp_port :)

tests/client/test_transport_stream_cleanup.py
maxisbey: using this now

tests/client/test_transport_stream_cleanup.py
```diff
@@ -0,0 +1,100 @@
...
+    """sse_client must close all 4 stream ends when the connection fails.
+
```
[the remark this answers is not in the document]
maxisbey: fixed

PR #2268, opened 2026-03-11
Fix stdio_client BrokenResourceError race condition during shutdown

### LINKED ISSUE #1960: BrokenResourceError race condition in stdio_client cleanup when context exits quickly

weiguangli-io:
## Summary

Fixes #1960. Supersedes #2219 (which was closed).

During `stdio_client` shutdown, the `finally` block closes `read_stream` (the receiving end of a zero-buffer memory channel) while `stdout_reader` may still be blocked on `send()`. When the receiving end closes underneath a pending `send()`, anyio raises `BrokenResourceError` — not `ClosedResourceError`. The existing `except` clause only caught `ClosedResourceError`, so `BrokenResourceError` escaped into an `ExceptionGroup`.

The fix catches `BrokenResourceError` alongside `ClosedResourceError` in both `stdout_reader` and `stdin_writer`. This is the minimal, correct change: `ClosedResourceError` means "you called send on a handle you already closed", while `BrokenResourceError` means "the other end was closed" — both are expected during shutdown and should be handled identically.

## Test plan

- Added regression test `test_stdio_client_no_broken_resource_error_on_shutdown` that reproduces the exact scenario: server sends a JSON-RPC message, client exits without consuming the read stream
- All 12 existing `tests/client/test_stdio.py` tests pass

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -158,7 +158,7 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
 
                         session_message = SessionMessage(message)
                         await read_stream_writer.send(session_message)
-        except anyio.ClosedResourceError:  # pragma: lax no cover
+        except (anyio.ClosedResourceError, anyio.BrokenResourceError):  # pragma: lax no cover
             await anyio.lowlevel.checkpoint()
 
     async def stdin_writer():
@@ -174,7 +174,7 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                             errors=server.encoding_error_handler,
                         )
                     )
-        except anyio.ClosedResourceError:  # pragma: no cover
+        except (anyio.ClosedResourceError, anyio.BrokenResourceError):  # pragma: no cover
             await anyio.lowlevel.checkpoint()
 
     async with anyio.create_task_group() as tg, process:
--- a/tests/client/test_stdio.py
+++ b/tests/client/test_stdio.py
@@ -509,6 +509,44 @@ class TestChildProcessCleanup:
 
 
 @pytest.mark.anyio
+@pytest.mark.filterwarnings("ignore::ResourceWarning")
+async def test_stdio_client_no_broken_resource_error_on_shutdown():
+    """Test that exiting stdio_client without consuming the read stream does not
+    raise BrokenResourceError.
+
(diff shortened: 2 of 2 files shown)


src/mcp/client/stdio.py
```diff
@@ -158,7 +158,7 @@ async def stdout_reader():
 
                         session_message = SessionMessage(message)
                         await read_stream_writer.send(session_message)
-        except anyio.ClosedResourceError:  # pragma: lax no cover
+        except (anyio.ClosedResourceError, anyio.BrokenResourceError):  # pragma: lax no cover
```
Bortlesboat, line 161: Noting for future readers: the SSE client (`sse.py`) avoids this class of race by calling `tg.cancel_scope.cancel()` in its `finally` block, which cancels tasks before closing their streams. The stdio client takes a different approach — letting stream closure propagate to tasks and catching the resulting exceptions. Both are valid; this PR correctly handles the full set of exceptions that anyio can raise in this scenario.

Bortlesboat (commented):
ACK. Clean minimal fix — `BrokenResourceError` is the correct counterpart to catch when the receiving end closes under a pending `send()`.

4444J99:
This would also resolve the semaphore leak reported in openai/openai-agents-python#618 — the SDK currently works around it by pre-closing the write stream before exit stack teardown (openai/openai-agents-python#2802), but that reaches into `BaseSession._write_stream` which is a private attribute. A proper fix here would let the SDK drop that workaround.

Happy to help test if useful.

weiguangli-io:
Friendly ping — CI is green. Would appreciate a review when you get a chance. Thanks!

maxisbey:
Closing as a duplicate of #2450, which is already in review for the same fix. Your root-cause analysis on #1960 was a big help in getting that PR right — thanks. If #2450 stalls we can reopen this.

PR #2280, opened 2026-03-11
Fix: preserve existing refresh_token when server omits it in refresh response

### LINKED ISSUE #2270: _handle_refresh_response discards existing refresh_token when server omits it

Jah-yee:
Fixes #2270

Per RFC 6749 Section 6, the server MAY issue a new refresh token. If the response omits it, preserve the existing one.

Many OAuth providers (Google, Auth0, Okta) omit refresh_token from refresh responses by default. The current behavior causes `can_refresh_token()` to return False after the first refresh, forcing full re-authentication.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -458,6 +458,17 @@ class OAuthClientProvider(httpx.Auth):
             content = await response.aread()
             token_response = OAuthToken.model_validate_json(content)
 
+            # Per RFC 6749 Section 6, the server MAY issue a new refresh token.
+            # If the response omits it, preserve the existing one.
+            if (
+                not token_response.refresh_token
+                and self.context.current_tokens
+                and self.context.current_tokens.refresh_token
+            ):
+                token_response = token_response.model_copy(
+                    update={"refresh_token": self.context.current_tokens.refresh_token}
+                )
+
             self.context.current_tokens = token_response
             self.context.update_token_expiry(token_response)
             await self.context.storage.set_tokens(token_response)


maxisbey:
fyi we have chosen to ban your account from mcp repos due to spamming PRs after I closed them with a warning

PR #2273, opened 2026-03-11
Handle connection errors in StreamableHTTPTransport

slykar:
## Motivation and Context

`StreamableHTTPTransport` is not handling connection/request errors properly, which leads to cancellation of the whole `anyio.TaskGroup`.

The fix should make connection/setup failures behave like normal request errors instead of letting them tear down the transport task group in a way that triggers cancellation churn.

Without the fix, a request collapses into an infinite AnyIO cancellation retry loop which may cause a sustained high CPU use.

## How Has This Been Tested?

I was primarily testing this with Agno and FastMCP. There are some changes I had to make to Agno itself as well.

Here's how I've tested it:

1. Start FastMCP server
2. Start HTTP app running Agno
3. Chat with the agent; ask it to use tools; everything good so far
4. Kill MCP server
5. Ask the agent to run tool
6. MCP client fails and I end up with a cancellation retry loop

## Breaking Changes

No code changes or configuration is needed.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #2282, opened 2026-03-12
Handle connection errors in StreamableHTTP post_writer

### LINKED ISSUE #2110: HTTP transport swallows non-2xx status codes causing client to hang

omar-y-abdi:
Fixes #2273.
Fixes #2110
When the server becomes unreachable mid-session, _handle_post_request raises a connection error (e.g. httpx.ConnectError). Without this fix, the exception propagates into the anyio task group and tears down the entire transport, causing an infinite cancellation retry loop with sustained high CPU use.

This wraps handle_request_async in a try/except that catches Exception (re-raising cancellation) and forwards the error to the read stream via read_stream_writer.send(exc). The client session then receives it as a normal error instead of crashing. This is the same approach used elsewhere in the transport (e.g. the outer post_writer catch).

Includes a test that connects to a dead port, sends a request, and verifies the connection error is forwarded to the read stream.

### DIFF
--- a/src/mcp/os/win32/utilities.py
+++ b/src/mcp/os/win32/utilities.py
@@ -123,6 +123,11 @@ class FallbackProcess:
         """Return the process ID."""
         return self.popen.pid
 
+    @property
+    def returncode(self) -> int | None:
+        """Return the exit code, or ``None`` if the process has not yet terminated."""
+        return self.popen.returncode
+
 
 # ------------------------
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #2124 as that predates this PR.

PR #2292, opened 2026-03-13
refactor: connect-first stream lifecycle for sse and streamable_http

maxisbey:
Follow-up to #2266 — applies the `websocket_client` pattern to the other two client transports.

## What changed

Instead of creating memory streams up front and cleaning them up in a `try`/`finally`, the transports now:

1. **Connect first** — `httpx_client_factory` / `aconnect_sse` / `AsyncExitStack` become the outer scopes
2. **Create streams only after the connection succeeds** — if `aconnect_sse()` raises `ConnectError`, no streams were ever created, nothing to clean up
3. **Own all four stream ends + task group in one merged `async with`** as the innermost scope — closes everything in reverse order on exit, task group waits for cancelled tasks before any stream end closes

Drops the `try`/`finally` wrapper and four explicit `aclose()` calls from each file. Net −6 lines.

## Why the previous attempt was reverted

Commit history in #2266: [`1e83583`](https://github.com/modelcontextprotocol/python-sdk/pull/2266/commits/1e835838) tried replacing `try:` with `async with streams:` at the **same nesting position**, leaving `httpx_client_factory` and `aconnect_sse` nested **inside** it. On Python 3.14, that triggered phantom branch arcs on the inner `async with` lines — each context manager's `__aexit__` compiles to a `POP_JUMP_IF_TRUE` ("did it suppress the exception?") that gets misattributed through 3.14's exception table. [`d297fcf`](https://github.com/modelcontextprotocol/python-sdk/pull/2266/commits/d297fcf) reverted to `try`/`finally`.

This PR inverts the nesting instead. Verified empirically across 3.11/3.13/3.14:

| Structure | 3.11 | 3.13 | 3.14 |
|---|---|---|---|
| 1 outer `async with` + multi-CM innermost (websocket, streamable_http) | ✅ | ✅ | ✅ |
| 2 outer `async with` + multi-CM innermost (sse) | ✅ | ✅ | 1 phantom on the multi-CM line |
| same + `# pragma: no branch` on that line | ✅ | ✅ | ✅ |

`streamable_http` has only one outer layer (`AsyncExitStack`) → clean on all versions, zero pragmas. `sse` has two unavoidable outer layers (`httpx_client_factory` feeds `client` into `aconnect_sse` — data dependency, can't merge into one line) → one documented pragma on the `async with (streams, tg)` line, which is the line we own.

## Behavior change

`sse_client`'s `ConnectError` is no longer wrapped in an `ExceptionGroup` — the task group is never entered when the connection fails, so there's nothing to group. Updated `test_sse_client_closes_all_streams_on_connection_error` to expect a bare `httpx.ConnectError`.

### DIFF
--- a/tests/client/test_transport_stream_cleanup.py
+++ b/tests/client/test_transport_stream_cleanup.py
@@ -71,6 +71,29 @@ async def test_sse_client_closes_all_streams_on_connection_error(free_tcp_port: 
 
 
 @pytest.mark.anyio
+async def test_sse_client_closes_all_streams_on_http_error() -> None:
+    """sse_client creates streams only after raise_for_status() passes, so an
+    HTTPStatusError from a 4xx/5xx response propagates bare (not wrapped in an
+    ExceptionGroup) with nothing to leak — the task group is never entered.
+    """
+
+    def return_403(request: httpx.Request) -> httpx.Response:
+        return httpx.Response(403)
+
+    def mock_factory(
+        headers: dict[str, str] | None = None,
+        timeout: httpx.Timeout | None = None,
+        auth: httpx.Auth | None = None,
+    ) -> httpx.AsyncClient:
+        return httpx.AsyncClient(transport=httpx.MockTransport(return_403))
+
+    with _assert_no_memory_stream_leak():
+        with pytest.raises(httpx.HTTPStatusError):
+            async with sse_client("http://test/sse", httpx_client_factory=mock_factory):
+                pytest.fail("should not reach here")  # pragma: no cover
+
+
+@pytest.mark.anyio
 async def test_streamable_http_client_closes_all_streams_on_exit() -> None:
     """streamable_http_client must close all 4 stream ends on exit.
 


src/mcp/client/sse.py
```diff
@@ -57,108 +57,107 @@ async def sse_client(
...
-                                        ),
+    logger.debug(f"Connecting to SSE endpoint: {remove_request_params(url)}")
+    async with httpx_client_factory(
+        headers=headers, auth=auth, timeout=httpx.Timeout(timeout, read=sse_read_timeout)
+    ) as client:
+        async with aconnect_sse(
+            client,
+            "GET",
+            url,
+        ) as event_source:
```
Kludex:
```suggestion
        async with aconnect_sse(client, "GET", url) as event_source:
```

src/mcp/client/sse.py
```diff
@@ -57,108 +57,107 @@ async def sse_client(
...
+                            case _:  # pragma: no cover
+                                logger.warning(f"Unknown SSE event: {sse.event}")  # pragma: no cover
+                except SSEError as sse_exc:  # pragma: lax no cover
+                    logger.exception("Encountered SSE exception")
+                    raise sse_exc
+                except Exception as exc:  # pragma: lax no cover
+                    logger.exception("Error in sse_reader")
+                    await read_stream_writer.send(exc)
+                finally:
+                    await read_stream_writer.aclose()
```
Kludex, line 122: Is this still neecessary?

maxisbey: yes, unless we rewrite this function to also use the `async with` thing but that'd also add a new layer of indentation.

PR #1979, opened 2026-01-30
feat: extract parameter descriptions from function docstrings

heath-hunnicutt-ruach-tov:
## Summary
- Parse Google, NumPy, and Sphinx-style docstrings to automatically populate parameter descriptions in tool/prompt/resource JSON schemas
- Previously, parameter descriptions required explicit `Field(description=...)` annotations; now they can come from standard docstrings
- `Field(description=...)` annotations take precedence when both are present, ensuring full backward compatibility

Closes #226

## Changes
- **New**: `src/mcp/server/mcpserver/utilities/docstring_utils.py` — docstring parsing utility using `griffe` with auto-detection of Google/NumPy/Sphinx formats
- **Modified**: `func_metadata()` accepts optional `param_descriptions` dict, injected as fallback when no `Field(description=...)` exists
- **Modified**: `Tool.from_function()`, `Prompt.from_function()`, `ResourceTemplate.from_function()` — parse docstrings and pass results through
- **Modified**: `pyproject.toml` — added `griffe>=1.0` as runtime dependency
- **New**: 8 unit tests for parsing utility + 7 integration tests covering all formats, precedence rules, and edge cases

## Test plan
- [x] All 8 `test_docstring_utils.py` tests pass
- [x] All 8 `test_parameter_descriptions.py` tests pass (1 existing + 7 new)
- [x] All 243 existing mcpserver tests pass with zero regressions
- [ ] CI passes on all supported Python versions (3.10–3.14)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2295, opened 2026-03-15
feat: extract parameter descriptions from docstrings into tool JSON schemas

### LINKED ISSUE #226: Improving the jsonschema created from the docstring in FastMCP
Currently, FastMCP does some function inspection to create the docstring here: https://github.com/modelcontextprotocol/python-sdk/blob/775f87981300660ee957b63c2a14b448ab9c3675/src/mcp/server/fastmcp/utilities/func_metadata.py#L105-L174

From my understanding, it creates a FuncMetadata model in pydantic which then gets converted to jsonschema. I could be wrong but I believe the argument descriptions do not get extracted properly.

It'd be nicer to follow one of the python docstring styles and parse out the argument descriptions from the docstring. 
```
def add_numbers(a: float, b: float) -> float:
    """
    Adds two numbers and returns the result.

    Args:
        a (float): The first number.
        b (float): The second number.

    Returns:
        float: The sum of a and b.
    """
    return a + b
```

should get parsed into:
```

```
{
  "name": "add_numbers",
  "description": "Adds two numbers and returns the sum.",
  "parameters": {
    "type": "object",
    "properties": {
      "a": {
        "type": "number",
        "description": "The first number to add."
      },
      "b": {
        "type": "number",
        "description": "The second number to add."
      }
    },
    "required": ["a", "b"]
  }
}
```

we used to do this in a previous python version of goose: https://github.com/block/goose/blob/eccb1b22614f39b751db4e5efd73d728d9ca40fc/packages/exchange/src/exchange/utils.py#L82-L107

here are some test examples: https://github.com/block/goose/blob/eccb1b22614f39b751db4e5efd73d728d9ca40fc/packages/exchange/tests/test_utils.py#L32-L136

**Additional context**
Add any other context or screenshots about the feature request here.

### PR DESCRIPTION
**Pull request by ameenalkhaldi**
## Summary

Closes #226.

Tool functions with standard Python docstrings now automatically get parameter descriptions included in their generated JSON schemas. Previously, parameter descriptions were only available through explicit `Field(description=...)` annotations — now Google, NumPy, and Sphinx-style docstrings are parsed and parameter descriptions are extracted into the schema.

**Before:**
```python
@mcp.tool()
def get_weather(city: str, units: str = "celsius") -> str:
    """Get the weather for a city.

    Args:
        city: The city to get weather for
        units: Temperature units (celsius or fahrenheit)
    """
    ...
```
```json
{"city": {"type": "string", "title": "City"}}
```

**After:**
```json
{"city": {"type": "string", "title": "City", "description": "The city to get weather for"}}
```

## Implementation

- Uses the [griffe](https://github.com/mkdocstrings/griffe) library (already a transitive dependency via mkdocstrings) for docstring parsing with automatic format detection
- Adapted the docstring style inference from pydantic-ai, as suggested by @Kludex in #226
- Explicit `Field(description=...)` annotations always take priority over docstring descriptions
- No changes to the public API or existing behavior for functions without docstrings

## Changes

- **`pyproject.toml`**: Added `griffe>=1.5.0` as a runtime dependency
- **`func_metadata.py`**: Added docstring parsing helpers and integrated them into the parameter schema generation loop
- **`test_func_metadata.py`**: 9 new tests covering Google/NumPy/Sphinx styles, priority logic, edge cases

## Test plan

- [x] All 42 tests pass (33 existing + 9 new)
- [x] pyright strict mode: 0 errors
- [x] ruff lint + format: clean
- [x] New code has 100% branch coverage
- [x] No changes to existing test assertions (full backwards compatibility)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=4.9",
+    "griffe>=1.5.0",
     "httpx>=0.27.1",
     "httpx-sse>=0.4",
     "pydantic>=2.12.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=4.9",
-    "griffe>=1.5.0",
     "httpx>=0.27.1",
     "httpx-sse>=0.4",
     "pydantic>=2.12.0",
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -1,14 +1,18 @@
 import functools
 import inspect
 import json
-from collections.abc import Awaitable, Callable, Sequence
+import logging
+import re
+from collections.abc import Awaitable, Callable, Iterator, Sequence
+from contextlib import contextmanager
 from itertools import chain
 from types import GenericAlias
-from typing import Annotated, Any, cast, get_args, get_origin, get_type_hints
+from typing import Annotated, Any, Literal, cast, get_args, get_origin, get_type_hints
 
 import anyio
 import anyio.to_thread
 import pydantic_core
+from griffe import Docstring, DocstringSectionKind, GoogleOptions
 from pydantic import BaseModel, ConfigDict, Field, WithJsonSchema, create_model
 from pydantic.fields import FieldInfo
 from pydantic.json_schema import GenerateJsonSchema, JsonSchemaWarningKind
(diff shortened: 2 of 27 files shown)


src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -167,6 +171,126 @@ def pre_parse_json(self, data: dict[str, Any]) -> dict[str, Any]:
...
+
+
+@contextmanager
+def _suppress_griffe_logging() -> Iterator[None]:
+    """Temporarily suppress griffe's verbose logging."""
+    old_level = logging.root.getEffectiveLevel()
+    logging.root.setLevel(logging.ERROR)
+    yield
+    logging.root.setLevel(old_level)
+
```
Bortlesboat, line 193: This sets the root logger level, which will suppress all logging globally (not just griffe's) for the duration of the context manager. It's also not thread-safe — concurrent code that logs during this window will silently lose messages.

Should target the griffe logger specifically:
```python
@contextmanager
def _suppress_griffe_logging() -> Iterator[None]:
    logger = logging.getLogger("_griffe")
    old_level = logger.getEffectiveLevel()
    logger.setLevel(logging.ERROR)
    yield
    logger.setLevel(old_level)
```

(`_griffe` is the internal logger name griffe uses — you can verify with `griffe.logger`)

pyproject.toml
```diff
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=4.9",
+    "griffe>=1.5.0",
```
Bortlesboat: `griffe` pulls in transitive deps (colorama, etc.) and is a non-trivial addition to the dependency tree for what's ultimately a convenience feature (docstring → description). Most users who care about tool descriptions are already using `Field(description=...)` or `Annotated`.

Worth considering:
- Making this an optional/extra dependency (e.g., `mcp[docstrings]`), with a graceful fallback when not installed
- Or using a simpler regex-based approach for the most common docstring formats (Google-style `Args:` covers the vast majority of cases)

ameenalkhaldi:
Good point — replaced griffe with a simple regex-based parser in 735bc50. Zero new dependencies now.

The regex approach covers Google (`Args:`), Sphinx (`:param:`), and NumPy (`Parameters\n---`) styles, which handles the vast majority of real-world docstrings. All 42 tests still pass.

Bortlesboat (commented):
The regex approach cleanly resolves both the dependency and logging concerns. Nice work.

src/mcp/server/mcpserver/utilities/func_metadata.py
pawamoy: Maybe on a old version? On latest it's `"griffe"`:

```python
>>> import griffe
>>> griffe.logger._logger.name
'griffe'
```

pyproject.toml
pawamoy: You can depend on `griffelib` to avoid pulling CLI dependencies such as colorama. Anyway you seem to have opted for a no-deps approach which is totally fine to me :smile: Just wanted to provide details :slightly_smiling_face: 

By the way most Griffe dependents missed that our `infer_docstring_style` function is now public: https://mkdocstrings.github.io/griffe/reference/api/docstrings/parsers/#griffe.infer_docstring_style. You could also get inspiration from our own [patterns](https://github.com/mkdocstrings/griffe/blob/97106e4f56c99146f23864c7777e5bfaec89bafe/packages/griffelib/src/griffe/_internal/docstrings/auto.py#L31).

maxisbey:
Closing for the same reason as #1979 — we're holding off on docstring parsing until Pydantic publishes a standalone package for it, rather than maintaining a parser in-repo. #226 stays open and `needs decision`. Thanks for the contribution.

Issue #1969, opened 2026-01-28
Propagate ContextVars to Transport Layer in MCP Clients


somaraani:
## Summary

Ensure that `contextvars` set in the calling thread are properly propagated to the transport layer at request time, enabling dynamic configuration of transport settings (e.g., HTTP headers, auth mechanism, etc.) on a per-request basis.

## Problem Statement

Currently, `contextvars` set in the main thread are not visible in the transport layer when making requests. This is because:

1. **Context isolation**: The `streamable_http_transport` runs in a different async context or thread, so context vars set after initialization are not visible
2. **One-time copy**: Context may be copied only at session/transport creation time, not per-request

This prevents common use cases:
- **Authentication**: Custom `httpx.Auth` implementations that read from context vars cannot access values set in the calling code
- **Token Exchange**: Middleware that performs OAuth token exchange and stores the result in a context var cannot make that token available to outbound MCP requests
- **Tracing**: Trace context set in request handlers isn't visible when the HTTP request is made
- **Multi-tenancy**: Tenant context set per-request isn't available to the transport

### Why ContextVars?

Python's `contextvars` module is the standard mechanism for request-scoped state:
- Already widely used in async frameworks (FastAPI, Starlette, aiohttp)
- Thread-safe and async-safe by design
- Values can be set in middleware/handlers and should be available downstream
- Zero coupling between application code and transport internals

## Proposed Solution

Ensure that context vars are properly propagated from the calling code to the transport layer at request time. This would allow:

1. Custom `httpx.Auth` implementations to read context vars
2. Any code in the transport layer to access current context

### Example Use Case: Custom Auth

```python
from contextvars import ContextVar
import httpx
from mcp.client.streamable_http import streamable_http_client

# Application defines context var for auth
auth_token_var: ContextVar[str] = ContextVar('auth_token')


class ContextVarAuth(httpx.Auth):
    """Auth that reads token from context var."""
    
    def auth_flow(self, request):
        try:
            token = auth_token_var.get()
            request.headers["Authorization"] = f"Bearer {token}"
        except LookupError:
            pass  # No token set
        yield request


# Create client with custom auth
client = httpx.AsyncClient(auth=ContextVarAuth())

async with streamable_http_client(
    "https://mcp.example.com",
    http_client=client,
) as (read, write, _):
    async with ClientSession(read, write) as session:
        await session.initialize()
        
        # Request 1: Set context for user A
        auth_token_var.set("user-a-token")
[...]

PR #1996, opened 2026-02-04
Propagate contextvars through anyio streams

aabmass:
<!-- Provide a brief summary of your changes -->
Fixes https://github.com/modelcontextprotocol/python-sdk/issues/1969
Part of https://github.com/modelcontextprotocol/python-sdk/issues/421

This PR propagates [`contextvars.Context`](https://docs.python.org/3/library/contextvars.html#contextvars.Context) through internal anyio memory streams and restores it in long running async tasks, so that they run with the correct context. Works for in memory, streamable http, and sse transports.

- For MCP clients, this captures the caller's active context and restores it before using the transport.
- For MCP servers, it captures the active context from the transport and restores it before using the Server's handlers

## Motivation and Context
<!-- Why is this change needed? What problem does it solve? -->
The goal is to make contextvars "work" with the MCP SDK as users expect. The main motivation is OpenTelemetry https://github.com/modelcontextprotocol/python-sdk/issues/421, but there are other use cases e.g. [per-request auth](https://github.com/modelcontextprotocol/python-sdk/issues/1969), [stdlib `decimal`](https://docs.python.org/3/library/decimal.html#decimal.localcontext), [starlette-context](https://github.com/tomwojcik/starlette-context), etc.

## How Has This Been Tested?
<!-- Have you tested this in a real application? Which scenarios were tested? -->
Added several integration tests which are failing at `main`. I also manually tested e2e with the some of the samples by using OpenTelemetry httpx instrumentation (I can share screenshots).

<details><summary>Tests failures at main</summary>
<p>

```console
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════ inline-snapshot ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── Fix snapshots ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── tests/test_context_propagation.py ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ @@ -83,7 +83,7 @@                                                                                                                                                                                                                                                                                                                                           │
│                                                                                                                                                                                                                                                                                                                                                             │
[...]

Issue #421, opened 2025-04-03
Adding Opentelemetry to MCP SDK


fali007:
**Is your feature request related to a problem? Please describe.**

I would like to see Opentelemetry traces and Metrics baked into SDK. 
E.g. 
1. Traces emitted on all Requests from client and server side of MCP SDK. 
2. Metrics for tool/prompt/resource calls

**Describe the solution you'd like**

I am trying to use Opentelemetry sdk to add traces and metrics features.
### Steps
1. Creating session in client and server.
2. Initialise tracer (Opentelemetry tracer instances) in `BaseSession` (`src/mcp/shared/session.py`).
3. Binding trace context (`traceparent`) to `_meta` of `RequestParams.Meta` before sending request from client (`send_request` in `src/mcp/shared/session.py`)
4. In server use the `traceparent` from received request to create new span with incoming span as parent span.

**Describe alternatives you've considered**

I tried to implement the same via traceloop. I was not able to get correct parent - child relationship between spans while doing so.

**Additional context**

I am able to achieve distributed tracing across my Agentic application and MCP server.

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/2165c7ca-6283-447b-9495-a9e9406fda4f" />

hk-bmi:
related Issue for feature request in traceloop/openllmetry to enable observability for MCP Server, https://github.com/traceloop/openllmetry/issues/2662

wenxuwan:
same needs

wenxuwan:
![Image](https://github.com/user-attachments/assets/a9997677-936a-41f9-8c54-ac911a5693e8)

![Image](https://github.com/user-attachments/assets/e62aec1f-2643-4d52-9b12-d7b963c50989)

It is also clear from the logs that the ContextVarsScope used for opentracing each time a request is processed is that of the first request, even though handle_post_message and SofaTracerMiddleware use the latest ContextVarsScope for each request.

![Image](https://github.com/user-attachments/assets/2a931f2b-fa1f-44ac-8234-08bb9d123c9d)

fali007:
Hi @wenxuwan, I was tinkering with the traceloop way of passing context by intercepting and wrapping over `_handle_request` function in `mcp.server.lowlevel.server.py` for server side and `send_request` function in `mcp.shared.session.py` for client side tracing. I am trying to pass `traceparent` in `params.meta.traceparent` of the `RequestParams` from `types.py`.
[...]

PR #2298, opened 2026-03-16
Propagate contextvars.Context through anyio streams without modifying SessionMessage

Kludex:
## Summary

MCP server handlers currently lose the sender's `contextvars.Context`. When a client sets context variables (e.g. OpenTelemetry trace/span IDs) before sending a request, the server handler runs in the receive loop's context instead, breaking trace propagation.

This PR propagates `contextvars.Context` through anyio memory streams **without adding any field to `SessionMessage`**, addressing [Kludex's review comment](https://github.com/modelcontextprotocol/python-sdk/pull/1996#discussion_r2813932367) on the previous approach (#1996).

### How it works

- `ContextSendStream` / `ContextReceiveStream` wrappers capture the sender's context at `send()` time and expose it via `last_context` on the receive side
- `BaseSession._receive_loop` reads `last_context` and passes it to `RequestResponder`, which carries it to the server handler boundary
- `lowlevel/server.py` uses `context.run(tg.start_soon, handler)` so handler tasks inherit the sender's context
- Client transport `post_writer` functions restore sender context the same way
- Protocol-based `ReadStream` / `WriteStream` replace concrete `MemoryObjectSendStream` / `MemoryObjectReceiveStream` types, so both raw and context-aware streams are accepted
- `create_context_streams[T](n)` mirrors anyio's bracket syntax with proper generic propagation

### Additional changes

- Switch coverage config from `exclude_lines` to `exclude_also` to inherit coverage.py's built-in defaults (which handle `...` stubs and `pragma: no cover`)

Fixes https://github.com/modelcontextprotocol/python-sdk/issues/1969
Part of https://github.com/modelcontextprotocol/python-sdk/issues/421

## Test plan

- [x] `test_context_propagation`: sets a contextvar via a context manager (which resets on exit), calls a tool, verifies the server handler sees the sender's value. Fails on `main`, passes with this PR.
- [x] All existing tests pass
- [x] `pyright` passes with 0 errors

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -350,4 +350,6 @@ async def test_context_propagation():
         with _set_test_contextvar("client_value"):
             result = await client.call_tool("check_context", {})
 
-    assert result.content[0].text == "client_value"  # type: ignore[union-attr]
+    assert result.content[0].text == "client_value", (  # type: ignore[union-attr]
+        "Server handler did not see the sender's contextvars.Context"
+    )


src/mcp/shared/_stream_protocols.py
```diff
@@ -0,0 +1,51 @@
...
+
+    async def send(self, item: T_contra, /) -> None: ...  # pragma: no branch
+    async def aclose(self) -> None: ...  # pragma: no branch
+    async def __aenter__(self) -> Self: ...  # pragma: no branch
+    async def __aexit__(  # pragma: no branch
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> bool | None: ...
```
maxisbey: why are there a bunch of no branch pragmas here

src/mcp/shared/session.py
```diff
@@ -406,6 +413,14 @@ async def _receive_loop(self) -> None:
                     else:  # Response or error
                         await self._handle_response(message)
 
+                async for message in self._read_stream:
+                    if isinstance(message, Exception):
+                        await self._handle_incoming(message)
+                        continue
+
+                    sender_ctx: contextvars.Context | None = getattr(self._read_stream, "last_context", None)
```
Kludex: Why are you using `getattr`? We shoudn't be using it!

maxisbey: Yea!

Kludex: Okay, I think we should.

Kludex: The point is that we have this "special" stream now, but the type should be `ReadStream` (to not be a breaking change).

maxisbey (approved):

PR #2323, opened 2026-03-20
fix: preserve real stdin/stdout after stdio server exits

owendevereaux:
## Summary

Fixes #1933

When running the MCP server with `transport='stdio'`, closing the server would also close the process's real stdin/stdout handles. This caused a `ValueError: I/O operation on closed file` when trying to use stdio after the server exits.

## Problem

The issue was that wrapping `sys.stdin.buffer`/`sys.stdout.buffer` with `TextIOWrapper` causes the underlying buffer to be closed when the wrapper is garbage collected, even if we don't use a context manager. The old comment said "Purposely not using context managers for these, as we don't want to close standard process handles" but this didn't prevent the closing.

## Solution

Use `os.dup()` to duplicate the file descriptors before wrapping. When the duplicated descriptors are closed (along with the TextIOWrapper), the original stdin/stdout remain intact and usable.

For streams without a `fileno()` (e.g., `BytesIO` in tests), we fall back to wrapping them directly (previous behavior).

## Test Plan

Added a regression test `test_stdio_server_does_not_close_real_stdio` that:
1. Creates real file descriptors (via temp files) to simulate stdin/stdout
2. Monkeypatches sys.stdin/stdout to use them
3. Runs the stdio server
4. Verifies that the original handles are NOT closed after the server exits

All existing tests continue to pass.

## Reproduction (from issue)

```python

mcp = FastMCP("Demo")
mcp.run(transport="stdio")
print("?")  # Previously raised: ValueError: I/O operation on closed file.
```

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,6 +17,8 @@ Example:
     ```
 """
 
+import io
+import os
 import sys
 from contextlib import asynccontextmanager
 from io import TextIOWrapper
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #2040 as that predates this PR.

PR #2324, opened 2026-03-21
fix: preserve client_metadata.scope if already set

### LINKED ISSUE #2317: Don't override client_metadata.scopes if they are already set
### Initial Checks

### Description

The scope selection strategy inside async_auth_flow unconditionally requests all available scopes. This overwrites the scope list that may have been explicitly set by the client. Being able to explicitly set the requested scopes is an important use case, either to reduce the permissions granted or because the server only permits certain scopes (despite advertising others).

From https://github.com/modelcontextprotocol/python-sdk/blob/v1.26.0/src/mcp/client/auth/oauth2.py#L553-L558:
```
                        extract_scope_from_www_auth(response),
                        self.context.protected_resource_metadata,
                        self.context.oauth_metadata,
                    )
```
This could be conditional on if `self.context.client_metadata.scope is None`.

I see that this behavior was previously suggested in https://github.com/modelcontextprotocol/python-sdk/pull/1324#discussion_r2410806282 and rejected, on the basis that "Requesting all available scopes allows the authorization server and end-user to determine appropriate permissions during the consent process". However, I think this is worth revisiting. The specific motivating example here is the official SalesForce MCP server: if the client requests scopes that are not authorized for the given client application, the server rejects the request entirely.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
python 3.12.12
sdk 1.26.0
```

owendevereaux:
## Summary

Don't override `client_metadata.scope` if it was explicitly set by the client.

## Problem

As reported in #2317, the scope selection strategy in `async_auth_flow` unconditionally overwrites any scope that was explicitly set by the client. This prevents clients from:
- Requesting fewer permissions than available
- Working with servers that reject requests for unauthorized scopes (e.g., SalesForce)

## Solution

Made the scope assignment conditional:

```python
# Step 3: Apply scope selection strategy (only if not already set)
if self.context.client_metadata.scope is None:
    self.context.client_metadata.scope = get_client_metadata_scopes(...)
```

This preserves the existing behavior when no scope is set, while respecting explicit scope configuration.

## Testing

- Existing tests continue to pass (scope is None by default)
- Clients that explicitly set `client_metadata.scope` will now have that value preserved

Fixes #2317

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -571,12 +571,13 @@ class OAuthClientProvider(httpx.Auth):
                         else:
                             logger.debug(f"OAuth metadata discovery failed: {url}")
 
-                    # Step 3: Apply scope selection strategy
-                    self.context.client_metadata.scope = get_client_metadata_scopes(
-                        extract_scope_from_www_auth(response),
-                        self.context.protected_resource_metadata,
-                        self.context.oauth_metadata,
-                    )
+                    # Step 3: Apply scope selection strategy (only if not already set)
+                    if self.context.client_metadata.scope is None:
+                        self.context.client_metadata.scope = get_client_metadata_scopes(
+                            extract_scope_from_www_auth(response),
+                            self.context.protected_resource_metadata,
+                            self.context.oauth_metadata,
+                        )
 
                     # Step 4: Register client or use URL-based client ID (CIMD)
                     if not self.context.client_info:


maxisbey:
Thanks for the PR! The linked issue (#2317) hasn't been triaged yet — per our [CONTRIBUTING.md](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md), please wait for maintainer feedback before starting work. Closing for now; feel free to reopen once the issue is labeled `ready for work`.

PR #2325, opened 2026-03-21
fix: return INVALID_PARAMS (-32602) for resource-not-found errors

### LINKED ISSUE #1579: FastMCP read_resource() returns incorrect error code when resource not found

leoneperdigao:
## Summary

- Fix `read_resource()` to raise `MCPError(code=INVALID_PARAMS)` instead of `ResourceError`, so resource-not-found errors return `-32602` instead of `0`
- Aligns with the TypeScript SDK behavior and emerging spec consensus ([modelcontextprotocol#1545](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1545))
- Adds regression test verifying the error code for both unknown resources and resource read failures

### Root cause

`ResourceError` extends `MCPServerError` which extends `Exception`, it is **not** an `MCPError`. When the low-level server's request handler catches it, it falls into the generic `except Exception` branch which returns `ErrorData(code=0, ...)`. The fix raises `MCPError` directly at the protocol boundary in `read_resource()`, following the same pattern used elsewhere in the codebase (validation errors, session errors, `UrlElicitationRequiredError`).

## Test plan

- [x] New test `test_1579_resource_error_code.py` verifies error code is `-32602` for unknown resources and read failures
- [x] Updated `test_141_resource_templates.py` to catch `MCPError` instead of `ResourceError`
- [x] All 364 existing tests pass with no regressions

Closes #1579

### DIFF
--- a/tests/issues/test_1579_resource_error_code.py
+++ b/tests/issues/test_1579_resource_error_code.py
@@ -23,10 +23,6 @@ async def test_unknown_resource_returns_invalid_params_error_code():
     """Reading an unknown resource returns INVALID_PARAMS (-32602), not 0."""
     mcp = MCPServer()
 
-    @mcp.resource("resource://exists")
-    def existing_resource():
-        return "data"
-
     async with Client(mcp) as client:
         with pytest.raises(MCPError) as exc_info:
             await client.read_resource("resource://does-not-exist")


leoneperdigao:
Closing in favor of #1582. As @maxisbey noted there, the spec discussion (modelcontextprotocol/modelcontextprotocol#1545) needs to settle first before merging a fix for the error code. Happy to revisit once that's resolved.

PR #2330, opened 2026-03-22
Fix ctx.log type to accept any JSON-serializable type

qianchongyang:
## Problem

The `ctx.info` and similar logging methods have `message: str` type, but the MCP spec defines `data: unknown` to allow any JSON-serializable type. This inconsistency causes type errors when trying to log non-string data.

## Solution

Updated type hints for logging methods to accept `unknown` instead of `str`:
- Changed `message: str` to `message: unknown` in `ctx.info`, `ctx.debug`, etc.
- Clarified `extra` parameter documentation
- Aligned with `LoggingMessageNotificationParams` and `send_log_message` signatures

## Validation

```python
# Now works without type error
ctx.info({"key": "value"})  # Dict logging
ctx.info([1, 2, 3])          # List logging
ctx.info(SomeObject())         # Object logging
```

Fixes #397

### DIFF
--- a/src/mcp/server/mcpserver/context.py
+++ b/src/mcp/server/mcpserver/context.py
@@ -187,7 +187,7 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
     async def log(
         self,
         level: Literal["debug", "info", "warning", "error"],
-        message: str,
+        data: Any,
         *,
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for your contribution! Closing this in favour of #1907 as that predates this PR.

PR #2334, opened 2026-03-22
[v1.x] fix: handle ClosedResourceError when transport closes mid-request

### LINKED ISSUE #2328: [Bug] Server crashes with anyio.ClosedResourceError when receiving raw invalid UTF-8 bytes
### Initial Checks

## Summary
MCP Python SDK servers crash with `anyio.ClosedResourceError` when a client sends raw invalid UTF-8 bytes in request parameters. A malicious client can crash any MCP server using a single request.

## Environment
- **MCP SDK Version**: 1.26.0 (latest)
- **Python**: 3.13
- **OS**: Windows 11

## Proof of Concept

Save as `crash_poc.py`:
```python
import sys
import time

# Initialize
sys.stdout.buffer.write(b'{"jsonrpc": "2.0", "id": 1, "method": "initialize", "params": {"protocolVersion": "2024-11-05", "capabilities": {}, "clientInfo": {"name": "test", "version": "1.0"}}}\n')
sys.stdout.flush()
time.sleep(1)

# Initialized notification
sys.stdout.buffer.write(b'{"jsonrpc": "2.0", "method": "notifications/initialized"}\n')
sys.stdout.flush()
time.sleep(0.5)

# Crash payload - raw bytes 0xFF 0xFE
sys.stdout.buffer.write(b'{"jsonrpc": "2.0", "id": 3, "method": "tools/call", "params": {"name": "fetch", "arguments": {"url": "http://x\xff\xfe"}}}\n')
sys.stdout.flush()
time.sleep(3)
```

Run:
```bash
python crash_poc.py | python -m mcp_server_fetch
```

## Stack Trace
```
File "...\mcp\server\lowlevel\server.py", line 781, in _handle_request
    await message.respond(response)
File "...\mcp\shared\session.py", line 349, in _send_response
File "...\anyio\streams\memory.py", line 218, in send_nowait
    raise ClosedResourceError
anyio.ClosedResourceError
```

## Key Observation
- String `"http://x[0xff][0xfe]"` → handled gracefully with validation error ✅
- Raw bytes `b"http://x\xff\xfe"` → server crash ❌

## Expected Behavior
The server should return a JSON-RPC error response for malformed input, not crash.

## Impact
**Denial of Service** - Any MCP client can crash any MCP server using the Python SDK with a single malformed request.

## Discovery
Found using [mcpsec](https://github.com/manthanghasadiya/mcpsec) evolutionary fuzzer.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
python --version
Python 3.13.1

pip show mcp | Select-String "Version."
Version: 1.26.0
```

owendevereaux:
## Summary

Backports fixes from #2306 to v1.x to address issue #2328.

When the transport closes while handlers are processing requests (e.g., stdin EOF during a long-running tool call, or a malicious client sending invalid UTF-8 bytes that crashes the transport), the server could crash with `ClosedResourceError` when trying to send a response through the already-closed write stream.

## Changes

1. **Wrap message loop in try/finally with cancel** - When the transport closes, in-flight handlers are now cancelled instead of being left to attempt responding through a closed stream
2. **Catch ClosedResourceError on respond** - If a response attempt fails due to closed transport, log and return gracefully instead of crashing
3. **Fix cancellation handling** - Distinguish between client-initiated cancellation (already responded) and transport-close cancellation (should re-raise)
4. **Fix dict iteration in finally** - Use `list()` snapshot when iterating `_response_streams` to avoid 'dictionary changed size during iteration' errors

## Testing

Added two new test cases that reproduce the crash scenarios:
- `test_server_cancels_in_flight_handlers_on_transport_close`
- `test_server_handles_transport_close_with_pending_server_to_client_requests`

All existing tests continue to pass.

Fixes #2328

### DIFF
--- a/tests/server/test_cancel_handling.py
+++ b/tests/server/test_cancel_handling.py
@@ -9,14 +9,21 @@ import mcp.types as types
 from mcp.server.lowlevel.server import Server
 from mcp.shared.exceptions import McpError
 from mcp.shared.memory import create_connected_server_and_client_session
+from mcp.shared.message import SessionMessage
 from mcp.types import (
+    LATEST_PROTOCOL_VERSION,
     CallToolRequest,
     CallToolRequestParams,
     CallToolResult,
     CancelledNotification,
     CancelledNotificationParams,
+    ClientCapabilities,
     ClientNotification,
     ClientRequest,
+    Implementation,
+    InitializeRequestParams,
+    JSONRPCNotification,
+    JSONRPCRequest,
     Tool,
 )
 
@@ -122,16 +129,6 @@ async def test_server_cancels_in_flight_handlers_on_transport_close():
     This drives server.run() with raw memory streams because InMemoryTransport
     wraps it in its own finally-cancel (_memory.py) which masks the bug.
     """
-    from mcp.shared.message import SessionMessage
-    from mcp.types import (
-        LATEST_PROTOCOL_VERSION,
-        ClientCapabilities,
-        Implementation,
-        InitializeRequestParams,
-        JSONRPCNotification,
-        JSONRPCRequest,
-    )
-
     handler_started = anyio.Event()
     handler_cancelled = anyio.Event()
     server_run_returned = anyio.Event()
@@ -139,7 +136,7 @@ async def test_server_cancels_in_flight_handlers_on_transport_close():
     server = Server("test")
 
     @server.call_tool()
-    async def handle_call_tool(name: str, arguments: dict | None) -> list[types.TextContent]:
+    async def handle_call_tool(name: str, arguments: dict[str, Any] | None) -> list[types.TextContent]:
         handler_started.set()
         try:
             await anyio.sleep_forever()
@@ -206,16 +203,6 @@ async def test_server_handles_transport_close_with_pending_server_to_client_requ
       2. The woken handler's MCPError is caught in _handle_request, which falls
          through to respond() against a write stream _receive_loop already closed.
     """
-    from mcp.shared.message import SessionMessage
-    from mcp.types import (
-        LATEST_PROTOCOL_VERSION,
-        ClientCapabilities,
-        Implementation,
-        InitializeRequestParams,
-        JSONRPCNotification,
-        JSONRPCRequest,
-    )
-
     handlers_started = 0
     both_started = anyio.Event()
     server_run_returned = anyio.Event()
@@ -223,7 +210,7 @@ async def test_server_handles_transport_close_with_pending_server_to_client_requ
     server = Server("test")
 
     @server.call_tool()
-    async def handle_call_tool(name: str, arguments: dict | None) -> list[types.TextContent]:
+    async def handle_call_tool(name: str, arguments: dict[str, Any] | None) -> list[types.TextContent]:
         nonlocal handlers_started
         handlers_started += 1
         if handlers_started == 2:


bbarwik:
@owendevereaux @maxisbey This issue was not correctly fixed. Bellow is report what is wrong, I'll add issue for it too:

# Bug: `AssertionError: Request already responded to` — cancellation race in v1.27.0

`AssertionError: Request already responded to` when CancelledNotification arrives after handler completes but before `respond()`

## Description

When a client sends a `notifications/cancelled` for a request whose handler has already finished executing but hasn't yet called `message.respond()`, the server crashes with `AssertionError: Request already responded to`.

PR #2334 (v1.27.0) fixed the `ClosedResourceError` crash path by catching `CancelledError` in `_handle_request` and guarding `respond()` against `BrokenResourceError`/`ClosedResourceError`. However, it left a race window between handler completion and `respond()` where a cancellation notification can set `_completed = True` first, causing the assert on line 129 of `session.py` to fire.

## Environment

- `mcp` 1.27.0
- Python 3.14
- anyio (asyncio backend)
- FastMCP stdio transport
- Client: Claude Code 2.1.92

## Reproduction scenario

1. Client sends a `tools/call` request with a long-running handler (e.g. polling with 600s timeout)
2. Handler completes and returns a result
3. Between the handler's `return` and the `await message.respond(response)` call in `_handle_request`, the client sends `notifications/cancelled` for that same request ID
4. The cancellation notification handler (`session.py:403-406`) calls `responder.cancel()`, which:
   - Calls `_cancel_scope.cancel()` 
   - Sets `_completed = True`
   - Sends an error response `"Request cancelled"`
5. Back in `_handle_request`, execution reaches `await message.respond(response)` at `server.py:800`
6. `respond()` hits `assert not self._completed` at `session.py:129` → **crash**

The `cancel_scope.cancel()` only raises `CancelledError` if the task is currently in an `await`. Since the handler already returned, the code path between the handler return and `respond()` is synchronous — no checkpoint where `CancelledError` can be delivered. The `except anyio.get_cancelled_exc_class()` guard at `server.py:773` never fires.

## Stack trace

```
    | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
      | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
        |   File "mcp/server/lowlevel/server.py", line 703, in _handle_message
        |     await self._handle_request(message, req, session, lifespan_context, raise_exceptions)
        |   File "mcp/server/lowlevel/server.py", line 800, in _handle_request
        |     await message.respond(response)
        |   File "mcp/shared/session.py", line 129, in respond
        |     assert not self._completed, "Request already responded to"
        | AssertionError: Request already responded to
```

## Analysis

The race window in `_handle_request` (`server.py:719`):

```python
# Line 770: handler completes, returns response
response = await handler(req)

# ... exception handling ...

# Line 799-800: GAP — between handler return and respond(),
# a CancelledNotification can arrive on another task and call
# responder.cancel(), setting _completed = True and sending
# an error response. No await in this gap means no CancelledError
# can be delivered.
try:
    await message.respond(response)  # <-- assert fires here
except (anyio.BrokenResourceError, anyio.ClosedResourceError):
    ...
```

The `except anyio.get_cancelled_exc_class()` at line 773 correctly handles the case where the cancellation arrives *during* handler execution. But it cannot handle cancellation that arrives *after* the handler returns, because there's no async checkpoint between the handler return and `respond()`.

## Suggested fix

In `session.py`, change `respond()` to handle the already-completed case gracefully instead of asserting:

```python
async def respond(self, response: SendResultT | ErrorData) -> None:
    if not self._entered:
    
    # If already completed (e.g. by a concurrent cancellation), skip silently.
    if self._completed:
        return

    if not self.cancelled:
        self._completed = True
            request_id=self.request_id, response=response
        )
```

Alternatively, the guard could be added in `_handle_request` before calling `respond()`:

```python
if not message._completed:
    try:
        await message.respond(response)
    except (anyio.BrokenResourceError, anyio.ClosedResourceError):
        logger.debug("Response for %s dropped - transport closed", message.request_id)
```

The first approach (in `respond()` itself) is more robust since it closes the race for all callers.

## Impact

This crashes the entire MCP server process, killing all in-flight requests. In our case, the server manages multiple long-running background agents, so a crash loses all active work. The crash is triggered by normal client behavior (user cancels an operation), making it a reliability issue rather than an edge case.

## Related

- #2328 — original ClosedResourceError report
- #2334 — v1.x fix (covers ClosedResourceError but not this assert race)
- #2306 — main branch transport-close cancellation

PR #2343, opened 2026-03-25
Expose optional stdin/stdout on MCPServer stdio entrypoints

mangelajo:
### Summary

`mcp.server.stdio.stdio_server` already supports optional `stdin` and `stdout`. This PR threads those parameters through **`MCPServer.run_stdio_async`** and the **`stdio`** branch of **`MCPServer.run`**, so applications can use the public API instead of calling `stdio_server` manually and depending on internal wiring.

### Motivation

MCP over stdio is JSON-RPC on a dedicated pair of streams. In real processes, fd 1 is also used for logging, `print()`, subprocesses, and native code. Any extra bytes on the protocol stream corrupt JSON-RPC and break clients.

A common pattern is to duplicate the real stdout fd for MCP-only I/O and redirect fd 1 (and often `sys.stdout`) to stderr so other output stays off the wire. That requires passing custom `stdin`/`stdout` into the same code path the server uses for stdio. **`stdio_server` already accepts them**; this exposes the same capability on **`MCPServer`**.

### Testing

Adds `tests/server/mcpserver/test_run_stdio_custom_streams.py`, which asserts custom streams are forwarded to `stdio_server` (with `stdio_server` and low-level `run` stubbed).

### Example consumer

Downstream servers https://github.com/jumpstarter-dev/jumpstarter/pull/383 need this to keep logs and tooling output off the MCP stream while staying on supported APIs.
We have a lot of logging originating from gRPC and other libraries that we want to keep out of the channel, and we are using async io.

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -238,7 +238,13 @@ class MCPServer(Generic[LifespanResultT]):
         return self._lowlevel_server.session_manager  # pragma: no cover
 
     @overload
-    def run(self, transport: Literal["stdio"] = ...) -> None: ...
+    def run(
+        self,
+        transport: Literal["stdio"] = ...,
+        *,
+        stdin: anyio.AsyncFile[str] | None = ...,
+        stdout: anyio.AsyncFile[str] | None = ...,
+    ) -> None: ...
 
     @overload
     def run(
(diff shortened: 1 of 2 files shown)


src/mcp/server/mcpserver/server.py
```diff
@@ -270,12 +276,19 @@ def run(
     def run(
         self,
         transport: Literal["stdio", "sse", "streamable-http"] = "stdio",
+        *,
+        stdin: anyio.AsyncFile[str] | None = None,
```
mangelajo: ok we don't need to modify on this one.

maxisbey:

PR #2344, opened 2026-03-25
fix(server): return -32602 for resource not found (SEP-2164)

pja-ant:
## Summary

Implements [SEP-2164](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164) — standardizes the resource-not-found error code to `-32602` (Invalid Params).

## Problem

The Python SDK currently returns error code `0` for resource-not-found, making it impossible for clients to reliably distinguish this from other errors. The spec now requires `-32602` with the URI in the `data` field.

## Changes

- Add `ResourceNotFoundError(ResourceError)` subclass in `mcpserver/exceptions.py`
- `MCPServer.read_resource()` now raises `ResourceNotFoundError` (instead of generic `ResourceError`) when the resource doesn't exist
- `_handle_read_resource()` catches `ResourceNotFoundError` and converts it to `MCPError(code=INVALID_PARAMS, ..., data={"uri": ...})`

The subclass approach keeps backward compatibility — existing code catching `ResourceError` continues to work unchanged.

## Wire format (before → after)

```diff
 {
   "jsonrpc": "2.0",
   "id": 2,
   "error": {
-    "code": 0,
+    "code": -32602,
     "message": "Unknown resource: file:///nonexistent.txt",
-    "data": null
+    "data": {"uri": "file:///nonexistent.txt"}
   }
 }
```

Spec: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164

### DIFF
--- a/src/mcp/server/mcpserver/__init__.py
+++ b/src/mcp/server/mcpserver/__init__.py
@@ -3,8 +3,7 @@
 from mcp.types import Icon
 
 from .context import Context
-from .exceptions import ResourceError, ResourceNotFoundError
 from .server import MCPServer
 from .utilities.types import Audio, Image
 
-__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon", "ResourceError", "ResourceNotFoundError"]
+__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon"]
--- a/src/mcp/server/mcpserver/__init__.py
+++ b/src/mcp/server/mcpserver/__init__.py
@@ -3,7 +3,8 @@
 from mcp.types import Icon
 
 from .context import Context
+from .exceptions import ResourceError, ResourceNotFoundError
 from .server import MCPServer
 from .utilities.types import Audio, Image
 
-__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon"]
+__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon", "ResourceError", "ResourceNotFoundError"]
--- a/src/mcp/server/mcpserver/resources/templates.py
+++ b/src/mcp/server/mcpserver/resources/templates.py
@@ -15,8 +15,11 @@ from mcp.server.mcpserver.exceptions import ResourceError
 from mcp.server.mcpserver.resources.types import FunctionResource, Resource
 from mcp.server.mcpserver.utilities.context_injection import find_context_parameter, inject_context
 from mcp.server.mcpserver.utilities.func_metadata import func_metadata
+from mcp.server.mcpserver.utilities.logging import get_logger
 from mcp.shared._callable_inspection import is_async_callable
 from mcp.types import Annotations, Icon
+
+logger = get_logger(__name__)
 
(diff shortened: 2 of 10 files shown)


src/mcp/server/mcpserver/resources/templates.py
```diff
@@ -126,5 +127,7 @@ async def create_resource(
                 meta=self.meta,
                 fn=lambda: result,  # Capture result in closure
             )
+        except ResourceError:
+            raise
         except Exception as e:
-            raise ValueError(f"Error creating resource from template: {e}")
+            raise ResourceError(f"Error creating resource from template: {e}")
```
Kludex, line 133: We keep leaking info when adding the exception in the message itself. We should drop it out.

This comment is not related to this PR.

src/mcp/server/mcpserver/__init__.py
```diff
@@ -3,7 +3,8 @@
 from mcp.types import Icon
 
 from .context import Context
+from .exceptions import ResourceError, ResourceNotFoundError
 from .server import MCPServer
 from .utilities.types import Audio, Image
 
-__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon"]
+__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon", "ResourceError", "ResourceNotFoundError"]
```
Kludex, line 10: Are those errors supposed to be user facing? I don't think that's the case in the MCPServer.

src/mcp/server/mcpserver/server.py
```diff
@@ -332,7 +334,12 @@ async def _handle_read_resource(
         self, ctx: ServerRequestContext[LifespanResultT], params: ReadResourceRequestParams
     ) -> ReadResourceResult:
         context = Context(request_context=ctx, mcp_server=self)
-        results = await self.read_resource(params.uri, context)
+        try:
+            results = await self.read_resource(params.uri, context)
+        except ResourceNotFoundError as err:
+            raise MCPError(code=INVALID_PARAMS, message=str(err), data={"uri": str(params.uri)})
+        except ResourceError as err:
+            raise MCPError(code=INTERNAL_ERROR, message=str(err), data={"uri": str(params.uri)})
```
Kludex, line 342: Don't we have something like MCPError.from something?

src/mcp/server/mcpserver/__init__.py
pja-ant: Reverted the export. `mcpserver/__init__.py` is back to its original `__all__`. Template handlers import from `mcp.server.mcpserver.exceptions` (same place `ToolError` lives).

src/mcp/server/mcpserver/server.py
pja-ant: only for inbound `JSONRPCError` and `ErrorData`, not from an exception.

maxisbey:
Thanks for the PR. This has since landed via #2920. I'm closing this out as part of a backlog cleanup now that v2 is out. Feel free to reopen if this is still relevant.

PR #2351, opened 2026-03-26
fix(security): remove shell=True to prevent command injection on Windows

### LINKED ISSUE #1257: Don’t use shell=True in mcp dev subprocess on Windows (command injection risk)

xr843:
## Summary

Fixes #1257 — removes `shell=True` from subprocess calls in `mcp dev` to prevent command injection on Windows.

- **`_get_npx_command()`**: Replaced `subprocess.run([cmd, "--version"], shell=True)` with `shutil.which(cmd)` to locate the npx executable without spawning a shell. This is both safer (no shell metacharacter interpretation) and more efficient (no subprocess overhead just to check existence).
- **`dev()` command**: Removed the conditional `shell=True` (`shell = sys.platform == "win32"`) from the `subprocess.run()` call that launches the MCP Inspector. Since `_get_npx_command()` already resolves the platform-specific executable name (e.g. `npx.cmd` on Windows), shell dispatch is unnecessary.
- **Tests**: Updated `test_get_npx_windows` and `test_get_npx_returns_none_when_npx_missing` to mock `shutil.which` instead of `subprocess.run`.

## Why this matters

On Windows, `shell=True` passes commands through `cmd.exe`, which interprets shell metacharacters (`&`, `|`, `%`, etc.). If any part of the command string contains these characters (e.g. a crafted file path like `server&calc.py`), arbitrary commands could be executed. The Python documentation [explicitly warns against this](https://docs.python.org/3/library/subprocess.html#security-considerations).

## Test plan

- [x] Existing tests updated to reflect new `shutil.which()` approach
- [ ] CI passes on all platforms
- [ ] Manual verification: `mcp dev server.py` still launches the inspector correctly on Windows

### DIFF
--- a/tests/cli/test_utils.py
+++ b/tests/cli/test_utils.py
@@ -91,5 +91,9 @@ def test_get_npx_windows(monkeypatch: pytest.MonkeyPatch):
 def test_get_npx_returns_none_when_npx_missing(monkeypatch: pytest.MonkeyPatch):
     """Should give None if every candidate is absent from PATH."""
     monkeypatch.setattr(sys, "platform", "win32", raising=False)
-    monkeypatch.setattr(shutil, "which", lambda cmd: None)
+
+    def fake_which(cmd: str) -> str | None:
+        return None
+
+    monkeypatch.setattr(shutil, "which", fake_which)
     assert _get_npx_command() is None


xr843:
Hi team, just a gentle ping on this PR. This fix removes the use of `shell=True` in subprocess calls, which prevents potential command injection vulnerabilities. Would appreciate a review when you get a chance. Thanks!

xr843:
Closing this as part of a cleanup — I want to focus my contributions on fewer projects rather than spreading thin. The fix is still valid if maintainers want to pick it up. Happy to reopen if there's interest!

Issue #2354, opened 2026-03-26
Design: future of `dependencies` parameter on MCPServer


maxisbey:
## Context

PR #1877 removed the `dependencies` parameter from `FastMCP` (now `MCPServer`). After investigation, we're restoring it for now because removing it without a replacement breaks high-profile users and leaves a functionality gap. This issue tracks the design decision for what to do long-term.

## What `dependencies` does

It's a list of pip/uv package names declared on the server instance:

```python
mcp = MCPServer("Screenshot Demo", dependencies=["pyautogui", "Pillow"])
```

The **only consumer** is the `mcp` CLI — `mcp dev` and `mcp install` read `server.dependencies` via `hasattr`/`getattr` and convert them to `uv run --with <pkg>` flags when launching the server or writing Claude Desktop config. See https://github.com/modelcontextprotocol/python-sdk/blob/7ba4fb881d85406f44a5af8169fb7200fa7c8e49/src/mcp/cli/cli.py#L261-L262 and https://github.com/modelcontextprotocol/python-sdk/blob/7ba4fb881d85406f44a5af8169fb7200fa7c8e49/src/mcp/cli/cli.py#L451-L453.

It has **no relationship to the MCP protocol** or the low-level `Server` class — it's purely a Python/uv packaging convenience that originated in jlowin's original FastMCP.

## Why we can't just remove it

**Real-world usage is significant** (~772 files on GitHub, ~95+ repos):

- **awslabs/mcp** (8.5k★) — 47 server files use it as their standard template
- **mindsdb/mindsdb** (38.8k★) — uses it
- **redis/mcp-redis** (461★), volcengine/mcp-server, camel-ai/camel, plus ~15–20 community servers

When jlowin's `fastmcp` package removed the same parameter, it caused 8+ breakage issues across awslabs/mcp, zotero-mcp, panther-labs, ClickHouse — all `TypeError: unexpected keyword argument 'dependencies'`.

## The replacement gap

The v2 examples were migrated to **PEP 723 inline script metadata**:

```python
# /// script
# dependencies = ["pyautogui", "Pillow"]
# ///
```

But this **does not currently work with `mcp install` / `mcp dev`**. Those commands generate `uv run --with mcp[cli] mcp run server.py` — `uv` only parses PEP 723 when the script is the direct target, not when it's an argument to another command. So the feature was removed without a working replacement.

### A. Keep `dependencies`, do nothing
[...]

PR #2358, opened 2026-03-26
feat(server): restore `dependencies` parameter on MCPServer

maxisbey:
Restores the `dependencies` parameter that was removed in #1877, pending a proper design decision on its future (tracked in #2354).

## Motivation and Context

The `dependencies` parameter is consumed by `mcp dev` and `mcp install` to auto-populate `uv run --with <pkg>` flags when launching servers or writing Claude Desktop config. Removing it without a working replacement creates two problems:

1. **Breaks high-profile users** — awslabs/mcp (8.5k★, 47 server files), mindsdb (38.8k★), redis/mcp-redis all use this parameter. When jlowin's `fastmcp` package removed the same parameter, it triggered 8+ breakage issues across the ecosystem.

2. **No working replacement** — the examples were migrated to PEP 723 inline script metadata, but `mcp install`/`mcp dev` can't read it. `uv` only parses PEP 723 when the script is the direct target; the CLI wraps it in `mcp run <file>`, so the metadata is ignored.

The CLI consumer code at [cli.py#L261](https://github.com/modelcontextprotocol/python-sdk/blob/7ba4fb881d85406f44a5af8169fb7200fa7c8e49/src/mcp/cli/cli.py#L261-L262) and [cli.py#L451](https://github.com/modelcontextprotocol/python-sdk/blob/7ba4fb881d85406f44a5af8169fb7200fa7c8e49/src/mcp/cli/cli.py#L451-L453) still checks for `server.dependencies` via `hasattr`/`getattr` — it was never removed, so this restoration makes those checks functional again.

Long-term options (deprecate → PEP 723, adopt a config file like FastMCP 2.0, or keep as-is) are laid out in #2354.

## How Has This Been Tested?

Added `test_dependencies` verifying the parameter round-trips through both `mcp.dependencies` and `mcp.settings.dependencies`, and defaults to `[]` when omitted. All 88 tests in `tests/server/mcpserver/test_server.py` pass.

## Breaking Changes

None — this restores previously-removed API.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Closes the gap left by #1877. See #2354 for the design discussion.

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -6,7 +6,7 @@ import base64
 import inspect
 import json
 import re
-from collections.abc import AsyncIterator, Awaitable, Callable, Collection, Iterable, Sequence
+from collections.abc import AsyncIterator, Awaitable, Callable, Iterable, Sequence
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
 from typing import Any, Generic, Literal, TypeVar, overload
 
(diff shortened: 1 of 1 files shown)


Kludex:
@maxisbey ???????

PR #2362, opened 2026-03-27
fix: terminate active streamable http sessions on shutdown

### LINKED ISSUE #2150: Active Streamable HTTP sessions are not terminated during shutdown

raashish1601:
## Summary
- terminate active stateful streamable HTTP transports before cancelling the session-manager task group during shutdown
- keep the manager's session registry consistent while transports are being closed
- add a regression that keeps one stateful session alive and verifies shutdown awaits transport termination

Fixes #2150.

## Testing
- python -m pytest tests/server/test_streamable_http_manager.py -k 'test_run_terminates_active_stateful_sessions_on_shutdown or test_stateful_session_cleanup_on_graceful_exit'
- python -m ruff check src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -130,11 +130,18 @@ class StreamableHTTPSessionManager:
                 yield  # Let the application run
             finally:
                 logger.info("StreamableHTTP session manager shutting down")
+                active_transports = list(self._server_instances.values())
+                self._server_instances.clear()
+
+                for transport in active_transports:
+                    try:
+                        await transport.terminate()
+                    except Exception:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2253, which also closes `_sse_stream_writers` in `terminate()` (the second half of the root cause analysis on #2150). Your regression test is worth carrying over. Thanks for the contribution.

Issue #1743, opened 2025-12-05
Extract OAuth flow logic into reusable components for proxy use cases


maxisbey:
## Summary

Refactor OAuth implementation so the flow logic and state machine are usable by server-side proxy services, not just client-side browser flows.

## Problem

The SDK's OAuth implementation is designed for local client-side flows (opening a browser locally). The business logic is embedded inside an httpx auth module, making it hard to reuse for other scenarios.

While individual helper functions have been extracted (PKCE utilities, token exchange, discovery), the core state machine that orchestrates the OAuth flow is not reusable. Proxy services that need to perform OAuth on behalf of users currently have to reimplement significant portions of the flow themselves.

## Goal

- Extract the OAuth state machine / flow orchestration into a reusable component
- Support server-side proxy patterns where OAuth happens on behalf of downstream clients
- Make it easier to correctly follow the OAuth specs without reimplementing from scratch
- Keep existing client-side flows working

## Related

- #1240 - Implement OAuth relying on Authlib

---

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

maxisbey:
I was playing around with some ideas this morning with Claude and got this draft which has _some_ interesting ideas for this. Although if there's a more common method of doing this (maybe authlib linked above? haven't looked into it yet) then we should do that.

Here's what Claude wrote:

<details><summary>Code</summary>

```python
"""
OAuth V2 Sketch - Action/Effect Pattern with Strategy Configuration

This separates:
- WHAT to do (Actions - pure data)
- HOW to execute (Adapters - httpx, fetch, etc.)
- WHEN to transition (Flow - pure state machine)
- CUSTOMIZATION (Config - strategies)
"""

from __future__ import annotations

import secrets
from dataclasses import dataclass, field
from enum import Enum, auto
from typing import Callable, Protocol
from urllib.parse import urlencode, urljoin, urlparse

import httpx
from pydantic import AnyUrl

# =============================================================================
# MODELS (existing, mostly unchanged)
# =============================================================================


@dataclass
class OAuthToken:
    access_token: str
    token_type: str = "Bearer"
    refresh_token: str | None = None
    expires_in: int | None = None
    scope: str | None = None


@dataclass
class ProtectedResourceMetadata:
    resource: str
    authorization_servers: list[str]
    scopes_supported: list[str] | None = None


@dataclass
class OAuthMetadata:
[...]

PR #2363, opened 2026-03-27
Fix stale auth context in stateful StreamableHTTP sessions

### LINKED ISSUE #2208: get_access_token() returns stale token in stateful streamable-HTTP sessions
## Summary

`get_access_token()` returns the bearer token from the **session-creating request** for the entire lifetime of a stateful streamable-HTTP session, regardless of what `Authorization` header later requests send.

## Repro

<details><summary>Open for code</summary>

```python
import multiprocessing
import socket
import time

import httpx
import pytest
import uvicorn
from starlette.middleware import Middleware
from starlette.middleware.authentication import AuthenticationMiddleware

from mcp.server import Server, ServerRequestContext
from mcp.server.auth.middleware.auth_context import AuthContextMiddleware, get_access_token
from mcp.server.auth.middleware.bearer_auth import BearerAuthBackend
from mcp.server.auth.provider import AccessToken
from mcp.types import (
    CallToolRequestParams,
    CallToolResult,
    ListToolsResult,
    PaginatedRequestParams,
    TextContent,
    Tool,
)


class _EchoTokenVerifier:
    """Accepts any bearer and echoes it back so we can tell tokens apart."""

        return AccessToken(token=token, client_id=token, scopes=[], expires_at=int(time.time()) + 3600)


async def _handle_whoami(ctx: ServerRequestContext, params: CallToolRequestParams) -> CallToolResult:
    # The user-facing contract of get_access_token(): call it from a handler,
    # get the token for the current request.
    access = get_access_token()
    text = access.token if access else "<none>"
    return CallToolResult(content=[TextContent(type="text", text=text)])


async def _handle_list_tools(ctx: ServerRequestContext, params: PaginatedRequestParams | None) -> ListToolsResult:
    return ListToolsResult(tools=[Tool(name="whoami", input_schema={"type": "object", "properties": {}})])


def _run_auth_server(port: int) -> None:
    server = Server(name="auth_test_server", on_call_tool=_handle_whoami, on_list_tools=_handle_list_tools)
    security = TransportSecuritySettings(allowed_hosts=["127.0.0.1:*"], allowed_origins=["http://127.0.0.1:*"])
    session_manager = StreamableHTTPSessionManager(app=server, security_settings=security, stateless=False)

    # Same middleware chain lowlevel Server.streamable_http_app builds when auth is on
    asgi_app = Starlette(
        routes=[Mount("/mcp", app=session_manager.handle_request)],
        middleware=[
            Middleware(AuthenticationMiddleware, backend=BearerAuthBackend(_EchoTokenVerifier())),
            Middleware(AuthContextMiddleware),
        ],
        lifespan=lambda app: session_manager.run(),
    )
    uvicorn.run(asgi_app, host="127.0.0.1", port=port, log_level="error")


class _MutableBearerAuth(httpx.Auth):
    """Reads the bearer from a mutable attribute at send-time so we can swap mid-session."""

    def __init__(self, token: str) -> None:
        self.token = token

    def auth_flow(self, request: httpx.Request):
        request.headers["Authorization"] = f"Bearer {self.token}"
        yield request


@pytest.mark.anyio
async def test_get_access_token_reflects_current_request_in_stateful_session() -> None:
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))
        port = s.getsockname()[1]

    proc = multiprocessing.Process(target=_run_auth_server, args=(port,), daemon=True)
    proc.start()
    try:
        # wait for server
        for _ in range(200):
            try:
                with socket.socket() as s:
                    s.connect(("127.0.0.1", port))
                break
            except OSError:
                time.sleep(0.01)

        url = f"http://127.0.0.1:{port}/mcp"
        auth = _MutableBearerAuth("token-A")

        async with httpx.AsyncClient(auth=auth, timeout=httpx.Timeout(30, read=30), follow_redirects=True) as http_client:
            async with streamable_http_client(url, http_client=http_client) as (read_stream, write_stream):
                    await session.initialize()

                    # Request 1: session created, _receive_loop spawns with token-A in its context
                    r1 = await session.call_tool("whoami", {})
                    assert isinstance(r1.content[0], TextContent)
                    assert r1.content[0].text == "token-A"

                    # Request 2: same session, different bearer — reuses the existing _receive_loop
                    auth.token = "token-B"
                    r2 = await session.call_tool("whoami", {})
                    assert isinstance(r2.content[0], TextContent)
                    # EXPECTED: "token-B" — handler should see the token sent with THIS request
                    # ACTUAL:   "token-A" — handler sees the session-creating request's token
                    assert r2.content[0].text == "token-B"
    finally:
        proc.kill()
        proc.join(timeout=2)
```

**Result on `main`:**

```
AssertionError: assert 'token-A' == 'token-B'
  - token-B
  ?       ^
  + token-A
  ?       ^
```

</details>

## Root cause

`AuthContextMiddleware` sets `auth_context_var` inside the ASGI request's task. But the tool handler doesn't run in that task — it runs in a task spawned by `Server.run()`'s `tg.start_soon(_handle_message, ...)`, which is itself inside `run_server`, which was spawned at session creation: https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/src/mcp/server/streamable_http_manager.py#L267

`tg.start()` copies the caller's `contextvars.Context` at call time. So `run_server` (and every task it spawns) carries a snapshot from request 1. Requests 2..N write to the transport's read stream from the new ASGI task, but the reader is `_receive_loop` — still running with the request-1 snapshot. The ContextVar set in request N's ASGI task never reaches the handler.

```
ASGI req 1 (auth_context_var=A)
  └─ tg.start(run_server)                           ← context copied: A
       └─ ServerSession.__aenter__
            └─ tg.start_soon(_receive_loop)         ← inherits A
       └─ async for msg in session.incoming_messages:
            └─ tg.start_soon(_handle_message)       ← inherits A
                 └─ tool handler: get_access_token() → A  ✓

ASGI req 2 (auth_context_var=B)
  └─ transport.handle_request(...)                  ← writes to read_stream
       ... _receive_loop (still ctx A) reads it ...
       ... tg.start_soon(_handle_message)           ← inherits A, not B
            └─ tool handler: get_access_token() → A  ✗
```

The existing unit test passes because `MockApp` runs inline in the same task as the middleware — no stream crossing: https://github.com/modelcontextprotocol/python-sdk/blob/528abfab86f1f3c003bd7d54a1f0bbd65d81c59c/tests/server/auth/middleware/test_auth_context.py#L33

## Impact

- **Token refresh mid-session**: refreshed token is invisible to handlers.
- **Compounds #2100**: if a session ID is hijacked, the attacker's requests execute with the victim's token visible to `get_access_token()`.
- **Stateless HTTP is unaffected** — new session per request means a fresh context snapshot each time.

## The correct path already exists

The Starlette `Request` is threaded explicitly through `ServerMessageMetadata.request_context` → `ServerRequestContext.request`. Inside a handler:

```python
request: Request = ctx.request_context.request
user = request.user                       # set by AuthenticationMiddleware
token = user.access_token                 # per-request, correct
```

## Proposed fix

Given `get_access_token()` has no callers in `src/` or `examples/` and isn't documented: remove `auth_context_var`, `get_access_token()`, and `AuthContextMiddleware`. Expose auth on `Context` as part of #2098 using the explicit `request` threading above.

Alternative (if we want to keep the API): thread the `AuthenticatedUser` alongside `request` on `ServerMessageMetadata` and set the contextvar at the `tg.start_soon` site in `Server.run()`. But that's re-inventing what `request.user` already provides.

## Related

- #2098 — expose session/auth/transport on `Context`
- #2100 — bind authenticated identity to sessions
- #1743 — auth rework
- #1969 — general user-ContextVar propagation through streams (different scope: arbitrary user vars, not just auth)

raashish1601:
## Summary
- rebind the auth ContextVar from the current transport request before low-level server dispatch
- add a stateful StreamableHTTP regression that switches from token-A to token-B within one session
- keep the existing request-context propagation behavior covered while fixing get_access_token() for later requests
Closes #2208.
## Testing
- .\\.venv\\Scripts\\python -m pytest tests\\server\\auth\\middleware\\test_auth_context.py tests\\server\\auth\\middleware\\test_auth_context_streamable_http.py -q
- .\\.venv\\Scripts\\python -m pytest tests\\shared\\test_streamable_http.py -q -k "request_context_propagation or request_context_isolation"

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -11,7 +11,7 @@ from mcp import types
 from mcp.client.experimental import ExperimentalClientFeatures
 from mcp.client.experimental.task_handlers import ExperimentalTaskHandlers
 from mcp.shared._context import RequestContext
-from mcp.shared.message import SessionMessage
+from mcp.shared.message import MessageMetadata, SessionMessage
 from mcp.shared.session import BaseSession, ProgressFnT, RequestResponder
 from mcp.shared.version import SUPPORTED_PROTOCOL_VERSIONS
 from mcp.types._types import RequestParamsMeta
@@ -461,6 +461,7 @@ class ClientSession(
     async def _handle_incoming(
         self,
         req: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
+        message_metadata: MessageMetadata = None,
     ) -> None:
         """Handle incoming messages by forwarding to the message handler."""
         await self._message_handler(req)
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -66,8 +66,8 @@ from mcp.server.streamable_http import EventStore
 from mcp.server.streamable_http_manager import StreamableHTTPASGIApp, StreamableHTTPSessionManager
 from mcp.server.transport_security import TransportSecuritySettings
 from mcp.shared.exceptions import MCPError
-from mcp.shared.message import ServerMessageMetadata, SessionMessage
-from mcp.shared.session import RequestResponder
+from mcp.shared.message import MessageMetadata, ServerMessageMetadata, SessionMessage
+from mcp.shared.session import NotificationWithMetadata, RequestResponder
 
 logger = logging.getLogger(__name__)
 
@@ -424,7 +424,9 @@ class Server(Generic[LifespanResultT]):
 
     async def _handle_message(
         self,
-        message: RequestResponder[types.ClientRequest, types.ServerResult] | types.ClientNotification | Exception,
+        message: RequestResponder[types.ClientRequest, types.ServerResult]
+        | NotificationWithMetadata[types.ClientNotification]
+        | Exception,
         session: ServerSession,
         lifespan_context: LifespanResultT,
         raise_exceptions: bool = False,
@@ -436,6 +438,13 @@ class Server(Generic[LifespanResultT]):
                         await self._handle_request(
                             message, responder.request, session, lifespan_context, raise_exceptions
                         )
+                case NotificationWithMetadata() as notification:
+                    await self._handle_notification(
+                        notification.notification,
+                        session,
+                        lifespan_context,
+                        notification.message_metadata,
+                    )
                 case Exception():
                     logger.error(f"Received exception from stream: {message}")
                     if raise_exceptions:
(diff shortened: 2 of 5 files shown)


maxisbey:
Closing — this overlaps with the broader auth rework planned in #1743 and currently has merge conflicts and a large CI failure count. #1998 covers the surgical v1.x fix. Thanks for the contribution.

PR #2366, opened 2026-03-28
fix: align Context logging methods with MCP spec data type

paikend:
## Summary
- Change `message: str` → `data: Any` in `Context.log()` and convenience methods (`debug`, `info`, `warning`, `error`)
- Remove the `extra` parameter — callers can now pass structured data directly as `data`
- Aligns with `ServerSession.send_log_message(data: Any)` which already accepts `Any`

## Motivation
The [MCP specification](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L833-L836) defines log data as `unknown` (any JSON-serializable type), but `Context.info()` and friends only accepted `str`.

**Before:**
```python
# Type error — spec allows this but SDK didn't
await ctx.info({"key": "value", "count": 42})

# Workaround via extra parameter
await ctx.log(level="info", message="msg", extra={"key": "value"})
```

**After:**
```python
# Now works — any JSON-serializable type
await ctx.info({"key": "value", "count": 42})
await ctx.info("plain string still works")
await ctx.info([1, 2, 3])
```

## Breaking Change
- `message` parameter renamed to `data` in `log()`, `debug()`, `info()`, `warning()`, `error()`
- `extra` parameter removed from `log()` and convenience methods

Callers using positional args (`ctx.info("msg")`) are unaffected. Callers using `message=` keyword or `extra=` need to update.

## Test plan
- [x] All 1143 existing tests pass (0 failures)
- [x] Updated `test_logging_callback.py` to use new API
- [x] `test_context_logging` in `test_server.py` passes (string data still works)

Fixes #397

### DIFF
--- a/README.v2.md
+++ b/README.v2.md
@@ -681,11 +681,11 @@ The Context object provides the following capabilities:
 - `ctx.mcp_server` - Access to the MCPServer server instance (see [MCPServer Properties](#mcpserver-properties))
 - `ctx.session` - Access to the underlying session for advanced communication (see [Session Properties and Methods](#session-properties-and-methods))
 - `ctx.request_context` - Access to request-specific data and lifespan resources (see [Request Context Properties](#request-context-properties))
-- `await ctx.debug(message)` - Send debug log message
-- `await ctx.info(message)` - Send info log message
-- `await ctx.warning(message)` - Send warning log message
-- `await ctx.error(message)` - Send error log message
-- `await ctx.log(level, message, logger_name=None)` - Send log with custom level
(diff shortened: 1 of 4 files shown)


maxisbey (approved):
Thanks for the PR! I added changed the log level to use the actual literal type with all the levels and added migration docs. Other than that looked good to me. Approved :)

Kludex:
I'd like to point out that the SDK doesn't need to have the same shape as the spec. The SDK should do what makes more sense for the language while staying compliant.

PR #2367, opened 2026-03-28
fix: allow Any type for logging message parameter

mac-agent:
## Summary

Change the type annotation of the `message` parameter in `Context.log()`, `Context.debug()`, `Context.info()`, `Context.warning()`, and `Context.error()` from `str` to `Any`.

## Problem

The MCP spec defines the log data field as:
```
data: unknown — "Any JSON serializable type is allowed here"
```

But the Python SDK types the message as `str`, preventing rich structured logging like:
```python
ctx.info({"event": "progress", "percent": 50})
```

## Solution

- Change `message: str` to `message: Any` in all 5 logging methods
- Update docstring to document the expanded type
- Keep parameter name as `message` for backward compatibility
- No behavioral changes — existing string calls continue to work identically

## Changes

- `src/mcp/server/mcpserver/context.py`: Type annotation changes only

Fixes #397


Kludex:

PR #2368, opened 2026-03-28
fix: handle None required_scopes in validate_scope

IgnazioDS:
## Summary
- Add early return in `validate_scope()` when `self.scope` is `None`
- `None` means "no scope restrictions" per the OAuth spec, so all client-requested scopes should be accepted
- Previously, `None` was treated as an empty allowed set, rejecting all scopes with `InvalidScopeError`

## Problem
When an MCP server does not specify required scopes (`None`), clients cannot authenticate because all requested scopes are rejected. See #2216.

## Test plan
- [x] Verify `validate_scope("read write admin")` accepts all scopes when `scope=None`
- [x] Verify `validate_scope("read write")` still rejects unauthorized scopes when `scope="read"`
- [x] Verify `validate_scope(None)` returns `None`
- [x] Verify `validate_scope("read write")` accepts when `scope="read write"`

Fixes #2216

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -71,11 +71,14 @@ class OAuthClientMetadata(BaseModel):
         if requested_scope is None:
             return None
         requested_scopes = requested_scope.split(" ")
-        allowed_scopes = [] if self.scope is None else self.scope.split(" ")
+        if self.scope is None:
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing — superseded by #2301, which removes the registration-scope check entirely. Thanks for the contribution.

PR #2369, opened 2026-03-28
Add stdio regression test for raw invalid UTF-8 tool arguments

raashish1601:
## Summary
- add an end-to-end stdio regression for issue #2328 using raw invalid UTF-8 bytes in a `tools/call` request
- assert the server returns a JSON-RPC error response instead of crashing or dropping the reply on stdin EOF
- exercise the existing stdio transport plus low-level server path without external server packages

## Testing
- `PYTHONPATH=src python -m pytest -o addopts='' tests/issues/test_2328_stdio_invalid_utf8.py -q`
- `PYTHONPATH=src python -m pytest -o addopts='' tests/server/test_stdio.py tests/issues/test_2328_stdio_invalid_utf8.py -q`
- `python -m ruff check tests/issues/test_2328_stdio_invalid_utf8.py`
- `git diff --check`

### DIFF
--- a/tests/issues/test_2328_stdio_invalid_utf8.py
+++ b/tests/issues/test_2328_stdio_invalid_utf8.py
@@ -1,0 +1,72 @@
+"""Regression test for issue #2328: raw invalid UTF-8 over stdio must not crash the server."""
+
+import io
+from io import TextIOWrapper
+
+import anyio
+import pytest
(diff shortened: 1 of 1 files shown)


maxisbey:
Thanks for the PR. This has since landed via #2306. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #2373, opened 2026-03-28
fix(auth): respect explicitly-set client_metadata.scope during discovery

### LINKED ISSUE #2317: Don't override client_metadata.scopes if they are already set

chasewhughes:
## Problem

`async_auth_flow` unconditionally overwrites `client_metadata.scope` with server-advertised scopes during OAuth discovery (Step 3). This prevents clients from limiting their requested permissions, which is required by some servers — for example, the official SalesForce MCP server rejects requests that include scopes not authorized for the given client application.

## Solution

Only apply automatic scope selection when `client_metadata.scope is None`. If the caller has explicitly set scopes, they are preserved through the discovery flow.

The 403 step-up flow (`insufficient_scope`) is intentionally left unchanged — when a server explicitly demands different scopes, those should override regardless.

## Tests

Two new tests:
- `test_auth_flow_preserves_explicit_scopes` — verifies that explicitly-set scopes survive the discovery flow
- `test_auth_flow_auto_selects_scopes_when_none` — verifies that auto-selection still works when no explicit scope is set

All 86 tests in `test_auth.py` pass.

Fixes #2317

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -572,11 +572,14 @@ class OAuthClientProvider(httpx.Auth):
                             logger.debug(f"OAuth metadata discovery failed: {url}")
 
                     # Step 3: Apply scope selection strategy
-                    self.context.client_metadata.scope = get_client_metadata_scopes(
-                        extract_scope_from_www_auth(response),
-                        self.context.protected_resource_metadata,
-                        self.context.oauth_metadata,
-                    )
+                    # Respect explicitly-set scopes; only auto-select
+                    # when the caller hasn't specified any.
+                    if self.context.client_metadata.scope is None:
+                        self.context.client_metadata.scope = get_client_metadata_scopes(
+                            extract_scope_from_www_auth(response),
+                            self.context.protected_resource_metadata,
+                            self.context.oauth_metadata,
+                        )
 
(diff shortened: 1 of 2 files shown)


maxisbey:

PR #2374, opened 2026-03-29
fix(session): log exceptions in default message_handler instead of silently swallowing

Aboudjem:
## Summary

Fixes #1401.

The default `_default_message_handler` in `ClientSession` silently discarded all messages — including exceptions. When a transport error (e.g. SSE read timeout) was delivered through the read stream, the handler did nothing but `await anyio.lowlevel.checkpoint()`, making it impossible to diagnose why requests were hanging.

This change adds a `logger.warning()` call with full exception traceback (`exc_info`) when the default handler receives an `Exception`. This provides immediate observability for transport errors without breaking backward compatibility — callers who want custom behavior (re-raising, suppression, etc.) can still pass their own `message_handler`.

### What changed

- `_default_message_handler` now logs a warning with the exception traceback when it receives an `Exception`, instead of silently discarding it
- Added test verifying the default handler logs exceptions with proper exc_info
- Added test verifying custom handlers can still suppress exceptions if desired

### Why not re-raise?

Re-raising in the default handler would terminate the receive loop and send `CONNECTION_CLOSED` errors to all pending requests. While this prevents hangs, it also affects non-critical exceptions routed through `_handle_incoming` (e.g. responses with unknown IDs). Logging provides observability without this side effect. Transport-level error propagation to response streams is addressed separately in #2122.

## Test plan

- [x] New test: `test_default_message_handler_logs_exceptions` — verifies the exception is logged at WARNING level with correct exc_info
- [x] New test: `test_custom_message_handler_can_suppress_exceptions` — verifies custom handlers can still suppress exceptions
- [x] All 13 client session tests pass
- [x] All 8 shared session tests pass
- [x] `ruff check` clean
- [x] `pyright` clean

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -57,6 +57,8 @@ class MessageHandlerFnT(Protocol):
 async def _default_message_handler(
     message: RequestResponder[types.ServerRequest, types.ClientResult] | types.ServerNotification | Exception,
 ) -> None:
+    if isinstance(message, Exception):
+        logger.warning("Unhandled exception in message handler", exc_info=message)
     await anyio.lowlevel.checkpoint()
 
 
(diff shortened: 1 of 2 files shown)


km-anthropic:
@claude review

maxisbey:
Thanks for the PR. Tracking this in #3087 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

PR #2377, opened 2026-03-29
fix: allow custom Content-Type header in StreamableHTTPTransport

### LINKED ISSUE #2375: Harcoded Content-Type in the client cannot be overriden
### Initial Checks

### Description

The StreamableHTTPTransport class used hardcodes the Content-Type to application/json which prevents overriding it to include, for example, custom charsets

The faulty code is on this line https://github.com/modelcontextprotocol/python-sdk/blob/fb2276b95fb5ac6631f9c160c9b1bd7a6d7312a9/src/mcp/client/streamable_http.py#L93

### Example Code

```Python
StreamableHTTPConnectionParams(
            url="....",
            headers={"Content-Type": "application/json; charset=utf-8"},
        ),

```

### Python & MCP Python SDK

```Text
1.26.0
```

I can try to provide a PR but want to confirm first if this is considered an issue by the maintainers or they want to have strict control over the Content-Type header

guoyangzhen:
Fixes #2375

## Problem

`StreamableHTTPTransport._prepare_headers()` hardcodes `Content-Type: application/json`, which overwrites any user-configured header (e.g. `application/json; charset=utf-8`). Users who pass custom headers via `httpx.AsyncClient(headers=...)` find them ignored because request-level headers take precedence over client defaults in httpx.

## Solution

Add an optional `headers` parameter to both `StreamableHTTPTransport.__init__` and `streamable_http_client`. User-supplied headers are merged **after** the MCP defaults, so they take precedence. This is backward-compatible — existing code with no `headers` argument behaves identically.

## Usage

```python
async with streamable_http_client(
    url,
    headers={'Content-Type': 'application/json; charset=utf-8'},
) as (read, write):
    ...
```

## Changes

- `StreamableHTTPTransport.__init__`: new `headers` dict param, stored as `self._user_headers`
- `_prepare_headers()`: calls `headers.update(self._user_headers)` after setting defaults
- `streamable_http_client()`: new `headers` kwarg, forwarded to transport

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -72,15 +72,18 @@ class RequestContext:
 class StreamableHTTPTransport:
     """StreamableHTTP client transport implementation."""
 
-    def __init__(self, url: str) -> None:
+    def __init__(self, url: str, headers: dict[str, str] | None = None) -> None:
         """Initialize the StreamableHTTP transport.
 
         Args:
             url: The endpoint URL.
+            headers: Optional extra headers included with every request.
+                User-supplied headers take precedence over built-in MCP defaults.
         """
         self.url = url
         self.session_id: str | None = None
         self.protocol_version: str | None = None
+        self._user_headers: dict[str, str] = headers or {}
 
     def _prepare_headers(self) -> dict[str, str]:
         """Build MCP-specific request headers.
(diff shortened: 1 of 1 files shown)


km-anthropic:
@claude review

dsabana-clgx:
There is a bigger issue of not allowing any custom headers/values at all. Maybe we need to enable forwarding custom headers. This will cover your use case as well.

maxisbey:
The linked #2375 was closed as not planned — `application/json` is UTF-8 by default per the JSON RFC, so a Content-Type override isn't something the client transport takes on; other headers already pass through a caller-supplied `http_client`. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

Issue #1839, opened 2026-01-07
fastmcp tool func block asyncio


archever:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

unlike fastapi, fastmcp run sync tool func directly. this may block async loop if the func uses block io like requests 

fastmcp
```python
class FuncMetadata(BaseModel):
    ...
    async def call_fn_with_arg_validation(
        self,
        fn: Callable[..., Any | Awaitable[Any]],
        fn_is_async: bool,
        arguments_to_validate: dict[str, Any],
        arguments_to_pass_directly: dict[str, Any] | None,
    ) -> Any:
        """Call the given function with arguments validated and injected.

        Arguments are first attempted to be parsed from JSON, then validated against
        the argument model, before being passed to the function.
        """
        arguments_pre_parsed = self.pre_parse_json(arguments_to_validate)
        arguments_parsed_model = self.arg_model.model_validate(arguments_pre_parsed)
        arguments_parsed_dict = arguments_parsed_model.model_dump_one_level()

        arguments_parsed_dict |= arguments_to_pass_directly or {}
        if fn_is_async:
            return await fn(**arguments_parsed_dict)
        else:
            return fn(**arguments_parsed_dict)
```

fastapi
```python
async def run_endpoint_function(
    *, dependant: Dependant, values: dict[str, Any], is_coroutine: bool
) -> Any:
    # Only called by get_request_handler. Has been split into its own function to
    # facilitate profiling endpoints, since inner functions are harder to profile.
    assert dependant.call is not None, "dependant.call must be a function"

    if is_coroutine:
        return await dependant.call(**values)
    else:
        return await run_in_threadpool(dependant.call, **values)
```

### Example Code

```Python
import time
from mcp.server.fastmcp import FastMCP

mcp = FastMCP("Demo", json_response=True, stateless_http=True)


@mcp.tool()
def wait() -> str:
    """Add two numbers"""
    time.sleep(10)
    return "ok"


mcp.run()

# call wait tool, any call will be blocked until the sleep is done
```

### Python & MCP Python SDK

```Text
python=3.12
mcp=latest
```

archever:
related issue https://github.com/microsoft/markitdown/issues/1517

jlowin:
[...]

Issue #1646, opened 2025-11-20
Why synchronous @mcp.tool definition is not supported for parallelism?


MikeBai523:
### Question

In `mcp.server.fastmcp.utilities.func_metadata.FuncMetadata.call_fn_with_arg_validation`

```python
if fn_is_async:
    return await fn(**arguments_parsed_dict)
else:
    return fn(**arguments_parsed_dict)
```

In FastMCP, if a tool is defined as a synchronous function, it will block the whole running loop instead of using to_thread().

May I know the consideration for it's different from FastAPI's concurrency model?

Kludex:
This is a bug.

It's not "FastAPI concurrency model".

maxisbey:
Fixed on `main` (v2): #1909 routes sync `@mcp.tool` functions through `anyio.to_thread.run_sync` so they no longer block the event loop, and #2380 extends the same to sync resource and prompt handlers. Closing — if you need this backported to the v1.x line, please say so.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

PR #2380, opened 2026-03-31
Don't block the event loop on sync resource and prompt functions

maxisbey:
Extends the #1909 fix to resources and prompts — sync `@mcp.resource` and `@mcp.prompt` handlers now run in worker threads instead of blocking the event loop.

## Motivation and Context

#1909 fixed tools by routing sync functions through `anyio.to_thread.run_sync`, but the same blocking pattern existed in three other places:

- `FunctionResource.read()`
- `ResourceTemplate.create_resource()`
- `Prompt.render()`

All three called `self.fn(...)` directly and checked `inspect.iscoroutine(result)` afterward. A blocking sync handler (file I/O, HTTP request, CPU-bound work) would freeze the entire event loop.

This applies the same fix: check `inspect.iscoroutinefunction(self.fn)` up front and dispatch sync functions to a worker thread.

Related: #1646, #1839

## How Has This Been Tested?

Added thread-identity regression tests for each of the three call sites. Each test captures `threading.get_ident()` inside a sync handler and asserts it differs from the event loop's thread.

Verified that `pydantic.validate_call` (which wraps the stored `self.fn` in templates and prompts) preserves async-ness — `inspect.iscoroutinefunction(validate_call(async_fn))` returns `True`, so the dispatch check works correctly on the wrapped function.

All 39 tests in the affected test files pass; pyright, ruff, and `strict-no-cover` are clean.

## Breaking Changes

None. Sync handlers that were previously starving the event loop now run concurrently.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The previous implementation happened to support callable objects with async `__call__` (by checking `iscoroutine(result)` after calling). That edge case is not preserved here — matching the approach taken in #1909, which relies on `_is_async_callable` being evaluated at registration time for tools. Resources and prompts have no equivalent pre-computed field; if that edge case matters, it's worth a separate discussion.

### DIFF
--- a/src/mcp/server/mcpserver/prompts/base.py
+++ b/src/mcp/server/mcpserver/prompts/base.py
@@ -2,10 +2,12 @@
 
 from __future__ import annotations
 
+import functools
 import inspect
 from collections.abc import Awaitable, Callable, Sequence
 from typing import TYPE_CHECKING, Any, Literal
 
+import anyio.to_thread
 import pydantic_core
 from pydantic import BaseModel, Field, TypeAdapter, validate_call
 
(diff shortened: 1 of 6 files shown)


Kludex (approved):
I'm not sure those tests are useful. I would prefer to not have tests in this case.

PR #2383, opened 2026-03-31
fix(streamable_http): allow Content-Type header override

### LINKED ISSUE #2375: Harcoded Content-Type in the client cannot be overriden

haroldfabla2-hue:
## Summary

Allow  and  to accept a custom  parameter instead of hardcoding `application/json`. This enables clients to override the Content-Type header to include custom charsets (e.g. `application/json; charset=utf-8`) or other attributes required by specific server implementations.

## Changes

- ****: Added  parameter and stored it as 
- ****: Changed hardcoded `"content-type": "application/json"` to use `self.content_type`
- ****: Added  parameter, passed through to 
- **Added test file**  with 3 test cases

## Fixes

Fixes #2375

## Example usage

```python
# Before (hardcoded, no way to override):
transport = StreamableHTTPTransport(url)

# After (custom charset supported):
transport = StreamableHTTPTransport(
    url,
    content_type="application/json; charset=utf-8",
)

# Or via the client helper:
async with streamable_http_client(
    url,
    content_type="application/json; charset=utf-8",
) as (read, write):
    ...
```

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -72,13 +72,17 @@ class RequestContext:
 class StreamableHTTPTransport:
     """StreamableHTTP client transport implementation."""
 
-    def __init__(self, url: str) -> None:
+    def __init__(self, url: str, content_type: str = "application/json") -> None:
         """Initialize the StreamableHTTP transport.
 
         Args:
             url: The endpoint URL.
+            content_type: The Content-Type header value for POST requests.
+                Defaults to "application/json". Can be overridden to include
+                custom charsets (e.g. "application/json; charset=utf-8").
         """
         self.url = url
+        self.content_type = content_type
         self.session_id: str | None = None
         self.protocol_version: str | None = None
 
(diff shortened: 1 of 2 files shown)


Kludex:
`utf-8` is implied on `"application/json"`.

See https://github.com/Kludex/starlette/issues/2342 and https://github.com/Kludex/starlette/discussions/1791 for more information.

---

Thanks for the PR.

PR #2391, opened 2026-04-04
Fix #1933: Prevent closing real stdio when server exits

717986230:
Fixes #1933

## Problem
When using `transport="stdio"`, the server closes `sys.stdin.buffer` and `sys.stdout.buffer` when exiting, causing subsequent stdio operations to fail with:
```
ValueError: I/O operation on closed file.
```

## Root Cause
The `stdio_server()` function wraps `sys.stdin.buffer` and `sys.stdout.buffer` directly. When these wrappers are closed (when the context manager exits), they also close the original system streams.

## Solution
Use `os.dup()` to create duplicate file descriptors before wrapping:
- Duplicate `sys.stdin.fileno()` and `sys.stdout.fileno()`
- Open new file handles from the duplicated descriptors
- Wrap these duplicated handles instead of the originals

This ensures the original stdin/stdout remain open when the wrapper streams are closed.

## Testing
Tested with the reproduction case from the issue:
```python

mcp = FastMCP("Demo")
mcp.run(transport="stdio")
print("?")  # This now works without error
```

Pressing Ctrl+D to exit the server no longer causes "I/O operation on closed file" error.

---
*This fix implements the proposed solution from the original issue.*

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -30,7 +30,10 @@ from mcp.shared.message import SessionMessage
 
 
 @asynccontextmanager
-async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.AsyncFile[str] | None = None):
+async def stdio_server(
+    stdin: anyio.AsyncFile[str] | None = None,
+    stdout: anyio.AsyncFile[str] | None = None,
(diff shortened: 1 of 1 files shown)


maxisbey:

PR #2394, opened 2026-04-05
feat: implement OTEL mcp.server.* metrics

verdie-g:
Implements the two server-side OTel metrics defined in the [MCP semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp), as discussed in #421.

`network.transport`, `network.protocol.name`, and `network.protocol.version` are not set. `ServerSession` does not currently know the transport type, adding it would require a new constructor parameter and a corresponding public API change. This can be addressed in a follow-up once the transport kind is plumbed through.


pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "pyjwt[crypto]>=2.10.1",
     "typing-extensions>=4.13.0",
     "typing-inspection>=0.4.1",
-    "opentelemetry-api>=1.28.0",
+    "opentelemetry-api>=1.30.0",
 ]
```
verdie-g, line 44: Bumped from 1.28.0: `explicit_bucket_boundaries_advisory` was added to `Meter.create_histogram()` in opentelemetry-api 1.30.0. The previous minimum caused a `TypeError` at runtime when the OTel proxy meter replayed histogram creation against the real meter.

pyproject.toml
```diff
@@ -72,7 +72,7 @@ dev = [
     "coverage[toml]>=7.10.7,<=7.13",
     "pillow>=12.0",
     "strict-no-cover",
-    "logfire>=3.0.0",
+    "logfire>=3.20.0",
 ]
```
verdie-g, line 76: Bumped from 3.0.0: logfire's `_ProxyMeter.create_histogram()` did not forward unknown kwargs (including `explicit_bucket_boundaries_advisory`) until 3.20.0, causing a `TypeError` when logfire was configured as the meter provider. Also replaced `capfire.get_collected_metrics()` (added in logfire 4.0.0) with `capfire.metrics_reader.get_metrics_data()` which works from 3.0.0, see `_get_mcp_metrics` in `test_otel.py`.

tests/shared/test_otel.py
```diff
@@ -1,20 +1,35 @@
...
+from mcp.server.context import ServerRequestContext
+from mcp.server.lowlevel.server import Server
 from mcp.server.mcpserver import MCPServer
+from mcp.shared.exceptions import MCPError
 
 pytestmark = pytest.mark.anyio
 
 
+def _get_mcp_metrics(capfire: CaptureLogfire) -> dict[str, Any]:
+    """Return collected metrics whose name starts with 'mcp.', keyed by name."""
```
verdie-g, line 21: `CaptureLogfire.get_collected_metrics()` was only added in logfire 4.0.0, but our minimum is 3.20.0. This helper replicates the same logic using `capfire.metrics_reader.get_metrics_data()`, which is available from logfire 3.0.0 (the `metrics_reader` field is part of the public `CaptureLogfire` dataclass).

maxisbey:
Thanks for the PR. The v2 rework changed the code this targets, so it no longer applies as-is. Closing this as part of a wider backlog cleanup following the v2 launch. If it's still relevant, feel free to reopen.

PR #2395, opened 2026-04-05
Fix infinite reconnection loop in StreamableHTTP client

### LINKED ISSUE #2393: StreamableHTTP client: _handle_reconnection resets attempt counter to 0, causing infinite retry loop
## Bug Description

`_handle_reconnection()` in `streamable_http.py` resets the `attempt` counter to `0` on line 494 when a reconnection succeeds but the stream ends without delivering a complete response. This makes `MAX_RECONNECTION_ATTEMPTS` ineffective — the counter only applies to consecutive exceptions, not total reconnection attempts. If the server accepts the connection but the stream drops repeatedly, the client retries forever.

## Reproduction

**MCP version:** 1.26.0 (also confirmed unpatched in 1.27.0 `main` branch)

### SSCE — Minimal reproducer

**Server** (`server.py`): A server that accepts SSE connections but closes them before sending a complete response, with a `last-event-id` header to trigger the reconnection path.

```python
"""Minimal MCP server that drops SSE streams to trigger infinite reconnect."""
import asyncio
from starlette.routing import Route
from starlette.responses import Response
from sse_starlette.sse import EventSourceResponse
import uvicorn, json, uuid

session_id = str(uuid.uuid4())

async def handle_mcp(request):
    if request.method == "POST":
        body = await request.json()
        method = body.get("method")

        if method == "initialize":
            resp = {
                "jsonrpc": "2.0",
                "id": body["id"],
                "result": {
                    "protocolVersion": "2025-06-18",
                    "capabilities": {"tools": {}},
                    "serverInfo": {"name": "drop-server", "version": "0.1.0"},
                },
            }
            return Response(
                json.dumps(resp),
                headers={"mcp-session-id": session_id},
            )

        if method == "notifications/initialized":
            return Response(status_code=202, headers={"mcp-session-id": session_id})

        if method == "tools/call":
            # Return SSE that sends a priming event with an ID, then drops
            async def event_generator():
                yield {"event": "message", "id": "evt-1", "data": ""}
                # Close without sending the actual response
                return

            return EventSourceResponse(
                event_generator(),
                headers={"mcp-session-id": session_id},
            )

        return Response(status_code=404)

    if request.method == "GET":
        # GET stream for server-initiated messages — also drop immediately
        async def get_stream():
            yield {"event": "message", "id": "evt-get-1", "data": ""}
            return

        return EventSourceResponse(
            get_stream(),
            headers={"mcp-session-id": session_id},
        )

    if request.method == "DELETE":
        return Response(status_code=200)

app = Starlette(routes=[Route("/mcp", handle_mcp, methods=["POST", "GET", "DELETE"])])

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=9999)
```

**Client** (`client.py`): Connects and calls a tool, demonstrating the infinite loop.

```python
"""Client that demonstrates infinite reconnection loop."""
import asyncio, logging
logging.basicConfig(level=logging.INFO)

from mcp import ClientSession

async def main():
    async with streamable_http_client("http://localhost:9999/mcp") as (read, write, _):
            await session.initialize()
            print("Initialized. Calling tool (will hang forever)...")
            # This call will trigger _handle_reconnection with attempt=0 reset
            try:
                result = await asyncio.wait_for(
                    session.call_tool("any_tool", {}),
                    timeout=30,
                )
            except asyncio.TimeoutError:
                print("CONFIRMED: call_tool hung for 30s (infinite reconnect loop)")

asyncio.run(main())
```

### Steps

1. `pip install mcp[cli] sse-starlette uvicorn`
2. Run server: `python server.py`
3. Run client: `python client.py`
4. Observe: client logs show repeated `"GET stream disconnected, reconnecting in 1000ms..."` messages. The `call_tool` never returns. After 30s the `wait_for` timeout fires, confirming the hang.

## Root Cause

In `streamable_http.py`, `_handle_reconnection` (line 437):

```python
async def _handle_reconnection(self, ctx, last_event_id, retry_interval_ms, attempt=0):
    if attempt >= MAX_RECONNECTION_ATTEMPTS:  # Only 2
        return

    # ... reconnects, iterates SSE ...

    # Line 494: Stream ended without response — resets attempt to 0!
```

When the reconnection *succeeds* (HTTP 200) but the stream ends without a complete `JSONRPCResponse`, line 494 recurses with `attempt=0`, restarting the counter. Only the exception path (line 498) increments the counter. A server that accepts connections but drops streams causes infinite recursion at 1-second intervals.

## Expected Behavior

After `MAX_RECONNECTION_ATTEMPTS` total reconnection attempts (regardless of whether they succeeded at the HTTP level), the client should give up and propagate an error to the caller.

## Suggested Fix

Track total attempts across the recursion rather than resetting on successful connect:

```python
# Line 494: increment instead of reset
await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, attempt + 1)
```

Or add a separate `max_total_reconnection_attempts` counter that is never reset.

## Impact

In production, this causes MCP client coroutines to hang forever when a server experiences transient stream drops. The calling application has no way to recover without wrapping every MCP call in `asyncio.wait_for()`. We discovered this when an agentquant research job hung for 5+ hours in a reconnection loop after an MCP server's SSE stream dropped.

Dharit13:
## Summary

Fixes #2393.

`_handle_reconnection()` reset the attempt counter to `0` when the SSE stream ended without delivering a complete response (only priming events). This made `MAX_RECONNECTION_ATTEMPTS` ineffective — a server that accepts connections but always drops streams caused the client to retry forever, hanging the caller indefinitely.

**Root cause:** Line 426 in `streamable_http.py` recursed with `attempt=0` whenever the stream ended without a `JSONRPCResponse`/`JSONRPCError`, regardless of whether any useful data was received.

**Fix:** Distinguish productive reconnections (ones that delivered actual message data like notifications) from unproductive ones (only priming events). Productive reconnections reset the counter so legitimate multi-close patterns (e.g., SSE polling with server-side disconnect) continue working. Unproductive reconnections increment the counter. When `MAX_RECONNECTION_ATTEMPTS` is reached, the client now sends a `JSONRPCError` back through the read stream so the caller gets unblocked instead of hanging forever.

## Changes

- **`src/mcp/client/streamable_http.py`**: Track whether actual message data (not just priming events) was received during a reconnection. Only reset the attempt counter when data was delivered. Send an error response to the caller when max attempts are exceeded.
- **`tests/shared/test_streamable_http.py`**: Add `tool_with_perpetual_stream_close` (server tool that always closes the stream without responding) and `test_reconnection_gives_up_after_max_attempts` (regression test proving the client gives up and raises `MCPError` instead of looping forever).

## How Has This Been Tested

- New regression test `test_reconnection_gives_up_after_max_attempts` — verifies the client raises `MCPError` after max attempts with a fail-after timeout of 30s (completes in ~2s)
- All 62 existing `test_streamable_http.py` tests pass, including the multi-reconnection tests (`test_streamable_http_multiple_reconnections`, `test_streamable_http_client_auto_reconnects`, `test_standalone_get_stream_reconnection`)
- `pyright` — 0 errors
- `ruff check` + `ruff format` — clean
- `strict-no-cover` — clean

## Breaking Changes

None. Productive reconnections (where actual data is delivered before the stream drops) continue to reset the counter, preserving existing behavior. Only unproductive reconnections (no message data, just priming events) now correctly increment toward the limit.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -380,8 +380,19 @@ class StreamableHTTPTransport:
     ) -> None:
         """Reconnect with Last-Event-ID to resume stream after server disconnect."""
         # Bail if max retries exceeded
-        if attempt >= MAX_RECONNECTION_ATTEMPTS:  # pragma: no cover
-            logger.debug(f"Max reconnection attempts ({MAX_RECONNECTION_ATTEMPTS}) exceeded")
+        if attempt >= MAX_RECONNECTION_ATTEMPTS:
+            logger.warning(f"Max reconnection attempts ({MAX_RECONNECTION_ATTEMPTS}) exceeded")
+            if isinstance(ctx.session_message.message, JSONRPCRequest):  # pragma: no branch
+                error_data = ErrorData(
+                    code=INTERNAL_ERROR,
+                    message=(
+                        f"SSE stream disconnected and max reconnection attempts ({MAX_RECONNECTION_ATTEMPTS}) exceeded"
+                    ),
+                )
+                error_msg = SessionMessage(
+                    JSONRPCError(jsonrpc="2.0", id=ctx.session_message.message.id, error=error_data)
+                )
+                await ctx.read_stream_writer.send(error_msg)
             return
 
         # Always wait - use server value or default
@@ -404,6 +415,7 @@ class StreamableHTTPTransport:
                 # Track for potential further reconnection
                 reconnect_last_event_id: str = last_event_id
                 reconnect_retry_ms = retry_interval_ms
+                received_data = False
 
                 async for sse in event_source.aiter_sse():
                     if sse.id:  # pragma: no branch
@@ -421,9 +433,15 @@ class StreamableHTTPTransport:
                         await event_source.response.aclose()
                         return
 
-                # Stream ended again without response - reconnect again (reset attempt counter)
+                    if sse.data:
+                        received_data = True
+
+                # Stream ended without response - reset counter only if we received
+                # actual message data (not just priming events), otherwise increment
+                # to prevent infinite reconnection loops when the server always drops.
+                next_attempt = 0 if received_data else attempt + 1
                 logger.info("SSE stream disconnected, reconnecting...")
-                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, 0)
+                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, next_attempt)
         except Exception as e:  # pragma: no cover
             logger.debug(f"Reconnection failed: {e}")
             # Try to reconnect again if we still have an event ID
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -225,6 +225,11 @@ async def _handle_list_tools(  # pragma: no cover
                 description="Tool that closes standalone GET stream mid-operation",
                 input_schema={"type": "object", "properties": {}},
             ),
+            Tool(
+                name="tool_with_perpetual_stream_close",
+                description="Tool that always closes the stream without sending a response",
+                input_schema={"type": "object", "properties": {}},
+            ),
         ]
     )
 
@@ -379,6 +384,16 @@ async def _handle_call_tool(  # pragma: no cover
         await ctx.session.send_resource_updated(uri="http://notification_2")
 
         return CallToolResult(content=[TextContent(type="text", text="Standalone stream close test done")])
+
+    elif name == "tool_with_perpetual_stream_close":
+        # Repeatedly close the stream without ever sending a response.
+        # Used to verify that _handle_reconnection gives up after MAX_RECONNECTION_ATTEMPTS.
+        for _ in range(10):
+            if ctx.close_sse_stream:
+                await ctx.close_sse_stream()
+            await anyio.sleep(0.3)
+        # This response should never be reached by the client because reconnection gives up
+        return CallToolResult(content=[TextContent(type="text", text="Should not reach")])
 
     return CallToolResult(content=[TextContent(type="text", text=f"Called {name}")])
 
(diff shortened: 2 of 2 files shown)


Dharit13:
## CI Green — Summary

All 25 checks pass (Python 3.10–3.14 × Ubuntu/Windows, conformance, pre-commit, readme-snippets).

### The bug

`_handle_reconnection()` recursed with `attempt=0` on line 426 when the SSE stream ended without a complete `JSONRPCResponse`/`JSONRPCError`. This reset the retry counter on every reconnection that succeeded at the HTTP level but dropped before delivering a response. A server that accepts connections and sends priming events but never completes caused the client to retry forever, hanging the caller indefinitely.

Discovered in production when an MCP client coroutine hung for 5+ hours in a reconnection loop after a server's SSE stream kept dropping.

### The fix (2 files, +62/−9 lines)

**`src/mcp/client/streamable_http.py`**
- Track whether actual message data (not just empty priming events) was received during each reconnection attempt.
- **Productive reconnections** (data delivered) → reset counter to 0. This preserves the existing SSE polling pattern where the server intentionally closes streams after delivering notifications.
- **Unproductive reconnections** (only priming events / no data) → increment counter toward `MAX_RECONNECTION_ATTEMPTS`.
- When max attempts is reached, send a `JSONRPCError` back through the read stream so the caller gets unblocked with `MCPError` instead of hanging forever.

**`tests/shared/test_streamable_http.py`**
- Added `tool_with_perpetual_stream_close` — a test tool that always closes the stream without ever sending a response.
- Added `test_reconnection_gives_up_after_max_attempts` — regression test that verifies the client raises `MCPError` after max attempts (completes in ~2s with a 30s safety timeout).

### What's NOT changed

Productive reconnections (where the server delivers notifications before closing the stream) still reset the counter. All existing reconnection tests pass unchanged: `test_streamable_http_multiple_reconnections` (3 consecutive close/reconnect cycles), `test_streamable_http_client_auto_reconnects`, `test_standalone_get_stream_reconnection`, `test_streamable_http_sse_polling_full_cycle`, and `test_streamable_http_events_replayed_after_disconnect`.

PR #2396, opened 2026-04-05
fix: make context logging methods accept Any data type per MCP spec

muhammadrashid4587:
## Summary

- Widens the `message` parameter type from `str` to `Any` in context logging methods (`log`, `debug`, `info`, `warning`, `error`) to match the MCP spec which defines logged data as `unknown` (any JSON-serializable type)
- When dict data is passed alongside `extra`, the dicts are now merged rather than wrapping in a `{"message": ...}` envelope
- Adds tests for non-string data types (dict, list, number) and dict merging with `extra`

## Context

Per the [MCP spec](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-03-26/schema.ts#L833), the `data` field in log message notifications accepts any JSON-serializable type. The underlying `ServerSession.send_log_message` already accepts `data: Any`, but the context convenience methods restricted input to `str` only.

This is a backward-compatible change — existing code passing strings continues to work identically.

Closes #397

## Test plan

- [x] Existing `test_context_logging` passes (string data still works)
- [x] New `test_context_logging_any_data` — verifies dict, list, and number data types
- [x] New `test_context_logging_dict_with_extra` — verifies dict merging behavior with `extra`

### DIFF
--- a/src/mcp/server/mcpserver/context.py
+++ b/src/mcp/server/mcpserver/context.py
@@ -187,22 +187,29 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
     async def log(
         self,
         level: Literal["debug", "info", "warning", "error"],
-        message: str,
+        message: Any,
         *,
         logger_name: str | None = None,
         extra: dict[str, Any] | None = None,
     ) -> None:
         """Send a log message to the client.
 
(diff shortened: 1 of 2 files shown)


Kludex:

PR #2399, opened 2026-04-07
Fix: Change context logging methods to accept Any type per MCP spec

internet-dot:
The MCP spec defines the logging data field as 'unknown' (any JSON serializable type), but the context logging methods (debug, info, warning, error, log) only accepted str. This prevented sending structured log data like dicts or lists.

Changed message parameter type from str to Any in all logging methods. Removed the extra parameter since callers can now pass structured data directly as the message.

Fixes #397


Kludex:

PR #2403, opened 2026-04-08
fix: validate_scope allows any scope when client has no registered scope restriction

enjoykumawat:
## Problem

`OAuthClientMetadata.validate_scope` incorrectly rejects all requested scopes when the client was registered without a `scope` restriction (`self.scope is None`).

The previous code built `allowed_scopes = []` when `self.scope is None`, so the membership check `scope not in allowed_scopes` was always `True`, raising `InvalidScopeError` for every token request — even though `None` should mean *no restrictions*.

Fixes #2216

## Fix

Return `requested_scopes` immediately when `self.scope is None`, bypassing the per-scope membership check. This matches the intended semantics: a client registered without scope restrictions may request any scope.

Also removed the `# pragma: no branch` and `# pragma: no cover` comments that were suppressing coverage for code paths that are now reachable with correct behavior.

## Tests

Added four regression tests to `tests/shared/test_auth.py`:

- `validate_scope` returns `None` when no scope is requested
- Valid registered scopes are returned as a list
- Unregistered scope raises `InvalidScopeError`
- **Any scope is allowed when client has no scope restriction** (the bug)

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -421,9 +421,9 @@ class StreamableHTTPTransport:
                         await event_source.response.aclose()
                         return
 
-                # Stream ended again without response - reconnect again (reset attempt counter)
+                # Stream ended again without response - reconnect again
                 logger.info("SSE stream disconnected, reconnecting...")
-                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, 0)
+                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, attempt + 1)
         except Exception as e:  # pragma: no cover
             logger.debug(f"Reconnection failed: {e}")
             # Try to reconnect again if we still have an event ID
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -92,10 +92,13 @@ class OAuthClientMetadata(BaseModel):
         if self.scope is None:
             # Client registered without scope restrictions — allow any requested scope
             return requested_scopes
-        allowed_scopes = self.scope.split(" ")
-        for scope in requested_scopes:
-            if scope not in allowed_scopes:
-                raise InvalidScopeError(f"Client was not registered with scope {scope}")
+        requested = set(requested_scopes)
+        allowed = set(self.scope.split())
+        if not requested.issubset(allowed):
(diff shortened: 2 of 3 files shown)


src/mcp/shared/auth.py
```diff
@@ -71,11 +71,14 @@ def validate_scope(self, requested_scope: str | None) -> list[str] | None:
...
             return None
         requested_scopes = requested_scope.split(" ")
-        allowed_scopes = [] if self.scope is None else self.scope.split(" ")
+        if self.scope is None:
+            # Client registered without scope restrictions — allow any requested scope
+            return requested_scopes
+        allowed_scopes = self.scope.split(" ")
         for scope in requested_scopes:
-            if scope not in allowed_scopes:  # pragma: no branch
+            if scope not in allowed_scopes:
```
Kludex: Since you are here, can we do this?

```py
requested = set(requested_scopes)
allowed = set(self.scope.split())

if not requested.issubset(allowed):
```

enjoykumawat:
Refactored to use set operations as suggested — cleaner and avoids the explicit loop. Thanks for the tip!

enjoykumawat:
Done! Updated to use `set` operations:

```python
requested = set(requested_scopes)
allowed = set(self.scope.split())

if not requested.issubset(allowed):
    raise InvalidScopeError(...)
```

Cleaner and handles duplicates correctly. Thanks for the suggestion!

src/mcp/client/streamable_http.py
```diff
@@ -421,9 +421,9 @@ async def _handle_reconnection(
                         await event_source.response.aclose()
                         return
 
-                # Stream ended again without response - reconnect again (reset attempt counter)
+                # Stream ended again without response - reconnect again
                 logger.info("SSE stream disconnected, reconnecting...")
-                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, 0)
+                await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, attempt + 1)
```
Kludex, line 426: What's this about?

maxisbey:
Closing — superseded by #2301, which removes the registration-scope check entirely. This PR also still has an unrelated `streamable_http.py` change that wasn't explained when asked. Thanks for the contribution.

PR #2404, opened 2026-04-08
fix(auth): coerce empty-string optional URL fields to None in OAuthClientMetadata

felixweinberger:
Coerce `""` to `None` for the five OPTIONAL URL fields on `OAuthClientMetadata` so that DCR responses from servers that echo omitted metadata as empty strings parse successfully.

## Motivation and Context

RFC 7591 §2 marks `client_uri`, `logo_uri`, `tos_uri`, `policy_uri`, and `jwks_uri` as OPTIONAL. Some authorization servers in the wild (observed: Udemy, deepsense.ai, Apify, windsor.ai, firecrawl, and others) echo the client's omitted metadata back as `""` instead of dropping the keys:

```json
{"client_id": "abc123", "client_uri": "", "logo_uri": "", "tos_uri": "", "policy_uri": "", ...}
```

`AnyHttpUrl` rejects `""`, so `handle_registration_response` raises `ValidationError` and callers discard an otherwise valid registration — the server returned 201 with a real `client_id`.

The servers are technically non-compliant (an empty string is not a valid URL), but the only reasonable interpretation of `""` for an OPTIONAL field is "absent". This applies Postel's law at the deserialization boundary, matching the existing `normalize_token_type` validator on `OAuthToken` in the same file. Because `OAuthClientInformationFull` inherits from `OAuthClientMetadata`, the coercion applies to parsed DCR responses as well.

## How Has This Been Tested?

New `TestOAuthClientMetadataEmptyUrlCoercion` covers per-field coercion (parametrized over all five fields), all-fields-together, valid-URL passthrough, inheritance through `OAuthClientInformationFull` (the DCR-response model), and a negative case confirming non-empty invalid URLs still raise.

## Breaking Changes

None. This only widens accepted input — previously-rejected `""` values now succeed as `None`. Valid URLs, `None`, and omitted keys are unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

RFC 7591 §2: https://datatracker.ietf.org/doc/html/rfc7591#section-2

v1.x backport: #2405

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,7 +45,9 @@
 
 - Framework: `uv run --frozen pytest`
 - Async testing: use anyio, not asyncio
-- Do not use `Test` prefixed classes, use functions
+- Do not use `Test` prefixed classes — write plain top-level `test_*` functions.
+  Legacy files still contain `Test*` classes; do NOT follow that pattern for new
+  tests even when adding to such a file.
 - IMPORTANT: Tests should be fast and deterministic. Prefer in-memory async execution;
   reach for threads only when necessary, and subprocesses only as a last resort.
 - For end-to-end behavior, an in-memory `Client(server)` is usually the
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -67,6 +67,24 @@ class OAuthClientMetadata(BaseModel):
     software_id: str | None = None
     software_version: str | None = None
 
+    @field_validator(
+        "client_uri",
+        "logo_uri",
+        "tos_uri",
+        "policy_uri",
+        "jwks_uri",
+        mode="before",
+    )
+    @classmethod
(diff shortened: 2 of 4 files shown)


maxisbey (commented):
Two test-organization asks:

1. **File location** — these are pure model-validation tests for `src/mcp/shared/auth.py`, so per the tests-mirror-source convention they belong in `tests/shared/test_auth.py` rather than `tests/client/test_auth.py`. I know the existing `OAuthClientMetadata` tests are already in the client file, but since this is a new self-contained block it's a good chance to land it in the right place rather than grow the misplacement.

2. **Drop the `Test...` class** — project convention is plain top-level functions instead of `Test`-prefixed classes. The parametrize and the docstring carry over fine as a module-level comment + standalone functions.

maxisbey (dismissed):

maxisbey (approved):

PR #2419, opened 2026-04-10
MCP Events: Client-side ProvenanceEnvelope and EventQueue utilities

elijahr:
## Summary
- Adds `ProvenanceEnvelope` dataclass for wrapping event payloads with client-generated provenance metadata (server identity, trust tier, topic, source, timestamp) before LLM context injection
- Adds `EventQueue` with priority-aware `drain()` API that returns events in strict priority order (urgent > high > normal > low)
- Both utilities live in `mcp.client.events` as standalone helpers

## Test plan
- [ ] 17 new tests in `tests/client/test_events.py` covering serialization, XML escaping, priority ordering, drain semantics
- [ ] All 67 existing event tests still pass
- [ ] `uv run pytest tests/ -k event -x`


elijahr:
Opened against wrong repo. This should be on axiomantic/python-sdk.

PR #2421, opened 2026-04-10
fix: increment reconnection attempt counter on clean stream disconnect

manjunathgujjar:
## Problem

`_handle_reconnection()` in `streamable_http.py` resets the `attempt` counter to `0` when a reconnection succeeds at the HTTP level but the stream closes without delivering a complete response (line 494 before this fix). This makes `MAX_RECONNECTION_ATTEMPTS` ineffective: only the exception path (network error) incremented the counter. A server that repeatedly accepts connections but drops them before sending a final response causes the client to retry forever.

This was reported in #2393 where production jobs hung for hours in a reconnection loop.

## Root Cause

```python
# Before: clean-close recursion resets attempt to 0

# Exception path correctly increments:
await self._handle_reconnection(ctx, last_event_id, retry_interval_ms, attempt + 1)
```

The asymmetry means `MAX_RECONNECTION_ATTEMPTS` only guards against consecutive HTTP-level failures, not against total reconnection attempts.

## Fix

Pass `attempt + 1` on clean-close recursion, matching the exception path. `MAX_RECONNECTION_ATTEMPTS` is also raised from 2 to 5 to give more headroom for legitimate transient disconnects while still bounding the retry loop.

The `# pragma: no cover` on the MAX guard is also removed since the new test now exercises that branch.

## Test

Added `test_reconnection_attempt_counter_increments_on_clean_disconnect` to `tests/shared/test_streamable_http.py`. The test spies on `_handle_reconnection` to record the `attempt` values seen, patches `MAX_RECONNECTION_ATTEMPTS=2`, then calls a tool that closes the SSE stream multiple times without completing. With the fix, the spy records `[0, 1, 2]` — the counter increments until the limit is hit. Without the fix it would record `[0, 0, 0, ...]` indefinitely.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -47,7 +47,7 @@ LAST_EVENT_ID = "last-event-id"
 
 # Reconnection defaults
 DEFAULT_RECONNECTION_DELAY_MS = 1000  # 1 second fallback when server doesn't provide retry
-MAX_RECONNECTION_ATTEMPTS = 2  # Max retry attempts before giving up
+MAX_RECONNECTION_ATTEMPTS = 5  # Max retry attempts before giving up
 
 
 class StreamableHTTPError(Exception):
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2395, which fixes the counter and propagates an error to the caller, and preserves the intentional-close polling pattern. CI is also failing on this branch. Thanks for the contribution.

PR #2425, opened 2026-04-12
feat: add reusable roots enforcement utilities to MCPServer

NeelakandanNC:
MCP Roots let clients declare filesystem boundaries to servers, but the SDK provides no helpers to enforce them — every server author has to write the same boundary-check logic by hand. This adds a small utility module so authors can drop in enforcement without reinventing it.

- get_roots(ctx): fetch root URIs declared by the client
- assert_within_roots(path, ctx): raise PermissionError if path is outside roots
- within_roots_check: decorator that auto-enforces roots on path parameters

Both sides of the comparison go through Path.resolve() so that platform firmlinks (e.g. macOS /tmp -> /private/tmp, /home -> /System/Volumes/Data/home) don't produce spurious denials.

## Motivation and Context

I learned about MCP Roots in an advanced MCP course. When I actually tried implementing it in my own servers, I realised you have to write the same enforcement logic every single time you build a new server — the course teaches the concept but the SDK gives you no utility to actually apply it without repeating yourself.
So I built one.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/mcpserver/__init__.py
+++ b/src/mcp/server/mcpserver/__init__.py
@@ -4,6 +4,16 @@ from mcp.types import Icon
 
 from .context import Context
 from .server import MCPServer
+from .utilities.roots import assert_within_roots, get_roots, within_roots_check
 from .utilities.types import Audio, Image
 
-__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon"]
+__all__ = [
+    "MCPServer",
+    "Context",
+    "Image",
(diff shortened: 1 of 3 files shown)


Kludex:
Please create an issue about it, with what you want to achieve. I don't think the proposed API is easy to use.

NeelakandanNC:
Sure, will create a issue and add proper docs to understand it and make it easier to use

PR #2426, opened 2026-04-12
Fix validate_scope rejecting scopes when client scope is None

Christian-Sidak:
## Summary

Fixes #2216

`OAuthClientMetadata.validate_scope()` incorrectly treats `self.scope = None` as an empty allowlist (`[]`), causing all requested scopes to be rejected with `InvalidScopeError`. Per OAuth 2.0 semantics, `None` means the client was registered without scope restrictions, so any requested scope should be permitted.

**Before:** `allowed_scopes = [] if self.scope is None else self.scope.split(" ")` -- any requested scope fails the `in` check against `[]`.

**After:** Skip the allowlist check entirely when `self.scope is None`, only validate against registered scopes when they are explicitly set.

## Test plan

- Added test: `None` client scope allows any requested scope
- Added test: set client scope correctly rejects unregistered scopes
- Added test: set client scope correctly allows registered scopes
- Added test: `None` requested scope returns `None` regardless of client scope

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -71,11 +71,12 @@ class OAuthClientMetadata(BaseModel):
         if requested_scope is None:
             return None
         requested_scopes = requested_scope.split(" ")
-        allowed_scopes = [] if self.scope is None else self.scope.split(" ")
-        for scope in requested_scopes:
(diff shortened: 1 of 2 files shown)


Christian-Sidak:
Closing -- just noticed existing PRs (#2224, #2246) already addressing this. Apologies for the duplicate!

PR #2427, opened 2026-04-12
Fix context logging methods to accept Any type per MCP spec

Christian-Sidak:
## Summary

Fixes #397

The MCP spec defines the logging notification `data` field as `unknown` (any JSON serializable type), but `Context.log()` and the convenience methods (`debug`, `info`, `warning`, `error`) only accepted `str`. This was inconsistent with both the spec and the existing `ServerSession.send_log_message(data: Any)` signature.

**Changes:**
- Changed parameter from `message: str` to `data: Any` in `Context.log()` and all convenience methods
- Removed the `extra` parameter, which was a workaround for the `str`-only limitation and caused confusion (calling with `extra=` kwargs raised `TypeError` since `Context.log()` did not accept them)
- Updated tests to cover structured data logging (dicts, lists, numbers, None)

**Before (broken):**
```python
# Type error - spec says this should work
ctx.info({"event": "user_login", "user_id": 123})
```

**After (fixed):**
```python
# All JSON serializable types now accepted
ctx.info("simple string message")
ctx.info({"event": "user_login", "user_id": 123})
ctx.debug(["step1", "step2", "step3"])
ctx.warning(42)
```

Backward compatible -- all existing code passing strings continues to work unchanged since the positional argument just changed name from `message` to `data`.

## Test plan
- [x] Existing `test_context_logging` passes (string messages)
- [x] New `test_context_logging_structured_data` passes (dict, list, int, None)
- [x] Updated `test_logging_callback` passes with new API

### DIFF
--- a/src/mcp/server/mcpserver/context.py
+++ b/src/mcp/server/mcpserver/context.py
@@ -187,28 +187,22 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
     async def log(
         self,
         level: Literal["debug", "info", "warning", "error"],
-        message: str,
+        data: Any,
         *,
         logger_name: str | None = None,
-        extra: dict[str, Any] | None = None,
     ) -> None:
         """Send a log message to the client.
 
         Args:
             level: Log level (debug, info, warning, error)
-            message: Log message
+            data: The data to be logged. Any JSON serializable type is allowed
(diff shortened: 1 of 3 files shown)


Christian-Sidak:
Closing -- I see this has been actively worked on by several contributors (#400, #1907). Apologies for the duplicate!

PR #2430, opened 2026-04-13
Fix SSE timeout hang by propagating transport exceptions to pending r…

widgetwalker:
Fixed SSE timeout hang in BaseSession._receive_loop. This fix propagates transport-level exceptions (like SSE read timeouts) to all pending request streams, ensuring RPC calls don't hang if the underlying connection fails. This specifically addresses issue #1401.


maxisbey:
Closing in favor of the transport-layer fixes (#2122, #2483, #2607). Modifying `shared/session.py` would also affect the server side; the fix belongs in the client transports. Thanks for the contribution.

PR #2438, opened 2026-04-14
fix: consume response body on unexpected content type to prevent client hang

### LINKED ISSUE #2432: Initialize call hangs forever if MCP server does not return a `Content-Type: text/plain`
### Initial Checks

### Description

We had some timeouts on our server while using MCP client from the SDK.
When users try to integrate our client with their MCP server, we do an `initialize` call to the server and we are waiting infinitely to the following:
```
sync with self:
    ...
```

After investigation, it seems the error happens when the MCP server is misconfigured, and returns something different than `Content-Type: text/event-stream` as a response from the `initialize` call.

Although it is on the server side, the client should be able to fail immediately with a understandable error to warn the user something went wrong with the server's `initialize` output.
   

### Example Code

```Python
# server.py

from http.server import BaseHTTPRequestHandler, HTTPServer

class PlainTextHandler(BaseHTTPRequestHandler):
    def do_POST(self):
        length = int(self.headers.get("Content-Length", 0))
        self.rfile.read(length)

        body = b"this is not json"
        self.send_response(200)
        self.send_header("Content-Type", "text/plain")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)

    def log_message(self, format, *args):
        print(f"[server] {format % args}")

server = HTTPServer(("localhost", 9000), PlainTextHandler)
print("[server] listening on http://localhost:9000")
server.serve_forever()

# client.py
import asyncio
import mcp

async def main():
    print("connecting…")
    async with streamable_http_client("http://localhost:9000/mcp") as (
        read_stream, write_stream, _,
    ):
        async with mcp.ClientSession(read_stream, write_stream) as session:
            await session.initialize()  # ← hangs here forever

    print("done (you should never see this)")
```

### Python & MCP Python SDK

```Text
- Python 3.12.12
- MCP - 1.27.0
```

vishal-vanam:
## Summary

Fixes #2432

When an MCP server returns an unexpected `Content-Type` (e.g. `text/plain` instead of `application/json` or `text/event-stream`), `session.initialize()` hangs indefinitely because the response body is never consumed, causing the httpx streaming context to block on exit.

**The fix:** Call `await response.aread()` to consume the response body before sending the error message through the read stream. This allows the streaming context to close cleanly and the error to propagate to the caller.

## Changes

- `src/mcp/client/streamable_http.py`: Added `await response.aread()` in the unexpected content-type branch to drain the response body
- `tests/shared/test_streamable_http.py`: Added test using httpx MockTransport to verify the client raises `MCPError` instead of hanging

## Test plan

- [x] New test `test_unexpected_content_type_does_not_hang` passes locally
- [ ] Verify existing streamable HTTP tests still pass

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -295,6 +295,7 @@ class StreamableHTTPTransport:
                 elif content_type.startswith("text/event-stream"):
                     await self._handle_sse_response(response, ctx, is_initialization)
                 else:
+                    await response.aread()  # consume body to prevent stream hang
                     logger.error(f"Unexpected content type: {content_type}")
                     error_data = ErrorData(code=INVALID_REQUEST, message=f"Unexpected content type: {content_type}")
                     error_msg = SessionMessage(JSONRPCError(jsonrpc="2.0", id=message.id, error=error_data))
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -2336,4 +2336,3 @@ async def test_unexpected_content_type_does_not_hang():
         async with ClientSession(read_stream, write_stream) as session:
             with pytest.raises(MCPError, match="Unexpected content type"):
                 await session.initialize()
-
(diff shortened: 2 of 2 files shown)


maxisbey:
Thanks for the PR. This has since landed via #2005. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2439, opened 2026-04-14
fix: make UrlElicitationRequiredError pickle-safe

### LINKED ISSUE #2431: McpError is not pickle-safe and fails to unpickle
### Initial Checks

## Summary

`mcp.shared.exceptions.McpError` does not survive a normal `cloudpickle.dumps()` / `cloudpickle.loads()` round-trip.

The failure appears to come from `McpError.__init__` expecting an `ErrorData` object, while exception unpickling reconstructs it with a plain string from `Exception.args`.

This is surfacing for us through background task execution, but the bug reproduces without Docket/FastMCP task machinery.

## Actual behavior

Unpickling fails with:

```text
AttributeError: 'str' object has no attribute 'message'
```

Traceback points at `McpError.__init__`:

```python
class McpError(Exception):
    error: ErrorData

    def __init__(self, error: ErrorData):
        super().__init__(error.message)
        self.error = error
```

## Expected behavior

`McpError(ErrorData(...))` should round-trip through pickle/cloudpickle without crashing.

At minimum, this should work:

- serialize `McpError`
- deserialize `McpError`
- preserve the message
- preserve the `error` payload, or at least degrade safely without raising during unpickle

## Suspected root cause

`McpError` stores `error.message` in `Exception.args` via `super().__init__(error.message)`.

On unpickle, exception reconstruction uses `args`, so `McpError` is effectively reconstructed as:

```python
McpError("Authentication Required")
```

But `McpError.__init__` assumes `error` is always an `ErrorData`, so it does:

```python
error.message
```

which crashes for `str`.

## Suggested fix

`McpError` likely needs to be pickle-safe by design. Any of these would probably fix it:

1. Make `__init__` accept both `ErrorData` and `str`, normalizing `str` into an `ErrorData`.
2. Implement `__reduce__` so pickle reconstructs using the full `ErrorData`.
3. Ensure constructor args and exception state are aligned with standard exception pickling behavior.

A robust version would probably do both `__reduce__` and tolerant initialization.

## Notes

This bug is easy to misattribute to `cloudpickle` or task runners, but the reproducer above shows it is local to `McpError` itself.

### Example Code

```Python
from importlib.metadata import version

import cloudpickle
from mcp.types import ErrorData


print("Versions:")
print(f"  mcp={version('mcp')}")
print(f"  cloudpickle={version('cloudpickle')}")

original = McpError(ErrorData(code=-32600, message="Authentication Required"))

print("\nOriginal exception:")
print(f"  type={type(original).__name__}")
print(f"  str={str(original)!r}")
print(f"  error_type={type(original.error).__name__}")
print(f"  error_message={original.error.message!r}")

payload = cloudpickle.dumps(original)

print("\nUnpickling:")
restored = cloudpickle.loads(payload)
print(f"  restored_type={type(restored).__name__}")
print(f"  restored_args={restored.args!r}")
print(f"  restored_error={getattr(restored, 'error', None)!r}")
```

### Python & MCP Python SDK

```Text
- `mcp==1.26.0`
- `fastmcp==3.2.3`
- `cloudpickle==3.1.2`
- Python 3.13
```

Christian-Sidak:
## Summary
- `UrlElicitationRequiredError` fails to unpickle because `MCPError` stores `(code, message, data)` in `Exception.args` but `UrlElicitationRequiredError.__init__` expects `(elicitations, message=None)`, causing a `TypeError` on unpickle
- Adds a `__reduce__` method to `UrlElicitationRequiredError` so pickle reconstructs with the correct `(elicitations, message)` arguments
- Adds pickle round-trip tests for both `MCPError` and `UrlElicitationRequiredError`

Closes #2431

## Test plan
- [x] Pickle round-trip test for `MCPError`
- [x] Pickle round-trip test for `UrlElicitationRequiredError` (single elicitation, custom message)
- [x] Pickle round-trip test for `UrlElicitationRequiredError` (multiple elicitations, default message)
- [ ] Verify with multiprocessing use case from issue

### DIFF
--- a/src/mcp/shared/exceptions.py
+++ b/src/mcp/shared/exceptions.py
@@ -94,6 +94,10 @@ class UrlElicitationRequiredError(MCPError):
         """The list of URL elicitations required before the request can proceed."""
         return self._elicitations
 
+    def __reduce__(self) -> tuple[type, tuple[list[ElicitRequestURLParams], str]]:
+        """Support pickling by reconstructing with the original __init__ signature."""
+        return (self.__class__, (self._elicitations, self.error.message))
+
     @classmethod
     def from_error(cls, error: ErrorData) -> UrlElicitationRequiredError:
         """Reconstruct from an ErrorData received over the wire."""
--- a/tests/shared/test_exceptions.py
+++ b/tests/shared/test_exceptions.py
@@ -1,5 +1,7 @@
 """Tests for MCP exception classes."""
 
+import pickle
+
 import pytest
 
 from mcp.shared.exceptions import MCPError, UrlElicitationRequiredError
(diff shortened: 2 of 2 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #2440, opened 2026-04-14
fix: add newline="" to stdio TextIOWrapper to prevent CRLF on Windows

### LINKED ISSUE #2433: Windows: TextIOWrapper in stdio_server() emits CRLF instead of LF, corrupting newline-delimited JSON messages
## Summary

## Affected file

```python
# Current (buggy on Windows)
if not stdin:
if not stdout:
```

## Reproduction

```python
import subprocess, sys

script = r'''
import sys
from io import TextIOWrapper
stdout.flush()
'''

print(repr(out))
```

Verified on:
- **Python:** 3.11
- **mcp:** 1.26.0

## Why this matters

## Fix

```python
if not stdin:
if not stdout:
```

With this fix:
```python
# newline="" result:
buf = io.BytesIO()
wrapper.flush()
repr(buf.getvalue())
```

## Context

Christian-Sidak:
## Summary
- Adds `newline=""` to both `TextIOWrapper` instances in `stdio_server()` to disable platform-dependent newline translation
- On Windows, Python's default `newline=None` translates `\n` to `\r\n` on write, corrupting NDJSON messages
- `newline=""` is a no-op on Linux/macOS, so this is safe cross-platform
- Adds test verifying raw output bytes use LF not CRLF

Closes #2433

## Test plan
- [x] New test asserts raw stdio output ends with `\n`, not `\r\n`
- [ ] Verify on Windows that JSON-RPC messages are no longer corrupted

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -39,9 +39,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     # python is platform-dependent (Windows is particularly problematic), so we
     # re-wrap the underlying binary stream to ensure UTF-8.
     if not stdin:
-        stdin = anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8", errors="replace"))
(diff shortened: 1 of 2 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:

PR #2446, opened 2026-04-15
feat(auth): add subject field to AccessToken

### LINKED ISSUE #1038: MCP server: AccessToken class should have field for subject claim ("sub")

Sagargupta16:
## Summary

Add an optional `subject` field to `AccessToken` for storing the JWT `sub` claim (user ID).

## Motivation

Currently, token verifiers that decode JWTs have no standard way to pass through the subject/user identity. Consumers must re-decode the JWT to retrieve `sub`, duplicating work.

With this change:

```python
class MyTokenVerifier:
        claims = decode_and_validate_jwt(token)
        return AccessToken(
            token=token,
            client_id=claims["client_id"],
            scopes=claims["scope"].split(),
            subject=claims["sub"],
        )
```

Then downstream code can access the user ID directly:

```python
```

## Changes

- Added `subject: str | None = None` to `AccessToken` in `provider.py`
- Added test for backward compatibility (no subject) and with subject

Backward compatible - field defaults to `None`.

Fixes #1038

### DIFF
--- a/src/mcp/server/auth/provider.py
+++ b/src/mcp/server/auth/provider.py
@@ -40,6 +40,7 @@ class AccessToken(BaseModel):
     scopes: list[str]
     expires_at: int | None = None
     resource: str | None = None  # RFC 8707 resource indicator
+    subject: str | None = None  # Subject identifier (typically the "sub" JWT claim / user ID)
 
 
 RegistrationErrorCode = Literal[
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing as a duplicate of #2209, which was claimed on the issue and implements the full scope discussed there (`subject` + `claims` + `Context.subject`). Thanks for the contribution.

Issue #2384, opened 2026-04-02
Inline $ref in tool inputSchema for LLM consumption (parity with typescript-sdk)


felixweinberger:
`tools/base.py:76` uses raw `model_json_schema()` which emits `$ref`/`$defs` for nested Pydantic models. LLM clients consuming `tools/list` often can't resolve `$ref`, leading to params being passed as stringified JSON instead of structured objects.

Repro / demand: anthropics/claude-code#18260 (Notion MCP server, `$ref` params get stringified by the model).

typescript-sdk addresses this in modelcontextprotocol/typescript-sdk#1563 by adding a `dereferenceLocalRefs()` step that inlines local `$ref` and throws on cycles (matching go-sdk precedent). Python should do the equivalent in the tool schema pipeline so all SDKs emit LLM-consumable schemas.

Relevant file: `src/mcp/server/mcpserver/tools/base.py:76`

RudrenduPaul:
I'd like to work on this.

The typescript-sdk already ships a `dereferenceLocalRefs()` pass (merged in #1563) and the go-sdk does the same. The Python SDK currently emits raw `$ref`/`$defs` from `model_json_schema()`, which LLM clients cannot resolve — causing nested-model parameters to be stringified.

My plan is to port the `dereferenceLocalRefs()` logic to the Python SDK and apply it in the tool `inputSchema` generation path (around `src/mcp/server/mcpserver/tools/base.py`). Would this approach work for maintainers? Happy to discuss before coding.

RudrenduPaul:
Hi — I'd like to take a look at implementing `$ref` dereferencing for the Python SDK, bringing it to parity with the TypeScript SDK (#1563).

My approach: when resolving JSON Schema tool definitions, recursively dereference `$ref` pointers against the `$defs` or `definitions` section before passing the schema to the tool handler. This avoids the tool receiving a schema with unresolved `$ref` strings.

Will open a PR with tests. Let me know if there's a specific design direction you prefer or if this is already in progress.

PR #1563, opened 2025-11-02
fix: Support string enum for elicitation

vincent0426:
## Add support for string Enum types in elicitation schemas

Related PR: https://github.com/modelcontextprotocol/inspector/pull/902

Changes Made
- Added `StrEnum` to `_ELICITATION_PRIMITIVE_TYPES` tuple
- Enhanced `_is_primitive_field()` to detect string enum types using `issubclass()` checks

## Motivation and Context

Per https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation#supported-schema-types, elicitation schemas support enum types.

Currently, `python-sdk` only supported basic primitive types (`str`, `int`, `float`, `bool`) in elicitation schemas, but did not recognize Enum types.

## How Has This Been Tested?

- Added test cases in `test_elicitation_schema_validation()` to verify enum schemas

- Manual test with
```python
from enum import Enum, StrEnum

from pydantic import BaseModel, Field

from mcp.server.fastmcp import Context, FastMCP
from mcp.server.session import ServerSession

mcp = FastMCP(name="Elicitation Example")


class FruitEnum(str, Enum):
    pear = "pear"
    apple = "apple"


class TestStrEnum(StrEnum):
    one = "1"
    two = "2"
    three = "3"


class BookingPreferences(BaseModel):
    """Schema for collecting user preferences."""

    checkAlternative: bool = Field(description="Would you like to check another date?")
    alternativeDate: str = Field(
        default="2024-12-26",
        description="Alternative date (YYYY-MM-DD)",
    )
    fruit: FruitEnum = FruitEnum.pear
    strenum: TestStrEnum = TestStrEnum.one


@mcp.tool()
async def book_table(date: str, time: str, party_size: int, ctx: Context[ServerSession, None]) -> str:
    """Book a table with date availability check."""
    # Check if date is available
    if date == "2024-12-25":
        # Date unavailable - ask user for alternative
        result = await ctx.elicit(
            message=(f"No tables available for {party_size} on {date}. Would you like to try another date?"),
            schema=BookingPreferences,
        )
        print(result)
        if result.action == "accept" and result.data:
            if result.data.checkAlternative:
                return f"[SUCCESS] Booked for {result.data.alternativeDate}"
            return "[CANCELLED] No booking made"
        return "[CANCELLED] Booking cancelled"

    # Date available
    return f"[SUCCESS] Booked for {date} at {time}"


@mcp.tool()
async def get_uuid(ctx: Context[ServerSession, None]) -> str:
[...]

PR #2448, opened 2026-04-15
feat(tools): inline local $ref in tool inputSchema (#2384)

MukundaKatta:
## Why

Pydantic's \`model_json_schema()\` emits \`\$ref\`/\`\$defs\` for nested models.
LLM clients consuming \`tools/list\` often cannot resolve \`\$ref\` — they
serialize referenced parameters as stringified JSON instead of structured
objects. (See e.g. the Notion MCP server case in
[anthropics/claude-code#18260](https://github.com/anthropics/claude-code/issues/18260).)

typescript-sdk addressed this in
[modelcontextprotocol/typescript-sdk#1563](https://github.com/modelcontextprotocol/typescript-sdk/pull/1563)
by inlining local \`\$ref\` pointers. This PR adds the Python equivalent
so all SDKs emit LLM-consumable schemas.

## What

- New module \`mcp/server/mcpserver/utilities/schema.py\` with
  \`dereference_local_refs(schema)\` that inlines \`\$ref\` pointers to local
  \`#/\$defs/...\` (and legacy \`#/definitions/...\`).
- Applied at \`Tool.from_function\` to the schema produced by pydantic.

## Behavior

| Case | Behavior |
|---|---|
| Simple local ref | Inlined; \`\$defs\` pruned |
| Diamond (A→D, B→D) | D resolved once, both uses share the inlined value |
| Cycle (Node → Node) | Ref left in place; cyclic \`\$defs\` entries preserved |
| Sibling keywords | Merged per JSON Schema 2020-12 (siblings override resolved) |
| Legacy \`definitions\` | Handled; \`\$defs\` wins if both present |
| External \`\$ref\` | Untouched |
| Unknown local ref | Untouched |
| No \`\$defs\` in schema | Returned unchanged |
| Input mutation | None — original schema preserved |

## Tests

11 new unit tests in \`tests/server/mcpserver/utilities/test_schema.py\`
covering each of the cases above. Existing \`tools/\` tests continue to pass.

## Related

- Issue: #2384
- TS reference: modelcontextprotocol/typescript-sdk#1563
- Downstream issue: anthropics/claude-code#18260

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -174,7 +174,12 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                             errors=server.encoding_error_handler,
                         )
                     )
-        except anyio.ClosedResourceError:  # pragma: no cover
+        except (
+            anyio.BrokenResourceError,
+            anyio.ClosedResourceError,
+            BrokenPipeError,
+            ConnectionResetError,
+        ):  # pragma: no cover
             await anyio.lowlevel.checkpoint()
 
     async with anyio.create_task_group() as tg, process:
--- a/src/mcp/server/mcpserver/tools/base.py
+++ b/src/mcp/server/mcpserver/tools/base.py
@@ -78,9 +78,7 @@ class Tool(BaseModel):
         # JSON instead of structured objects. Inline local refs so tool schemas
         # are self-contained and LLM-consumable. Matches behavior of
         # typescript-sdk (#1563) and go-sdk.
-        parameters = dereference_local_refs(
-            func_arg_metadata.arg_model.model_json_schema(by_alias=True)
-        )
+        parameters = dereference_local_refs(func_arg_metadata.arg_model.model_json_schema(by_alias=True))
 
         return cls(
             fn=fn,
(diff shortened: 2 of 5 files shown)


keeltrace:

I think the sibling-keyword merge changes JSON Schema 2020-12 semantics when a sibling collides with a validation keyword in the referenced schema.

The new resolver does:

```python
return {**resolved_schema, **resolved_siblings}
```

That treats a sibling with the same key as a replacement. But JSON Schema 2020-12 Core §8.2.3.1 defines `$ref` as an applicator whose results are the results of the referenced schema, and §10.1 says schema keywords typically operate independently. Keywords adjacent to `$ref` are therefore evaluated alongside the referenced schema; they do not replace same-named constraints inside it.

For example:

```json
{
  "$defs": {
    "Base": {
      "type": "string",
      "maxLength": 5
    }
  },
  "properties": {
    "x": {
      "$ref": "#/$defs/Base",
      "maxLength": 10
    }
  }
}
```

Under Draft 2020-12, `{"x": "1234567"}` is invalid because the referenced `maxLength: 5` still applies. This PR dereferences the property to:

```json
{
  "type": "string",
  "maxLength": 10
}
```

so the same instance becomes valid. I reproduced that difference with `jsonschema.Draft202012Validator`: the original schema rejects the 7-character value, while the merged schema accepts it.

The current sibling regression test only collides on `description`, which is an annotation, so it does not expose this validation broadening. It also currently comments that “siblings override resolved”, which is the behavior causing the mismatch.

Could sibling-bearing refs preserve the conjunction instead—for example by retaining both schemas through an `allOf`-style representation, or another transformation that does not overwrite referenced validation constraints? A regression test with colliding `maxLength` (or `minimum`) would pin the semantic case.

I also checked the TypeScript PR this change cites for parity (#1563). Its current implementation uses the same object-spread merge, and the review discussion there covers integration/cycle handling rather than this collision case, so I don't think this concern is already accounted for upstream.

Checked against PR head `a5351e8`. I did not run this repository's full suite locally.

PR #2450, opened 2026-04-15
fix(stdio): handle BrokenResourceError in stdout_reader race (#1960)

### LINKED ISSUE #1960: BrokenResourceError race condition in stdio_client cleanup when context exits quickly

ddullah:
## Summary

Fixes #1960. Paired with #2449 (the matching `[v1.x]` backport).

`stdio_client` has a shutdown race: `read_stream_writer.aclose()` in the
context's `finally` block can close the receiver while the background
`stdout_reader` task is mid-`send`. anyio raises `BrokenResourceError`,
which the outer `except` does not cover (`ClosedResourceError` is the
sibling class, raised on **already-closed** streams, not streams closed
**during** an in-flight send). The exception propagates through the task
group as `ExceptionGroup` and fails every caller that exits the context
while the subprocess is still writing to stdout.

## Trigger

Any MCP server that writes a burst of output (startup banners, log-level
`notifications/message` frames, scheduler init messages, etc.) paired with
a caller that opens a short-lived `stdio_client` per invocation. The task
group exits before the subprocess drains, and the race fires.

Reproduced against `jules-mcp-server` — which emits three
`notifications/message` frames on init — driven by `mcp2cli`. Every tool
call surfaced as:

\`\`\`
    |   File '.../mcp/client/stdio/__init__.py', line 162, in stdout_reader
    |   File '.../anyio/streams/memory.py', line 213, in send_nowait
    | anyio.BrokenResourceError
\`\`\`

## Fix

Wrap both `read_stream_writer.send(...)` sites in `stdout_reader` with
`try`/`except (ClosedResourceError, BrokenResourceError): return`, and
widen the outer `except` to the same union for defense in depth.
`stdout_reader` now shuts down cleanly no matter how abruptly the caller
exits the context. No API changes.

## Alternative considered

#1960 also suggests `tg.cancel_scope.cancel()` before the stream closes in
`finally`. That works but changes shutdown ordering (cancels user-defined
notification handlers, etc.). The chosen fix is narrower: treat the
shutdown race as a graceful exit from `stdout_reader` specifically, without
altering the task group lifecycle.

## Tests

Added `test_stdio_client_exits_cleanly_while_server_still_writing` in
`tests/client/test_stdio.py`. Spawns a subprocess that writes a thousand
valid JSONRPC notifications, exits the `stdio_client` context immediately,
asserts no exception propagates. Wrapped in `anyio.fail_after(5.0)` per
AGENTS.md. Fails before the fix (ExceptionGroup / BrokenResourceError),
passes after.

Closes #1960.

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -153,12 +153,24 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                             message = types.jsonrpc_message_adapter.validate_json(line, by_name=False)
                         except Exception as exc:  # pragma: no cover
                             logger.exception("Failed to parse JSONRPC message from server")
-                            await read_stream_writer.send(exc)
+                            try:
+                                await read_stream_writer.send(exc)
+                            except (anyio.ClosedResourceError, anyio.BrokenResourceError):
+                                return
                             continue
 
                         session_message = SessionMessage(message)
-                        await read_stream_writer.send(session_message)
-        except anyio.ClosedResourceError:  # pragma: lax no cover
+                        try:
+                            await read_stream_writer.send(session_message)
+                        except (anyio.ClosedResourceError, anyio.BrokenResourceError):
+                            # The caller exited the stdio_client context while the
+                            # subprocess was still writing; the reader stream has been
+                            # closed (ClosedResourceError) or the closure raced our
+                            # in-flight send (BrokenResourceError). Either way there
+                            # is nowhere to deliver this message, so shut down
+                            # cleanly. Fixes #1960.
+                            return
+        except (anyio.ClosedResourceError, anyio.BrokenResourceError):  # pragma: lax no cover
             await anyio.lowlevel.checkpoint()
 
     async def stdin_writer():
--- a/tests/client/test_stdio.py
+++ b/tests/client/test_stdio.py
@@ -38,6 +38,45 @@ async def test_stdio_context_manager_exiting():
 
 
 @pytest.mark.anyio
+async def test_stdio_client_exits_cleanly_while_server_still_writing():
+    """Regression test for #1960.
+
+    Exiting the ``stdio_client`` context while the subprocess is still writing to
+    stdout used to surface ``anyio.BrokenResourceError`` through the task group
+    (as an ``ExceptionGroup``). The ``finally`` block closes
+    ``read_stream_writer`` while the background ``stdout_reader`` task is
+    mid-``send``.
+
+    The fix makes ``stdout_reader`` catch both ``ClosedResourceError`` and
+    ``BrokenResourceError`` and return cleanly, so exiting the context is a
+    no-op no matter what the subprocess is doing.
+    """
+    # A server that emits a large burst of valid JSON-RPC notifications without
+    # ever reading stdin. When we exit the context below, the subprocess is
+    # still in the middle of that burst, which is the exact shape of the race.
+    noisy_script = textwrap.dedent(
+        """
+        import sys
+        for i in range(1000):
+            sys.stdout.write(
+                '{"jsonrpc":"2.0","method":"notifications/message",'
+                '"params":{"level":"info","data":"line ' + str(i) + '"}}\\n'
+            )
+            sys.stdout.flush()
+        """
+    )
+
+    server_params = StdioServerParameters(command=sys.executable, args=["-c", noisy_script])
+
+    # The ``async with`` must complete without an ``ExceptionGroup`` /
+    # ``BrokenResourceError`` propagating. ``anyio.fail_after`` prevents a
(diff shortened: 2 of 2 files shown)


src/mcp/client/stdio.py
```diff
@@ -153,12 +153,24 @@ async def stdout_reader():
                             message = types.jsonrpc_message_adapter.validate_json(line, by_name=False)
                         except Exception as exc:  # pragma: no cover
                             logger.exception("Failed to parse JSONRPC message from server")
-                            await read_stream_writer.send(exc)
+                            try:
+                                await read_stream_writer.send(exc)
+                            except (anyio.ClosedResourceError, anyio.BrokenResourceError):
+                                return
```
maxisbey, line 159:
```suggestion
                            await read_stream_writer.send(exc)
```

src/mcp/client/stdio.py
```diff
@@ -153,12 +153,24 @@ async def stdout_reader():
...
+                            await read_stream_writer.send(session_message)
+                        except (anyio.ClosedResourceError, anyio.BrokenResourceError):
+                            # The caller exited the stdio_client context while the
+                            # subprocess was still writing; the reader stream has been
+                            # closed (ClosedResourceError) or the closure raced our
+                            # in-flight send (BrokenResourceError). Either way there
+                            # is nowhere to deliver this message, so shut down
+                            # cleanly. Fixes #1960.
+                            return
+        except (anyio.ClosedResourceError, anyio.BrokenResourceError):  # pragma: lax no cover
```
maxisbey, line 173:
```suggestion
                        await read_stream_writer.send(session_message)
        except (anyio.ClosedResourceError, anyio.BrokenResourceError):
            # BrokenResourceError: read_stream (receiver) was closed during cleanup while
            # send() was blocked waiting for a consumer.
```

tests/client/test_stdio.py
```diff
@@ -37,6 +37,45 @@ async def test_stdio_context_manager_exiting():
...
+    )
+
+    server_params = StdioServerParameters(command=sys.executable, args=["-c", noisy_script])
+
+    # The ``async with`` must complete without an ``ExceptionGroup`` /
+    # ``BrokenResourceError`` propagating. ``anyio.fail_after`` prevents a
+    # regression from hanging CI.
+    with anyio.fail_after(5.0):
+        async with stdio_client(server_params) as (_, _):
+            pass
```
maxisbey, line 76:
```suggestion
async def test_stdio_client_exits_cleanly_with_unread_server_output():
    """Exiting the stdio_client context while stdout_reader is blocked in send() must
    not raise. Cleanup closes read_stream (the receiver) first, which wakes the
    blocked send() with BrokenResourceError; stdout_reader should swallow it.
    """
    script = textwrap.dedent(
        """
        import sys
        sys.stdout.write('{"jsonrpc":"2.0","method":"notifications/message","params":{}}\\n')
        sys.stdout.flush()
        sys.stdin.read()
        """
    )
    server_params = StdioServerParameters(command=sys.executable, args=["-c", script])

    with anyio.fail_after(5.0):
        async with stdio_client(server_params) as (read_stream, _):
            # Wait until stdout_reader is parked inside send() on the zero-buffer
            # stream. statistics() makes this a positive check rather than a sleep.
            while read_stream.statistics().tasks_waiting_send == 0:
                await anyio.sleep(0)
```

maxisbey (changes requested):
Thanks for this — the diagnosis and fix are right. I couldn't push to your branch (org-owned fork), so I've left the simplifications as suggestions you can apply directly.

Summary:
- The outer `except` already covers both `send()` calls, so the inner try/except blocks aren't needed — widening the outer one is the whole fix. Dropping them also lets us remove the `# pragma: lax no cover` since the new test now hits that line directly.
- Reworked the test to wait on `read_stream.statistics().tasks_waiting_send` so we know `stdout_reader` is actually parked in `send()` before exiting — deterministic across platforms and avoids the terminate-timeout path.

maxisbey:
Thanks for the PR. This has since landed via #2773. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2452, opened 2026-04-16
feat: add Dispatcher Protocol and DirectDispatcher

maxisbey:
Design doc: https://gist.github.com/maxisbey/1e14e741d774acf52b80e69db292c5d7

First in a stack of PRs reworking the SDK internals to decouple MCP request handling from JSON-RPC framing. This PR lands only the abstraction and an in-memory implementation; nothing is wired into `BaseSession`/`Server` yet.

## Motivation and Context

The current `BaseSession` hard-couples three concerns: JSON-RPC framing, request/response correlation, and MCP semantics. That makes alternative transports (gRPC, in-process) impossible without smuggling JSON through them, and makes the receive loop hard to follow and test.

The `Dispatcher` Protocol is the call/return boundary: `send_request(method, params) -> dict`, `notify(method, params)`, and `run(on_request, on_notify)` to drive the inbound side. Method names are strings, params/results are dicts. MCP types live above it; wire encoding lives below it.

`Outbound` is a two-method base Protocol (`send_request` + `notify`) that both `Dispatcher` and `DispatchContext` extend. It names the surface that `PeerMixin` (a later PR) will wrap to provide typed MCP methods — one Protocol covers both top-level outbound and the per-request back-channel, with no aliases or adapter classes.

`DirectDispatcher` is an in-memory implementation that wires two peers with no transport. It serves as the second-impl proof for the Protocol and as a fast test substrate for the layers that will sit above the dispatcher in later PRs.

### Files

- `shared/dispatcher.py` — `Outbound`, `Dispatcher`, `DispatchContext` Protocols; `CallOptions`, `OnRequest`/`OnNotify`, `ProgressFnT`, `DispatchMiddleware`
- `shared/transport_context.py` — `TransportContext` base dataclass
- `shared/direct_dispatcher.py` — in-memory `Dispatcher` impl
- `shared/exceptions.py` — `NoBackChannelError(MCPError)` for transports without a server-to-client request channel
- `types/` — `REQUEST_CANCELLED` SDK error code

## How Has This Been Tested?

`tests/shared/test_dispatcher.py` — 13 behavioral tests covering the `send_request` contract (round-trip, MCPError passthrough, exception normalization, timeout), `notify`, the `DispatchContext` back-channel (`send_request`, `notify`, `progress`, `NoBackChannelError`), and the `run`/`close` lifecycle. 100% coverage on the new modules.

## Breaking Changes

None. New code only; nothing existing is touched beyond adding `NoBackChannelError` and the `REQUEST_CANCELLED` constant.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

Stack:
1. **(this PR)** Dispatcher Protocol + DirectDispatcher
2. JSONRPCDispatcher
3. BaseContext / Context / Peer / Connection
4. ServerRunner

`send_request` contract: returns `dict[str, Any]` or raises `MCPError`. Implementations normalize all handler exceptions to `MCPError` so callers see one exception type. Timeout maps to `MCPError(REQUEST_TIMEOUT)`.


tests/shared/test_dispatcher.py
```diff
@@ -0,0 +1,251 @@
...
+        tg.start_soon(client.run, client_on_request or c_req, client_on_notify or c_notify)
+        tg.start_soon(server.run, server_on_request or s_req, server_on_notify or s_notify)
+        try:
+            yield client, server, client_rec, server_rec
+        finally:
+            client.close()
+            server.close()
+
+
+@pytest.mark.anyio
```
maxisbey, line 69: mark this once at the top of the file

Kludex:
@maxisbey I need more context here. The reference design document is huge, and I was not present in the discussions about this. What is the motivation for it, and what problem does it try to solve?

---

Also, please remove the double `.

maxisbey:
> @maxisbey I need more context here. The reference design document is huge, and I was not present in the discussions about this. What is the motivation for it, and what problem does it try to solve?
> 
> Also, please remove the double `.

This started from looking into replacing SessionMessage and allowing for pluggable transports. The main problem being that everything in the SDK has been tied into the BaseSession/ServerSession/ClientSession class monolith with three main issues:

- It's very spaghetti
- It is hard coupled to JSONRPC, making it not possible to do pluggable transports
- Hard to test with a lot of state in one class

So, with the initial draft design I put up of a dispatcher I was wanting to continue from there and do it properly and add actual distinct layers to the SDK for Transport (shttp, stdio, etc.), wire (jsonrpc, which will be the Dispatcher here, and hopefully much easier to swap out for something like gRPC), and then MCP layer (which will be the ServerRunner and Server class layer).

The other thing coming along with this in a later PR is a cleaned up Context object which can properly represent the different paths of communication server -> client and what not. So you can send via the response stream to a tool call, or via the GET stream in the current spec.

That's the main motivation though :)

With the design doc it is big, but the WALKTHROUGH in that link is the better file to read TBH. Shorter and more clearly explains what I was thinking.

Kludex:
> It is hard coupled to JSONRPC, making it not possible to do pluggable transports

This seems the real motivation from the bullet list. Why not having a parameter onto the transport that would be passed down to the `BaseSession`?

src/mcp/shared/dispatcher.py
```diff
@@ -0,0 +1,146 @@
...
+
+It is deliberately *not* MCP-aware. Method names are strings, params and
+results are ``dict[str, Any]``. The MCP type layer (request/result models,
+capability negotiation, ``Context``) sits above this; the wire encoding
+(JSON-RPC, gRPC, in-process direct calls) sits below it.
+
+See ``JSONRPCDispatcher`` for the production implementation and
+``DirectDispatcher`` for an in-memory implementation used in tests and for
+embedding a server in-process.
+"""
```
Kludex, line 17: Please single `, not double.

src/mcp/shared/transport_context.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@dataclass(kw_only=True, frozen=True)
+class TransportContext:
+    """Base transport metadata for an inbound message.
+
+    Subclass per transport and add fields as needed. Instances are immutable.
+    """
+
+    kind: str
```
Kludex, line 21:
```suggestion
    kind: Literal['stdio', 'streamable-http'] | str
```

src/mcp/shared/direct_dispatcher.py
```diff
@@ -0,0 +1,173 @@
...
+) -> tuple[DirectDispatcher, DirectDispatcher]:
+    """Create two `DirectDispatcher` instances wired to each other.
+
+    Args:
+        can_send_request: Sets `TransportContext.can_send_request` on both
+            sides. Pass ``False`` to simulate a transport with no back-channel.
+
+    Returns:
+        A ``(left, right)`` pair. Conventionally ``left`` is the client side
+        and ``right`` is the server side, but the wiring is symmetric.
```
Kludex, line 166: So there is no validation of messages according to this?

It's far easier to understand to have client/server instead of left/right.

src/mcp/shared/transport_context.py
```diff
@@ -0,0 +1,30 @@
...
+    """Short identifier for the transport (e.g. ``"stdio"``, ``"streamable-http"``)."""
+
+    can_send_request: bool
+    """Whether the transport can deliver server-initiated requests to the peer.
+
+    ``False`` for stateless HTTP and HTTP with JSON response mode; ``True`` for
+    stdio, SSE, and stateful streamable HTTP. When ``False``,
+    `DispatchContext.send_raw_request` raises `NoBackChannelError`.
+    """
```
Kludex, line 30: This is not used in this PR, can you remove it, please?

tests/shared/test_dispatcher.py
```diff
@@ -0,0 +1,251 @@
...
+async def test_send_raw_request_wraps_non_mcperror_exception_as_internal_error():
+    async def on_request(
+        ctx: DispatchContext[TransportContext], method: str, params: Mapping[str, Any] | None
+    ) -> dict[str, Any]:
+        raise ValueError("oops")
+
+    async with running_pair(server_on_request=on_request) as (client, *_):
+        with anyio.fail_after(5), pytest.raises(MCPError) as exc:
```
Kludex, line 100: 5 seconds is too much to be a limit.

Kludex (commented):
I've stopped reviewing. I don't think it makes sense for me to continue without proper context.

PR #2456, opened 2026-04-16
fix(stdio_client): tolerate invalid UTF-8 from child stdout

shaun0927:
Fixes #2454

## Summary

`stdio_client()` currently decodes child stdout with `encoding_error_handler="strict"`, so a malformed UTF-8 line can raise during `TextReceiveStream(...)` iteration and crash the transport task group.

This PR makes the client follow the same strategy that `stdio_server()` already uses after #2302:

- decode with `errors="replace"`
- let malformed lines fail JSON validation
- surface those failures as in-stream exceptions instead of killing the transport

It also treats abrupt child shutdowns (`BrokenResourceError` / `ConnectionResetError`) as normal shutdown-time conditions in the background stdio tasks.

## Motivation and Context

PR #2302 fixed the server side so invalid UTF-8 on stdin becomes a parse error instead of a transport crash. The client side still behaved asymmetrically: a buggy or non-compliant child process could kill the Python client with a single malformed stdout line even if the next line was valid JSON-RPC.

That makes the client less robust than the server for the same class of malformed stdio input.

## How Has This Been Tested?

Added a focused regression test in `tests/client/test_stdio.py` that spawns a child process which writes:

1. one malformed UTF-8 line (`b"\xff\xfe\n"`)
2. one valid JSON-RPC line

Without this patch, the transport fails before the valid line is delivered.

With this patch:
- the first item from the read stream is an `Exception`
- the second item is the valid `SessionMessage`

Local verification:
- `uv run --frozen pytest tests/client/test_stdio.py`
- `uv run --frozen ruff check src/mcp/client/stdio.py tests/client/test_stdio.py`
- `uv run --frozen pyright src/mcp/client/stdio.py`

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -151,7 +151,7 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                     for line in lines:
                         try:
                             message = types.jsonrpc_message_adapter.validate_json(line, by_name=False)
-                        except Exception as exc:  # pragma: no cover
+                        except Exception as exc:
                             logger.exception("Failed to parse JSONRPC message from server")
                             await read_stream_writer.send(exc)
                             continue
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -92,7 +92,7 @@ class StdioServerParameters(BaseModel):
     Defaults to utf-8.
     """
 
-    encoding_error_handler: Literal["strict", "ignore", "replace"] = "strict"
+    encoding_error_handler: Literal["strict", "ignore", "replace"] = "replace"
     """
     The text encoding error handler.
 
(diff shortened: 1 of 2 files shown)


shaun0927:
for additional signal: `mcp-claude[bot]` independently reproduced this on both `main` and `v1.x` and arrived at the same one-line default change (issue #2454 comment). all 25 CI checks are green; happy to address review feedback.

Bartok9:
Heads up @shaun0927 — this PR has gone conflicting after the stdio transport was refactored on `main`. I rebuilt the same root-cause fix (defaulting `encoding_error_handler` to `"replace"`) cleanly against current `main` with a mid-session regression test in #2873, crediting your original approach here. Happy to defer if you'd rather refresh this one instead.

PR #2468, opened 2026-04-18
feat(mcpserver): add Context.assert_within_roots for server-side roots enforcement

### LINKED ISSUE #2453: feat: Add roots enforcement utility to FastMCP (get_roots, assert_within_roots, @within_roots_check)
## Problem

FastMCP has no built-in way to enforce that file paths or resource URIs stay
within client-declared roots. The MCP spec defines a Roots capability so clients
can advertise which filesystem boundaries a server should respect — but nothing
in the SDK enforces those boundaries.

Server authors who want to honour roots today must:

- Manually call `ctx.session.list_roots()` on every tool invocation
- Write their own path normalisation and prefix-checking logic
- Repeat this across every tool that handles file paths

This makes it easy to accidentally bypass the security intent of the Roots
capability.

### Example of the gap

```python
# Client declares roots: ["/home/user/project"]
# But a tool call arrives with an arbitrary path:

@mcp.tool()
async def read_file(path: str, ctx: Context) -> str:
    with open(path) as f:           # No enforcement.
        return f.read()              # Happily reads /etc/passwd
                                     # or ../../.ssh/id_rsa.
```

The client declared a boundary. The SDK exposed that boundary via `ctx.session.list_roots()`. But nothing checked the incoming `path` against it — because that check is the server author's responsibility, and the SDK ships no helper for it.

## Proposed Solution

Add a utility module at `src/mcp/server/fastmcp/utilities/roots.py` that exposes:

**`get_roots(ctx)`** — async helper that fetches the current roots from the session.

**`assert_within_roots(path, roots)`** — validates a path is within at least one
declared root; raises `PermissionError` with a clear message if not.

**`@within_roots_check`** — decorator for tool functions that runs the above two
automatically, so a tool can be protected with one line.

## Example

```python
from mcp.server.fastmcp.utilities.roots import within_roots_check

mcp = FastMCP("file-server")

@mcp.tool()
@within_roots_check
async def read_file(path: str, ctx: Context) -> str:
    with open(path) as f:
        return f.read()
```

If `path` is outside any client-declared root, the decorator raises before the
body runs.

## Implementation Notes

- Roots are fetched via `ctx.session.list_roots()` (the existing SDK pattern)
- Path comparison uses `pathlib.Path.resolve()` to normalise symlinks and
  relative segments before prefix-checking
- The decorator uses `functools.wraps` to preserve the tool function's signature
- Follows the same pattern as other FastMCP utilities (`logging`, `dependencies`)

## Checklist

- [x] Searched existing issues — no duplicate found
- [ ] Implementation and tests written locally on branch `feat/roots-utility`
- [x] Will submit a PR referencing this issue

### References

_No response_

NeelakandanNC:
Closes #2453.

Adds `assert_within_roots(path)` as an async method on the `Context` class, so any FastMCP/MCPServer tool can enforce that a user-provided path stays within the filesystem boundaries the client declared via the Roots capability.

```python
@mcp.tool()
async def read_file(path: str, ctx: Context) -> str:
    await ctx.assert_within_roots(path)  # raises PermissionError if outside roots
    with open(path) as f:
        return f.read()
```

## Motivation and Context

MCP clients can declare roots, but the SDK never enforces them server-side. Any tool can open any path regardless of what the client declared — the check is left entirely to the tool author. This is a real security gap for filesystem-exposing servers and easy to forget.

The reference `@modelcontextprotocol/server-filesystem` has enforcement baked in, but anyone building a custom FastMCP/MCPServer doesn't get that for free.

**On the API design — responding to feedback on #2425:** an earlier attempt at this was closed with the feedback that the proposed API (a `@within_roots_check` decorator in a separate utility module) was "not easy to use." This PR takes a different approach based on that feedback:

- No decorator, no magic — the check is an explicit line at the top of the tool body
- Method lives on `Context` alongside `ctx.report_progress`, `ctx.read_resource`, `ctx.elicit`, so it's discoverable by anyone already writing tools that take a `Context`
- Single call site, clear failure mode — raises `PermissionError` with the offending path, no configuration, no opt-in

**Behavior:**

- Resolves the target path with `pathlib.Path.resolve()` — symlinks and relative segments are normalized before comparison
- Iterates declared roots via `self.request_context.session.list_roots()` and accepts the path if it falls within any of them
- Raises `PermissionError` if the path is outside every declared root, **or** if no roots are declared (fail-closed)
- Uses `urllib.request.url2pathname` to convert `file://` URIs — works on both POSIX and Windows

## How Has This Been Tested?

Five new tests in `tests/server/mcpserver/test_roots.py`, exercising the real client/server path (no mocks on `list_roots`):

- Path inside a declared root passes
- Path outside all declared roots raises
- Empty roots list raises
- Symlink inside a root pointing outside raises (validates `.resolve()` behavior)
- Path inside any one of multiple declared roots passes

All five pass locally. Full suite (`pytest tests/`) is also green — 1166 passed, 98 skipped, 1 xfailed (existing state, unchanged by this PR).

## Breaking Changes

None. This is a purely additive change — a new method on an existing class. Existing tools are unaffected; the check is opt-in per tool.

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the MCP Documentation
- [x] I have added or updated documentation as needed (docstring on the new method with usage example)

## Additional context

The docstring on `assert_within_roots` includes a full usage example that matches the patterns already used in `report_progress`, `read_resource`, etc., so the method is self-documenting for anyone using hover tooltips or generated API docs.

### DIFF
--- a/src/mcp/server/mcpserver/context.py
+++ b/src/mcp/server/mcpserver/context.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Generic
+from urllib.parse import urlparse
+from urllib.request import url2pathname
 
 from pydantic import AnyUrl, BaseModel
 
@@ -116,6 +119,42 @@ class Context(BaseModel, Generic[LifespanContextT, RequestT]):
         """
         assert self._mcp_server is not None, "Context is not available outside of a request"
         return await self._mcp_server.read_resource(uri, self)
+
+    async def assert_within_roots(self, path: str | Path) -> None:
+        """Assert that a filesystem path is within the client's declared roots.
+
+        Provides server-side enforcement of the filesystem boundaries declared by
+        the client via the Roots capability. Call this at the start of any tool
+        that accepts a user-provided path, to prevent the tool from accessing
+        files outside the client's declared scope.
+
+        Args:
+            path: The filesystem path to validate. Accepts a string or Path.
+                Relative paths and symlinks are resolved before comparison.
+
+        Raises:
+            PermissionError: If the path is outside every declared root, or if
+                the client has declared no roots.
+
+        Example:
+            ```python
+            @server.tool()
+            async def read_file(path: str, ctx: Context) -> str:
+                await ctx.assert_within_roots(path)
+                with open(path) as f:
+                    return f.read()
+            ```
+        """
+        target = Path(path).resolve()
+
+        result = await self.request_context.session.list_roots()
+
+        for root in result.roots:
+            root_path = Path(url2pathname(urlparse(str(root.uri)).path)).resolve()
+            if target.is_relative_to(root_path):
+                return
+
+        raise PermissionError(f"Path {target} is not within any declared root")
 
     async def elicit(
         self,
--- a/tests/server/mcpserver/test_roots.py
+++ b/tests/server/mcpserver/test_roots.py
@@ -52,7 +52,7 @@ async def test_path_outside_roots_raises(tmp_path: Path):
     @server.tool("check")
     async def check(context: Context, path: str) -> bool:
         await context.assert_within_roots(path)
-        return True
+        return True  # pragma: no cover
 
     callback = _make_callback([Root(uri=FileUrl(f"file://{root_dir}"))])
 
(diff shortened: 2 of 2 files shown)


maxisbey:
Thanks for the PR. The v2 rework changed the code this targets, so it no longer applies as-is. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2472, opened 2026-04-19
test: regression test for initialize hang on unexpected content-type

### LINKED ISSUE #2432: Initialize call hangs forever if MCP server does not return a `Content-Type: text/plain`

Christian-Sidak:
## Summary

- Adds a regression test (`test_initialize_does_not_hang_on_unexpected_content_type`) that directly reproduces the hang described in issue #2432
- The test creates a minimal HTTP server that returns `Content-Type: text/plain` for **all** requests (including `initialize`) and asserts that `session.initialize()` raises `MCPError` immediately instead of blocking forever

The underlying fix is already present on `main` — `_handle_post_request` sends a proper `JSONRPCError` (with the matching request ID) when it encounters an unexpected content type, which lets `ClientSession` resolve the pending request instead of waiting indefinitely. This PR adds the missing test that exercises that code path for the `initialize` call specifically.

## Test plan

- [ ] `tests/client/test_notification_response.py::test_initialize_does_not_hang_on_unexpected_content_type` — new test passes
- [ ] All existing `test_notification_response.py` tests continue to pass

Fixes #2432

### DIFF
--- a/tests/client/test_notification_response.py
+++ b/tests/client/test_notification_response.py
@@ -71,6 +71,21 @@ def _create_unexpected_content_type_app() -> Starlette:
     return Starlette(debug=True, routes=[Route("/mcp", handle_mcp_request, methods=["POST"])])
 
 
+def _create_plain_text_server_app() -> Starlette:
+    """Create a server that returns text/plain for all requests, including initialize.
+
+    This reproduces the scenario from issue #2432 where a misconfigured server
+    returns an unexpected content type on the initialize call, causing the client
+    to hang forever.
+    """
+
+    async def handle_mcp_request(request: Request) -> Response:
+        # Always return text/plain — never a valid MCP response.
+        return Response(content="this is not json", status_code=200, media_type="text/plain")
+
+    return Starlette(debug=True, routes=[Route("/mcp", handle_mcp_request, methods=["POST"])])
+
+
 async def test_non_compliant_notification_response() -> None:
     """Verify the client ignores unexpected responses to notifications.
 
@@ -114,6 +129,20 @@ async def test_unexpected_content_type_sends_jsonrpc_error() -> None:
 
                 with pytest.raises(MCPError, match="Unexpected content type: text/plain"):  # pragma: no branch
                     await session.list_tools()
+
+
+async def test_initialize_does_not_hang_on_unexpected_content_type() -> None:
+    """Verify that initialize() raises MCPError immediately when server returns wrong content type.
+
+    Regression test for issue #2432: when a misconfigured server returns a content type
+    other than application/json or text/event-stream in response to the initialize request,
+    the client must raise MCPError right away instead of hanging forever.
+    """
+    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=_create_plain_text_server_app())) as client:
+        async with streamable_http_client("http://localhost/mcp", http_client=client) as (read_stream, write_stream):
+            async with ClientSession(read_stream, write_stream) as session:  # pragma: no branch
+                with pytest.raises(MCPError, match="Unexpected content type: text/plain"):  # pragma: no branch
+                    await session.initialize()
 
 
 def _create_http_error_app(error_status: int, *, error_on_notifications: bool = False) -> Starlette:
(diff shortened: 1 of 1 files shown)


Christian-Sidak:

Christian-Sidak:

MukundaKatta:
Does the same fix path also cover a server returning 200 with an empty body, or 200 with valid JSON for a different request id? Both have the same "initialize never resolves" shape, so if the fix is at the right layer they should unblock without new code. If those cases are tested elsewhere already, ignore.

Christian-Sidak:
No — this test is scoped to the unexpected content-type path only. The hang in issue #2432 is triggered when the SDK reaches the `if content_type not in ("application/json", "text/event-stream")` guard and raises `McpError`; the `initialize()` future is resolved by that error propagation.

The two cases you describe follow different paths:
- **200 with empty body**: content-type would still be `application/json` (or absent), so it bypasses the content-type check. The body JSON-parse would fail or produce an empty response, which is a separate code path — likely a different hang or error class.
- **200 with valid JSON for a different request id**: the response matcher would never complete the outstanding `initialize` future since the id doesn't match, so the session would hang differently — at the response-matching layer, not the content-type layer.

Both are valid "initialize never resolves" failure modes but would need separate tests against the relevant code paths. Happy to add them here or in a follow-up PR if that's useful.

Christian-Sidak:

Christian-Sidak:
Rebased onto current `main`: the streamable-HTTP test module has since moved from `httpx`/`mcp.types` to `httpx2`/`mcp_types` (repo-wide rename), so the new regression test needed the same update to avoid a `NameError` once merged. Content/behavior of the added test is unchanged.

Christian-Sidak:
Re cubic's review: the `httpx2`/`mcp_types` imports (flagged P1) are correct as-is — `main` already renamed these (see `pyproject.toml`'s `httpx2>=2.5.0` dependency and `src/mcp-types/mcp_types`), so this matches the current tree, not a regression.

Addressed the P2 point: wrapped the `initialize()` await in `test_initialize_does_not_hang_on_unexpected_content_type` with `anyio.fail_after(5)` so a reintroduced hang fails the test deterministically instead of stalling the run. Pushed as 5ef8194.

Christian-Sidak:

maxisbey:
Thanks for the PR. Equivalent coverage is already on `main`: `test_unexpected_content_type_sends_jsonrpc_error` landed with the #2005 fix, and the branch here has since conflicted with the httpx2/mcp_types renames. Closing this as part of a wider backlog cleanup following the v2 launch. If you think the initialize-triggered variant covers a genuinely distinct path, feel free to reopen and say so.

PR #2475, opened 2026-04-19
Return 405 for GET in stateless mode

MukundaKatta:
## Summary
- return 405 with `Allow: POST` for standalone GET requests when the server is running in stateless mode
- keep the existing stateful GET SSE behavior unchanged
- add an end-to-end regression test for stateless GET requests

## Testing
- `python3 -m py_compile src/mcp/server/streamable_http.py tests/shared/test_streamable_http.py`
- `pytest tests/shared/test_streamable_http.py -k "stateless or method_not_allowed or get_validation"` *(not run locally: `pytest` is not installed in this environment)*

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -658,6 +658,18 @@ class StreamableHTTPServerTransport:
         if writer is None:  # pragma: no cover
             raise ValueError("No read stream writer available. Ensure connect() is called first.")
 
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2509, which short-circuits at the manager layer (before a transport is created), also handles DELETE, and has green CI. Thanks for the contribution.

PR #1436, opened 2025-10-06
[DOC] simple mcp client with sampling capability

yarnabrina:
## Motivation and Context

The is no documented MCP client example how to support MCP servers using sampling capability. This PR attempts to add a basic example.

## How Has This Been Tested?

This changes are subset of [my personal repository](https://github.com/yarnabrina/learn-model-context-protocol). It is not tested in a production system, but thoroughly tested locally.

## Breaking Changes

No

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

Related #1205 
Closes #539

Issue #1205, opened 2025-07-27
Controlling Context in Client for Sampling Requests


yarnabrina:
### Question

I am exploring the `sampling` components in MCP. Based on my understanding of the protocol, this should allow the server to utilise the LLM(s) available on the client side. Unfortunately, none of the examples provide a working demonstration of this.

I have managed to get something working [here](https://github.com/yarnabrina/learn-model-context-protocol/blob/770728148a8aa423a1467785bdb695927cf3428d/src/mcp_learning/mcp_client/orchestrator.py#L253-L334), but I had to invoke the LLM directly using only part of the information from the `CreateMessageRequestParams` object. I have not been able to determine the purpose of `RequestContext` or how it can be leveraged.

In summary, these are my questions:

1. How can I access the LLM(s) available to the client directly, rather than making an explicit call?
2. How can I use `modelPreferences` to suggest that the client changes the choice of LLM based on priorities or name hints?
3. How can I use the `RequestContext` to pass available context to the LLM for improved generation?
4. How can I use the `includeContext` attribute of `CreateMessageRequestParams` to filter what information is included in the context?

Any guidance or examples would be greatly appreciated.

### Additional Context

- MCP Python SDK version: `1.12.2`
- Python version: `3.13.5`
- Operating System: Ubuntu 22.04.5 LTS on Windows 11 Home via WSL 2

yarnabrina:
@dsp-ant @ihrpr @Kludex may I kindly request you to have a look at this question when you have a moment? I'd appreciate your inputs.

Kludex:
> I am exploring the sampling components in MCP. Based on my understanding of the protocol, this should allow the server to utilise the LLM(s) available on the client side. Unfortunately, none of the examples provide a working demonstration of this.

Check https://ai.pydantic.dev/mcp/server/#mcp-sampling.

yarnabrina:
Thanks for the response. Sorry if I was unclear, what I meant was none of the examples actually show how does the MCP client callback calls already configured LLM (or use hints to switch the model). I am not using `pydantic-ai` yet, but in that example also assumes a LLM response as a TODO comment only.

As linked, I do have a sample implementation, but I am not sure if this is the intended way to use as I failed to use some properties, some guidance will be much appreciated.

https://github.com/yarnabrina/learn-model-context-protocol/blob/770728148a8aa423a1467785bdb695927cf3428d/src/mcp_learning/mcp_client/orchestrator.py#L253-L334

[...]

PR #2476, opened 2026-04-19
feat(examples): add simple-sampling server and client

trentisiete:
Adds an end-to-end sampling example to `examples/`, addressing #1205.

Two new workspace members:

- `examples/servers/simple-sampling/` — minimal server with one tool (`write_story`) that calls `session.create_message(...)` populating every advisory field of `CreateMessageRequestParams`: model preferences with hints + cost/speed/intelligence priorities, system prompt, temperature, stop sequences, include_context, metadata.
- `examples/clients/simple-sampling-client/` — wires `sampling_callback` to a real LLM via any OpenAI-compatible endpoint. Default is Groq (free tier). Switching to OpenAI / OpenRouter / Ollama (`/v1`) / vLLM is 1-3 env vars; no extra dependencies.

## Context

There's already #1436 from @yarnabrina addressing this issue. Their PR has been waiting on review for months and they've explicitly given the green light for an independent PR (see the recent comment thread on #1205). This one tries to land the same goal while incorporating the feedback their PR received and never got a chance to resolve:

- No dependency on the `openai` SDK; uses `httpx` against the OpenAI-compatible `/chat/completions` schema.
- `LLMClient` abstraction matching the `simple-chatbot` pattern that was requested in their review — the sampling callback knows nothing about httpx, and the LLM wrapper knows nothing about MCP.

## Design notes

- Model selection: first usable `ModelHint.name` is treated as a soft override, falling back to `LLM_MODEL`. The numeric priorities (cost/speed/intelligence) get logged but not used for routing — picking a model from those would require a provider-specific catalog, which is out of scope for an example.
- LLM failures return `types.ErrorData(code=INTERNAL_ERROR, ...)` rather than raising, so the server gets a readable error instead of a transport-level one.
- Non-text content blocks (image/audio) become `[<type> content omitted]` placeholders rather than being silently dropped. A production client would forward or explicitly reject them.

## Verification

End-to-end check against a local `ThreadingHTTPServer` stub of `/chat/completions` that captures the request body and returns a canned story. Asserts that:

- The selected model is the first hint (`llama-3.1-8b`).
- `max_tokens=200`, `temperature=0.8`, `stop=["THE END"]`, the system prompt, the user message and the metadata round-trip correctly.
- The canned response surfaces in the client stdout.

The verification harness itself is not part of the PR. Repo checks pass:

- `uv run ruff format --check`: clean
- `uv run ruff check`: 0 issues
- `uv run pyright`: 0 errors, 0 warnings, 0 informations

## Try it

    export LLM_API_KEY=<your_key>
    uv sync --all-packages
    uv run mcp-simple-sampling-client --topic "a lighthouse keeper"

Full env var table in `examples/clients/simple-sampling-client/README.md`.

### DIFF
--- a/examples/clients/simple-sampling-client/README.md
+++ b/examples/clients/simple-sampling-client/README.md
@@ -1,0 +1,67 @@
+# mcp-simple-sampling-client
+
+Companion client for
+[`examples/servers/simple-sampling`](../../servers/simple-sampling).
+Implements a `sampling_callback` that fulfils server-initiated LLM
+requests and honours every advisory field in
+`CreateMessageRequestParams`.
+
+## What this example shows
+
+The MCP Python SDK already provides the wiring for sampling, but there
+was no end-to-end example of a client that plugs a real LLM into
+`ClientSession(sampling_callback=...)`. This client does that and
+illustrates:
+
+- Mapping `SamplingMessage` role/content into an OpenAI-compatible chat
+  payload.
+- Treating `modelPreferences.hints` as soft overrides for model
+  selection while logging the numeric priorities.
+- Forwarding `systemPrompt`, `temperature`, `maxTokens`,
+  `stopSequences` and `metadata` to the LLM provider.
+- Logging `includeContext` so users can see the hook point where a
+  multi-server client would inject session context.
+- Returning `ErrorData` on provider failure instead of letting the
+  exception propagate.
+
+## LLM provider
+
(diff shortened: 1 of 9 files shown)


maxisbey:
Thanks for the PR. The v2 rework changed the code this targets, so it no longer applies as-is. I'm closing this out as part of a backlog cleanup now that v2 is out. Feel free to reopen if this is still relevant.

PR #2478, opened 2026-04-20
fix(shared): preserve MCPError payload on pickle

MukundaKatta:
## Summary
- make MCPError pickle-safe by reconstructing exceptions from the full ErrorData payload instead of bare Exception args
- preserve the specialized UrlElicitationRequiredError shape during round-trip reconstruction
- add regression tests for pickling both MCPError and UrlElicitationRequiredError

## Testing
- python3 -m py_compile src/mcp/shared/exceptions.py tests/shared/test_exceptions.py
- python3 -m pytest tests/shared/test_exceptions.py -k pickle_roundtrip
- PYTHONPATH=src python3 -m pytest tests/shared/test_exceptions.py -k pickle_roundtrip

Local note: pytest is blocked in this desktop environment because the available interpreter bottoms out before the repo's required typing features (ImportError on TypeAlias from Python 3.9).

### DIFF
--- a/src/mcp/shared/exceptions.py
+++ b/src/mcp/shared/exceptions.py
@@ -40,7 +40,7 @@ class MCPError(Exception):
 
     @property
     def data(self) -> Any:
-        return self.error.data  # pragma: no cover
+        return self.error.data
 
     @classmethod
     def from_jsonrpc_error(cls, error: JSONRPCError) -> MCPError:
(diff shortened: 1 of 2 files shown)


MukundaKatta:
CI was failing on coverage (99.98% vs required 100%). The gap was from the defensive `hasattr(exc_type, 'from_error_data')` / `exc_type.__new__(...)` fallback branch — unreachable in practice because every `MCPError` subclass inherits `from_error_data` from the base class.

Simplified the reconstructor to two lines. Behavior is unchanged for all three cases verified locally (MCPError, UrlElicitationRequiredError, and an arbitrary MCPError subclass); the existing 11 tests in `tests/shared/test_exceptions.py` all pass.

maxisbey:
Thanks for the PR. Tracking this in #2555 instead. Closing this as part of a wider backlog cleanup following the v2 launch. If it's still relevant, feel free to reopen.

PR #2479, opened 2026-04-20
fix: auto-send notifications/cancelled on anyio task cancellation

### LINKED ISSUE #1410: Cancelling a Call Tool Request from Client
### Question

Hey Team,
I have a question regarding cancelling the tool call from client when the MCP server is created from `FastMCP` class. 

I see the test - [test_cancel_handling](https://github.com/modelcontextprotocol/python-sdk/blob/main/tests/server/test_cancel_handling.py) but this one uses low level server.

My question is specifically on how we can get the `request_id` which the client needs for cancelling the call tool request on server. 
```
await client.send_notification(
        ClientNotification(
            CancelledNotification(
                    requestId={request_id},
                    reason="Reson",
              ),
          )
     )
)
```

Do we have to get the request id from `call_tool` request? like below?

```py
  request_id = None
  mcp = FastMCP(
      name="Simple gRPC Server",
  )
  @mcp.tool()
  def get_temperature(city: str, ctx: Context[ServerSession, None]) -> float:
      # fetching request id from context 
      request_id = ctx.request_context.request_id
      return 22.5
  
```

Or is there another way?

### Additional Context

_No response_

cubaseuser123:
Fixes #1410.

## Motivation and Context

The friction here is pretty specific: when you use high-level request wrappers like `client.call_tool()`, the generated `request_id` stays hidden internally inside `send_request`. So if you need to abort that task from the outside using standard Python async patterns (like `anyio.fail_after` or `CancelScope.cancel()`), the client aborts locally, but it has no way to tell the server. The server continues executing an orphaned request that the client has already stopped waiting for. 

So I moved the cancellation hook into `BaseSession.send_request`. The idea is simple: intercept the `anyio` cancellation before it bubbles up, read the `request_id` we just generated, and automatically dispatch a `notifications/cancelled` straight to the peer. 

## How Has This Been Tested?

I made sure it runs cleanly end-to-end. Added `test_anyio_cancel_scope_sends_cancelled_notification` to `tests/server/test_cancel_handling.py` which:
- Kicks off `call_tool`
- Triggers an outer `CancelScope.cancel()` 
- Asserts the server handler actually got aborted

(Side note: Verified locally against the full regression suite using `anyio` 4.10.0 on Windows.)

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

A few quick implementation details:
- **It uses a `request_sent` flag:** Cancels are only dispatched if the request actually hit the wire (so we don't spam the server for requests that failed on `_write_stream.send()`).
- **It's shielded:** The dispatch runs inside `with anyio.CancelScope(shield=True)` so it survives the very cancellation it's reacting to.
- **It has a deadlock timeout:** Added `with anyio.move_on_after(2.0)` inside the shield. If the transport buffer is completely full, we don't want to hang the user's cancellation handler forever. 
- **It's bidirectional:** Since this lives in `BaseSession`, it automatically works for both client→server and server→client directions.

Happy to tweak the timeout value or the notification reason string if there's a preference. Let me know if I should squash anything.

Just a heads-up for reviewers — the CI failures here are pre-existing and not related to the changes in this PR. 

The `pyright` hook in `.pre-commit-config.yaml` runs against the whole repo. Currently, the upstream `main` branch (commit `3d7b311`) has 21 Pyright errors located entirely in `src/mcp/os/win32/utilities.py`. 

Because pre-commit is configured with `fail_fast: true`, it aborts early and cascades all 20 test matrix jobs (`checks / test`) to fail with exit code 2 before Pytest even gets a chance to run. (I noticed this is happening on other recent PRs like #2475 as well).

I have verified the test suite locally using parallel workers (`pytest -n auto`):
- **1170 passed, 101 skipped, 1 xfailed**, exit 0.
- `uv run pyright src/mcp/shared/session.py` returns **0 errors**.

Happy to rebase to trigger a clean run once the upstream Pyright issue on `main` is resolved, or if a maintainer wants to merge this as-is!

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -23,6 +23,7 @@ from mcp.types import (
     INVALID_PARAMS,
     REQUEST_TIMEOUT,
     CancelledNotification,
+    CancelledNotificationParams,
     ClientNotification,
     ClientRequest,
     ClientResult,
@@ -269,6 +270,7 @@ class BaseSession(
             # Store the callback for this request
             self._progress_callbacks[request_id] = progress_callback
 
+        request_sent = False
         try:
             target = request_data.get("params", {}).get("name")
             span_name = f"MCP send {request.method} {target}" if target else f"MCP send {request.method}"
@@ -284,6 +286,7 @@ class BaseSession(
 
                 jsonrpc_request = JSONRPCRequest(jsonrpc="2.0", id=request_id, **request_data)
                 await self._write_stream.send(SessionMessage(message=jsonrpc_request, metadata=metadata))
+                request_sent = True
 
                 # request read timeout takes precedence over session read timeout
                 timeout = request_read_timeout_seconds or self._session_read_timeout_seconds
@@ -300,6 +303,26 @@ class BaseSession(
                     raise MCPError.from_jsonrpc_error(response_or_error)
                 else:
                     return result_type.model_validate(response_or_error.result, by_name=False)
+
+        except anyio.get_cancelled_exc_class():
+            # Automatically notify the other side when a task/scope is cancelled,
+            # so the peer can abort work for a request nobody is waiting for.
+            if request_sent:
+                with anyio.CancelScope(shield=True):
+                    try:
+                        # Add a short timeout to prevent deadlock if the transport buffer is full
+                        with anyio.move_on_after(2.0):
+                            await self.send_notification(
+                                CancelledNotification(  # type: ignore[arg-type]
+                                    params=CancelledNotificationParams(
+                                        request_id=request_id,
+                                        reason="Task cancelled",
+                                    )
+                                )
+                            )
+                    except Exception:
+                        pass  # Transport may already be closed
+            raise
 
         finally:
             self._response_streams.pop(request_id, None)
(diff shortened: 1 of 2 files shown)


christianjohnmorral18:
The `request_id` visibility issue is a real friction point for anyone writing custom cancellation logic on top of the high-level wrappers. Having the SDK handle `notifications/cancelled` automatically on anyio cancellation makes the cancel path much more reliable — previously you had to inject your own cleanup hook at the task group level to get any notification out. One question: for servers that don't handle `notifications/cancelled`, is there a risk the auto-notification triggers a confusing error log on their side?

cubaseuser123:
This was something I thought about during implementation so:

`notifications/cancelled` is part of the base MCP spec, and in this SDK it's handled internally by `BaseSession`'s message loop (not routed to user-level handlers). When the notification arrives, the receiver checks `if cancelled_id in self._in_flight`. So if the request already finished or was never tracked, it's a no-op. No log, no error, nothing fires.

For servers built on other SDKs or custom implementations, JSON-RPC notifications are fire-and-forget by design,the spec explicitly says a server must not reply to a notification, so even if a server doesn't recognize `notifications/cancelled`, there's no error response path. A well-behaved implementation should just silently ignore it.

(Side note: this is exactly why I threw in that 2.0s `move_on_after` timeout. If the transport gets choked up or starts closing, we don't want to hang the whole thing waiting to send a notification. We just drop the attempt and let the original cancellation bubble up. So at worst, the notification just never makes it, rather than blowing anything up on the other end.)

maxisbey:
Closing in favor of #2514, which covers both the timeout path and the caller-cancellation path with regression tests. Thanks for the contribution.

PR #2481, opened 2026-04-20
fix: stop sending JSON-RPC response on request cancellation

### LINKED ISSUE #2480: RequestResponder.cancel sends JSON-RPC response, violating cancellation spec
## Description

`RequestResponder.cancel` in `mcp/shared/session.py` sends an `ErrorData(code=0, message="Request cancelled")` JSON-RPC response back to the sender after a `notifications/cancelled` is received. The cancellation spec explicitly says receivers SHOULD NOT do this:

> Receivers of cancellation notifications **SHOULD**:
> - Stop processing the cancelled request
> - Free associated resources
> - **Not send a response for the cancelled request**

— https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/cancellation.mdx#L33-L36

## Reproduction

`mcp` 1.27.0, Python 3.11. Driving the server over stdio:

1. Send `initialize` + `notifications/initialized`.
2. Send any `tools/call` whose handler enters an `await` (e.g. `await asyncio.sleep`).
3. Send `notifications/cancelled` with the request id.
4. Observe stdout: an unsolicited response is sent for the cancelled request.

```json
{"jsonrpc": "2.0", "id": 99, "error": {"code": 0, "message": "Request cancelled"}}
```

Source:
https://github.com/modelcontextprotocol/python-sdk/blob/v1.27.0/src/mcp/shared/session.py#L137-L150

```python
async def cancel(self) -> None:
    ...
    self._cancel_scope.cancel()
    self._completed = True
    # Send an error response to indicate cancellation
        request_id=self.request_id,
        response=ErrorData(code=0, message="Request cancelled", data=None),
    )
```

## Impact

Strict clients treat the unexpected response as an unknown message id and drop the transport. Claude Code, for example, logs `Received a response for an unknown message ID`, closes the stdio transport, and reconnects. With long-running tools (custom polling tools, long DB queries, etc.) every cancellation costs a reconnect cycle.

## Suggested fix

Drop the `_send_response` call from `RequestResponder.cancel`. The cancel scope is already cancelled and `_completed` is already set, so the request is correctly cleaned up server-side without violating the spec.

## Workaround

Until this lands, downstream servers can monkey-patch `RequestResponder.cancel`:

```python
from mcp.shared.session import RequestResponder

async def _cancel_without_response(self) -> None:
    if not self._entered:
    if not self._cancel_scope:
    self._cancel_scope.cancel()
    self._completed = True

RequestResponder.cancel = _cancel_without_response
```

## Environment
- mcp 1.27.0
- Python 3.11
- stdio transport

flobo3:
## Summary

`RequestResponder.cancel()` in `mcp/shared/session.py` sent an `ErrorData(code=0, message="Request cancelled")` JSON-RPC response back to the sender after a `notifications/cancelled` was received. The [MCP cancellation spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/cancellation.mdx#L33-L36) explicitly states:

> Receivers of cancellation notifications SHOULD NOT send a response for the cancelled request.

This unexpected response caused strict clients (e.g. Claude Code) to treat it as an unknown message ID, close the transport, and reconnect — making every cancellation of a long-running tool cost a full reconnect cycle.

## Changes

- **`src/mcp/shared/session.py`**: Removed `_send_response()` call from `RequestResponder.cancel()`. Added spec reference comment.
- **`src/mcp/server/lowlevel/server.py`**: Updated log message and comment to reflect that no response is sent (was "duplicate response suppressed", now "no response sent per spec").
- **`tests/shared/test_session.py`**: Rewrote `test_request_cancellation` to verify the server handler's cancel scope is cancelled, rather than expecting an `MCPError` response.
- **`tests/server/test_cancel_handling.py`**: Added `request_read_timeout_seconds` to the cancelled request so it times out cleanly (no response is sent).

Fixes #2480

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -502,9 +502,9 @@ class Server(Generic[LifespanResultT]):
                     response = err.error
                 except anyio.get_cancelled_exc_class():
                     if message.cancelled:
-                        # Client sent CancelledNotification; responder.cancel() already
-                        # sent an error response, so skip the duplicate.
-                        logger.info("Request %s cancelled - duplicate response suppressed", message.request_id)
+                        # Client sent CancelledNotification; per the MCP spec the
+                        # receiver MUST NOT send a response for the cancelled request.
+                        logger.info("Request %s cancelled - no response sent per spec", message.request_id)
                         return
                     # Transport-close cancellation from the TG in run(); re-raise so the
                     # TG swallows its own cancellation.
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -148,11 +148,8 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
 
         self._cancel_scope.cancel()
         self._completed = True  # Mark as completed so it's removed from in_flight
-        # Send an error response to indicate cancellation
-        await self._session._send_response(  # type: ignore[reportPrivateUsage]
-            request_id=self.request_id,
-            response=ErrorData(code=0, message="Request cancelled"),
-        )
+        # Per the MCP spec, receivers of cancellation notifications SHOULD NOT
+        # send a response for the cancelled request.
 
     @property
     def in_flight(self) -> bool:  # pragma: no cover
--- a/tests/server/test_cancel_handling.py
+++ b/tests/server/test_cancel_handling.py
@@ -66,10 +66,11 @@ async def test_server_remains_functional_after_cancel():
                 await client.session.send_request(
                     CallToolRequest(params=CallToolRequestParams(name="test_tool", arguments={})),
                     CallToolResult,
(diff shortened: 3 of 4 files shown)


dwis-hacks:
This is the right fix — sending an `ErrorData` response after `notifications/cancelled` violates JSON-RPC 2.0 which says you should send at most one response per request. Having both a notification and an error response for the same request ID is the kind of thing that makes conformant clients confused about whether to expect more data. Does this change affect any of the existing integration tests for the cancellation flow?

maxisbey:
Closing in favor of #2493, which also resolves the sender's local pending response (matching the TypeScript SDK), so the caller doesn't hang until timeout when the response is no longer sent. Thanks for the contribution.

PR #2487, opened 2026-04-21
docs(sampling): clarify client-side sampling context

### LINKED ISSUE #1205: Controlling Context in Client for Sampling Requests

MukundaKatta:
## Summary
- explain that ctx.session.create_message(...) routes sampling through the connected client rather than exposing a direct server-side model handle
- document how to use model_preferences, include_context, and the client-side sampling callback together
- update the README callback example to use ClientRequestContext and sync the snippet source files with the clarified guidance

## Testing
- not run (docs/snippets only)

Closes #1205

### DIFF
--- a/README.md
+++ b/README.md
@@ -930,13 +930,16 @@ The `elicit()` method returns an `ElicitationResult` with:
 
 ### Sampling
 
-Tools can interact with LLMs through sampling (generating text):
+Tools can interact with LLMs through sampling (generating text). The server does
+not call a model directly through the SDK here; instead it sends a
+`sampling/createMessage` request to the connected client, and the client handles
+that request via its registered sampling callback:
 
 <!-- snippet-source examples/snippets/servers/sampling.py -->
 ```python
 from mcp.server.fastmcp import Context, FastMCP
 from mcp.server.session import ServerSession
-from mcp.types import SamplingMessage, TextContent
+from mcp.types import ModelHint, ModelPreferences, SamplingMessage, TextContent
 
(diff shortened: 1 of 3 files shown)


MukundaKatta:
Closing out — pre-commit failures and I'm not pushing updates. Thanks.

PR #2490, opened 2026-04-21
fix: dispatch client request handlers concurrently (#2489)

### LINKED ISSUE #2489: Concurrent server-side requests are serialized end-to-end by BaseSession
### Description

When a server tool issues multiple concurrent server to client requests over a single MCP session — for example, fanning out several `ServerSession.create_message(...)` (sampling) calls via `asyncio.gather` — only one is ever in flight on the receiving side: `BaseSession._receive_loop` awaits each incoming request handler inline before reading the next message off the stream, so subsequent responses can't even be dequeued until the current handler returns. A minimal reproducer using `mcp.server.fastmcp.FastMCP` + `ClientSession` over the SDK's own in-memory transport (5 concurrent `create_message` calls, sampling callback sleeps 0.5s) shows peak in-flight = 1 and elapsed ≈ 2.5s instead of ≈ 0.5s.

Concrete use case: a tool that fans out dozens of independent sampling requests per invocation against an LLM deployment with provisioned capacity sitting idle, the work is embarrassingly parallel and there's token budget to burn, but wall-clock scales linearly with the fan-out because every request blocks the next one. The same shape applies to elicitation, list-roots, and any other server to client request type, since they all flow through the same loop. 

I'd like the SDK to support concurrent handling of incoming requests on a single session, likely as an opt-in knob to preserve today's strict-ordering semantics for handlers that depend on them, with whatever bounding/back-pressure mechanism the maintainers prefer.

### References

```python
"""Reproducer: concurrent server->client requests are serialized end-to-end.

A FastMCP tool fans out N concurrent ``ServerSession.create_message`` calls
via ``asyncio.gather``. The client's sampling callback records peak in-flight.

Expected with concurrent dispatch:  peak == N,  elapsed ~= SLEEP
Observed today:                     peak == 1,  elapsed ~= N * SLEEP

Root cause: ``BaseSession._receive_loop`` awaits each incoming request
handler inline before reading the next message off the stream
(mcp/shared/session.py).

Run:  python repro_concurrent_sampling.py
"""

import asyncio

from mcp.types import (
    CreateMessageRequestParams,
    CreateMessageResult,
    SamplingMessage,
    TextContent,
)

N = 5
SLEEP = 0.5


async def main() -> None:
    inflight = 0
    peak = 0

    async def sampling_callback(
        _ctx: RequestContext,
        params: CreateMessageRequestParams,
    ) -> CreateMessageResult:
        nonlocal inflight, peak
        inflight += 1
        peak = max(peak, inflight)
        try:
            await asyncio.sleep(SLEEP)
        finally:
            inflight -= 1
        msg = params.messages[0].content
        echo = msg.text if isinstance(msg, TextContent) else ""
        return CreateMessageResult(
            role="assistant",
            content=TextContent(type="text", text=f"echo:{echo}"),
            model="test-model",
        )

    server = FastMCP("repro")

    @server.tool()
    async def fanout(ctx: Context) -> str:
        async def one(i: int) -> str:
            r = await ctx.session.create_message(
                messages=[
                    SamplingMessage(
                        role="user",
                        content=TextContent(type="text", text=str(i)),
                    )
                ],
                max_tokens=8,
            )
            return r.content.text if isinstance(r.content, TextContent) else ""

        return ",".join(await asyncio.gather(*(one(i) for i in range(N))))

    loop = asyncio.get_running_loop()
    t0 = loop.time()
    async with create_connected_server_and_client_session(
        server, sampling_callback=sampling_callback
    ) as session:
        await session.call_tool("fanout", {})
    elapsed = loop.time() - t0

    print(f"N={N}, per-call sleep={SLEEP}s")
    print(f"peak in-flight: {peak}    (concurrent: {N}, serialized: 1)")
    print(f"elapsed:        {elapsed:.2f}s  (concurrent: ~{SLEEP}s, serialized: ~{N * SLEEP}s)")

    assert peak == N, (
        f"server->client requests were serialized: peak in-flight={peak}, expected {N}"
    )


if __name__ == "__main__":
    asyncio.run(main())
```

demoray:
Closes #2489.

## Problem

`BaseSession._receive_loop` awaited each incoming request handler inline, so concurrent server→client requests (e.g. N `create_message` sampling callbacks launched via `asyncio.gather`) ran strictly one-at-a-time. Peak in-flight handlers = 1 regardless of fan-out.

## Fix

Add an opt-in `_dispatch_requests_concurrently` flag on `BaseSession` that spawns each request handler in the session's task group via `start_soon` instead of awaiting it inline.

`ClientSession` enables the flag. `ServerSession` stays serial because its `_received_request` has an `InitializeRequest` state machine (`Initializing → respond → Initialized`) that would race if a follow-up request were dispatched concurrently before the state transitioned.

## Collateral fixes

Concurrent dispatch widens two latent `RequestResponder` races:

- **Cancel-before-enter.** `__enter__` used to replace the cancel scope created in `__init__`. With concurrent dispatch, a `CancelledNotification` can arrive before the handler task enters `with responder:` — the pre-entry `cancel()` would then target a scope that is about to be thrown away. Fixed by not replacing the scope in `__enter__`.
- **Double cancel / pre-entry cancel.** `cancel()` is now idempotent (early return when already completed) and no longer requires `_entered`.

Handler exceptions are translated into a JSON-RPC error response so a raising handler can't wedge the peer (otherwise the exception would propagate from the dispatch task and tear down the session's task group).

## Tests

Four regression tests in `tests/shared/test_session.py`:

- `test_concurrent_server_to_client_requests_run_in_parallel` — N=4 fan-out, asserts peak concurrency = 4.
- `test_sampling_callback_exception_returns_error_response` — handler that raises still gets a JSON-RPC error reply for its id.
- `test_double_cancel_does_not_send_second_response` — idempotent cancel.
- `test_cancel_before_context_entered_marks_scope_cancelled` — cancel is safe before `__enter__`.
- `test_handler_that_responds_then_raises_emits_no_duplicate_error` — responded-then-raised handlers don't produce a second error for the same id.

All 1174 tests pass; 100% branch coverage maintained; `strict-no-cover`, ruff, and pyright clean.

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -123,6 +123,9 @@ class ClientSession(
         experimental_task_handlers: ExperimentalTaskHandlers | None = None,
     ) -> None:
         super().__init__(read_stream, write_stream, read_timeout_seconds=read_timeout_seconds)
+        # Dispatch incoming server->client requests concurrently so a slow
+        # sampling/elicitation callback doesn't serialize other in-flight requests.
+        self._dispatch_requests_concurrently = True
         self._client_info = client_info or DEFAULT_CLIENT_INFO
         self._sampling_callback = sampling_callback or _default_sampling_callback
         self._sampling_capabilities = sampling_capabilities
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -99,7 +99,8 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
     def __enter__(self) -> RequestResponder[ReceiveRequestT, SendResultT]:
         """Enter the context manager, enabling request cancellation tracking."""
         self._entered = True
-        self._cancel_scope = anyio.CancelScope()
+        # Enter the scope created in __init__ so pre-entry cancel() targets
+        # the same scope the handler will later run under.
         self._cancel_scope.__enter__()
         return self
 
@@ -140,11 +141,12 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
             )
 
     async def cancel(self) -> None:
-        """Cancel this request and mark it as completed."""
-        if not self._entered:  # pragma: no cover
-            raise RuntimeError("RequestResponder must be used as a context manager")
-        if not self._cancel_scope:  # pragma: no cover
-            raise RuntimeError("No active cancel scope")
+        """Cancel this request and mark it as completed.
+
+        Safe to call before the context manager has been entered.
+        """
+        if self._completed:
+            return
 
         self._cancel_scope.cancel()
         self._completed = True  # Mark as completed so it's removed from in_flight
@@ -157,6 +159,10 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
     @property
     def in_flight(self) -> bool:  # pragma: no cover
         return not self._completed and not self.cancelled
+
+    @property
+    def completed(self) -> bool:
+        return self._completed
 
     @property
     def cancelled(self) -> bool:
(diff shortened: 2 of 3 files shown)


demoray:
Happy to see the underlying issue upleveled by PR 2838.

PR #2493, opened 2026-04-22
fix: don't send a response for cancelled requests

### LINKED ISSUE #2480: RequestResponder.cancel sends JSON-RPC response, violating cancellation spec

Zelys-DFKH:
## Summary

- Remove the `_send_response` call from `RequestResponder.cancel()`. The MCP cancellation spec says receivers must not send a response for a cancelled request.
- When `send_notification` is called with a `CancelledNotification`, immediately resolve the pending response stream so `send_request()` returns instead of waiting indefinitely.
- Update a now-stale comment in `lowlevel/server.py` that said "responder.cancel() already sent an error response."

## Motivation and Context

`RequestResponder.cancel()` was sending `ErrorData(code=0, message="Request cancelled")` back to the sender after receiving `notifications/cancelled`. The [cancellation spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/cancellation.mdx#L33-L36) says receivers SHOULD NOT do this. Claude Code, for example, logs "Received a response for an unknown message ID" and reconnects every time a long-running tool gets cancelled.

The TypeScript SDK already gets this right: the sender rejects its own pending promise locally when it sends the cancellation ([source](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L1184-L1186)), and the receiver just calls `controller.abort()` without sending anything back. This PR brings the Python SDK in line with that behavior.

## How Has This Been Tested?

- `uv run --frozen pytest tests/shared/test_session.py` — all 9 tests pass, including `test_request_cancellation` (unchanged) and the new `test_cancel_notification_for_unknown_request_id`
- `uv run --frozen pytest tests/server/test_cancel_handling.py` — all 3 tests pass
- `./scripts/test` — 1172 tests, 100% branch coverage

## Breaking Changes

Code that inspects `e.code` on a cancellation `MCPError` will see `-32001` (`REQUEST_TIMEOUT`) instead of `0`. The message `"Request cancelled"` is unchanged. The TypeScript SDK uses `ErrorCode.RequestTimeout` for this same case, so the behavior now matches across SDKs.

## Types of changes

- [ ] Documentation update

## Checklist

Fixes #2480

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -502,9 +502,9 @@ class Server(Generic[LifespanResultT]):
                     response = err.error
                 except anyio.get_cancelled_exc_class():
                     if message.cancelled:
-                        # Client sent CancelledNotification; responder.cancel() already
-                        # sent an error response, so skip the duplicate.
-                        logger.info("Request %s cancelled - duplicate response suppressed", message.request_id)
+                        # Client sent CancelledNotification; per spec receivers must
+                        # not send a response for a cancelled request, so return.
+                        logger.info("Request %s cancelled", message.request_id)
                         return
                     # Transport-close cancellation from the TG in run(); re-raise so the
                     # TG swallows its own cancellation.
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -148,11 +148,8 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
 
         self._cancel_scope.cancel()
         self._completed = True  # Mark as completed so it's removed from in_flight
-        # Send an error response to indicate cancellation
-        await self._session._send_response(  # type: ignore[reportPrivateUsage]
-            request_id=self.request_id,
-            response=ErrorData(code=0, message="Request cancelled"),
-        )
+        # Per spec, receivers of notifications/cancelled must NOT send a response.
+        # The sender is responsible for resolving its own pending request.
 
     @property
     def in_flight(self) -> bool:  # pragma: no cover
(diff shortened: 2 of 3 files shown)


Zelys-DFKH:
Closing this. #3188 landed this in v2 (2.0.0), and per #2480 the 1.x line only takes critical fixes.

PR #2495, opened 2026-04-23
fix: default encoding_error_handler to 'replace' in StdioServerParameters

Christian-Sidak:
## Summary

`stdio_client` crashed with an `ExceptionGroup` when the child process wrote invalid UTF-8 bytes to stdout because `StdioServerParameters.encoding_error_handler` defaulted to `"strict"`.

This was already fixed on the server side in PR #2302; this PR applies the same one-line fix to the client side.

**Change:** default `encoding_error_handler` from `"strict"` to `"replace"` so malformed bytes are replaced with U+FFFD, which then fails JSON parsing and is delivered as an in-stream `Exception` — keeping the transport alive for subsequent valid messages.

## Test

Added `test_stdio_client_invalid_utf8` to `tests/client/test_stdio.py`: spawns a child that writes `b"\xff\xfe\n"` followed by a valid JSON-RPC line, and asserts the first stream item is an `Exception` and the second is a `SessionMessage`.

Fixes #2454

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -151,7 +151,7 @@ async def stdio_client(server: StdioServerParameters, errlog: TextIO = sys.stder
                     for line in lines:
                         try:
                             message = types.jsonrpc_message_adapter.validate_json(line, by_name=False)
-                        except Exception as exc:  # pragma: no cover
+                        except Exception as exc:
                             logger.exception("Failed to parse JSONRPC message from server")
                             await read_stream_writer.send(exc)
                             continue
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -92,7 +92,7 @@ class StdioServerParameters(BaseModel):
     Defaults to utf-8.
     """
 
-    encoding_error_handler: Literal["strict", "ignore", "replace"] = "strict"
+    encoding_error_handler: Literal["strict", "ignore", "replace"] = "replace"
     """
     The text encoding error handler.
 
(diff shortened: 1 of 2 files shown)


Christian-Sidak:

Christian-Sidak:

Christian-Sidak:
Fixed the CI failures: the new test exercises the `except Exception as exc` path in `stdio_client` (invalid UTF-8 is now decoded via `replace` handler and then fails JSON parsing), which made the `# pragma: no cover` annotation stale. Removed it so `strict-no-cover` no longer flags those lines.

Christian-Sidak:
Fixed the coverage gap: branch `532->537` was the async-for exhaustion path, never reached because the loop always exited via `break`. Removed `time.sleep(1)` from the child script so the process exits immediately after writing the two lines, then dropped the early-break guard. The loop now terminates by exhaustion, covering both branches and restoring 100% coverage.

Christian-Sidak:
Fixed the ruff PERF401 lint error: replaced the `async for` + `items.append` pattern with `items.extend([item async for item in read_stream])`. The Windows/Python 3.14 coverage failure (lines 290-296, 850-851 of `server.py`) appears to be a pre-existing platform-specific gap unrelated to this PR — all 1172 tests pass on that platform and the uncovered lines are in `server.py`, which this PR does not touch.

maxisbey:
Closing as a duplicate of #2456, which was opened first with the same change plus broader teardown handling. Appreciated the iteration on CI here — thanks for the contribution.

PR #2498, opened 2026-04-23
feat(security): add subdomain wildcard support to allowed_hosts and allowed_origins

### LINKED ISSUE #2141: Add wildcard pattern support for `allowed_hosts` in transport security
## Problem

Currently, the `allowed_hosts` configuration in `TransportSecuritySettings` only supports exact string matching [link](https://github.com/modelcontextprotocol/python-sdk/blob/6f0e8e78514b4aab7bdf8fd2eb7438bbae90100e/src/mcp/server/transport_security.py#L43). This is too restrictive for real-world scenarios where you need to allow multiple subdomains under the same parent domain.

For example, if you want to allow `app.mysite.com`, `api.mysite.com`, `admin.mysite.com`, etc., you currently need to list each subdomain explicitly:

```python
allowed_hosts=["app.mysite.com", "api.mysite.com", "admin.mysite.com", ...]
```

This becomes impractical when you have many subdomains or dynamic subdomain generation.

## Proposed Solution

Add wildcard pattern support similar to [Starlette's TrustedHostMiddleware](https://github.com/encode/starlette/blob/master/starlette/middleware/trustedhost.py) or [Django's ALLOWED_HOSTS](https://docs.djangoproject.com/en/6.0/ref/settings/#allowed-hosts) , allowing patterns like:

- `*.mysite.com` - matches any subdomain of `mysite.com` (e.g., `app.mysite.com`, `api.mysite.com`) as well as the base domain `mysite.com`
- `example.com:*` - matches `example.com` with any port (already supported)

## Example Usage

```python

settings = TransportSecuritySettings(
    enable_dns_rebinding_protection=True,
    allowed_hosts=["*.mysite.com", "localhost:*"],
)
```

I'd be happy to follow up and create a PR. 

### References

_No response_

Zelys-DFKH:
Closes #2141.

`TransportSecuritySettings` previously supported two matching modes for `allowed_hosts` and `allowed_origins`:
- Exact: `"example.com"`
- Port wildcard: `"example.com:*"`

This adds subdomain wildcards:
- `"*.example.com"` in `allowed_hosts` — matches `example.com` and any of its subdomains; port is stripped before matching
- `"https://*.example.com"` in `allowed_origins` — same, but scheme must also match

Behavior follows Django's `ALLOWED_HOSTS` and Starlette's `TrustedHostMiddleware`: `*.example.com` covers the base domain too, so no separate entry for `example.com` is needed.

Also fixes a coverage gap: `_validate_host`, `_validate_origin`, and the DNS-rebinding path in `validate_request` were all `# pragma: no cover`. Those paths were only exercised by subprocess-based integration tests, which coverage never sees. This PR adds direct unit tests for `TransportSecurityMiddleware` (28 tests, 100% branch coverage) and removes all the no-cover markers.

Files changed:
- `src/mcp/server/transport_security.py`
- `tests/server/test_transport_security.py` (new)

### DIFF
--- a/src/mcp/server/transport_security.py
+++ b/src/mcp/server/transport_security.py
@@ -1,6 +1,7 @@
 """DNS rebinding protection for MCP server transports."""
 
 import logging
+from urllib.parse import urlparse
 
 from pydantic import BaseModel, Field
 from starlette.requests import Request
@@ -22,11 +23,23 @@ class TransportSecuritySettings(BaseModel):
     allowed_hosts: list[str] = Field(default_factory=list)
     """List of allowed Host header values.
 
+    Supports exact matches, port wildcards, and subdomain wildcards:
+
+    - ``"example.com"`` — exact match
+    - ``"example.com:*"`` — any port on that host
+    - ``"*.example.com"`` — any subdomain (or the base domain itself)
+
     Only applies when `enable_dns_rebinding_protection` is `True`.
     """
 
     allowed_origins: list[str] = Field(default_factory=list)
     """List of allowed Origin header values.
+
+    Supports exact matches, port wildcards, and subdomain wildcards:
+
+    - ``"https://example.com"`` — exact match
+    - ``"https://example.com:*"`` — any port on that origin
+    - ``"https://*.example.com"`` — any subdomain (or the base domain itself) with HTTPS
 
     Only applies when `enable_dns_rebinding_protection` is `True`.
     """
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. Per our CONTRIBUTING.md we ask that PRs start from a linked, triaged issue with maintainer buy-in, which this one doesn't have yet. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #2499, opened 2026-04-23
fix(mcpserver): advertise capabilities only for registered primitives

### LINKED ISSUE #2473: MCPServer unconditionally declares `prompts`, `resources`, `tools` capabilities on `initialize`
### Pre-flight

- [x] Using the latest MCP Python SDK (`v1.27.0`) and also verified against `main @ 3d7b311`
- [x] Searched existing issues and PRs — no duplicate found

### Description

Reproduced against `main @ 3d7b311` (2026-04-14).

Per [`schema/2025-11-25/schema.ts` L388-L431](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts#L388-L431), each entry in `ServerCapabilities` is documented as **"Present if the server offers any [prompt templates | resources | tools]"**. [`lifecycle.mdx`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/basic/lifecycle.mdx#L215-L218) also requires both peers to **only use capabilities that were successfully negotiated**.

Current behavior is inconsistent with the documented schema semantics: `MCPServer` announces `prompts`, `resources`, and `tools` capabilities on every `initialize` response, regardless of whether any prompt / resource / tool has actually been registered. A server that only exposes one `tool` still advertises `prompts` and `resources` it doesn't have, and an entirely empty server advertises all three.

**Root cause (current `main`):**

1. [`src/mcp/server/mcpserver/server.py#L178-L184`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/mcpserver/server.py#L178-L184) — `MCPServer.__init__` unconditionally wires `on_list_tools=self._handle_list_tools`, `on_list_resources=...`, `on_list_prompts=...` into the lowlevel `Server(...)`.
2. [`src/mcp/server/lowlevel/server.py#L205-L224`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L205-L224) — `Server.__init__` unconditionally populates `self._request_handlers` with `"tools/list"`, `"resources/list"`, `"prompts/list"` entries.
3. [`src/mcp/server/lowlevel/server.py#L283-L328`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L283-L328) — `get_capabilities()` derives capability entries from key presence in `_request_handlers`, so all three are always included.

This behavior was carried over verbatim from `FastMCP._setup_handlers()` in the #1951 rename — the symptom pre-dates the refactor.

The maintainer TODO at [`lowlevel/server.py#L249-L253`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/lowlevel/server.py#L249-L253) (`Rethink capabilities API ... Consider deriving capabilities entirely from server state`) points to the same area — happy to align a fix with whatever direction you have in mind there.

**Expected:** capability entries should only appear when the corresponding primitive is actually offered (registered via decorator or `add_*()`).

### Example Code

```python
# empty_server.py
mcp = MCPServer("empty")
if __name__ == "__main__":
    mcp.run()
```

```python
# one_tool_server.py
mcp = MCPServer("hello")

@mcp.tool()
def echo(text: str) -> str:
    return text

if __name__ == "__main__":
    mcp.run()
```

```python
# probe.py — sends initialize over stdio and prints the advertised capabilities
import json, subprocess, sys
for path in ["empty_server.py", "one_tool_server.py"]:
    proc = subprocess.Popen([sys.executable, path],
        stdin=subprocess.PIPE, stdout=subprocess.PIPE)
    req = {"jsonrpc":"2.0","id":1,"method":"initialize","params":{
        "protocolVersion":"2025-11-25","capabilities":{},
        "clientInfo":{"name":"probe","version":"0"}}}
    proc.stdin.write((json.dumps(req)+"\n").encode()); proc.stdin.flush()
    print(path)
    print(json.dumps(json.loads(proc.stdout.readline())["result"]["capabilities"], indent=2))
    proc.terminate()
```

**Actual output (both servers, identical):**

```json
{
  "experimental": {},
  "prompts":   {"listChanged": false},
  "resources": {"subscribe": false, "listChanged": false},
  "tools":     {"listChanged": false}
}
```

**Expected:**

- `empty_server.py` → no `prompts`, `resources`, or `tools` entries.
- `one_tool_server.py` → `tools` present, `prompts` and `resources` absent.

### Possible fixes (open to maintainer preference)

**A. Filter capabilities in `MCPServer` based on manager state.**
Post-process the `ServerCapabilities` returned by the lowlevel `Server` — omit `tools` / `resources` / `prompts` when the respective manager is empty. Minimal diff, keeps the decorator-after-init pattern. Could be done via a new optional `capability_filter` callable on the lowlevel `Server`, or via an override at the `MCPServer` layer.

**B. Register list handlers lazily.**
Pass `on_list_tools=None` etc. from `MCPServer.__init__`; register via the existing `Server._add_request_handler` on the first `add_tool()` / `add_resource()` / `add_prompt()`. Matches the documented "offers any X" schema semantics directly, no post-processing needed. This shifts one usage contract: primitives added via `add_*()` *after* `run()` would not appear in capabilities. In practice primitives are registered before `run()`, and post-handshake capability changes aren't supported by the spec regardless, so the regression surface is narrow — but it's worth calling out.

I'd like to take this on if you're open to a fix — please assign to me after triage, and let me know if you'd prefer a direction other than A or B.

### Python & MCP Python SDK

```
Python 3.13
Reproduced on: v1.27.0 (released), main @ 3d7b311 (2026-04-14)
```

Zelys-DFKH:
## Summary

- `MCPServer.__init__` unconditionally passes non-`None` list handlers to
  the lowlevel `Server`, causing `get_capabilities()` to always advertise
  `tools`, `resources`, and `prompts`, even when none are registered.
- Adds an optional `capability_filter` parameter to the lowlevel `Server`
  that post-processes the computed `ServerCapabilities`.
- `MCPServer` passes a filter that suppresses each entry when the
  corresponding manager is empty at capability-computation time.

## Motivation and Context

Per the [MCP schema spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-11-25/schema.ts#L388-L431) and [lifecycle docs](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-11-25/basic/lifecycle.mdx#L215-L228), a `ServerCapabilities` entry should appear only when the server actually offers that primitive. Currently, an empty `MCPServer` advertises all three, and a server with only one tool still advertises `resources` and `prompts` it doesn't have.

The filter runs when `create_initialization_options()` is called at the start of `run()`, after all `@mcp.tool()` / `@mcp.resource()` / `@mcp.prompt()` decorators have been applied. Every transport path (stdio, SSE, streamable HTTP) calls `server.create_initialization_options()` → `get_capabilities()`, so the filter applies without any per-transport changes.

## How Has This Been Tested?

Six new tests in `tests/server/mcpserver/test_server.py` cover: empty server, tool-only, resource-only, resource-template-only, prompt-only, and all-registered cases.

```bash
uv run --frozen pytest tests/server/mcpserver/test_server.py -v
```

Full suite: `./scripts/test` — 100% branch coverage.

## Breaking Changes

None. Servers that previously over-advertised will now send correct capability sets.

## Types of changes

- [ ] Documentation update

## Checklist

Closes #2473

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -173,6 +173,7 @@ class Server(Generic[LifespanResultT]):
             [ServerRequestContext[LifespanResultT], types.RequestParams | None],
             Awaitable[types.EmptyResult],
         ] = _ping_handler,
+        capability_filter: Callable[[types.ServerCapabilities], types.ServerCapabilities] | None = None,
         # Notification handlers
         on_roots_list_changed: Callable[
             [ServerRequestContext[LifespanResultT], types.NotificationParams | None],
@@ -198,6 +199,7 @@ class Server(Generic[LifespanResultT]):
             str, Callable[[ServerRequestContext[LifespanResultT], Any], Awaitable[None]]
         ] = {}
         self._experimental_handlers: ExperimentalHandlers[LifespanResultT] | None = None
+        self._capability_filter = capability_filter
         self._session_manager: StreamableHTTPSessionManager | None = None
         logger.debug("Initializing server %r", name)
 
@@ -325,6 +327,8 @@ class Server(Generic[LifespanResultT]):
         )
         if self._experimental_handlers:
             self._experimental_handlers.update_capabilities(capabilities)
+        if self._capability_filter is not None:
+            capabilities = self._capability_filter(capabilities)
         return capabilities
 
     @property
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -62,6 +62,7 @@ from mcp.types import (
     PaginatedRequestParams,
     ReadResourceRequestParams,
     ReadResourceResult,
+    ServerCapabilities,
     TextContent,
     TextResourceContents,
     ToolAnnotations,
@@ -167,6 +168,16 @@ class MCPServer(Generic[LifespanResultT]):
             resources=resources, warn_on_duplicate_resources=self.settings.warn_on_duplicate_resources
         )
         self._prompt_manager = PromptManager(warn_on_duplicate_prompts=self.settings.warn_on_duplicate_prompts)
+
+        def _filter_capabilities(caps: ServerCapabilities) -> ServerCapabilities:
+            if not self._tool_manager.list_tools():
+                caps.tools = None
+            if not (self._resource_manager.list_resources() or self._resource_manager.list_templates()):
+                caps.resources = None
+            if not self._prompt_manager.list_prompts():
+                caps.prompts = None
+            return caps
+
         self._lowlevel_server = Server(
             name=name or "mcp-server",
             title=title,
@@ -182,6 +193,7 @@ class MCPServer(Generic[LifespanResultT]):
             on_list_resource_templates=self._handle_list_resource_templates,
             on_list_prompts=self._handle_list_prompts,
             on_get_prompt=self._handle_get_prompt,
+            capability_filter=_filter_capabilities,
             # TODO(Marcelo): It seems there's a type mismatch between the lifespan type from an MCPServer and Server.
             # We need to create a Lifespan type that is a generic on the server type, like Starlette does.
             lifespan=(lifespan_wrapper(self, self.settings.lifespan) if self.settings.lifespan else default_lifespan),  # type: ignore
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -303,7 +303,7 @@ class MCPServer(Generic[LifespanResultT]):
         if transport not in TRANSPORTS.__args__:  # type: ignore  # pragma: no cover
             raise ValueError(f"Unknown transport: {transport}")
 
-        match transport:
+        match transport:  # pragma: lax no cover
             case "stdio":
                 anyio.run(self.run_stdio_async)
             case "sse":  # pragma: no cover
@@ -857,7 +857,7 @@ class MCPServer(Generic[LifespanResultT]):
 
         return decorator  # pragma: no cover
 
-    async def run_stdio_async(self) -> None:
+    async def run_stdio_async(self) -> None:  # pragma: lax no cover
         """Run the server using stdio transport."""
         async with stdio_server() as (read_stream, write_stream):
             await self._lowlevel_server.run(
--- a/tests/server/mcpserver/test_server.py
+++ b/tests/server/mcpserver/test_server.py
@@ -1516,3 +1516,102 @@ async def test_report_progress_passes_related_request_id():
         message="halfway",
         related_request_id="req-abc-123",
     )
+
+
+# ---------------------------------------------------------------------------
+# Capability filtering: MCPServer only advertises capabilities for primitives
+# that are actually registered, per the MCP schema spec.
+# ---------------------------------------------------------------------------
+
+
+def _get_caps(mcp: MCPServer) -> Any:
+    """Return the ServerCapabilities advertised by this MCPServer at run time."""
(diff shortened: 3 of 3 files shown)


keeltrace:

I think this filter now conflates “the server supports this primitive” with “the current list is non-empty”, which breaks the dynamic-list lifecycle that `MCPServer` itself exposes.

At this PR's head, capability filtering is based on current manager contents:

```python
if not self._tool_manager.list_tools():
    caps.tools = None
```

Capabilities are negotiated during initialization. If an `MCPServer` supports tools but happens to have zero tools at that moment, the client is told that the server has no `tools` capability at all. Adding a tool later cannot repair that negotiation.

That matters for two reasons:

1. The current MCP tools spec explicitly says servers that support tools MUST declare `tools`, and that the set returned by `tools/list` **MAY be empty and MAY change over time**. `listChanged` exists to describe exactly that changing-set lifecycle.
2. This SDK documents runtime mutation itself. `docs_src/context/tutorial003.py` has a handler that does:

```python
mcp.add_tool(recommend_book)
await ctx.session.send_tool_list_changed()
```

So “no tools right now” is not equivalent to “this server does not implement tools”.

A concrete regression shape is a server that initially exposes only a resource/prompt (or otherwise starts with an empty tool manager), then adds its first tool at runtime. Before this PR it negotiates the `tools` feature and a later tool can be discovered. With this PR, `tools` is omitted at initialize; after the first `add_tool()`, the client still has no negotiated tools capability and should not start issuing `tools/list`/`tools/call` merely because it receives or suspects a list change.

The same concern applies to resources: the current spec explicitly allows a declared resource set to be empty and to change over time.

I think the safer distinction is **support vs current cardinality**. If `MCPServer` is meant to allow disabling tools/resources/prompts entirely, that probably needs an explicit support/configuration signal (or handler presence), rather than inferring support from whether the list happens to be empty at initialization.

Could the regression coverage include “capability supported with an initially empty list, item added later”? The six new tests currently only assert static pre-initialization cardinality, so they encode the problematic equivalence rather than exercising the list-changed lifecycle.

One historical note: #2473 itself mentioned post-`run()` additions but dismissed them as unsupported. The protocol text has since become explicit that these sets may start empty and change over time, so I don't think that assumption is safe for current `main`.

Checked against PR head `d1c8560`. This is a static protocol/code-path review; I did not run the repository suite locally.

Zelys-DFKH:
You're right, and thanks for laying it out so clearly.

The MUST prose settles it. `tools.mdx` 38, `resources.mdx` 32 and `prompts.mdx` 30 all tie the requirement to whether the server supports the primitive, not to how many it holds. I leaned on the `schema.ts` field comment ("offers any tools to call") and treated it as the whole story.

And I half-saw this at the time, then talked myself out of it. In #2473 I waved off post-`run()` additions with "post-handshake capability changes aren't supported by the spec regardless", which conflates two things. The capability set is fixed at handshake; the list isn't, and `listChanged` exists precisely so it can change. My filter switches that off. It only bites a server starting with zero of a primitive, so `tutorial003` survives, but that's a narrower regression rather than none.

The two artifacts genuinely disagree, so I'd rather a maintainer called it. `schema.ts` says "offers any", the prose says "supports". If the prose governs, the fix for #2473 isn't a cardinality filter at all. If the filter stays, it should be opt-in instead of wired unconditionally.

I'll rework it whichever way it lands. Needs a rebase regardless, so I'll wait on the shape.

Issue #2422, opened 2026-04-10
Specific ToolNotFoundError for easier handling than ToolError in multi-tenant environments


DevonFulcher:
### Description

We are running a remote MCP server and we are seeing errors like:
```
mcp.server.fastmcp.exceptions.ToolError: Unknown tool: <unknown-tool-name>
```
I'm unsure why our users are trying to call our remote MCP server with the name of a tool that we don't manage. Regardless, we would like to handle these errors separately from a generic `ToolError`. For instance:
- `ToolNotFoundError`: user error that shouldn't alert us.
- `ToolError`: an error that could indicate an actual issue with our system that we should alert on and fix.

### References

I couldn't find any issues related to this.

RudrenduPaul:
I'd like to implement this.

My plan: add `ToolNotFoundError(ToolError)` to `src/mcp/server/fastmcp/exceptions.py` and raise it in the tool-dispatch path instead of a bare `ToolError` when the requested tool name isn't found. This allows callers to distinguish between "tool not found" and other tool errors.

Small, self-contained change with a test. Would maintainers like a PR for this?

RudrenduPaul:
Hi — I'd like to implement this. Making `ToolNotFoundError` a subclass of `McpError` (or a specific base exception) would let callers catch tool-not-found specifically without having to inspect error messages.

My plan:
1. Add `ToolNotFoundError(McpError)` to `exceptions.py`
2. Raise it in the tool dispatch path where `tool_name not in tools`
3. Add a test that catches `ToolNotFoundError` specifically

Will keep the change minimal. Let me know if there's a preferred exception hierarchy or if you'd like it to carry additional fields (e.g. the tool name that was requested).

maxisbey:
What does adding a separate exception type get you here? Are you somehow catching it explicitly?

zeweihan:
Multi-tenant environments really need this. We run an MCP-based agent orchestration platform for legal workflows where different law firms have different tool sets (some have OCR, some have due diligence plugins, etc.).

Currently, catching generic `ToolError` means we can't distinguish between:
1. A tool that doesn't exist for this tenant's configuration
2. A tool that exists but had a runtime failure

Having `ToolNotFoundError` would let us gracefully fall back to alternative tools or prompt the user to enable a plugin — instead of surfacing a confusing error. +1 on this proposal.

The multi-tenant pattern is becoming more common as MCP adoption grows beyond single-user setups.

Issue #1967, opened 2026-01-27
ClosedResourceError During MCP Stateless Server Shutdown When Sending Log Messages


jonfinerty:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

I'm seeing occasional errors from our MCP server, which I think are a race condition during shutdown. 

I think this is a similar but distinct issue to the race condition in #1190 and #1384 as this is during `send_log_message`

I haven't been able to get a consistent reproduce yet, will update when I can.

## Stack Trace

```python
+ Exception Group Traceback (most recent call last):
  |   File "/app/.venv/lib/python3.13/site-packages/mcp/server/streamable_http_manager.py", line 180, in run_stateless_server
  |     await self.app.run(
  |     ...<4 lines>...
  |     )
  |   File "/app/.venv/lib/python3.13/site-packages/fastmcp/server/low_level.py", line 202, in run
  |     async with AsyncExitStack() as stack:
  |                ~~~~~~~~~~~~~~^^
  |   File "/usr/local/lib/python3.13/contextlib.py", line 768, in __aexit__
  |     raise exc
  |   File "/usr/local/lib/python3.13/contextlib.py", line 751, in __aexit__
  |     cb_suppress = await cb(*exc_details)
  |                   ^^^^^^^^^^^^^^^^^^^^^^
  |   File "/app/.venv/lib/python3.13/site-packages/mcp/shared/session.py", line 238, in __aexit__
  |     return await self._task_group.__aexit__(exc_type, exc_val, exc_tb)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/app/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 783, in __aexit__
  |     raise BaseExceptionGroup(
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Exception Group Traceback (most recent call last):
    |   File "/app/.venv/lib/python3.13/site-packages/fastmcp/server/low_level.py", line 214, in run
    |     async with anyio.create_task_group() as tg:
[...]

Issue #1219, opened 2025-07-31
_handle_stateless_request  ClosedResourceError


281743556:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description
Every request will try an exception

use `_handle_stateless_request` > `run_stateless_server `> `http_transport.connect()` > `for session_message in write_stream_reader`
It will go in twice, and an exception(yio.ClosedResourceError) will be thrown when the second request has been closed

### Example Code
```Python
app = Server("mcp-server")
def main(port: int, transport: str, dev: bool, limit_concurrency: int) -> int:
    # Create the session manager with true stateless mode
    from starlette.applications import Starlette
    from starlette.routing import Mount, Route
    session_manager = StreamableHTTPSessionManager(
        app=app,
        event_store=None,
        json_response=True,
        stateless=True, 
    )

    @contextlib.asynccontextmanager
    async def lifespan(app: Starlette) -> AsyncIterator[None]:
        """Context manager for managing session manager lifecycle."""
        async with session_manager.run():
            logger.info("Application started with StreamableHTTP session manager!")
            try:
                yield
            finally:
                logger.info("Application shutting down...")

    async def handle_streamable_http(
        scope: Scope, receive: Receive, send: Send
    ) -> None:
        await session_manager.handle_request(scope, receive, send)

    starlette_app = Starlette(
        debug=dev,
        routes=[
            Mount("/message", app=handle_streamable_http), 
        ],
        lifespan=lifespan,
    )
    import uvicorn
    uvicorn.run(starlette_app, host="0.0.0.0", port=port, http="httptools", limit_concurrency=limit_concurrency)

if __name__ == "__main__":
    asyncio.run(main())
```
view logs:
```shell
[2025-07-31 15:10:41,884][MainThread:19320][task_id:mcp.server.streamable_http][streamable_http.py:630][INFO][Terminating session: None]
[2025-07-31 15:10:41,885][MainThread:19320][task_id:mcp.server.streamable_http][streamable_http.py:880][ERROR][Error in message router]
Traceback (most recent call last):
  File "d:\project\bpaas\apihub-mcp-server\.venv\Lib\site-packages\mcp\server\streamable_http.py", line 831, in message_router
    async for session_message in write_stream_reader:
    ...<46 lines>...
            )
[...]

PR #2502, opened 2026-04-24
[v1.x] Drop responses/notifications when write stream is already closed

bobbyo:
## Audience / Scope / Status
| If you are… | Read this section | Why |
| --- | --- | --- |
| reviewing the runtime fix | Summary, Topology, Invariants, Changes | shows why the catch belongs in `BaseSession`, not only in outer server layers |
| reviewing the test/CI follow-up | Validation | explains the coverage and `pyright` failures on the first push |
| looking for broader transport redesign | Non-goals | this PR intentionally hardens closed-write semantics only |

Last verified against: `v1.x` on 2026-04-24

## Relevant upstream issues
- #1219: stateless request handling can still trip `ClosedResourceError` after client disconnect. Relevant because the failing write path here is also in the shared session layer.
- #1967: closed-stream writes during shutdown/logging should be treated as expected drops. Same class of failure, different send site.
- #2422: request lifecycle races still exist around response timing. Relevant because this PR makes the final write idempotent with respect to disconnect.
- PrefectHQ/fastmcp#823: downstream repro of the same symptom from a StreamableHTTP/FastMCP stack. Relevant because it demonstrates this is not just a local application bug.

## Summary
This PR makes `BaseSession.send_notification()` and `BaseSession._send_response()` treat writes to an already-closed transport as expected no-ops rather than exceptions.

The goal is narrow: if the peer has already gone away, a late response or notification should be dropped cleanly at the session send boundary.

## Failure topology
```mermaid
flowchart LR
  H["Request handler\n`mcp.server.lowlevel.server`"]:::server
  R["RequestResponder.respond()\n`src/mcp/shared/session.py:121`"]:::fyr
  S["BaseSession._send_response() / send_notification()\n`src/mcp/shared/session.py:318-357`"]:::fyr
  W["write_stream.send(SessionMessage)\nanyio memory/object stream"]:::transport
  C["Client transport\nstream already closed"]:::client

  H -- "handler result / late progress" --> R
  R -- "JSON-RPC response" --> S
  S -- "SessionMessage write" --> W
  W -. "peer disconnected earlier" .-> C

  classDef client fill:#bfdbfe,stroke:#1e3a8a,stroke-width:2px,color:#0f172a
  classDef transport fill:#e5e7eb,stroke:#6b7280,stroke-width:1px,color:#374151
  classDef server fill:#dcfce7,stroke:#166534,stroke-width:2px,color:#0f172a
  classDef fyr fill:#ddd6fe,stroke:#5b21b6,stroke-width:2px,color:#0f172a
```

### Why the catch belongs here
| Layer | Existing behavior | Gap |
| --- | --- | --- |
| `mcp.server.lowlevel.server` | already has an outer closed-transport catch around `await message.respond(response)` at `src/mcp/server/lowlevel/server.py:799-809` | only covers the `RequestResponder` path that reaches this exact outer frame |
| `BaseSession.send_notification()` | previously wrote directly to `_write_stream.send(...)` | late notifications could still bubble `ClosedResourceError` / `BrokenResourceError` |
| `BaseSession._send_response()` | previously wrote directly to `_write_stream.send(...)` | any caller below the outer server frame still relied on higher layers to catch transport-close races |

## Invariants
- INV-1: A closed peer transport is not a recoverable write target, but it is also not a server bug. Late writes should be dropped, not escalated.
- INV-2: The session send boundary is the narrowest shared seam that sees both responses and notifications. If violated: one call site gets hardened while sibling send paths still leak the same failure.
- INV-3: Higher-layer transport-close catches may remain for defense in depth. If violated: removing all outer guards in the same PR would expand scope from behavior hardening into control-flow refactoring.

## Changes
- Catch `anyio.BrokenResourceError` / `anyio.ClosedResourceError` in `BaseSession.send_notification()` at `src/mcp/shared/session.py:318-339`.
- Catch `anyio.BrokenResourceError` / `anyio.ClosedResourceError` in `BaseSession._send_response()` at `src/mcp/shared/session.py:343-357`.
- Add session-layer regression tests for closed and open write streams at `tests/server/test_session.py:223-282`.
- Add a low-level regression test that exercises the existing outer server catch at `tests/server/test_lowlevel_exception_handling.py:78-96` and `src/mcp/server/lowlevel/server.py:799-809`.

### Functional
```bash
PYTHONPATH=src python3 -m pytest -o addopts='' \
  tests/server/test_session.py \
  tests/server/test_lowlevel_exception_handling.py -q
```
Local result after the CI-fix follow-up:
- `17 passed`

### CI follow-up
The first push exposed two CI-only problems, not a runtime regression:

| Check | Initial failure | Follow-up in this PR |
| --- | --- | --- |
| `test` matrix | coverage dropped because the new session-layer catch made the old outer catch in `lowlevel/server.py` unexercised | added `test_handle_request_drops_response_when_transport_is_closed()` to cover the existing outer branch |
| `pre-commit` | strict `pyright` rejected loosely typed fake session/write-stream helpers | tightened annotations and cast sites in `tests/server/test_session.py` |

Targeted local follow-up verification:
```bash
/tmp/python-sdk-upstream/.venv-ci/bin/pyright \
  tests/server/test_session.py \
  tests/server/test_lowlevel_exception_handling.py
```
Local result:
- `0 errors, 0 warnings, 0 informations`

## Non-goals
- This PR does not redesign cancellation or stateless request lifecycle.
- This PR does not remove the outer guard in `lowlevel/server.py`; it keeps defense in depth.
- This PR does not claim to solve every disconnect race in the stack. It narrows one shared failure seam so downstream servers stop surfacing expected closed-write conditions as crashes.

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -38,6 +38,8 @@ SendRequestT = TypeVar("SendRequestT", ClientRequest, ServerRequest)
 SendResultT = TypeVar("SendResultT", ClientResult, ServerResult)
 SendNotificationT = TypeVar("SendNotificationT", ClientNotification, ServerNotification)
 ReceiveRequestT = TypeVar("ReceiveRequestT", ClientRequest, ServerRequest)
+
+logger = logging.getLogger(__name__)
 ReceiveResultT = TypeVar("ReceiveResultT", bound=BaseModel)
 ReceiveNotificationT = TypeVar("ReceiveNotificationT", ClientNotification, ServerNotification)
 
@@ -332,13 +334,16 @@ class BaseSession(
             message=JSONRPCMessage(jsonrpc_notification),
             metadata=ServerMessageMetadata(related_request_id=related_request_id) if related_request_id else None,
         )
-        await self._write_stream.send(session_message)
+        try:
+            await self._write_stream.send(session_message)
+        except (anyio.BrokenResourceError, anyio.ClosedResourceError):
+            logger.debug("Notification dropped - transport closed")
+            return
 
     async def _send_response(self, request_id: RequestId, response: SendResultT | ErrorData) -> None:
         if isinstance(response, ErrorData):
             jsonrpc_error = JSONRPCError(jsonrpc="2.0", id=request_id, error=response)
             session_message = SessionMessage(message=JSONRPCMessage(jsonrpc_error))
-            await self._write_stream.send(session_message)
         else:
             jsonrpc_response = JSONRPCResponse(
                 jsonrpc="2.0",
@@ -346,7 +351,12 @@ class BaseSession(
                 result=response.model_dump(by_alias=True, mode="json", exclude_none=True),
             )
             session_message = SessionMessage(message=JSONRPCMessage(jsonrpc_response))
+
+        try:
             await self._write_stream.send(session_message)
+        except (anyio.BrokenResourceError, anyio.ClosedResourceError):
+            logger.debug("Response for %s dropped - transport closed", request_id)
+            return
 
     async def _receive_loop(self) -> None:
         async with (
--- a/tests/server/test_lowlevel_exception_handling.py
+++ b/tests/server/test_lowlevel_exception_handling.py
@@ -1,5 +1,6 @@
 from unittest.mock import AsyncMock, Mock
 
+import anyio
 import pytest
 
 import mcp.types as types
@@ -72,3 +73,24 @@ async def test_normal_message_handling_not_affected():
 
     # Verify _handle_request was called
     server._handle_request.assert_called_once()
+
+
+@pytest.mark.anyio
+async def test_handle_request_drops_response_when_transport_is_closed():
+    """Closed write streams during respond should be treated as expected drops."""
+    server = Server("test-server")
+    session = Mock(spec=ServerSession)
+
+    responder = Mock(spec=RequestResponder)
+    responder.request_id = 1
+    responder.respond = AsyncMock(side_effect=anyio.ClosedResourceError())
+
+    await server._handle_request(
+        responder,
+        types.PingRequest(method="ping"),
+        session,
+        {},
+        raise_exceptions=False,
+    )
+
+    responder.respond.assert_called_once()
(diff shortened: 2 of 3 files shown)


aminghadersohi:
We hit this in production with mcp 1.28.1 / FastMCP on StreamableHTTP stateless mode:

**18 occurrences across 4 workspaces over 2026-07-19–20**, always the same stack:

```
ExceptionGroup(
  anyio.ClosedResourceError
)
  mcp/server/streamable_http_manager.py:180 run_stateless_server
    mcp/server/lowlevel/server.py:694 _handle_message
      await session.send_log_message(...)   # Exception-path arm — client already gone
        mcp/server/session.py:213 send_log_message → send_notification
          mcp/shared/session.py:335 send_notification
              anyio.ClosedResourceError
```

The trigger: an `Exception` arrives in the session's read stream (the client disconnected mid-request), `_handle_message` hits the `case Exception()` arm and tries to `send_log_message` back — but the write stream is already closed.

This PR (#2502) directly addresses it by hardening `send_notification` and `_send_response` at the `BaseSession` layer. As a stopgap we've applied a local monkey-patch in our downstream server, but we'd very much like to see this merged so we can pin to a clean upstream release.

Happy to help with review, testing, or rebasing if that would unblock this.

aminghadersohi:
Friendly bump on this one 🙂

Wanted to share a quick status check in case it's helpful for review prioritization:

- **CI is fully green** — all 21 checks pass (lint, pyright, full test matrix across Python 3.10–3.13 on Linux/Windows, client/server conformance).
- **Still applies cleanly to current `v1.x`.** I just re-verified with `git merge-tree` against the latest `v1.x` tip — zero conflicts across all three changed files, and GitHub still reports it as mergeable. No rebase needed on your end.
- The only thing standing between this and merging appears to be the required-review gate — there are no open review threads or unresolved comments to address.

Given the production impact we're seeing (18 crashes across 4 workspaces in 2 days on 1.28.1, per the report above), we'd really appreciate a maintainer look when you have a moment. Happy to help with anything — additional test coverage, a rebase, or answering questions — if that would help move it along. Thanks so much for the work on this fix! 🙏

PR #2505, opened 2026-04-26
Fix stdio_server closing process stdio

GitAashishG:
Summary
- duplicate stdin/stdout file descriptors before wrapping them in stdio_server
- close only the duplicated async wrappers during teardown so process stdio stays usable
- add a regression test covering post-shutdown writes to sys.stdout

Testing
- uv run --frozen pytest tests/server/test_stdio.py
- uv run --frozen ruff check src/mcp/server/stdio.py tests/server/test_stdio.py

Closes #1933

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,9 +17,10 @@ Example:
     ```
 """
 
+import os
 import sys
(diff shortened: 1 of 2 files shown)


maxisbey:

PR #2509, opened 2026-04-27
fix(server): return 405 on GET/DELETE in stateless HTTP mode

### LINKED ISSUE #2474: Return 405 on GET when stateless_http=True
### Initial Checks

### Description

When a Streamable HTTP server is configured with `stateless_http=True`, `GET /mcp` is still accepted and opens an SSE stream that has no session context and can never receive server-initiated messages — a dead-end.

The MCP Streamable HTTP spec explicitly permits returning `405 Method Not Allowed` when the server does not offer an SSE stream at the endpoint
([spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports)):

> The server **MUST** return HTTP 405 Method Not Allowed if an SSE stream is
> not offered at the endpoint.

A stateless server cannot offer one — there is no session to push to — so GET should 405. The TypeScript SDK already behaves this way in its stateless example.

A prior PR (#2262) proposed essentially this change but was closed for lack of a corresponding issue per CONTRIBUTING.md. 
Opening this issue to agree on scope so the fix can be re-submitted.

### Example Code

```Python

mcp = FastMCP("my-server", stateless_http=True, json_response=True)

@mcp.tool()
def ping() -> str:
    return "pong"

```

A couple of clarifications for triage:

### Current vs expected behavior

Today (v1.27.0):
```
$ curl -i -X GET http://localhost:8000/mcp
HTTP/1.1 200 OK
content-type: text/event-stream
... (stream idles until platform timeout)
```

Expected:
```
$ curl -i -X GET http://localhost:8000/mcp
HTTP/1.1 405 Method Not Allowed
Allow: POST
```

### Concrete impact
The idle GET SSE stream holds a long-lived connection per client with no useful payload, exhausting concurrent connection limits on serverless platforms like Cloud Run. Background previously raised in #2232 / #1941.

### Proposed resolution

PR #2262 already has a working implementation. Once this issue is labeled `ready for work`, reopening it (narrowed to GET per this issue's scope) should be sufficient — no new PR needed.

### Python & MCP Python SDK

```Text
1.27.0
```

faridun-ag2:
## Summary

In stateless HTTP mode (`stateless_http=True`), GET requests to the MCP endpoint were creating a transport and opening an SSE stream that could never receive server-initiated messages, idling until timeout. DELETE had the same shape (no session to terminate). This wastes connections, especially on serverless platforms (Cloud Run, Lambda).

This PR makes `StreamableHTTPSessionManager._handle_stateless_request` short-circuit GET and DELETE with HTTP 405 before any transport is spawned. POST is unchanged. Stateful mode is unchanged.

The MCP spec permits this: *"Servers MAY return HTTP 405 Method Not Allowed if an SSE stream is not offered at the endpoint."* The TypeScript SDK already implements this behavior.

Closes #2474.

## What changed

- `src/mcp/server/streamable_http_manager.py` — early-return guard at the top of `_handle_stateless_request`. If `request.method` is `GET` or `DELETE`, return a JSON-RPC formatted 405 with `Allow: POST` and bail out before transport creation.
- `tests/server/test_streamable_http_manager.py` — three new tests:
  - `test_stateless_get_returns_405` — status, `Allow` header, JSON-RPC error body
  - `test_stateless_delete_returns_405` — same checks for DELETE
  - `test_stateless_get_does_not_create_transport` — asserts no `StreamableHTTPServerTransport` is instantiated for a stateless GET

## Design notes

- **Layer placement.** The guard lives in the manager, not the transport. The manager already owns the stateless/stateful routing decision; `StreamableHTTPServerTransport` stays mode-agnostic.
- **Response format.** Mirrors the existing 404 "Session not found" handler (`JSONRPCError(jsonrpc="2.0", id=None, error=ErrorData(code=INVALID_REQUEST, ...))` serialized via `model_dump_json(by_alias=True, exclude_unset=True)`), with an added `Allow: POST` header.
- **`Allow` header.** Returns `Allow: POST` (not `GET, POST, DELETE` like the stateful transport's `_handle_unsupported_request`). The difference is intentional — stateless mode genuinely only supports POST.
- **HEAD/OPTIONS in stateless.** Not caught by this guard; they fall through to the transport's `_handle_unsupported_request` and still return 405. The `Allow` header from that path advertises GET/DELETE that stateless doesn't actually support — minor inconsistency, happy to extend the guard if reviewers prefer.
- **No new imports.** `Request`, `Response`, `HTTPStatus`, `INVALID_REQUEST`, `ErrorData`, `JSONRPCError` were all already imported in this module.

## Test plan

- [x] `uv run --frozen pytest tests/server/test_streamable_http_manager.py` — 14/14 pass
- [x] `uv run --frozen pytest tests/ -k stateless` — 13/13 pass
- [x] `uv run --frozen pytest tests/server/` — 477/477 pass (no regressions)
- [x] `uv run --frozen ruff format` + `ruff check` — clean
- [x] `uv run --frozen pyright` — 0 errors
- [x] Targeted coverage on `streamable_http_manager.py` — 100% (branch-covered)
- [x] `strict-no-cover` — clean
- [x] `pre-commit run` on changed files — all applicable hooks pass

### DIFF
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -152,6 +152,29 @@ class StreamableHTTPSessionManager:
 
     async def _handle_stateless_request(self, scope: Scope, receive: Receive, send: Send) -> None:
         """Process request in stateless mode - creating a new transport for each request."""
+        # In stateless mode, only POST is meaningful. GET (SSE stream) and DELETE
+        # (session termination) both require session state that stateless mode does
+        # not maintain, so reject them with 405 before creating a transport.
+        request = Request(scope, receive)
+        if request.method in ("GET", "DELETE"):
+            logger.debug(f"Stateless mode: rejecting {request.method} with 405")
+            error_response = JSONRPCError(
+                jsonrpc="2.0",
+                id=None,
+                error=ErrorData(
+                    code=INVALID_REQUEST,
+                    message=(f"Method Not Allowed: {request.method} is not supported in stateless mode"),
+                ),
+            )
+            response = Response(
+                content=error_response.model_dump_json(by_alias=True, exclude_unset=True),
+                status_code=HTTPStatus.METHOD_NOT_ALLOWED,
+                headers={"Allow": "POST"},
+                media_type="application/json",
+            )
+            await response(scope, receive, send)
+            return
+
         logger.debug("Stateless mode: Creating new transport for this request")
         # No session ID needed in stateless mode
         http_transport = StreamableHTTPServerTransport(
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -413,3 +413,116 @@ def test_session_idle_timeout_rejects_non_positive():
 def test_session_idle_timeout_rejects_stateless():
     with pytest.raises(RuntimeError, match="not supported in stateless"):
         StreamableHTTPSessionManager(app=Server("test"), session_idle_timeout=30, stateless=True)
+
+
+async def _collect_stateless_response(
+    method: str,
+) -> tuple[Message | None, bytes]:
+    """Send a request of the given method to a stateless manager and return
+    (response.start message, response body)."""
+    app = Server("test-stateless-method")
+    manager = StreamableHTTPSessionManager(app=app, stateless=True)
+
+    sent_messages: list[Message] = []
+    response_body = b""
+
+    async def mock_send(message: Message):
+        nonlocal response_body
+        sent_messages.append(message)
+        if message["type"] == "http.response.body":
+            response_body += message.get("body", b"")
+
+    scope = {
+        "type": "http",
+        "method": method,
+        "path": "/mcp",
+        "headers": [
+            (b"content-type", b"application/json"),
+            (b"accept", b"application/json, text/event-stream"),
+        ],
+    }
+
+    async def mock_receive():  # pragma: no cover
+        return {"type": "http.request", "body": b"", "more_body": False}
+
+    async with manager.run():
+        await manager.handle_request(scope, mock_receive, mock_send)
+
+    response_start = next(
+        (msg for msg in sent_messages if msg["type"] == "http.response.start"),
+        None,
+    )
+    return response_start, response_body
+
+
+@pytest.mark.anyio
+async def test_stateless_get_returns_405():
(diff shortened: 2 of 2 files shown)


Explorer-64:
Adding a real-world data point in support of merging this.

We hosted an MCP server on Cloud Run with `stateless_http=True` and hit the exact issue this PR fixes. Cloud Run bills per-second for active requests, so idle GET SSE connections cost money even when the server is doing nothing. With a single connected client: Cloud Run's 301-second request timeout triggers a disconnect, the client immediately reconnects, and the cycle runs continuously at roughly $1-2/day. Zero tool calls, constant billing.

The silent failure makes it worse. Clients show "Connected" but tools never populate because the SSE stream never carries anything useful. There is no visible signal that billing is running. The only way to spot it is to check Cloud Run request logs and notice the 301-second request durations cycling every few minutes.

Our workaround is a manual ASGI guard that returns 405 before the request reaches the MCP app, which is exactly what this PR does at the right layer. Our response body is plain text with actionable instructions so agents and users know what to do:

```
The Imagcon hosted MCP server runs in stateless mode and does not support
persistent SSE connections. Your client is trying to open a long-lived GET
stream, which is not supported here.

Tool calls (POST requests) work fine, but most IDE clients require SSE to
connect, which means you need the local package instead.

Switch to the local package (takes ~30 seconds):
  1. Get a free API key: https://imagcon.app/api-keys
  2. Claude Code: claude mcp add imagcon -- uvx imagcon-mcp --api-key ic_live_...
  3. Cursor: add {"command":"uvx","args":["imagcon-mcp","--api-key","ic_live_..."]}
     to .cursor/mcp.json

Full setup for all clients: https://imagcon.app/.well-known/blueprint.txt
```

The message tells clients: SSE is not supported, POST tool calls still work, and here is exactly what to do instead. Without this, clients silently reconnect in a loop with no signal that anything is wrong.

The JSON-RPC formatted response in this PR is better than our plain-text approach since MCP-aware clients can parse and surface it directly. One suggestion: consider including a `message` field in the JSON-RPC error body that is specific enough for an agent runtime to act on, not just log. Something like "SSE not supported in stateless mode. Use POST for tool calls, or switch to a local stdio transport." would give agents enough context to either retry correctly or surface a useful error to the user.

dbt-labs already cherry-picked this into their fork. The billing angle gives another concrete reason to resolve the conflicts and get this merged.

PR #2457, opened 2026-04-16
fix(streamable-http): expose session_idle_timeout and pause idle reaping during active requests

shaun0927:
Fixes #2455

## Summary

This PR fixes two linked `session_idle_timeout` problems in the StreamableHTTP stack:

1. expose `session_idle_timeout` through the public `streamable_http_app()` entrypoints
2. suspend idle reaping while one or more requests are actively in flight

## Motivation and Context

PR #1994 / #2022 describe `session_idle_timeout` as a way to reap sessions that receive **no HTTP requests** for the configured duration.

Current `main` has two mismatches with that contract:

- `StreamableHTTPSessionManager(...)` accepts `session_idle_timeout`, but `Server.streamable_http_app(...)` and `MCPServer.streamable_http_app(...)` do not, so the canonical high-level API cannot configure the feature
- the idle deadline can fire while a request is still being processed, terminating the session mid-handler even though the session is not idle in any practical sense

## Changes

- add `session_idle_timeout` passthrough to:
  - `mcp.server.lowlevel.server.Server.streamable_http_app()`
  - `mcp.server.mcpserver.server.MCPServer.streamable_http_app()`
- track active request count on `StreamableHTTPServerTransport`
- suspend idle reaping while requests are active
- restore the idle deadline only after the last active request finishes
- initialize the idle scope before the first request can complete

## How Has This Been Tested?

Added focused regression coverage in `tests/server/test_streamable_http_manager.py`:

- `test_streamable_http_app_exposes_session_idle_timeout`
- `test_session_idle_timeout_does_not_cancel_in_flight_request`

The second test uses a handler that runs longer than the configured idle timeout and verifies that the request still completes successfully.

Existing idle reaping coverage (`test_idle_session_is_reaped`) continues to pass, so finished sessions are still cleaned up normally.

Local verification:
- `uv run --frozen pytest tests/server/test_streamable_http_manager.py`
- `uv run --frozen ruff check src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py src/mcp/server/streamable_http.py src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py`
- `uv run --frozen pyright src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py src/mcp/server/streamable_http.py src/mcp/server/streamable_http_manager.py`

## Breaking Changes

None.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
[...]

PR #2510, opened 2026-04-27
fix(streamable_http): expose session_idle_timeout and don't reap active sessions

### LINKED ISSUE #2455: session_idle_timeout is not exposed via streamable_http_app() and can cancel active requests

BuildWithAbid:
## Closes #2455

Fixes both problems described in the issue:

### Problem 1: `session_idle_timeout` not exposed by `streamable_http_app()`

`StreamableHTTPSessionManager(...)` accepted `session_idle_timeout`, but neither `Server.streamable_http_app(...)` nor `MCPServer.streamable_http_app(...)` did. So the canonical high-level API could not configure a supported session-manager feature.

This PR threads `session_idle_timeout: float | None = None` through both wrappers and forwards it to the underlying `StreamableHTTPSessionManager`.

### Problem 2: active requests cancelled by the idle timeout

The previous logic reset the deadline only at the *start* of each incoming request. A handler whose work outlived `session_idle_timeout` would be cancelled mid-flight; the client waited until session termination instead of receiving the tool result.

Replaced the deadline-bumping logic with a per-transport in-flight counter and a `_suspend_idle_timeout()` async context manager:

- on entry: `idle_active_requests += 1`. If this is the first concurrent request, set `idle_scope.deadline = math.inf`.
- on exit: `idle_active_requests -= 1`. If this was the last in-flight request, restore `now + session_idle_timeout`.

Both the new-session and existing-session code paths now wrap `transport.handle_request(...)` with `_suspend_idle_timeout`, so the deadline is suspended for the lifetime of every request.

## Test plan

All in `tests/server/test_streamable_http_manager.py`:

- [x] `test_session_idle_timeout_passthrough_lowlevel` — verifies the kwarg propagates from `Server.streamable_http_app()` to the manager.
- [x] `test_session_idle_timeout_passthrough_mcpserver` — same for `MCPServer.streamable_http_app()`.
- [x] `test_suspend_idle_timeout_sets_deadline_inf_then_restores` — direct unit test of the helper: deadline is `inf` during the request, finite (≈ `now + timeout`) after.
- [x] `test_suspend_idle_timeout_only_restores_after_last_concurrent_request` — nested suspensions keep the deadline at `inf` until the outermost exit; counter logic correctness.
- [x] `test_suspend_idle_timeout_no_op_without_timeout` — helper is inert when no timeout is configured.
- [x] `test_long_running_request_outlives_idle_timeout` — **regression for problem 2**: end-to-end via `httpx.ASGITransport` + `Client`, a tool whose handler sleeps 4× the timeout still completes successfully.
- [x] `test_idle_session_reaped_after_request_completes` — the session is still reaped once the request finishes, so existing reaping behavior is preserved.

Local results:

\`\`\`
tests/server/test_streamable_http_manager.py    18 passed
tests/server/ + tests/shared/test_streamable_http.py    542 passed
ruff check + ruff format    clean
pyright    0 errors, 0 warnings
\`\`\`

## Notes

- No public API removed — `session_idle_timeout` is a new kwarg with the same `None` default everywhere.
- The validation guards (`session_idle_timeout > 0`, mutually exclusive with `stateless`) still live on `StreamableHTTPSessionManager.__init__` and continue to raise correctly when invoked through either wrapper.

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -567,6 +567,7 @@ class Server(Generic[LifespanResultT]):
         stateless_http: bool = False,
         event_store: EventStore | None = None,
         retry_interval: int | None = None,
+        session_idle_timeout: float | None = None,
         transport_security: TransportSecuritySettings | None = None,
         host: str = "127.0.0.1",
         auth: AuthSettings | None = None,
@@ -591,6 +592,7 @@ class Server(Generic[LifespanResultT]):
             json_response=json_response,
             stateless=stateless_http,
             security_settings=transport_security,
+            session_idle_timeout=session_idle_timeout,
         )
         self._session_manager = session_manager
 
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -1050,6 +1050,7 @@ class MCPServer(Generic[LifespanResultT]):
         stateless_http: bool = False,
         event_store: EventStore | None = None,
         retry_interval: int | None = None,
+        session_idle_timeout: float | None = None,
         transport_security: TransportSecuritySettings | None = None,
         host: str = "127.0.0.1",
     ) -> Starlette:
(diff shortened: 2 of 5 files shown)


maxisbey:
Closing as a duplicate of #2457, which was opened first with a smaller diff. Thanks for the contribution.

PR #2511, opened 2026-04-27
feat: add custom content support to ToolError

### LINKED ISSUE #348: No way to set isError=True for arbitrary tool result content

blackwell-systems:
## Summary

Adds an optional `content` parameter to `ToolError`, allowing tools to return arbitrary content blocks (images, embedded resources, multiple text blocks) with `isError=True`.

Previously, `ToolError` only carried a string message, which was always converted to a single `TextContent`. This made it impossible to return rich error content like a screenshot of a failed render or structured diagnostics alongside the error flag.

Follows the approach suggested by @Kludex in #348.

```python
# Existing behavior (unchanged)
raise ToolError("Something went wrong")

# New: error with custom content
raise ToolError(
    "Image processing failed",
    content=[
        ImageContent(type="image", data="...", mime_type="image/png"),
        TextContent(type="text", text="Additional error details"),
    ],
)
```

## Changes

- **`src/mcp/server/mcpserver/exceptions.py`**: Add optional `content: list[ContentBlock] | None` parameter to `ToolError.__init__`
- **`src/mcp/server/mcpserver/tools/base.py`**: Re-raise `ToolError` as-is in `Tool.run()` to preserve custom content (without this, `ToolError` is caught by the generic `except Exception` and wrapped in a new `ToolError`, losing the content)
- **`src/mcp/server/mcpserver/server.py`**: Handle `ToolError` with custom content in `_handle_call_tool`, falling back to `str(e)` when no content is provided
- **`src/mcp/server/mcpserver/__init__.py`**: Export `ToolError` for convenient imports

## Tests

Added `tests/server/mcpserver/test_tool_error_content.py` with 5 tests:
1. Text-only ToolError still works as before (backwards compatibility)
2. ToolError with image content returns custom content with `isError=True`
3. ToolError with explicit TextContent list uses that content
4. ToolError with `content=None` falls back to message string
5. Generic exceptions (non-ToolError) still handled correctly (regression check)

All existing tests pass (94 total). Pyright and ruff clean.

## Context

I maintain [mcp-assert](https://github.com/blackwell-systems/mcp-assert), an MCP server testing tool. We scanned 54 servers and found `isError` handling is the most common failure pattern: servers return `-32603` instead of `isError: true`, leaving agents unable to recover. This SDK fix enables server authors to return richer error content that agents can act on.

Note: #1824 addressed the same issue but has been open since January with merge conflicts and no review. This PR is a fresh implementation rebased on current `main`.

Closes #348

### DIFF
--- a/src/mcp/server/mcpserver/__init__.py
+++ b/src/mcp/server/mcpserver/__init__.py
@@ -3,7 +3,8 @@
 from mcp.types import Icon
 
 from .context import Context
+from .exceptions import ToolError
 from .server import MCPServer
 from .utilities.types import Audio, Image
 
-__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon"]
+__all__ = ["MCPServer", "Context", "Image", "Audio", "Icon", "ToolError"]
--- a/src/mcp/server/mcpserver/exceptions.py
+++ b/src/mcp/server/mcpserver/exceptions.py
@@ -1,5 +1,12 @@
 """Custom exceptions for MCPServer."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from mcp.types import ContentBlock
+
 
 class MCPServerError(Exception):
     """Base error for MCPServer."""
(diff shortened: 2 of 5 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #1824.

PR #2513, opened 2026-04-28
fix: move logging configuration from MCPServer.__init__ to run()

### LINKED ISSUE #1656: FastMCP configures logging on init, which messes up application-level logging

dragogargo:
## Motivation and Context

Fixes #1656. `MCPServer.__init__()` calls `configure_logging()` which invokes `logging.basicConfig()`, adding handlers and setting the level on the root logger. This violates the [Python logging best practice](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) that library code should never configure the root logger.

Applications embedding `MCPServer` as a library had their logging config silently overwritten at construction time.

## Changes

- Move `configure_logging()` call from `MCPServer.__init__()` to `MCPServer.run()` — the actual application entrypoint for standalone server processes
- Add regression test verifying that constructing `MCPServer` does not modify the root logger

## How Has This Been Tested?

- Regression test (`test_mcpserver_init_does_not_configure_logging`) that captures root logger state before/after `MCPServer()` construction and asserts no handlers were added and log level was not changed
- All 90 existing `test_server.py` tests pass

## Breaking Changes

No. `run()` still configures logging before starting the server. Only consumers who instantiate `MCPServer` without calling `run()` (library usage) are affected — they now keep full control of their logging setup, which is the desired behavior.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -203,9 +203,6 @@ class MCPServer(Generic[LifespanResultT]):
             self._token_verifier = ProviderTokenVerifier(auth_server_provider)
         self._custom_starlette_routes: list[Route] = []
 
-        # Configure logging
-        configure_logging(self.settings.log_level)
-
     @property
     def name(self) -> str:
         return self._lowlevel_server.name
@@ -290,6 +287,8 @@ class MCPServer(Generic[LifespanResultT]):
         TRANSPORTS = Literal["stdio", "sse", "streamable-http"]
         if transport not in TRANSPORTS.__args__:  # type: ignore  # pragma: no cover
             raise ValueError(f"Unknown transport: {transport}")
+
+        configure_logging(self.settings.log_level)
 
         match transport:
             case "stdio":
--- a/tests/server/mcpserver/test_server.py
+++ b/tests/server/mcpserver/test_server.py
@@ -1516,3 +1516,22 @@ async def test_report_progress_passes_related_request_id():
         message="halfway",
         related_request_id="req-abc-123",
     )
+
+
+def test_mcpserver_init_does_not_configure_logging() -> None:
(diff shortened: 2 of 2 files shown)


maxisbey:
Closing in favor of #2503. Per the discussion on #1656, moving `configure_logging()` to `run()` doesn't fully solve the problem — `run()` is still library code, so embedding apps would still get their logging configured. The accepted direction is to remove the call entirely. Thanks for the contribution.

PR #2514, opened 2026-04-28
fix: send notifications/cancelled on request timeout and caller cancellation

### LINKED ISSUE #2507: `ClientSession` never sends `notifications/cancelled` when `call_tool` is cancelled — server-side coroutines leak
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK (1.29.0 verified)

### Description

`ClientSession.send_request()` (and therefore every `call_tool`, `list_tools`, etc.) never emits a `notifications/cancelled` message when its in-flight `await` is interrupted, regardless of whether the interruption comes from the SDK's own timeout or from the caller's `asyncio.wait_for`. The MCP spec (cancellation.mdx) requires the sender to issue this notification on timeout, and any cooperative cancellation likewise leaves the server with an orphan request.

Empirical impact: server-side tool coroutines remain suspended after a client cancellation. They hold whatever resources they had acquired (DB connections, cursors, locks, file handles) until the session itself ends. With long-lived sessions and cancellable workloads, every cancelled call is a silent leak.

This was previously raised in #1458 ("Missing Cancellation Notifications on Request Timeout") and closed as `DUPLICATE`, but the proposed fix never landed and the underlying behavior is still present in `1.29.0`. The current report adds a second uncovered path (external `CancelledError`) and concrete evidence of the resource-leak impact, so I'm filing rather than commenting on the closed issue.

### Two paths, neither sends the notification

**Path A — SDK-internal timeout (`anyio.fail_after`):** Already documented in #1458. The `except TimeoutError` branch raises `McpError` and falls into the `finally` block that closes the local response stream. No `CancelledNotification` is sent.

`mcp/shared/session.py:290-303` (1.29.0):

```python
try:
except TimeoutError:
    raise McpError(
        ErrorData(
            code=httpx.codes.REQUEST_TIMEOUT,
            message=(
                f"Timed out while waiting for response to "
                f"{request.__class__.__name__}. Waited "
                f"{timeout} seconds."
            ),
        )
    )
```

**Path B — external cancellation:** When the caller wraps `session.call_tool(...)` in `asyncio.wait_for(...)` (or any other cancellation source), `asyncio.CancelledError` is raised at the `await response_stream_reader.receive()` point. There is no `except` for `CancelledError` / `anyio.get_cancelled_exc_class()` anywhere in `send_request`. The exception flows up through the `finally`, which only cleans up the *client-local* response stream:

`mcp/shared/session.py:310-313`:

```python
finally:
    self._response_streams.pop(request_id, None)
    self._progress_callbacks.pop(request_id, None)
    await response_stream.aclose()
    await response_stream_reader.aclose()
```

The server is never told the request is gone. Its in-flight tool task continues to completion, then sends back a response that gets dropped because no one is reading the response stream.

### Reproduction

Minimal repro script (full version: https://github.com/sherman94062/databricks-ai-steward/blob/main/stress/probe_a1_leak.py):

```python
import asyncio

# stress.server is a tiny FastMCP server with two tools:
#   hangs_forever_async_guarded — `await asyncio.sleep(300)`
#   task_count                  — returns len(asyncio.all_tasks()) - 1
async def main():
        command="python", args=["-m", "stress.server"]
    )
            await session.initialize()

            baseline = await session.call_tool("task_count", {})
            print("baseline:", baseline)

            for _ in range(50):
                try:
                    await asyncio.wait_for(
                        session.call_tool("hangs_forever_async_guarded", {}),
                        timeout=0.1,
                    )
                except asyncio.TimeoutError:
                    pass

            await asyncio.sleep(0.5)  # let any cleanup settle
            after = await session.call_tool("task_count", {})
            print("after 50 cancels:", after)

asyncio.run(main())
```

Output (mcp 1.29.0, Python 3.14):

```
baseline:        4 tasks
after 50 cancels: 54 tasks
```

50 cancelled `call_tool` invocations → 50 leaked server-side coroutines, persistent until the session closes. Same numbers under stdio and `streamable_http`.

If the client sent `notifications/cancelled`, the server's existing handler at `mcp/shared/session.py:402-406` would cancel each leaked coroutine immediately:

```python
if isinstance(notification.root, CancelledNotification):
    cancelled_id = notification.root.params.requestId
    if cancelled_id in self._in_flight:
```

The server side already does the right thing on receipt. Only the client-side emit is missing.

### Spec citations

> Implementations **SHOULD** establish timeouts for all sent requests… When the request has not received a success or error response within the timeout period, the sender **SHOULD** issue a cancellation notification for that request and stop waiting for a response.

> Either side can cancel an in-progress request by sending a cancellation notification.

— https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/cancellation.mdx

### Proposed fix

Two small additions to `BaseSession.send_request`. The internal-timeout branch is exactly what #1458 proposed; the external-cancellation branch is new and uses `anyio.get_cancelled_exc_class()` so it works under both asyncio and trio backends:

```python
try:
except TimeoutError:
    await self._send_cancelled_notification(request_id, "request timed out")
    raise McpError(...)
except anyio.get_cancelled_exc_class():
    await self._send_cancelled_notification(request_id, "request cancelled by caller")
    raise


async def _send_cancelled_notification(self, request_id, reason):
    try:
        await self.send_notification(
            ClientNotification(
                CancelledNotification(
                    method="notifications/cancelled",
                        requestId=request_id, reason=reason
                    ),
                )
            )
        )
    except Exception:
        # Best-effort: if the transport is already gone, nothing to do.
        logger.warning(
            "failed to send cancellation notification for request %s",
            request_id,
        )
```

The notification *must* be sent before re-raising — once the cancellation propagates out of `send_request`, the caller may close the session and the write stream becomes unusable. A small async-shielded wrapper around the `send_notification` call may be needed to guarantee delivery on the cancellation path; happy to put that into a PR.

### Why this matters in practice

Most production MCP servers acquire external resources inside tool handlers — DB connections, HTTP clients, transactions, file locks. Without the cancellation notification, every aborted client call wastes one such resource for the lifetime of the session. We discovered this while building a Databricks-facing MCP server: a connection pool of 10 plus 10 cancelled tool calls = pool exhausted.

A server-side per-tool timeout (`asyncio.wait_for` inside the tool wrapper) bounds the leak window, but it shouldn't be load-bearing. The client should tell the server when a request is dead.

### Environment

- `mcp` 1.29.0 (latest at time of writing)
- Python 3.14.3, macOS 14
- Same behavior reproduced with `streamable_http` transport (different transport, identical client cancellation path)

### Related

- #1458 — same bug, closed as DUPLICATE, fix not shipped
- #1420 — closed; missing cancellation reason logging (server-side)
- #1419, #2480 — closed/open; server-side spec gap (sending response after cancellation)
- #2416 — open; server-side cancellation race

dragogargo:
## Motivation and Context

Fixes #2507. `ClientSession.send_request()` never emits a `notifications/cancelled` message when a request is interrupted — either by the SDK's own `anyio.fail_after` timeout or by external cancellation (e.g. `asyncio.wait_for`). The [MCP spec requires](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/basic/utilities/cancellation.mdx) the sender to issue this notification on timeout, and cooperative cancellation likewise needs it.

**Impact:** Server-side tool coroutines remain suspended after a client cancellation, holding DB connections, cursors, locks, and file handles until the session ends. With long-lived sessions, every cancelled call is a silent resource leak. The issue author demonstrated 50 cancelled `call_tool` invocations → 50 leaked server-side coroutines.

The server already handles `CancelledNotification` correctly (cancels the in-flight request). Only the client-side emit is missing.

## Changes

1. **Timeout path** (`TimeoutError`): send `notifications/cancelled` before raising `MCPError`
2. **External cancellation path** (`CancelledError`): catch `anyio.get_cancelled_exc_class()`, send notification, re-raise
3. **`_send_cancelled_notification` helper**: builds and sends the notification inside a shielded cancel scope to guarantee delivery even when called from a cancelled task
4. **Two regression tests**: one for each cancellation path, using in-memory streams

## How Has This Been Tested?

- `test_send_request_sends_cancelled_notification_on_timeout` — verifies the notification is sent when `read_timeout_seconds` triggers
- `test_send_request_sends_cancelled_notification_on_caller_cancel` — verifies the notification is sent when the caller cancels the task externally
- All 10 existing `test_session.py` tests pass

## Breaking Changes

No. The only behavioral change is that the client now sends an additional notification that the server already expects and handles.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -144,7 +144,7 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
         """Cancel this request and mark it as completed."""
         if not self._entered:
             raise RuntimeError("RequestResponder must be used as a context manager")
-        if not self._cancel_scope:
+        if not self._cancel_scope:  # pragma: no cover
             raise RuntimeError("No active cancel scope")
 
         self._cancel_scope.cancel()
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -23,6 +23,7 @@ from mcp.types import (
     INVALID_PARAMS,
     REQUEST_TIMEOUT,
     CancelledNotification,
+    CancelledNotificationParams,
     ClientNotification,
     ClientRequest,
     ClientResult,
@@ -292,9 +293,13 @@ class BaseSession(
                     with anyio.fail_after(timeout):
                         response_or_error = await response_stream_reader.receive()
                 except TimeoutError:
+                    await self._send_cancelled_notification(request_id, "request timed out")
                     class_name = request.__class__.__name__
                     message = f"Timed out while waiting for response to {class_name}. Waited {timeout} seconds."
                     raise MCPError(code=REQUEST_TIMEOUT, message=message)
+                except anyio.get_cancelled_exc_class():
+                    await self._send_cancelled_notification(request_id, "request cancelled by caller")
+                    raise
 
                 if isinstance(response_or_error, JSONRPCError):
                     raise MCPError.from_jsonrpc_error(response_or_error)
@@ -539,6 +544,26 @@ class BaseSession(
         message: str | None = None,
     ) -> None:
         """Sends a progress notification for a request that is currently being processed."""
+
+    async def _send_cancelled_notification(self, request_id: RequestId, reason: str) -> None:
+        """Send a cancellation notification to the remote side (best-effort).
+
+        Uses a shielded cancel scope so the notification is delivered even
+        when called from inside a cancelled task.
+        """
+        try:
+            with anyio.CancelScope(shield=True):
+                notification = CancelledNotification(
+                    method="notifications/cancelled",
+                    params=CancelledNotificationParams(request_id=request_id, reason=reason),
+                )
+                jsonrpc_notification = JSONRPCNotification(
+                    jsonrpc="2.0",
+                    **notification.model_dump(by_alias=True, mode="json", exclude_none=True),
+                )
+                await self._write_stream.send(SessionMessage(message=jsonrpc_notification))
+        except Exception:
+            logging.warning("Failed to send cancellation notification for request %s", request_id)
 
     async def _handle_incoming(
         self, req: RequestResponder[ReceiveRequestT, SendResultT] | ReceiveNotificationT | Exception
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -548,20 +548,22 @@ class BaseSession(
     async def _send_cancelled_notification(self, request_id: RequestId, reason: str) -> None:
         """Send a cancellation notification to the remote side (best-effort).
 
-        Uses a shielded cancel scope so the notification is delivered even
-        when called from inside a cancelled task.
+        Uses a shielded cancel scope with a timeout so the notification is
+        delivered even when called from inside a cancelled task, but does not
+        block shutdown if the write stream is unavailable.
         """
         try:
             with anyio.CancelScope(shield=True):
-                notification = CancelledNotification(
-                    method="notifications/cancelled",
-                    params=CancelledNotificationParams(request_id=request_id, reason=reason),
-                )
-                jsonrpc_notification = JSONRPCNotification(
-                    jsonrpc="2.0",
-                    **notification.model_dump(by_alias=True, mode="json", exclude_none=True),
-                )
-                await self._write_stream.send(SessionMessage(message=jsonrpc_notification))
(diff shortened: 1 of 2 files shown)


dragogargo:
CI note: Ubuntu checks all pass (10/10). Windows failures are from `test_notification_validation_error` in `tests/issues/test_88_random_error.py` — unrelated to this PR's changes (`session.py` + `test_session.py`).

Issue #1464, opened 2025-10-10
Improve docs on how to use `instructions` on `InitializeResult`


felixweinberger:
### Question

Lots of people want tool bundling / grouping / namespaces to guide servers how to use  tools together.

We should make `instructions` more obvious and have examples for how to use it - maybe this can address a lot of requests for bundling and namespaces if it's more obvious this is something you can do.

https://modelcontextprotocol.io/specification/2025-06-18/schema#initializeresult-instructions

ChenyangLi4288:
Hi Felix,

If I am trying to provide with more obvious examples. Should I open a PR to [modelcontextprotocol](https://github.com/modelcontextprotocol/modelcontextprotocol) and edit the 'draft' version?

Thanks!

maxisbey:
> Hi Felix,
> 
> If I am trying to provide with more obvious examples. Should I open a PR to [modelcontextprotocol](https://github.com/modelcontextprotocol/modelcontextprotocol) and edit the 'draft' version?
> 
> Thanks!

I think this is more about how to do it in the Python SDK rather than the protocol itself. If you'd like to make a PR for this repo with examples for how to use instructions with the SDK then feel free!

ryanwilliams90:
Hi — happy to take this on. Plan:

1. Add an example snippet `examples/snippets/servers/instructions_example.py` showing `instructions=` on `MCPServer` used to describe how a small set of related tools work together (the bundling/workflow case).
2. Add a short `Server instructions` section to `README.v2.md` (and reference it from the snippet) explaining when to use `instructions`, with concrete examples for tool-bundling and domain-context use cases.
3. No public-API changes — purely docs and an example.

I'll target `main`. Could you assign the issue to me?

Joosboy:
Hi! I have implemented this documentation improvement. Since pull request creation is currently restricted on this repository, I have pushed the changes to my fork branch: https://github.com/Joosboy/python-sdk/tree/docs-server-instructions-1464

Please let me know if a maintainer can open the PR or pull these changes. Thanks!

PR #2515, opened 2026-04-28
Document FastMCP server instructions

MukundaKatta:
Adds a README section showing how to use FastMCP `instructions` to communicate workflow guidance and tool ordering/grouping context to clients through `InitializeResult`.\n\nCloses #1464.\n\nVerification:\n- Documentation-only change

### DIFF
--- a/README.v2.md
+++ b/README.v2.md
@@ -370,6 +370,35 @@ async def long_running_task(task_name: str, ctx: Context, steps: int = 5) -> str
 _Full example: [examples/snippets/servers/tool_progress.py](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/snippets/servers/tool_progress.py)_
 <!-- /snippet-source -->
 
+#### Server Instructions
+
+Use `instructions` to tell clients how the server's tools and resources fit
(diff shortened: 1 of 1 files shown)


ryanwilliams90:
Hi @MukundaKatta — I started on #1464 today and posted a plan there before noticing this PR was already open. Apologies for the overlap.

Happy to defer if this lands. If it would be useful, I have a slightly broader take that I'm happy to either fold into this PR (with credit) or open separately if maintainers prefer:

- A runnable snippet file (`examples/snippets/servers/instructions_example.py`) wired through `update_readme_snippets.py` so the README block stays in sync the same way as the other server examples.
- A worked example covering the bundling / sequencing case the original issue calls out (`search → fetch → confirm → update`), plus explicit guidance on what belongs in `instructions` vs. tool docstrings — that's the design call I find people get wrong most often.

@maxisbey — you green-lit the docs work on #1464; happy to be guided on which direction is most useful. I can also just leave review comments here if you'd rather land this PR as-is.

maxisbey:
Closing in favor of your #2598, which uses the snippet machinery so the example stays in sync with the code.

PR #2524, opened 2026-05-01
fix(cli): avoid Windows shell for mcp dev

### LINKED ISSUE #1257: Don’t use shell=True in mcp dev subprocess on Windows (command injection risk)

Genmin:
## Summary

- remove Windows `shell=True` from the `mcp dev` inspector subprocess path
- resolve the Windows `npx` executable with `shutil.which()` instead of probing through a shell
- add a regression test that exercises a path containing `&` and verifies the subprocess receives an argument list without shell dispatch

Fixes #1257.

## Validation

- `uv run pytest tests/cli -q`
- `uv run ruff check src/mcp/cli/cli.py tests/cli/test_utils.py`
- `uv run pyright src/mcp/cli/cli.py tests/cli/test_utils.py`
- `git diff --check`

### DIFF
--- a/src/mcp/cli/cli.py
+++ b/src/mcp/cli/cli.py
@@ -3,6 +3,7 @@
 import importlib.metadata
 import importlib.util
 import os
+import shutil
 import subprocess
 import sys
 from pathlib import Path
@@ -44,11 +45,8 @@ def _get_npx_command():
     if sys.platform == "win32":
         # Try both npx.cmd and npx.exe on Windows
         for cmd in ["npx.cmd", "npx.exe", "npx"]:
-            try:
-                subprocess.run([cmd, "--version"], check=True, capture_output=True, shell=True)
-                return cmd
-            except subprocess.CalledProcessError:
-                continue
+            if resolved := shutil.which(cmd):
+                return resolved
         return None
     return "npx"  # On Unix-like systems, just use npx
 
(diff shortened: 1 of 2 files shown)


yyzquwu:
Tested this on Windows.

Environment:
- Windows 11
- Python `3.13.13`
- PR head `a176ccc07a5c48d439a909f1d9dcb9c71287dc81`

What passed:

```text
pytest tests/cli -q
# 27 passed

ruff check src/mcp/cli/cli.py tests/cli/test_utils.py
# passed
```

One thing still fails locally:

```text
pyright src/mcp/cli/cli.py tests/cli/test_utils.py
# tests/cli/test_utils.py:87:29 - "shutil" is not exported from module ".cli"
# tests/cli/test_utils.py:98:29 - "shutil" is not exported from module ".cli"
# tests/cli/test_utils.py:125:29 - "subprocess" is not exported from module ".cli"
# tests/cli/test_utils.py:157:29 - "subprocess" is not exported from module ".cli"
# tests/cli/test_utils.py:204:29 - "subprocess" is not exported from module ".cli"
# tests/cli/test_utils.py:228:29 - "subprocess" is not exported from module ".cli"
```

The fix itself looks right to me. The remaining issue seems test-side: the new tests monkeypatch `cli.shutil` and `cli.subprocess`, which trips `reportPrivateImportUsage`. Importing `shutil` and `subprocess` in the test module and monkeypatching those module objects directly should keep the same behavior without the private import warnings, because `cli.py` holds the same imported modules.

PR #2525, opened 2026-05-01
fix(server): skip output schema validation for tool errors

### LINKED ISSUE #2429: [v1.x] Server-side outputSchema validation blocks tool error reporting (isError: true)
### Initial Checks

### Description

When a tool handler returns unstructured content for an error case, the low-level server's call_tool decorator validates the output against outputSchema before checking if the response is an error. 

This prevents tools with outputSchema from reporting errors via isError: true, as the validation fails with "outputSchema defined but no structured output returned", replacing the original error message.

This was already fixed in the TypeScript SDK via modelcontextprotocol/typescript-sdk#654 / PR #655 (2025-06-24), but the equivalent fix has not been applied to the Python SDK v1.x branch.

The issue is in src/mcp/server/lowlevel/server.py, the call_tool decorator handler:

1. Line ~560: outputSchema validation runs unconditionally — should skip when the result is an error
2. Line ~575: isError=False is hardcoded — there is no path for the handler to signal an error through unstructured content while outputSchema is defined

### Example Code

```Python
from mcp import types

server = Server("test")

@server.call_tool()
async def call_tool(name: str, arguments: dict):
    # Tool has outputSchema but needs to return an error
    # This gets blocked by outputSchema validation
    return [types.TextContent(type="text", text="Resource not found")]
```

### Python & MCP Python SDK

```Text
- Python: 3.11 / 3.12
- MCP SDK: v1.26.0+ (v1.x branch)
```

Genmin:
## Summary

Fixes #2429.

When a typed FastMCP tool returns a `CallToolResult` with `isError=True`, the v1.x server should preserve that error result. Previously, `FuncMetadata.convert_result()` still validated `structuredContent` against the inferred output schema, so error results without structured content raised a Pydantic validation error and replaced the intended tool error.

This gates schema validation to non-error `CallToolResult` values while preserving validation for successful structured tool results.

## Validation

- `uv run pytest tests/server/fastmcp/test_func_metadata.py -q`
- `uv run pytest tests/server/fastmcp -q`
- `uv run ruff check src/mcp/server/fastmcp/utilities/func_metadata.py tests/server/fastmcp/test_func_metadata.py`
- `uv run pyright src/mcp/server/fastmcp/utilities/func_metadata.py tests/server/fastmcp/test_func_metadata.py`
- `git diff --check`

### DIFF
--- a/src/mcp/server/fastmcp/utilities/func_metadata.py
+++ b/src/mcp/server/fastmcp/utilities/func_metadata.py
@@ -112,7 +112,7 @@ class FuncMetadata(BaseModel):
         the structured output.
         """
         if isinstance(result, CallToolResult):
-            if self.output_schema is not None:
+            if self.output_schema is not None and not result.isError:
                 assert self.output_model is not None, "Output model must be set if output schema is defined"
                 self.output_model.model_validate(result.structuredContent)
             return result
--- a/tests/server/fastmcp/test_func_metadata.py
+++ b/tests/server/fastmcp/test_func_metadata.py
@@ -878,6 +878,20 @@ def test_tool_call_result_annotated_is_structured_and_invalid():
         meta.convert_result(func_returning_annotated_tool_call_result())
 
 
+def test_tool_call_result_annotated_error_skips_structured_validation():
+    class PersonClass(BaseModel):
+        name: str
+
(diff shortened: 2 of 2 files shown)


maxisbey:
Closing as a duplicate of #2467 (the issue author's v1.x PR with the same change). Thanks for the contribution.

PR #2526, opened 2026-05-01
fix(client): forward User-Agent through OAuth auth flow

### LINKED ISSUE #1664: User-Agent header in sHTTP transport is not forwarded to auth flow

Genmin:
## What
- Forward the original transport `User-Agent` header onto OAuth-generated discovery, registration, refresh, and token requests.
- Add a regression that walks the auth generator through PRM discovery, OASM discovery, dynamic client registration, and token exchange while asserting the custom `User-Agent` is preserved.

## Why
Streamable HTTP callers can configure a custom `User-Agent` on their HTTP client, but OAuth requests are created inside the auth provider as fresh `httpx.Request` instances. Those generated requests dropped the caller's user agent, which breaks deployments behind WAF rules that require it.

Fixes #1664

## Tests
- `uv run pytest tests/client/test_auth.py -k user_agent -q`
- `uv run pytest tests/client/test_auth.py -q`
- `uv run ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `uv run ruff format --check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `git diff --check`

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -273,6 +273,12 @@ class OAuthClientProvider(httpx.Auth):
         self._validate_resource_url_callback = validate_resource_url
         self._initialized = False
 
+    def _copy_user_agent_header(self, request: httpx.Request, source_request: httpx.Request) -> httpx.Request:
+        user_agent = source_request.headers.get("User-Agent")
+        if user_agent and "User-Agent" not in request.headers:
+            request.headers["User-Agent"] = user_agent
+        return request
+
     async def _handle_protected_resource_response(self, response: httpx.Response) -> bool:
         """Handle protected resource metadata discovery response.
 
@@ -515,6 +521,7 @@ class OAuthClientProvider(httpx.Auth):
             if not self.context.is_token_valid() and self.context.can_refresh_token():
                 # Try to refresh token
                 refresh_request = await self._refresh_token()  # pragma: no cover
+                self._copy_user_agent_header(refresh_request, request)  # pragma: no cover
                 refresh_response = yield refresh_request  # pragma: no cover
 
                 if not await self._handle_refresh_response(refresh_response):  # pragma: no cover
@@ -539,6 +546,7 @@ class OAuthClientProvider(httpx.Auth):
 
                     for url in prm_discovery_urls:  # pragma: no branch
                         discovery_request = create_oauth_metadata_request(url)
+                        self._copy_user_agent_header(discovery_request, request)
 
                         discovery_response = yield discovery_request  # sending request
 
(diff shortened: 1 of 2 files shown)


Genmin:
Pushed a follow-up for the coverage blocker. The full-flow exception path is now covered directly, so the auth flow still logs `OAuth flow error` and re-raises while keeping the repo-wide coverage gate strict.

Validation:
- `uv run pytest tests/client/test_auth.py -q` -> 98 passed, 1 xfailed
- `uv run ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py` -> passed
- `uv run ruff format --check src/mcp/client/auth/oauth2.py tests/client/test_auth.py` -> passed
- `uv run coverage run -m pytest -n auto` -> 1174 passed, 98 skipped, 1 xfailed
- `uv run coverage combine`
- `uv run coverage report` -> 100.00%
- `uv run strict-no-cover` -> passed
- `git diff --check` -> passed

Genmin:
Follow-up pushed for the pre-commit/pyright failure: the new `caplog` fixture is now typed as `pytest.LogCaptureFixture`.

Validated locally after the update:
- `uv run pyright tests/client/test_auth.py`
- `uv run pytest tests/client/test_auth.py::TestAuthFlow::test_auth_flow_logs_and_reraises_oauth_errors -q`
- `uv run pytest tests/client/test_auth.py -q`
- `uv run ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `uv run ruff format --check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `git diff --check`

One local note: `uv run pre-commit run pyright --files tests/client/test_auth.py` currently exits before running pyright because the cached `pre-commit-hooks` manifest has an unsupported `stages: [pre-commit]` entry for this local pre-commit version. Direct pyright is clean and matches the CI error this commit addresses.

maxisbey:
Closing as a duplicate of #2077, which was opened first with the same fix at the same call sites. Thanks for the contribution.

PR #2528, opened 2026-05-01
fix: avoid closing stdio server streams

HenryLee789:
## Summary
- wrap default stdin/stdout with duplicated file descriptors so stdio_server does not take ownership of the real process streams
- close only the duplicated wrappers on exit
- add a regression test that verifies sys.stdin/sys.stdout remain open after stdio_server exits

Fixes #1933

## Testing
- `uv run pytest tests/server/test_stdio.py -q`
- `uv run ruff check src/mcp/server/stdio.py tests/server/test_stdio.py`

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -19,8 +19,10 @@ Example:
 
 import os
 import sys
+from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


maxisbey:

PR #2531, opened 2026-05-01
feat: allow overriding SSE messages endpoint

### LINKED ISSUE #2255: Allow explicit `message_url` override in `mcp.client.sse.sse_client`
### Initial Checks

### Description

When using the MCP Python client with SSE transport, the `sse_client` helper only accepts the SSE endpoint:

```python
async with sse_client(url=mcp_server_sse_url, headers=headers, timeout=timeout) as streams:
    async with ClientSession(*streams) as session:
        await session.initialize()
        return await session.call_tool(name=name, arguments=arguments)
```

Internally, the MCP client derives the **messages endpoint automatically from the SSE URL and sse.data**.

However, this assumption breaks in deployments where **reverse proxies (e.g., Nginx) inject path prefixes** that the backend service is unaware of.

**Example deployment**

Externally exposed endpoints after reverse proxying:

```
https://example.com/api/v1/sse
https://example.com/api/v1/messages/{session_id}
```

Internally, the MCP server only defines:

* **SSE endpoint:** `/v1/sse`
* **Messages POST endpoint:** `/v1/messages/`

The `/api` prefix is injected by the reverse proxy.

In this situation:

* The client connects sse endpoint correct (Because we are providing the url): `https://example.com/api/v1/sse`
* But the MCP client internally derives the messages endpoint as:

```
https://example.com/v1/messages/{session_id} -> Based on sse.data
```

instead of:

```
https://example.com/api/v1/messages/{session_id}
```

Because the client assumes a fixed base path.

**Current limitation**

The `sse_client` interface does not allow explicitly providing a `messages_url`, which makes it impossible to support deployments where:

* the base path is rewritten by a reverse proxy
* the backend service is unaware of the external prefix

**Proposed improvement**

Allow explicitly specifying the messages endpoint when creating the SSE client:

```python
async with sse_client(
    url="https://example.com/api/v1/sse",
    messages_url="https://example.com/api/v1/messages",
    headers=headers,
    timeout=timeout
) as streams:
```

Behavior:

* If `messages_url` is provided → use it directly
* Otherwise → fall back to the current behavior of deriving it from the SSE URL

This change would allow the MCP Python client to work correctly in **reverse-proxy deployments, versioned APIs, and gateway setups**, while maintaining backward compatibility.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Python 3.13.12
mcp>=1.26.0
```

HenryLee789:
## Summary
- add an optional `messages_url` override to `sse_client()` for reverse-proxy deployments
- preserve session query parameters from the server endpoint event when an override is used
- cover absolute and relative overrides plus end-to-end client POST routing

Fixes #2255

## Checks
- uv run pytest tests/shared/test_sse.py -q
- uv run ruff format --check src/mcp/client/sse.py tests/shared/test_sse.py
- uv run ruff check src/mcp/client/sse.py tests/shared/test_sse.py
- uv run pyright src/mcp/client/sse.py tests/shared/test_sse.py
- uv run coverage combine && uv run coverage report --include='src/mcp/client/sse.py' (100%)
- git diff --check

### DIFF
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -27,6 +27,20 @@ def _extract_session_id_from_endpoint(endpoint_url: str) -> str | None:
     return query_params.get("sessionId", [None])[0] or query_params.get("session_id", [None])[0]
 
 
+def _resolve_endpoint_url(sse_url: str, endpoint_data: str, messages_url: str | None = None) -> str:
+    if messages_url is None:
+        return urljoin(sse_url, endpoint_data)
+
+    endpoint_url = urljoin(sse_url, messages_url)
+    endpoint_query = urlparse(endpoint_data).query
+    if endpoint_query:
+        endpoint_parsed = urlparse(endpoint_url)
+        query = "&".join(filter(None, [endpoint_parsed.query, endpoint_query]))
+        endpoint_url = endpoint_parsed._replace(query=query).geturl()
+
+    return endpoint_url
+
+
 @asynccontextmanager
 async def sse_client(
     url: str,
@@ -36,6 +50,7 @@ async def sse_client(
     httpx_client_factory: McpHttpClientFactory = create_mcp_http_client,
     auth: httpx.Auth | None = None,
     on_session_created: Callable[[str], None] | None = None,
+    messages_url: str | None = None,
 ):
     """Client transport for SSE.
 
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. Per our CONTRIBUTING.md we ask that PRs start from a linked, triaged issue with maintainer buy-in, which this one doesn't have yet. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #2536, opened 2026-05-02
fix: resolve lost-wakeup races in InMemoryTaskStore.wait_for_update

### LINKED ISSUE #2535: InMemoryTaskStore.wait_for_update has lost-wakeup races (concurrent waiters, notify-before-wait)
`wait_for_update` overwrites `_update_events[task_id]` with a fresh event on every call. This causes two races:

## Race 1: Concurrent waiters

If two callers poll the same task simultaneously, the second overwrites the first's event. `notify_update` only sets the latest event, so the first waiter hangs forever.

## Race 2: Notify before wait

If `update_task` completes (calling `notify_update`) before `wait_for_update` is called, the signal is lost because no event exists yet. The waiter blocks until the next update, which never comes for a terminal task.

Both are reachable via `task_result_handler.py:126`, which calls `_wait_for_task_update` in a polling loop.

## Reproducer

```python
import anyio
from mcp.shared.experimental.tasks.in_memory_task_store import InMemoryTaskStore
from mcp.types import TaskMetadata

async def main():
    store = InMemoryTaskStore()
    task = await store.create_task(TaskMetadata())

    # Race 1: concurrent waiters
    woke = {"a": False, "b": False}
    async def waiter(name):
        await store.wait_for_update(task.task_id)
        woke[name] = True
    async def updater():
        await anyio.sleep(0.05)
        await store.update_task(task.task_id, status="completed")
    try:
        with anyio.fail_after(2):
                tg.start_soon(waiter, "a")
                await anyio.sleep(0.01)
                tg.start_soon(waiter, "b")
                tg.start_soon(updater)
    except TimeoutError:
        pass
    print(f"a: {'woke' if woke['a'] else 'HUNG'}, b: {'woke' if woke['b'] else 'HUNG'}")

    # Race 2: notify before wait
    store2 = InMemoryTaskStore()
    task2 = await store2.create_task(TaskMetadata())
    await store2.update_task(task2.task_id, status="completed")
    try:
        with anyio.fail_after(1):
            await store2.wait_for_update(task2.task_id)
            print("wait returned")
    except TimeoutError:
        print("HUNG: signal lost")

anyio.run(main)
```

Output:
```
a: HUNG, b: woke
HUNG: signal lost
```

blackwell-systems:
## Summary

Two races in `InMemoryTaskStore.wait_for_update` that can cause indefinite hangs when polling task status.

## Race 1: Concurrent waiters

`wait_for_update` overwrites `_update_events[task_id]` with a fresh event on every call. If two clients poll the same task, the second overwrites the first's event. `notify_update` sets only the latest event, so the first waiter hangs forever.

**Fix:** Use a list of events per task_id. Each waiter appends its own event. `notify_update` sets and removes all events for the task.

## Race 2: Notify before wait

If `update_task` calls `notify_update` before any `wait_for_update` is active, the signal is lost (no event exists to set). The next `wait_for_update` creates a fresh event and waits for an update that already happened.

**Fix:** Track pending updates in a `set`. `notify_update` adds to the set when no waiters exist. `wait_for_update` checks and consumes the pending flag before creating an event.

## Reproducer

Both races are reproducible with a 30-line script (included in issue #2535). Output before fix:

```
Race 1: concurrent waiters
  waiter a: HUNG
  waiter b: woke
  FAIL: lost wakeup

Race 2: notify before wait
  FAIL: signal lost, waiter hung forever
```

After fix:

```
Race 1: concurrent waiters
  waiter a: woke
  waiter b: woke
  PASS

Race 2: notify before wait
  PASS
```

## Test plan

- [x] Reproducer script confirms both races are fixed
- [x] All 217 existing task store tests pass (`pytest tests/experimental/tasks/ -v`)
- [x] 0 regressions

Fixes #2535

### DIFF
--- a/src/mcp/shared/experimental/tasks/in_memory_task_store.py
+++ b/src/mcp/shared/experimental/tasks/in_memory_task_store.py
@@ -46,7 +46,8 @@ class InMemoryTaskStore(TaskStore):
     def __init__(self, page_size: int = 10) -> None:
         self._tasks: dict[str, StoredTask] = {}
         self._page_size = page_size
-        self._update_events: dict[str, anyio.Event] = {}
+        self._update_events: dict[str, list[anyio.Event]] = {}
+        self._pending_updates: set[str] = set()
 
     def _calculate_expiry(self, ttl_ms: int | None) -> datetime | None:
         """Calculate expiry time from TTL in milliseconds."""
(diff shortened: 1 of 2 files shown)


Kludex:
We are dropping the experimental.tasks from the sdk. Thanks for the PR.

PR #2544, opened 2026-05-05
fix: exclude None optional fields from task result serialization

### LINKED ISSUE #2539: Bug: task_result_handler.py serializes None optional fields as JSON null, breaking Node SDK Zod validation
### Initial Checks

### Description

 Bug: task_result_handler.py serializes None optional fields as JSON null,
  breaking Node SDK Zod validation

  What the spec says

  The MCP spec defines TextContent with optional fields:

  interface TextContent {
    type: "text";
    text: string;
    annotations?: Annotations;  // optional — may be absent
    _meta?: Record<string, unknown>;  // optional — may be absent
  }

  "Optional" in the spec means the field may be absent from the JSON. Both SDKs
  agree on this:
  - Python SDK's TextContentSchema defines annotations and _meta with
  Optional[...] = None
  - Node SDK's TextContentSchema defines them with .optional() — accepts
  undefined (absent) but not null

  What the Python SDK does

  The Python SDK has two serialization paths for CallToolResult:

  1. Normal responses (_send_response in session.py): uses
  model_dump(by_alias=True, mode="json", exclude_none=True). This correctly omits
   None fields from the JSON. Works fine.
  2. Task result delivery (handle in task_result_handler.py, line 131): uses
  result.model_dump(by_alias=True) without exclude_none=True. This serializes
  None fields as explicit JSON null:

  # task_result_handler.py line 131
  result_data = result.model_dump(by_alias=True)  # ← missing exclude_none=True

  Produces:
  {
    "content": [{
      "type": "text",
      "text": "counted to 20",
      "annotations": null,
      "_meta": null
    }],
    "isError": false
  }

  What breaks

  The Node SDK's requestStream polls tasks/result when a task completes, then
  parses the response with CallToolResultSchema. The Zod discriminated union for
  content blocks uses:

  annotations: AnnotationsSchema.optional()  // accepts undefined, rejects null
  _meta: z.record(z.string(), z.unknown()).optional()  // accepts undefined,
  rejects null

  null ≠ undefined in Zod. The parse fails:
  "expected object, received null" at path ["annotations"]
  "expected record, received null" at path ["_meta"]

  This kills the task result stream. The runtime falls back to polling tasks/get
  + tasks/result, but by then the result may already be consumed.

  Fix

  One line — add exclude_none=True to the model_dump call in
  task_result_handler.py:

  # Before (line 131):
  result_data = result.model_dump(by_alias=True)

  # After:
  result_data = result.model_dump(by_alias=True, exclude_none=True)

  This matches the pattern used everywhere else in the SDK (_send_response,
  send_notification).

  Reproduction

  1. Python MCP server with a task-aware tool that returns
  CallToolResult(content=[TextContent(type="text", text="hello")])
  2. Node MCP client connects and calls the tool with task: {}
  3. Task completes → client calls tasks/result → Zod parse fails on annotations:
   null

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
- mcp (Python): 1.27.0
  - @modelcontextprotocol/sdk (Node): 1.29.0
```

abhinavmathur-atlan:
## Summary

Fixes #2539

`task_result_handler.py` called `model_dump(by_alias=True)` without `exclude_none=True`, causing optional fields like `TextContent.annotations` to be emitted as JSON `null`. The Node SDK Zod schema marks these fields as optional (absent), not nullable, so receiving `null` fails validation.

**Root cause:** The `result_data` dict built at line 119 carried `null` values into `GetTaskPayloadResult.model_validate()`. Since `GetTaskPayloadResult` uses `extra="allow"`, the nested content is stored as raw dict values — the transport-level `exclude_none=True` does not recursively clean them.

**Fix:** Add `exclude_none=True, mode="json"` to both `model_dump()` calls in the terminal-state path, consistent with every other serialization site in `shared/session.py` and `server/session.py`.

## Changes

- `src/mcp/server/experimental/task_result_handler.py`: add `exclude_none=True, mode="json"` to `result.model_dump()`, and `exclude_none=True` to `related_task.model_dump()`
- `tests/experimental/tasks/server/test_task_result_handler.py`: regression test asserting `annotations` and `_meta` are absent (not null) in the serialized wire output

### DIFF
--- a/src/mcp/server/experimental/task_result_handler.py
+++ b/src/mcp/server/experimental/task_result_handler.py
@@ -114,9 +114,11 @@ class TaskResultHandler:
                 # The stored result contains the actual payload data
                 # Per spec: tasks/result MUST include _meta with related-task metadata
                 related_task = RelatedTaskMetadata(task_id=task_id)
-                related_task_meta: dict[str, Any] = {RELATED_TASK_METADATA_KEY: related_task.model_dump(by_alias=True)}
+                related_task_meta: dict[str, Any] = {
+                    RELATED_TASK_METADATA_KEY: related_task.model_dump(by_alias=True, exclude_none=True)
+                }
                 if result is not None:
-                    result_data = result.model_dump(by_alias=True)
+                    result_data = result.model_dump(by_alias=True, mode="json", exclude_none=True)
                     existing_meta: dict[str, Any] = result_data.get("_meta") or {}
                     result_data["_meta"] = {**existing_meta, **related_task_meta}
                     return GetTaskPayloadResult.model_validate(result_data)
--- a/tests/experimental/tasks/server/test_task_result_handler.py
+++ b/tests/experimental/tasks/server/test_task_result_handler.py
@@ -292,6 +292,33 @@ async def test_deliver_skips_resolver_registration_when_no_original_id(
 
 
 @pytest.mark.anyio
+async def test_handle_omits_none_optional_fields_in_result(
+    store: InMemoryTaskStore, queue: InMemoryTaskMessageQueue, handler: TaskResultHandler
+) -> None:
+    """None optional fields (e.g. TextContent.annotations) must be omitted, not serialized as null.
+
+    The Node SDK Zod schema marks these fields as optional (absent), not nullable,
+    so sending null breaks validation.
+    """
+    task = await store.create_task(TaskMetadata(ttl=60000), task_id="test-task")
+    result = CallToolResult(content=[TextContent(type="text", text="hello")])
+    await store.store_result(task.task_id, result)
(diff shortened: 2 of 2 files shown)


abhinavmathur-atlan:
@maxisbey @Kludex  @felixweinberger  , when you get a chance, this is a small fix (2 lines + test) for #2539. Happy to adjust if anything looks off.

abhinavmathur-atlan:
> @maxisbey @Kludex @felixweinberger , when you get a chance, this is a small fix (2 lines + test) for #2539. Happy to adjust if anything looks off.

ICYMI @maxisbey @Kludex @felixweinberger

Kludex:
We are dropping the experimental.tasks from the SDK, so closing this. Thanks for the PR.

PR #2549, opened 2026-05-07
fix: use blocking stdin read instead of anyio.wrap_file() in stdio_server

celstnblacc:
## Summary

Replaces `anyio.wrap_file()` wrapping of stdin in `stdio_server()` with a blocking `run_sync(raw_stdin.readline)` loop.

## Problem

`anyio.wrap_file()` wraps `sys.stdin.buffer` as an async iterator. When the MCP client closes its end of stdin between connection cycles, the async iterator sees EOF and exits the `async for` loop, killing the read stream and the server.

This manifests as `-32000: Connection closed` errors in MCP clients like Claude Code, OpenCode, and Codex CLI, requiring a client restart every time the connection resets.

## Fix

The default stdin path now uses `anyio.to_thread.run_sync(raw_stdin.readline)` in a `while True` loop. Blocking `readline()` waits until the OS delivers more data, surviving transient client disconnects. An empty read (real process EOF) still cleanly exits the loop.

The existing custom-stdin path (used by tests and when callers pass an `anyio.AsyncFile`) is preserved unchanged.

## Changes

- **src/mcp/server/stdio.py**: Default stdin path → blocking `run_sync(readline)` loop (37 insertions, 11 deletions)
- **tests/server/test_stdio.py**: Added `test_stdio_server_survives_stdin_eof` regression test (46 insertions)

## Testing

All existing and new tests pass:

```
uv run --frozen pytest tests/server/test_stdio.py -v
tests/server/test_stdio.py::test_stdio_server PASSED
tests/server/test_stdio.py::test_stdio_server_invalid_utf8 PASSED
tests/server/test_stdio.py::test_stdio_server_survives_stdin_eof PASSED
```

Ruff lint: clean.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -23,6 +23,7 @@ from io import TextIOWrapper
 
 import anyio
 import anyio.lowlevel
+from anyio.to_thread import run_sync
 
 from mcp import types
 from mcp.shared._context_streams import create_context_streams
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing — for the stdio transport, EOF on stdin is the protocol's shutdown signal, so blocking past it would prevent clean shutdown. CI is also failing. If you've hit a real case where stdin closes between connection cycles, that's likely a client-side bug worth reporting separately. Thanks for the contribution.

Issue #1873, opened 2026-01-15
Bug: String parameters starting with digits coerced to numbers, causing UUID data loss


TommyVee:
# FastMCP GitHub Issue: Parameter Type Coercion Ignores `str` Type Annotations

## Title
**Parameter Type Coercion Ignores `str` Type Annotations, Causing Data Loss for UUIDs**

## Description

FastMCP is coercing string parameters that start with digits to `int` or `float`, **ignoring explicit `str` type annotations**. This causes permanent data loss for UUIDs and other string identifiers that begin with numeric characters.

### Impact

- **Data Loss**: Original UUID values cannot be recovered once coerced
- **Functionality Broken**: Tools cannot process UUIDs starting with digits
- **Silent Corruption**: Coercion happens before function code runs, making it hard to detect

### Minimal Example

```python
from mcp.server.fastmcp import FastMCP, Context

mcp = FastMCP("TestServer")

@mcp.tool()
async def update_task(
    ctx: Context,
    task_id: str | None = None  # Explicitly typed as str
) -> str:
    """Update a task by UUID."""
    print(f"Received task_id: {task_id!r}, type: {type(task_id).__name__}")
    return f"Task ID: {task_id}"

if __name__ == "__main__":
    mcp.run(transport="streamable-http")
```

### Test Cases

**Case 1: UUID Starting with Digits**
- **Input**: `task_id = "58aa9efd-faad-4901-89e8-99e807a1a2d6"`
- **Expected**: Function receives `task_id` as `str` with full UUID
- **Actual**: Function receives `task_id` as `str` with value `"58"` (truncated to first digits)

**Case 2: UUID with Scientific Notation Pattern**
- **Input**: `task_id = "3400e37e-b251-49d9-91b0-f8dd8602ff7e"`
- **Expected**: Function receives `task_id` as `str` with full UUID
- **Actual**: Function receives `task_id` as `str` with value `"3.4e+40"` (scientific notation)

### Actual Behavior

When calling the tool with a UUID starting with digits:

```python
# Tool call
await update_task(task_id="58aa9efd-faad-4901-89e8-99e807a1a2d6")

# Function receives:
# task_id = "58"  (type: str, but value is corrupted)
# Original UUID is permanently lost
```

### Evidence from Logs

```
[UUID DEBUG] manage_task CALLED: action=update, task_id=3.4e+40, task_id_type=str
[...]

PR #2554, opened 2026-05-07
fix: handle Optional[str] in pre_parse_json to prevent incorrect type coercion

### LINKED ISSUE #381: Fastmcp tool parameter parsing type error
When I define the tool parameter as string, the parameter parsing function of fastmcp parses the numeric string I pass in into a numeric value.

- mcp/server/fastmcp/utilities/func_metadata.py

`
    def pre_parse_json(self, data: dict[str, Any]) -> dict[str, Any]:
        """Pre-parse data from JSON.

        Return a dict with same keys as input but with values parsed from JSON
        if appropriate.

        This is to handle cases like `["a", "b", "c"]` being passed in as JSON inside
        a string rather than an actual list. Claude desktop is prone to this - in fact
        it seems incapable of NOT doing this. For sub-models, it tends to pass
        dicts (JSON objects) as JSON strings, which can be pre-parsed here.
        """
        new_data = data.copy()  # Shallow copy
        for field_name, field_info in self.arg_model.model_fields.items():
            import logging
            logging.info((field_name, field_info))
            logging.info(new_data)
            if field_name not in data.keys():
                continue
            if isinstance(data[field_name], str):
                try:
                    pre_parsed = json.loads(data[field_name])
                except json.JSONDecodeError:
                    continue  # Not JSON - skip
                if isinstance(pre_parsed, str):
                    # This is likely that the raw value is e.g. `"hello"` which we
                    # Should really be parsed as '"hello"' in Python - but if we parse
                    # it as JSON it'll turn into just 'hello'. So we skip it.
                    continue
                new_data[field_name] = pre_parsed
        assert new_data.keys() == data.keys()
        return new_data
`

test output:
`
video-1  | ('cid', FieldInfo(annotation=str, required=True, description='xxx.'))
video-1  | before parse: {'cid': '1.2'}
video-1  | after parse: {'cid': 1.2}
video-1  | Error calling tool: Error executing tool getinfo: 1 validation error for getinfoArguments
video-1  | cid
video-1  |   Input should be a valid string [type=string_type, input_value=1.2, input_type=float]
video-1  |     For further information visit https://errors.pydantic.dev/2.10/v/string_type
`

Here, json.loads will parse the incoming string "1.2" into the floating point number 1.2

So, when I use pydantic to define parameters in tool, a validation error occurs
`
from pydantic import validate_call, Field

@server.tool(description="xxx")
@validate_call
async def getinfo(
        cid: str = Field(..., description="xxx."),
) -> bool | None:
`

Is my understanding correct?

Maanik23:
Fixes #381

## Summary

The `pre_parse_json` method in `func_metadata.py` uses `field_info.annotation is not str` to decide whether to attempt JSON parsing of string values. This check fails for union types like `Optional[str]` (`str | None`) because the annotation object is `str | None`, not `str`.

This causes string values like `"1.2"` or `'{"key": "value"}'` to be parsed via `json.loads` into `float`/`dict`, which then fails Pydantic validation with `Input should be a valid string`.

## Reproducer (from issue)

```python
@server.tool()
async def getinfo(cid: Optional[str] = Field(..., description="xxx.")) -> bool:
    ...

# Client calls with: {"cid": "1.2"}
# Result: ValidationError - Input should be a valid string [input_value=1.2, input_type=float]
```

## Changes

**`src/mcp/server/mcpserver/utilities/func_metadata.py`**
- Added `_annotation_accepts_str()` helper that correctly identifies str-only unions (`Optional[str]`, `str | None`) as string-accepting annotations
- Uses `is_union_origin` (already imported) and `get_origin`/`get_args` (already imported) for introspection
- Preserves existing behavior for unions with other complex types (e.g. `str | list[str]` still gets JSON pre-parsing)

**`tests/server/mcpserver/test_func_metadata.py`**
- Added `test_optional_str_preserves_json_string`: numeric strings, JSON objects, and JSON arrays all stay as `str` when annotation is `Optional[str]`
- Added `test_pipe_none_str_preserves_json_string`: same coverage for `str | None` syntax
- Added `test_optional_str_runtime_validation`: end-to-end `call_fn_with_arg_validation` with `Optional[str]` param

## Backward compatibility

- No change for `str` annotations (already handled by PR #1113)
- No change for `str | list[str]` unions (still parses JSON arrays into lists)
- Fixes `Optional[str]` and `str | None` to correctly preserve string values
- All 36 existing tests in `test_func_metadata.py` continue to pass

### DIFF
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -29,6 +29,22 @@ from mcp.types import CallToolResult, ContentBlock, TextContent
 logger = get_logger(__name__)
 
 
+def _annotation_accepts_str(annotation: Any) -> bool:
+    """Check if an annotation only accepts str (or None), not other complex types.
+
+    Returns True for `str` and `Optional[str]` (str | None), where JSON parsing
+    a string value would be incorrect. Returns False for unions like `str | list[str]`
+    where the other types benefit from JSON pre-parsing.
+    """
+    if annotation is str:
+        return True
+    if is_union_origin(get_origin(annotation)):
+        args = get_args(annotation)
+        non_none_args = [a for a in args if a is not type(None)]
+        return non_none_args == [str]
+    return False
+
+
 class StrictJsonSchema(GenerateJsonSchema):
     """A JSON schema generator that raises exceptions instead of emitting warnings.
 
@@ -148,7 +164,7 @@ class FuncMetadata(BaseModel):
                 continue
 
             field_info = key_to_field_info[data_key]
-            if isinstance(data_value, str) and field_info.annotation is not str:
+            if isinstance(data_value, str) and not _annotation_accepts_str(field_info.annotation):
                 try:
                     pre_parsed = json.loads(data_value)
                 except json.JSONDecodeError:
--- a/tests/server/mcpserver/test_func_metadata.py
+++ b/tests/server/mcpserver/test_func_metadata.py
@@ -551,6 +551,74 @@ async def test_str_annotation_runtime_validation():
     assert result == f"Handled payload of length {len(json_array_payload)}"
 
 
+def test_optional_str_preserves_json_string():
+    """Regression test for issue #381: Ensure that when a parameter is annotated
+    as Optional[str] or str | None, valid JSON strings are NOT parsed into Python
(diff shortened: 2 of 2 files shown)


maxisbey:
Closing as a duplicate of #2032, which was opened first for #1873 by the contributor who claimed it on the issue. Thanks for the contribution.

PR #2555, opened 2026-05-07
fix(shared): make UrlElicitationRequiredError pickle-safe (refs #2431)

MukundaKatta:
References #2431.

## Bug

The default `Exception` unpickle path reconstructs via `cls(*self.args)`. `MCPError.__init__` sets `args` to `(code, message, data)` — three positionals. That doesn't match `UrlElicitationRequiredError`'s signature `(elicitations, message=None)`, so unpickling raises:

```text
TypeError: UrlElicitationRequiredError.__init__() takes from 2 to 3 positional arguments but 4 were given
```

Reproduces on stdlib `pickle` and `cloudpickle` alike. Repro on current main:

```python
import pickle
from mcp.shared.exceptions import UrlElicitationRequiredError
from mcp.types import ElicitRequestURLParams

err = UrlElicitationRequiredError([
    ElicitRequestURLParams(mode="url", message="auth", url="https://e.com", elicitation_id="x"),
])
pickle.loads(pickle.dumps(err))  # ↑ TypeError
```

## Fix

Add `__reduce__` that reconstructs from the typed `(elicitations, message)` pair, so the round-trip preserves the high-level field rather than the wire-format `data` dict.

## Tests

- New `test_url_elicitation_required_error_pickle_round_trip` — exercises a 2-elicitation error, asserts class identity, message, elicitations list, and reconstructed `error.code`.
- New `test_mcp_error_pickle_round_trip` — pins the parent class's already-working round trip so a future `__init__` refactor can't silently break it.
- 11/11 passing in `tests/shared/test_exceptions.py`.

## Note on issue scope

#2431 describes a related but slightly different constructor shape (`McpError(error: ErrorData)`) that no longer exists on main — the parent class was refactored to `(code, message, data=None)` since the report. That refactor accidentally fixed the parent's pickle round-trip but left the URL-elicitation subclass broken on a different code path. This PR closes that remaining gap.

### DIFF
--- a/src/mcp/shared/exceptions.py
+++ b/src/mcp/shared/exceptions.py
@@ -104,3 +104,19 @@ class UrlElicitationRequiredError(MCPError):
         raw_elicitations = cast(list[dict[str, Any]], data.get("elicitations", []))
         elicitations = [ElicitRequestURLParams.model_validate(e) for e in raw_elicitations]
         return cls(elicitations, error.message)
+
+    def __reduce__(self) -> tuple[Any, ...]:
+        """Make the class pickle-safe.
+
+        Without this, the default ``Exception.__reduce_ex__`` reconstructs by
+        calling ``cls(*self.args)`` — and ``self.args`` was set by
+        :class:`MCPError`'s ``__init__`` to ``(code, message, data)``, which
+        does not match this subclass's constructor signature
+        ``(elicitations, message=None)``. The mismatch raises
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. Tracking this in #3262 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

PR #2566, opened 2026-05-10
Support is_error flag for tool results

### LINKED ISSUE #348: No way to set isError=True for arbitrary tool result content

zuhabul:
Fixes #348 — allow tools to return (unstructured_content, structured_content, is_error) so non-text results (e.g., images) can be marked as errors. Also document the behavior in docs/migration.md.


maxisbey:
This targets #348 rather than #2429. The discussion on #348 favors the raise-an-exception approach (#1824) over a new tuple return shape. CI is also failing across the matrix. Closing — happy to revisit if the direction on #348 changes.

PR #2568, opened 2026-05-10
Avoid closing real stdio on stdio transport

zuhabul:
Fixes #1933 — duplicate underlying stdio file descriptors before wrapping so closing wrappers doesn't close the real process stdio. See issue: https://github.com/modelcontextprotocol/python-sdk/issues/1933


maxisbey:
Closing as a duplicate of #2040, which uses the `closefd=False` approach discussed in review there. This PR also bundles unrelated changes for #348 and #1564 that should be filed separately. Thanks for the contribution.

PR #2569, opened 2026-05-10
Fix MCPServer.call_tool return type

### LINKED ISSUE #1251: Incorrect typing hint for FastMCP.call_tool

FU-max-boop:
## Motivation and Context
Closes #1251.

`MCPServer.call_tool` always requests converted tool results from the tool manager, so its return annotation should match the actual converted shapes: content-only sequences, structured `(content, structured_content)` tuples, or direct `CallToolResult` values. The previous `dict[str, Any]` return type also left an unreachable handler branch in `_handle_call_tool`.

## How Has This Been Tested?
- `python3 -m py_compile src/mcp/server/mcpserver/server.py tests/server/mcpserver/test_server.py`
- `git diff --check`
- `uv run --frozen --no-default-groups --with pytest --with inline-snapshot pytest -c /dev/null tests/server/mcpserver/test_server.py -k 'call_tool_direct_returns_structured_tuple or tool_structured_output_primitive'`

## Breaking Changes
None. This is a type annotation and cleanup change with direct coverage for the runtime return shape.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
I kept the scope limited to the incorrect direct `call_tool` type and the unreachable code path it created. I attempted the full project-style `uv` dev-tool run locally, but dependency downloads for dev tooling stalled; the targeted regression test above passed.

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -4,11 +4,10 @@ from __future__ import annotations
 
 import base64
 import inspect
-import json
 import re
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterable, Sequence
 from contextlib import AbstractAsyncContextManager, asynccontextmanager
-from typing import Any, Generic, Literal, TypeAlias, TypeVar, overload
+from typing import Any, Generic, Literal, TypeAlias, TypeVar, cast, overload
 
 import anyio
 import pydantic_core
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2175, which removes the tuple return shape rather than annotating it — the new test here would pin the API shape we're trying to retire. Thanks for the contribution.

PR #2585, opened 2026-05-13
fix(auth): normalize URL paths before hierarchical resource check

Choppaaahh:
## Summary

`check_resource_allowed` in `src/mcp/shared/auth_utils.py` compares URL paths via `startswith` after only trailing-slash normalization. Paths that contain unresolved dot-segments or their percent-encoded equivalents can satisfy the prefix check while resolving to a path outside the configured scope on any downstream server that normalizes the URL — which is the standard behavior of every common web framework router.

Examples that pre-fix return `True`:

```python
check_resource_allowed("https://example.com/api/../admin", "https://example.com/api")     # True (resolves to /admin)
check_resource_allowed("https://example.com/api/%2e%2e/admin", "https://example.com/api") # True (resolves to /admin)
check_resource_allowed("https://example.com/api//victim",  "https://example.com/api/victim2")  # True (fake boundary)
```

This is a confused-deputy class issue: the auth check passes against the literal-string prefix while the downstream router resolves the request to a different scope than the one the token was authorized for. While exploitability depends on whether the downstream resource server is configured with a smaller scope than the literal prefix the token presents, the defense-in-depth principle is to normalize before comparison.

## The fix

`_normalize_url_path` decodes percent-encoding (`unquote`) and resolves dot-segments (`posixpath.normpath`) before the `startswith` comparison runs. Trailing-slash semantics are preserved so existing legitimate matches still work.

```python
def _normalize_url_path(path: str) -> str:
    decoded = unquote(path)
    normalized = posixpath.normpath(decoded) if decoded else decoded
    if path.endswith("/") and normalized != "/" and not normalized.endswith("/"):
        normalized += "/"
    return normalized
```

## Test plan

- [x] All 7 existing `check_resource_allowed` tests still pass (identical / different-scheme / different-domain / different-port / hierarchical / trailing-slash / case-insensitive-origin / empty-paths)
- [x] 4 new test functions added covering: dot-segment escape (`/api/../admin`), percent-encoded escape (`/api/%2e%2e/admin` + uppercase variant + nested), double-slash collapse (`/api//victim` not matching `/api/victim2`), self-reference dot-segment (`/api/./v1` still matches `/api`)
- [x] 33/33 cases pass with the patch applied (25 existing + 8 new traversal cases)

## Notes

- Found while auditing the OAuth Resource Indicators (RFC 8707) implementation path. The rest of the OAuth surface (validate_redirect_uri strict-listing, PKCE-S256 verification, client_secret/state constant-time compare via `hmac.compare_digest` / `secrets.compare_digest`) is well-defended — this was the one normalization gap in an otherwise solid implementation.
- I did not find a concrete exploit chain in the SDK's own call-sites (the direction of the `startswith` check in `IntrospectionTokenVerifier` and `oauth2.py` happens to not align with the typical attack shape on this primitive), but the function is a public utility that downstream MCP server implementations may use in either direction. Normalization removes the footgun regardless.

---

*AI-assisted.*

### DIFF
--- a/src/mcp/shared/auth_utils.py
+++ b/src/mcp/shared/auth_utils.py
@@ -1,9 +1,36 @@
 """Utilities for OAuth 2.0 Resource Indicators (RFC 8707) and PKCE (RFC 7636)."""
 
+import posixpath
 import time
-from urllib.parse import urlparse, urlsplit, urlunsplit
+from urllib.parse import unquote, urlparse, urlsplit, urlunsplit
 
 from pydantic import AnyUrl, HttpUrl
+
+
+def _normalize_url_path(path: str) -> str:
+    """Decode percent-encoding and resolve dot-segments in a URL path.
+
+    Used by check_resource_allowed to ensure hierarchical matching operates on
+    normalized paths. Without normalization, a path like "/api/../admin" would
+    satisfy startswith("/api") even though the resolved path is "/admin",
+    enabling a confused-deputy attack at downstream resource servers that DO
+    normalize paths.
+
+    - unquote() decodes percent-encoded segments so "%2e%2e" -> ".." cannot
+      bypass the normalization step.
+    - posixpath.normpath() resolves "." and ".." segments to canonical form.
+
+    Returns a path that ends with "/" if the original did, so trailing-slash
+    semantics are preserved.
+    """
+    decoded = unquote(path)
+    normalized = posixpath.normpath(decoded) if decoded else decoded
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. Per our CONTRIBUTING.md we ask that PRs start from a linked, triaged issue with maintainer buy-in, which this one doesn't have yet. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2587, opened 2026-05-13
feat(utilities): add dereference_json_schema helper

### LINKED ISSUE #2586: Add dereference helper for tool inputSchema with nested Pydantic models
## Problem

When a tool's `inputSchema` is built from a Pydantic model containing nested `BaseModel` types, `model_json_schema()` emits a `\$defs` block with named definitions and `\$ref` pointers at each use site. The output is valid JSON Schema 2020-12, but some MCP clients (notably Claude Desktop during the `tools/list` discovery flow) don't resolve internal `\$ref` references — tools with nested types in their inputs become invisible or unusable in those clients.

The current workaround that downstream MCP servers reach for is "flatten every input model by hand" — this leaks a Pydantic limitation into the public API surface of every MCP author who hits the issue. In our case (`research-mcp`, ~14 tools), we hand-flattened all input models.

## Proposed shape

A small, additive helper that post-processes the output of `model_json_schema()`:

\`\`\`python
from mcp.server.mcpserver.utilities.json_schema import dereference_json_schema

flat = dereference_json_schema(MyToolInput.model_json_schema())
# \`flat\` has no \$defs; every internal \$ref has been inlined.
\`\`\`

- Pure function. Doesn't mutate the input.
- Only handles internal \`#/\$defs/\` refs. External refs (URLs, pointers into other documents) preserved verbatim.
- Self-referential and mutually recursive definitions are handled safely: \`\$ref\` is preserved at the cycle boundary and the relevant \`\$defs\` entries are retained so the output schema stays valid.
- Sibling keys to \`\$ref\` (allowed by JSON Schema 2020-12; emitted by Pydantic for some types) are merged into the resolved object, with sibling values overriding the resolved definition's values (matches the JSON Schema 2020-12 semantics for sibling annotations).

Default behavior of the SDK is **unchanged** — this is purely additive. Callers opt in. Existing tools/users who prefer the compact \`\$ref\` form are unaffected.

## Why a helper rather than a flag on \`model_json_schema\`

A flag on \`model_json_schema\` would have to live in Pydantic. A helper in the SDK is a one-line change at the call site for users who need it, and doesn't require coordinating across projects.

## Implementation

I have a working implementation with 21 tests covering: flat schemas, single-level inlining, transitive resolution, arrays, anyOf, sibling key merging, external refs, unknown internal refs, direct self-reference, mutual recursion, and idempotency. All existing \`tests/server/mcpserver/test_func_metadata.py\` and \`test_tool_manager.py\` tests still pass (no regressions).

Happy to open a PR once this is \`ready for work\`.

Burton-David:
Closes #2586

## Summary

Add `dereference_json_schema(schema: dict[str, Any]) -> dict[str, Any]` in `mcp.server.mcpserver.utilities.json_schema`. Post-processes the output of Pydantic's `model_json_schema()` by inlining internal `#/\$defs/` refs while preserving external refs and self-referential definitions.

Pure function, no mutation of input. Opt-in only — no existing SDK behavior changes. Callers who want flat schemas for clients that don't resolve `\$ref` (e.g. Claude Desktop's tools/list discovery) apply this helper once after building the schema.

## Test plan

- [x] 21 new tests in `tests/server/mcpserver/test_json_schema.py`
- [x] Existing `test_func_metadata.py` (66 tests) — no regressions
- [x] Existing `test_tool_manager.py` (17 tests) — no regressions
- [x] `ruff check` clean
- [x] `pyright` strict clean

Test coverage:
- Empty schemas, no-`\$defs` schemas (no-op + strip)
- Single-level inlining
- Multiple refs to same def
- Arrays of nested models, `anyOf` / `oneOf` / `allOf`
- Transitive resolution (def A contains ref to def B)
- Sibling keys to `\$ref` (Pydantic emits these for some types)
- External `\$ref` (URL) preserved
- Unknown internal `\$ref` preserved (don't silence producer bugs)
- Non-`\$defs` internal refs (`#/properties/x`) preserved
- Direct self-reference: `\$ref` kept at boundary, `\$defs` entry retained
- Mutual recursion (A → B → A): `\$ref` at innermost boundary
- Idempotency (apply twice = apply once)
- Real Pydantic `model_json_schema()` outputs (nested, list of nested, self-referencing).

## Notes on design choices

- Lives in `mcp.server.mcpserver.utilities` alongside `func_metadata` since it's the natural sibling to the existing `model_json_schema` call sites.
- Cycle handling preserves `\$ref` at the boundary rather than raising — the alternative (raising on cycles) would force users with legitimate recursive schemas to fall back to the current `\$ref` form anyway.
- Sibling-key merge order: resolved body first, siblings second (siblings override). Matches JSON Schema 2020-12 sibling-annotation semantics where `\$ref` siblings provide overrides/refinements.
- No call-site changes in `func_metadata.py` or `tools/base.py` — the helper is purely additive. If maintainers want it folded in as a flag on existing APIs, happy to update.

Happy to iterate on the approach if the maintainers prefer a different shape.

### DIFF
--- a/src/mcp/server/mcpserver/utilities/json_schema.py
+++ b/src/mcp/server/mcpserver/utilities/json_schema.py
@@ -1,0 +1,127 @@
+"""JSON Schema post-processing utilities.
+
+Pydantic's :meth:`pydantic.BaseModel.model_json_schema` emits a ``$defs``
+block with named definitions for nested models, and ``$ref`` pointers
+at each use site. The output is fully valid JSON Schema, but some MCP
+clients don't resolve internal references when discovering tools via
+``tools/list``. For those clients, fields whose types are nested models
+become unusable.
+
+This module provides :func:`dereference_json_schema`, an opt-in helper
+that inlines ``$defs`` references into the schema body. It's safe to
+apply at any point in the schema lifecycle (the function does not
+mutate its input) and conservative about edge cases — external
+references and self-referential definitions are preserved unchanged.
+"""
+
+from __future__ import annotations
+
+import copy
+from typing import Any, cast
+
+_INTERNAL_DEFS_PREFIX = "#/$defs/"
+
+
+def dereference_json_schema(schema: dict[str, Any]) -> dict[str, Any]:
+    """Return ``schema`` with internal ``$defs`` references inlined.
+
+    Replaces each ``{"$ref": "#/$defs/<name>"}`` node with the body of
+    the corresponding ``$defs`` entry, recursing into nested objects
+    and arrays. Self-referential definitions (a model whose schema
+    contains a ``$ref`` back to itself, directly or transitively) are
+    preserved as ``$ref`` at the cycle boundary; the ``$defs`` entries
+    they depend on are retained so the output schema remains valid.
+
+    Only internal ``#/$defs/`` references are resolved. External
+    references (URLs, pointers into other documents) are preserved
+    verbatim. Sibling keys alongside a ``$ref`` (allowed in JSON Schema
+    2020-12 and emitted by Pydantic for some types) are merged into
+    the resolved object, with sibling values taking precedence.
+
+    Args:
+        schema: A JSON Schema dict. Not mutated.
+
+    Returns:
+        A new schema dict with internal refs inlined. The top-level
+        ``$defs`` block is removed when fully dereferenced, and
+        retained (containing only the cycle definitions) otherwise.
+
+    Example::
+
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing in favor of #2448, which applies the dereferencing by default in the tool pipeline, matching #2384 (the maintainer-filed issue) and the TypeScript/Go SDKs. CI is also failing on this branch. Thanks for the contribution — the helper itself may still be useful as part of that work.

PR #2592, opened 2026-05-14
fix: support PEP 604 structured output unions

he-yufeng:
﻿## Summary
- wrap FastMCP structured output annotations with Pydantic's explicit `(annotation, ...)` field form
- add regression coverage for `dict | list | str` return annotations

Fixes #2591.

## To verify
- `uv run pytest tests\server\mcpserver\test_func_metadata.py -q --basetemp .tmp\pytest -p no:cacheprovider`
- `uv run ruff check src\mcp\server\mcpserver\utilities\func_metadata.py tests\server\mcpserver\test_func_metadata.py`
- `uv run ruff format --check src\mcp\server\mcpserver\utilities\func_metadata.py tests\server\mcpserver\test_func_metadata.py`
- `git diff --check`

PR #2599, opened 2026-05-15
fix(fastmcp): support PEP 604 union return types

### LINKED ISSUE #2591: FastMCP crashes when tool return type uses Python 3.10+ `A | B | C` union syntax
### Initial Checks

### Description

`FastMCP` crashes on startup when a tool function's return type annotation uses Python 3.10+ union syntax (`A | B | C`, i.e. `types.UnionType`). The error originates from `_create_wrapped_model` in `func_metadata.py`, which passes the union annotation directly to Pydantic's `create_model()` as a bare keyword value, causing `PydanticUserError`.

### Example Code

```Python

mcp = FastMCP("test")

@mcp.tool()
async def my_tool(flag: bool) -> dict | list | str:  # <-- triggers the bug
    if flag:
        return {"key": "value"}
    return ["item"]
```

### Python & MCP Python SDK

```Text
- mcp (Python): latest (>=1.0.0,<2.0.0)
- pydantic: >=2.10 (observed on 2.10.x)
- Python: >=3.10
```

MukundaKatta:
## Summary
- define wrapped structured-output fields using Pydantic's explicit required field tuple
- add regression coverage for FastMCP tools returning container/primitive PEP 604 unions like `dict | list | str`

Fixes #2591

## Test plan
- `uv run pytest tests/server/mcpserver/test_func_metadata.py -k "structured_output"`
- `uv run pytest tests/server/mcpserver/test_func_metadata.py`

### DIFF
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -478,7 +478,7 @@ def _create_wrapped_model(func_name: str, annotation: Any) -> type[BaseModel]:
     """
     model_name = f"{func_name}Output"
 
-    return create_model(model_name, result=annotation)
+    return create_model(model_name, result=(annotation, ...))
 
 
 def _create_dict_model(func_name: str, dict_annotation: Any) -> type[BaseModel]:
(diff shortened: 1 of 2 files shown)


maxisbey:
Closing as a duplicate of #2592, which was opened first with the same source change. Thanks for the contribution.

PR #2606, opened 2026-05-15
fix: reject changed duplicate initialize

### LINKED ISSUE #2605: Duplicate `initialize` with changed parameters can overwrite `ServerSession.client_params`
### Initial Checks

### Description

After a normal MCP initialization flow, a Python SDK server accepts another `initialize` request on the same live connection/session and returns a normal `initialize` result.

I agree that the spec does not explicitly say every repeated `initialize` must be rejected. My concern is narrower: after the session has already moved into normal operation, a later same-session `initialize` with different parameters can update the server's stored client initialization parameters.

The MCP lifecycle documentation describes initialization as the first client/server interaction and says the operation phase uses the negotiated protocol version and capabilities. In that model, silently changing the stored client identity/capabilities/protocol version after operation has begun seems surprising unless the SDK intentionally supports same-session renegotiation.

- `src/mcp/server/session.py` stores the client initialization request in `ServerSession._client_params`.
- `ServerSession.client_params` exposes that value.
- In both the stable release and the `main` snapshot I tested, the `InitializeRequest` handler assigns `self._client_params = params`.
- `ServerSession.check_client_capability()` later reads `self._client_params.capabilities`, so a duplicate `initialize` with different `capabilities` can affect capability-gated server logic.

### What I observed

Reproduced on stable `v1.27.1` (`77431ebe7dda9ed0c61451b22d3e7f8d981bc092`):

- `stdio`: a second `initialize` was accepted after `notifications/initialized`; a later `ping` still succeeded.
- stateful Streamable HTTP: a second `initialize` was accepted after `notifications/initialized`; a later `ping` still succeeded.
- stateless Streamable HTTP: a second `initialize` was accepted after `notifications/initialized`; a later `ping` still succeeded.

For stateful Streamable HTTP on `v1.27.1`, I added a local diagnostic tool that reads `ctx.session.client_params` before and after the duplicate `initialize` on the same session:

| Duplicate `initialize` parameters | Observed result |
|---|---|
| Same parameters | No state change |
| Changed `clientInfo` and `capabilities` | `client_params.clientInfo` and `client_params.capabilities` changed |
| Older `protocolVersion` (`2024-11-05`) | `client_params.protocolVersion` and `client_params.clientInfo` changed |

On the frozen `main` snapshot (`161834d4aee2633c42d3976c8f8751b6c4d947d5`), the same duplicate-initialize acceptance still reproduced. My diagnostic snapshot directly observed `client_params.capabilities` changing after a duplicate `initialize` with changed capabilities on both `stdio` and stateful Streamable HTTP.

### Expected behavior

If same-session duplicate initialization is not intended, I would expect the server to reject the later `initialize` after normal operation has begun, for example with a JSON-RPC `-32600 Invalid Request` error.

If this tolerant behavior is intentional, I think it should be either:

- a no-op once the session has already initialized, especially for changed parameters, or
- documented as same-session renegotiation, including whether `ServerSession.client_params`, client capabilities, client identity, and protocol version are allowed to change.

### Example Code

```Python
import json

try:
    # Current main branch naming.
    from mcp.server.mcpserver import Context, MCPServer
except ImportError:
    # Latest stable release naming.
    from mcp.server.fastmcp import Context, FastMCP as MCPServer


mcp = MCPServer("dup-init-repro")


def field(obj, *names):
    if obj is None:
        return None
    for name in names:
        if hasattr(obj, name):
            return getattr(obj, name)
    return None


def dump(value):
    if value is None:
        return None
    if hasattr(value, "model_dump"):
        return value.model_dump(mode="json", by_alias=True)
    return value


@mcp.tool()
async def session_snapshot(ctx: Context) -> str:
    params = ctx.session.client_params
    return json.dumps(
        {
            "protocolVersion": field(params, "protocolVersion", "protocol_version"),
            "clientInfo": dump(field(params, "clientInfo", "client_info")),
            "clientCapabilities": dump(field(params, "capabilities")),
        },
        separators=(",", ":"),
        default=str,
    )


if __name__ == "__main__":

'''
Request sequence:

1. Start the server.
2. Send a normal `initialize`.
3. Send `notifications/initialized`.
4. Call the `session_snapshot` tool and record the returned JSON.
5. Send another `initialize` on the same live session, but change `clientInfo` and/or `capabilities`.
6. Call `session_snapshot` again on the same session.

On `v1.27.1`, the second snapshot reflects the changed `clientInfo` / `capabilities`. On the `main` snapshot I tested, the second snapshot still reflects changed `capabilities`.
'''
```

### Python & MCP Python SDK

```Text
- Python: `3.12.3`
- MCP Python SDK: `mcp==1.27.1` (`v1.27.1`, `77431ebe7dda9ed0c61451b22d3e7f8d981bc092`)
- Also reproduced on `main` snapshot `161834d4aee2633c42d3976c8f8751b6c4d947d5` (`v1.25.0-159-g161834d4`, pinned 2026-05-11)
```

he-yufeng:
﻿## Summary
- reject a second `initialize` request if it tries to change the session's client parameters
- keep the first `client_params` intact so later capability checks cannot be spoofed by a duplicate handshake
- add a regression test covering the changed-params path

## To verify
- `uv run pytest tests/server/test_session.py -q -k "duplicate_initialize or server_session_initialize"`
- `uv run ruff check src/mcp/server/session.py tests/server/test_session.py`
- `uv run pyright src/mcp/server/session.py tests/server/test_session.py`
- `uv run python -m py_compile src/mcp/server/session.py tests/server/test_session.py`
- `git diff --check`

Closes #2605

### DIFF
--- a/src/mcp/server/runner.py
+++ b/src/mcp/server/runner.py
@@ -398,7 +398,12 @@ class ServerRunner(Generic[LifespanT]):
 
     def _handle_initialize(self, params: Mapping[str, Any] | None) -> InitializeResult:
         """Build the `initialize` result; state commits later in `_on_request`."""
-        _, negotiated = self._negotiate_initialize(params)
+        init, negotiated = self._negotiate_initialize(params)
+        if self.connection.client_params is not None and init != self.connection.client_params:
+            raise MCPError(
+                code=INVALID_PARAMS,
+                message="Session already initialized with different parameters",
+            )
         opts = self.init_options if self.init_options is not None else self.server.create_initialization_options()
         return InitializeResult(
             protocol_version=negotiated,
--- a/tests/server/test_runner.py
+++ b/tests/server/test_runner.py
@@ -173,6 +173,25 @@ async def test_runner_handles_initialize_and_populates_connection(server: SrvT):
 
 
 @pytest.mark.anyio
+async def test_runner_rejects_changed_duplicate_initialize(server: SrvT):
+    async with connected_runner(server, initialized=False) as (client, runner):
+        first_params = _initialize_params()
+        await client.send_raw_request("initialize", first_params)
+        assert runner.connection.client_params is not None
+        assert runner.connection.client_params.client_info.name == "test-client"
+
+        second_params = _initialize_params()
+        second_params["clientInfo"] = {"name": "second-client", "version": "1.0"}
+        with pytest.raises(MCPError) as exc:
+            await client.send_raw_request("initialize", second_params)
+
+        assert exc.value.error.code == INVALID_PARAMS
+        assert "already initialized" in exc.value.error.message
+        assert runner.connection.client_params is not None
+        assert runner.connection.client_params.client_info.name == "test-client"
+
+
+@pytest.mark.anyio
 async def test_runner_initialize_opens_gate_but_event_fires_only_after_initialized_notification(server: SrvT):
     """`initialize` commits the gate flag and peer info, but the public
     `connection.initialized` event waits for `notifications/initialized` (the


BossChaos:
## Code Review

**PR:** [fix: reject changed duplicate initialize](https://github.com/modelcontextprotocol/python-sdk/pull/2606) by @he-yufeng

- ✅ Bug fix

he-yufeng:
Rebased this PR onto the current runner/connection architecture instead of keeping the old `ServerSession._received_request` patch. The duplicate-initialize guard now lives in `ServerRunner._handle_initialize()`, so a second initialize with changed params returns `INVALID_PARAMS` and leaves the original `connection.client_params` intact.

Validation:
- `python -m py_compile src\mcp\server\runner.py tests\server\test_runner.py`
- `uv run pytest tests/server/test_runner.py -q -k "initialize" --basetemp .tmp\pytest-2606-20260610 -p no:cacheprovider` (16 passed)

he-yufeng:
Rebased onto current `main` and resolved the conflict in `_handle_initialize`: the duplicate-initialize rejection now layers on top of the new `opts` resolution (the `create_initialization_options()` fallback) instead of the old `assert self.init_options is not None`. `test_runner_rejects_changed_duplicate_initialize` still passes against the rebased base.

he-yufeng:
Bumping this. A second initialize with changed params can still shift session state after the handshake, so rejecting the changed duplicate keeps initialization once-only per the spec. Rebases cleanly on main if there is interest.

he-yufeng:
Still relevant: runner.py:231-239 overwrite the state on a repeated initialize instead of rejecting it. Would be nice to get a maintainer's take on this.

PR #2611, opened 2026-05-15
fix: suppress completed request cancellation leak

### LINKED ISSUE #2610: RequestResponder.__exit__ leaks CancelledError on cancelled request, killing the stdio receive loop
## Summary

When a client sends `notifications/cancelled` for an in-flight request handled over stdio, the server's receive loop task group dies. The process stays alive but stops reading stdin, so every subsequent request hangs and the client eventually reports `MCP error -32000: Connection closed`. The bug is racy — it reproduces in roughly 40–80% of attempts depending on platform timing.

Affects: `mcp` 1.26.0 and 1.27.1 (latest at time of writing). Confirmed with FastMCP 3.1.x and the in-tree stdio server.

## Reproduction

A standalone subprocess test is in [mlorentedev/hive `tests/test_transport_recovery.py`](https://github.com/mlorentedev/hive/blob/master/tests/test_transport_recovery.py). The relevant flow:

1. `initialize` → ack
2. `notifications/initialized`
3. `tools/call id=2`
4. `notifications/cancelled` for `requestId=2` (within a few ms of step 3)
5. Receive: `{"id": 2, "error": {"code": 0, "message": "Request cancelled"}}` ← OK
6. `tools/call id=3`
7. **No response.** Server is alive but `proc.stdout.readline()` hangs.

## Root cause

`mcp.shared.session.RequestResponder.__exit__`:

```python
def __exit__(self, exc_type, exc_val, exc_tb):
    try:
        if self._completed:
            self._on_complete(self)
    finally:
        self._entered = False
        ...
        self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)  # ← (A)
```

When `notifications/cancelled` arrives, `RequestResponder.cancel()` calls `self._cancel_scope.cancel()` and sends the error response. The handler task catches the `CancelledError` in `mcp/server/lowlevel/server.py` (around line 766) and `return`s. The `with responder:` block then exits with `exc_type=None` while the cancel scope is still in cancelled state — at line (A), anyio's `CancelScope.__exit__` re-raises `CancelledError`.

That exception bubbles up to:

```python
        tg.start_soon(self._handle_message, ...)
```

…in `Server._run`. anyio task groups cancel all sibling tasks and propagate the cancellation. The receive loop is one of those sibling tasks, so it dies.

## Suggested fix

Swallow the spurious cancellation when the responder has already sent its response:

```python
def __exit__(self, exc_type, exc_val, exc_tb):
    try:
        if self._completed:
            self._on_complete(self)
    finally:
        self._entered = False
        if not self._cancel_scope:
        try:
        except BaseException as exc:
            if self._completed and isinstance(exc, anyio.get_cancelled_exc_class()):
                # cancel() already sent the error response — the scope's
                # re-raised cancellation is spurious.
                return
            raise
```

Mirrors what we ship in [hive `src/hive/_compat.py`](https://github.com/mlorentedev/hive/blob/master/src/hive/_compat.py).

## Workaround used downstream

We monkey-patch `RequestResponder.__exit__` at import time. The patch is self-gated (only fires when `_completed=True` AND the leaking exception is `anyio.get_cancelled_exc_class()`), so it stays inert once a fix lands upstream.

## Environment

- Reproduced on Windows 11 (`mcp` 1.26.0, 1.27.1) and via Claude Code as the host.
- Tracked downstream in [mlorentedev/hive#75](https://github.com/mlorentedev/hive/issues/75).
- Regression test passes 5/5 with the patch applied on Python 3.12; fails 2/5 — 4/5 without it.
- Python 3.13 appears to have additional uncancel semantics that the patch does not yet fully cover — verification in progress.

MukundaKatta:
## Summary
- return the underlying CancelScope.__exit__ suppression result from RequestResponder.__exit__
- swallow a late cancellation from an already-cancelled/completed responder so cancelled requests do not tear down sibling tasks
- add regressions for completed responder cancellations

Fixes #2610.

## Tests
- uv run pytest tests/shared/test_session.py -q -k "request_responder or request_cancellation"
- uv run pytest tests/server/test_cancel_handling.py tests/shared/test_session.py -q
- uv run ruff check src/mcp/shared/session.py tests/shared/test_session.py
- uv run ruff format --check src/mcp/shared/session.py tests/shared/test_session.py
- uv run pyright src/mcp/shared/session.py tests/shared/test_session.py

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -108,8 +108,9 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
         exc_type: type[BaseException] | None,
         exc_val: BaseException | None,
         exc_tb: TracebackType | None,
-    ) -> None:
+    ) -> bool | None:
         """Exit the context manager, performing cleanup and notifying completion."""
+        suppress = False
         try:
             if self._completed:
                 self._on_complete(self)
@@ -117,7 +118,19 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
             self._entered = False
             if not self._cancel_scope:  # pragma: no cover
                 raise RuntimeError("No active cancel scope")
-            self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            try:
+                suppress = self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            except BaseException as exc:
+                if (
+                    self._completed
+                    and self._cancel_scope.cancel_called
+                    and isinstance(exc, anyio.get_cancelled_exc_class())
+                ):
+                    suppress = True
+                else:
+                    raise
+
+        return suppress
 
     async def respond(self, response: SendResultT | ErrorData) -> None:
         """Send a response for this request.
--- a/tests/shared/test_session.py
+++ b/tests/shared/test_session.py
@@ -1,3 +1,5 @@
+from typing import Any, cast
+
 import anyio
 import pytest
 
@@ -21,6 +23,11 @@ from mcp.types import (
     ServerNotification,
     ServerRequest,
 )
+
+
+class _DummySession:
+    async def _send_response(self, request_id: int | str, response: ClientResult | ErrorData) -> None:
+        pass
 
 
 @pytest.mark.anyio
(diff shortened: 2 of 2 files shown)


BossChaos:
## Code Review

**PR:** [fix: suppress completed request cancellation leak](https://github.com/modelcontextprotocol/python-sdk/pull/2611) by @MukundaKatta

- ✅ Bug fix

maxisbey:
Closing as a duplicate of #2624, which the #2610 reporter has been validating. Thanks for the contribution.

PR #2624, opened 2026-05-17
Fix completed request cancellation cleanup

### LINKED ISSUE #2610: RequestResponder.__exit__ leaks CancelledError on cancelled request, killing the stdio receive loop

FU-max-boop:
## Summary

Fixes #2610.

- Suppress the cancellation raised by `CancelScope.__exit__` only after a `RequestResponder` has completed and sent its response.
- Keep cancellation propagation unchanged for incomplete responders.
- Add focused regression coverage for the completed and incomplete cleanup paths.

The implementation is rebased onto the maintained `v1.x` line (v1.29.0), where `BaseSession` and this cleanup path remain active.

## Validation

- `pytest tests/shared/test_session.py tests/server/test_cancel_handling.py -q` — 11 passed
- `ruff format --check src/mcp/shared/session.py tests/shared/test_session.py`
- `ruff check src/mcp/shared/session.py tests/shared/test_session.py`
- `pyright --pythonpath .venv/bin/python src/mcp/shared/session.py tests/shared/test_session.py`
- `git diff --check`

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -117,7 +117,11 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
             self._entered = False
             if not self._cancel_scope:  # pragma: no cover
                 raise RuntimeError("No active cancel scope")
-            self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            try:
+                self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            except BaseException as exc:
+                if not (self._completed and isinstance(exc, anyio.get_cancelled_exc_class())):
+                    raise
 
     async def respond(self, response: SendResultT | ErrorData) -> None:
         """Send a response for this request.
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -114,7 +114,11 @@ class RequestResponder(Generic[ReceiveRequestT, SendResultT]):
             self._entered = False
             if not self._cancel_scope:  # pragma: no cover
                 raise RuntimeError("No active cancel scope")
-            self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            try:
+                self._cancel_scope.__exit__(exc_type, exc_val, exc_tb)
+            except BaseException as exc:
+                if not (self._completed and isinstance(exc, anyio.get_cancelled_exc_class())):
+                    raise
 
     async def respond(self, response: SendResultT | ErrorData) -> None:
         """Send a response for this request.
--- a/tests/shared/test_session.py
+++ b/tests/shared/test_session.py
@@ -1,3 +1,5 @@
+from typing import Any, cast
+
 import anyio
 import pytest
 
@@ -21,6 +23,18 @@ from mcp.types import (
     ServerNotification,
     ServerRequest,
 )
+
+
+class _CancelScopeThatRaisesOnExit:
+    cancel_called = True
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: object | None,
+    ) -> None:
+        raise anyio.get_cancelled_exc_class()()
 
 
 @pytest.mark.anyio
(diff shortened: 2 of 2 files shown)


BossChaos:
## Code Review

**PR:** [Fix completed request cancellation cleanup](https://github.com/modelcontextprotocol/python-sdk/pull/2624) by @FU-max-boop

- ✅ Bug fix

BossChaos:
## Code Review

**PR:** [Fix completed request cancellation cleanup](https://github.com/modelcontextprotocol/python-sdk/pull/2624) by @FU-max-boop

- ✅ Bug fix / error handling

*Reviewing under MCP ecosystem*

FU-max-boop:
CI note: I inspected the two failed Ubuntu jobs. They fail in existing tests that are not changed by this PR: `tests/interaction/lowlevel/test_timeouts.py::test_session_level_timeout_applies_to_every_request` timed out during initialize, and `tests/shared/test_streamable_http.py::test_client_crash_handled` hit an `httpx.ConnectError` while connecting to the local test server.

Both pass locally on this branch:

```bash
uv run --frozen --no-sync pytest tests/interaction/lowlevel/test_timeouts.py::test_session_level_timeout_applies_to_every_request -q
uv run --frozen --no-sync pytest tests/shared/test_streamable_http.py::test_client_crash_handled -q
```

I do not have permission to rerun Actions on this repository, but these look like transient CI timing/port failures. Happy to patch if they reproduce consistently.

FU-max-boop:
Rebased this fix onto v1.29.0 / `v1.x` and retargeted the PR because `main` is now v2 and no longer contains this `BaseSession` path. The PR is back to one commit and two files. Local validation passes: 11 focused session/cancellation tests, Ruff, Pyright, and `git diff --check`.

PR #2627, opened 2026-05-18
fix: prevent stdio_server from closing real stdin/stdout

EvanYao826:
## Summary

Fixes #1933

`stdio_server()` wraps `sys.stdin.buffer` / `sys.stdout.buffer` in `TextIOWrapper`. When the server exits, these wrappers close the underlying binary streams, which also closes the real process stdin/stdout. Any subsequent `print()` or `input()` in the caller raises `ValueError: I/O operation on closed file`.

### Root cause

`TextIOWrapper.__exit__` closes the wrapped buffer. When that buffer is `sys.stdin.buffer`, the close propagates to the real file descriptor.

### Fix

Use `os.dup()` to duplicate the file descriptors before wrapping. Closing the wrapper now only closes the duplicate, leaving the original process handles intact.

When stdin/stdout lack real file descriptors (e.g. `io.BytesIO` in tests), fall back to the original behavior.

### Changes

- `src/mcp/server/stdio.py`: Duplicate fd with `os.dup()` before wrapping
- `tests/server/test_stdio.py`: Add regression test using real file descriptors via `os.pipe()`

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,6 +17,8 @@ Example:
     ```
 """
 
+import io
+import os
 import sys
 from contextlib import asynccontextmanager
 from io import TextIOWrapper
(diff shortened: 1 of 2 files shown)


maxisbey:

Issue #1446, opened 2025-10-07
Question: MCP Python SDK Redirect URI Validation - Implementation Granularity?


younaman:
## Question

Does the current `validate_redirect_uri()` implementation in `OAuthClientMetadata` fully comply with OAuth 2.1 Section 7.12.2 requirement that "Authorization servers **MUST** take precautions to prevent redirecting user agents to untrusted URI's"?

## Current Implementation Analysis

**File**: `src/mcp/shared/auth.py`

```python
# Lines 82-91: validate_redirect_uri()
def validate_redirect_uri(self, redirect_uri: AnyUrl | None) -> AnyUrl:
    if redirect_uri is not None:
        # Validate redirect_uri against client's registered redirect URIs
        if redirect_uri not in self.redirect_uris:
            raise InvalidRedirectUriError(f"Redirect URI '{redirect_uri}' not registered for client")
        return redirect_uri
    elif len(self.redirect_uris) == 1:
        return self.redirect_uris[0]
    else:
        raise InvalidRedirectUriError("redirect_uri must be specified when client has multiple registered URIs")
```

## Current Security Measures
- ✅ Validates redirect URI against client's registered URIs
- ✅ Uses Pydantic's AnyUrl for basic URL format validation
- ✅ Handles single vs. multiple registered URIs correctly

## Key Security Concern
**What if the registered URIs themselves are untrusted?**

The current implementation only checks if a redirect URI is in the client's registered list, but **does not validate whether those registered URIs are themselves trustworthy**. For example:

- A client could register `http://evil.com/callback` as a valid redirect URI
- The current validation would **allow** this redirect
- This could lead to open redirect attacks or phishing

## Potential Security Gaps
- ❌ No validation of URI protocol (HTTP vs HTTPS)
- ❌ No domain whitelist validation
- ❌ No STRONG restrictions protection against open redirect attacks?
- ❌ No validation of URI path components
- ❌ No check for internal/local network redirects
- ❌ **No validation that registered URIs themselves are trustworthy**

## Question
Is the current implementation sufficient for OAuth 2.1 Section 7.12.2 compliance, or should additional security measures be added to prevent redirection to untrusted URIs beyond just checking against registered URIs?

**P.S.** I am not very confident that it is a real issue, which seems just a "implementation choice". I will be very appreciative if you maintainers, have any time to review it and tell me your opinion.

PR #2630, opened 2026-05-18
fix(auth): validate redirect_uri schemes at DCR registration (RFC 9700 §4.1.1, RFC 7591 §2)

### LINKED ISSUE #2629: DCR registration accepts redirect_uris with non-HTTPS / non-loopback / fragmented schemes
## Summary

The DCR handler (`mcp.server.auth.handlers.register.RegistrationHandler.handle`) does not validate the scheme of submitted `redirect_uris`. A client registered via DCR can supply `javascript:`, `data:`, `vbscript:`, `file:`, `ftp:`, or cleartext `http://` (non-loopback) values, and they pass through to the provider's `register_client`. The SDK already enforces an HTTPS-or-loopback policy on the Issuer URL (`routes.validate_issuer_url`); the same policy is missing for registered `redirect_uris`. RFC 9700 §4.1.1 and RFC 7591 §2 require it.

## Reproduction

The underlying field, `mcp.shared.auth.OAuthClientMetadata.redirect_uris` (`src/mcp/shared/auth.py:40`), is typed `list[AnyUrl] | None`. Pydantic's `AnyUrl` accepts any well-formed URL with a scheme. Verified on `main` at `161834d4ae`:

```python
from pydantic import AnyUrl, BaseModel, Field
from typing import List

class M(BaseModel):
    redirect_uris: List[AnyUrl] = Field(..., min_length=1)

for uri in [
    "javascript:alert(1)",
    "data:text/html,<script>alert(1)</script>",
    "file:///etc/passwd",
    "vbscript:msgbox(1)",
    "ftp://attacker.example/cb",
    "http://attacker.example/cb",
    "https://example.com/cb#frag",
    "https://example.com/cb#",
]:
    M(redirect_uris=[uri])  # all accepted, no ValidationError
```

Against a running MCP server with the default DCR handler, `POST /register` with any of the above values returns 201 and stores the URI. After registration, `OAuthClientMetadata.validate_redirect_uri` does exact-equality match against the registered list, so the bad URI is accepted as the authorization callback target.

## Existing parallel logic to mirror

`src/mcp/server/auth/routes.py:24–42` (`validate_issuer_url`):

```python
if url.scheme != "https" and url.host not in ("localhost", "127.0.0.1", "[::1]"):
    raise ValueError("Issuer URL must be HTTPS")

if url.fragment:
    raise ValueError("Issuer URL must not have a fragment")
```

## Related

- #1446 (closed as `duplicate`, no cross-reference recorded) — raised the same concern in question form.
- #1934 (open) — fixes RFC 8252 §7.3 loopback port matching at authorize-time. Adjacent surface but orthogonal; does not add scheme validation at register-time.
- TypeScript SDK PR `modelcontextprotocol/typescript-sdk#1738` covers the same authorize-time loopback gap on the TS side.

## Proposed fix

Add `validate_registered_redirect_uri(url: AnyUrl) -> None` next to `validate_issuer_url`:

- Reject schemes other than `https`, or `http` with host in `{"localhost", "127.0.0.1", "[::1]"}`.
- Reject URIs with a fragment (including empty fragments, e.g. `https://example.com/cb#` — note: this is also a latent bug in `validate_issuer_url`'s current `if url.fragment:` check, which I have NOT touched here to keep scope tight).
- Permit query strings (RFC 7591 §2 explicitly allows them).

Call it once per URI in `RegistrationHandler.handle` immediately after `model_validate_json` succeeds. On failure return `400 invalid_redirect_uri` per RFC 7591 §3.2.2.

PR with the patch + tests: #<PR_NUM_HERE>.

## Notes on severity

Browsers no longer navigate `javascript:` / `data:` schemes received in `Location` headers, which neutralises those vectors for browser-mediated flows. The realistic exploitable residue is (a) cleartext-HTTP redirect_uris to attacker-controlled hosts, and (b) custom-scheme deep links on devices where the MCP client uses a system handler. Defense-in-depth, not a critical exploit chain — happy to be downgraded if maintainers see it differently.

CrypticCortex:
Closes #2629.

## Summary

Adds `validate_registered_redirect_uri(url: AnyUrl)` and wires it into `RegistrationHandler.handle` so each submitted `redirect_uri` is validated immediately after `model_validate_json` succeeds. On failure, returns `400 invalid_redirect_uri` per RFC 7591 §3.2.2.

The helper mirrors the existing `validate_issuer_url` policy:

- `https` is allowed.
- `http` is allowed only for loopback hosts (`localhost`, `127.0.0.1`, `[::1]`) per RFC 8252 §7.3.
- Fragments are rejected per RFC 7591 §2 — including empty trailing-`#` fragments.
- Query strings are permitted (explicitly allowed by RFC 7591 §2).

The helper lives in a new `mcp.server.auth._redirect_uri` module to avoid a circular import with `handlers.register` (the existing `mcp.server.auth.routes` re-exports it so the public import path stays stable).

## Why now

`OAuthClientMetadata.redirect_uris` is typed `list[AnyUrl]`, which Pydantic accepts for any well-formed URL — including `javascript:`, `data:`, `file:`, `vbscript:`, `ftp:`, cleartext `http://` to non-loopback hosts, and `https://...#`. The handler stores those values verbatim and authorize-time validation accepts them by exact match. Issue #<ISSUE_NUM> has the full reproduction.

## Tests

`tests/server/auth/test_routes.py`: 13 unit tests covering each accept/reject path of the new helper, mirroring the existing `test_validate_issuer_url_*` cases (includes an explicit empty-fragment test).

`tests/server/auth/test_error_handling.py`: 5 end-to-end tests posting against `/register` — three rejection cases (`javascript:`, cleartext-`http://`, fragment) and two acceptance cases (HTTPS, loopback HTTP).

Local: `./scripts/test` green; 1190 passed, 100.00% coverage, strict-no-cover clean.

## Compatibility

No public-API removal. The change tightens what `/register` accepts; clients submitting non-spec-compliant URIs will now get `400 invalid_redirect_uri` where they previously got `201`. Existing stored client records are not re-validated.

## Related

- Adjacent to #1934 (RFC 8252 loopback port relaxation at authorize-time) — no conflict; touches different code paths.
- Question raised earlier in #1446 (closed as duplicate).

## Checks

- [x] `uv run --frozen ruff format .`
- [x] `uv run --frozen ruff check .`
- [x] `uv run --frozen pyright`
- [x] `./scripts/test` — green, 100% coverage, strict-no-cover clean
```

### DIFF
--- a/src/mcp/server/auth/_redirect_uri.py
+++ b/src/mcp/server/auth/_redirect_uri.py
@@ -1,0 +1,41 @@
+"""Internal helpers for validating registered OAuth redirect URIs.
+
+Lives outside :mod:`mcp.server.auth.routes` to avoid a circular import:
+``routes`` imports :class:`~mcp.server.auth.handlers.register.RegistrationHandler`,
+and the handler in turn needs the redirect-URI validator, so the validator
+must sit in a module that neither side has to depend on transitively.
+:mod:`mcp.server.auth.routes` re-exports :func:`validate_registered_redirect_uri`
+so callers (including tests) keep the public import path.
+"""
+
+from pydantic import AnyUrl
+
+from mcp.shared.auth import InvalidRedirectUriError
+
+
+def validate_registered_redirect_uri(url: AnyUrl) -> None:
+    """Validate that a registered redirect_uri meets OAuth 2.0 + RFC 7591 requirements.
+
+    Mirrors the policy that :func:`mcp.server.auth.routes.validate_issuer_url`
+    applies to issuer URLs: redirect URIs must use ``https``, except that
+    ``http`` is permitted for loopback hosts (``localhost``, ``127.0.0.1``,
+    ``[::1]``) per RFC 8252 §7.3, and they MUST NOT carry a fragment component
+    per RFC 7591 §2.
+
+    Args:
+        url: A registered redirect_uri value from
+            :class:`mcp.shared.auth.OAuthClientMetadata`.
+
+    Raises:
+        InvalidRedirectUriError: If the URI uses a scheme other than ``https``
+            or loopback ``http``, or if it contains a fragment.
+    """
+    # RFC 9700 §4.1.1 (OAuth 2.0 Security BCP): https-only, with the RFC 8252
+    # native-app loopback exception.
+    if url.scheme not in ("https", "http"):
+        raise InvalidRedirectUriError(f"redirect_uri must use https (or http for loopback); got scheme {url.scheme!r}")
+    if url.scheme == "http" and url.host not in ("localhost", "127.0.0.1", "[::1]"):
+        raise InvalidRedirectUriError(f"redirect_uri must use https for non-loopback hosts; got {str(url)!r}")
+    # RFC 7591 §2: redirect_uri MUST NOT contain a fragment component.
+    if url.fragment is not None:
+        raise InvalidRedirectUriError(f"redirect_uri must not have a fragment; got {str(url)!r}")
--- a/src/mcp/server/auth/handlers/register.py
+++ b/src/mcp/server/auth/handlers/register.py
@@ -8,11 +8,12 @@ from pydantic import BaseModel, ValidationError
 from starlette.requests import Request
 from starlette.responses import Response
 
+from mcp.server.auth._redirect_uri import validate_registered_redirect_uri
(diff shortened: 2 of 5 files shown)


maxisbey:
Thanks for the PR. Tracking this in #2999 instead. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2633, opened 2026-05-18
fix(streamable-http): close SSE responses on errors

### LINKED ISSUE #1450: Response leak in SSE handlers

Epochex:
﻿SSE response objects can be leaked when streaming fails or ends unexpectedly.

This ensures HTTP responses are always closed in:
- `_handle_sse_response`
- `_handle_resumption_request`

Fixes #1450.

Tests:
- uv run --frozen python -m ruff check src/mcp/client/streamable_http.py tests/client/test_streamable_http_response_cleanup.py
- uv run --frozen python -m ruff format src/mcp/client/streamable_http.py tests/client/test_streamable_http_response_cleanup.py
- uv run --frozen pytest tests/client/test_streamable_http_response_cleanup.py -q

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -270,7 +270,7 @@ class StreamableHTTPTransport:
                     if is_complete:
                         break
         finally:
-            if event_source is not None:
+            if event_source is not None:  # pragma: no branch
                 await event_source.response.aclose()
 
     def _consume_modern_cancellation(self, session_message: SessionMessage) -> bool:
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -253,20 +253,25 @@ class StreamableHTTPTransport:
         if isinstance(ctx.session_message.message, JSONRPCRequest):  # pragma: no branch
             original_request_id = ctx.session_message.message.id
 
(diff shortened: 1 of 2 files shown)


Epochex:
CI was failing only on Python 3.14 due to branch coverage reporting 256->exit in _handle_resumption_request cleanup. Added # pragma: no branch on the defensive if event_source is not None to keep branch coverage stable across 3.14 sysmon runs. All checks are green now.

PR #2643, opened 2026-05-19
fix(stdio): use os.dup to avoid closing real stdin/stdout on server exit

jrlprost:
Fixes #1933.

## Root cause

`stdio_server()` wraps `sys.stdin.buffer` in a `TextIOWrapper`. When that wrapper is garbage-collected after the context manager exits, Python's finalizer calls `close()` on it, which propagates down and closes `sys.stdin.buffer` (fd 0). Any subsequent `print()` or `sys.stdin.read()` then raises `ValueError: I/O operation on closed file.`

The existing comment ("Purposely not using context managers for these, as we don't want to close standard process handles") shows the intent was correct, but the wrapper's `__del__` closes the buffer regardless.

## Fix

Duplicate the file descriptor with `os.dup()` before wrapping. Our `TextIOWrapper` now owns a private copy of the fd; closing it leaves the original `sys.stdin`/`sys.stdout` untouched.

Falls back to the original buffer when the stream is not backed by a real fd (e.g. `BytesIO` in tests), matching prior behaviour in that case. The dup'd wrappers are explicitly closed in a `finally` block after the task group exits.

## Changes

- `src/mcp/server/stdio.py`: `os.dup()` the fd; track ownership with `stdin_created`/`stdout_created`; close in `finally`.
- `tests/server/test_stdio.py`: regression test using real OS pipes — asserts `sys.stdin.buffer` and `sys.stdout.buffer` are still open after `stdio_server()` exits.

## Testing

```
uv run --frozen pytest tests/server/test_stdio.py -v  # 3/3 pass
uv run --frozen ruff check .                           # no issues
uv run --frozen pyright src/mcp/server/stdio.py tests/server/test_stdio.py  # 0 errors
```

Branch coverage for `src/mcp/server/stdio.py`: 100 %.

The `except io.UnsupportedOperation` branch is covered by the existing `test_stdio_server_invalid_utf8` test (BytesIO stdin). The new test covers the `try` success path with real pipe fds.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,6 +17,8 @@ Example:
     ```
 """
 
+import io
+import os
 import sys
 from contextlib import asynccontextmanager
 from io import TextIOWrapper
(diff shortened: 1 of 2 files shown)


maxisbey:
closing as there's already an existing PR for this issue

PR #2652, opened 2026-05-21
feat: add protocol version override support for client session initialization

STiFLeR7:
## Description
This PR adds support for client session protocol version overrides during initialization, enabling clients to negotiate older or custom protocol versions (e.g. `2024-11-05`) with MCP servers.

### Changes
- Updated `ClientSession.initialize(...)` to accept a custom `protocol_version`.
- Added `protocol_version` field to the high-level `Client` dataclass.
- Added `protocol_version` field to `ClientSessionParameters` in `ClientSessionGroup`.
- Implemented corresponding unit and integration tests.

### DIFF
--- a/src/mcp/client/_probe.py
+++ b/src/mcp/client/_probe.py
@@ -44,7 +44,7 @@ def _parse_supported(data: Any) -> list[str] | None:
         return None
 
 
-async def negotiate_auto(session: ClientSession) -> None:
+async def negotiate_auto(session: ClientSession, protocol_version: str | None = None) -> None:
     """Drive the ``mode='auto'`` connect-time policy on ``session``.
 
     Probes ``server/discover`` once (twice if the server names a mutual
@@ -73,7 +73,10 @@ async def negotiate_auto(session: ClientSession) -> None:
                 if supported is not None and not any(v in HANDSHAKE_PROTOCOL_VERSIONS for v in supported):
                     raise  # server is modern-only and disjoint — real incompatibility
             try:
-                await session.initialize()  # every other rpc-error → legacy (the denylist)
+                if protocol_version is not None:
+                    await session.initialize(protocol_version=protocol_version)
+                else:
+                    await session.initialize()  # every other rpc-error → legacy (the denylist)
             except MCPError as handshake_exc:
                 if handshake_exc.code != UNSUPPORTED_PROTOCOL_VERSION or attempt != 0:
                     raise
@@ -94,7 +97,10 @@ async def negotiate_auto(session: ClientSession) -> None:
         try:
             result = types.DiscoverResult.model_validate(raw)
         except ValidationError:
-            await session.initialize()  # unparseable result → not modern evidence
+            if protocol_version is not None:
+                await session.initialize(protocol_version=protocol_version)
+            else:
+                await session.initialize()  # unparseable result → not modern evidence
             return
         session.adopt(result)
         return
--- a/src/mcp/client/_probe.py
+++ b/src/mcp/client/_probe.py
@@ -49,7 +49,7 @@ def _parse_supported(data: Any) -> list[str] | None:
         return None
 
 
-async def negotiate_auto(session: ClientSession) -> None:
+async def negotiate_auto(session: ClientSession, protocol_version: str | None = None) -> None:
     """Drive the ``mode='auto'`` connect-time policy on ``session``.
 
     Probes ``server/discover`` once (twice if the server names a mutual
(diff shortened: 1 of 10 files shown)


keeltrace:

I think the high-level override is silently ignored on the normal/default `mode="auto"` success path.

`Client.mode` defaults to `"auto"`. This PR passes `protocol_version_override` into `negotiate_auto()`, but `negotiate_auto()` only uses that value in its **legacy fallback** calls to `session.initialize(...)`.

If `server/discover` succeeds and returns a valid modern `DiscoverResult`, the function does:

```python
session.adopt(result)
return
```

without consulting `protocol_version` at all.

So this configuration:

```python
Client(server, protocol_version_override="2024-11-05")
```

does not actually request `2024-11-05` from a server that supports the modern discovery path. With the default `mode="auto"`, it can negotiate/adopt a 2026-era version instead. The override only takes effect when discovery fails (or is unparseable) and auto mode falls back to `initialize`.

The new tests line up with that gap:

- the high-level `Client` test explicitly uses `mode="legacy"`;
- the two `negotiate_auto` tests exercise only `MCPError` / invalid-discover fallback;
- there is no test where discovery succeeds while an override is supplied.

That makes the public field especially easy to misuse because its docstring says:

> The protocol version to request during initialization. Defaults to the latest version.

while the default connection mode may never initialize at all.

Could we make the contract explicit in one of these ways?

1. If `protocol_version_override` is a handshake-era pin, have `Client` force the handshake / bypass successful modern discovery; or
2. reject `protocol_version_override` with `mode="auto"` and require `mode="legacy"` so it cannot be silently ignored; or
3. if the intent really is “fallback handshake version only”, rename/document it that way and add a successful-discovery test proving that behavior is intentional.

A regression test worth adding is a server that successfully answers `server/discover`, with:

```python
async with Client(server, protocol_version_override="2024-11-05") as client:
    ...
```

and an assertion for whichever behavior the API intends.

Related consistency point: although the PR description says this enables “older or custom protocol versions”, `ClientSession.initialize()` still rejects any server-selected version not in `HANDSHAKE_PROTOCOL_VERSIONS`, so a truly arbitrary custom version cannot successfully negotiate even if a server echoes it. Tightening the advertised contract would avoid that ambiguity too.

Checked against PR head `abb1d5b`. This is a static API/control-flow review; I did not run the repository suite locally.

STiFLeR7:
Confirmed and fixed in `f55e795a`. You're right: `negotiate_auto`'s successful-discovery path (`session.adopt(result); return`) never consulted `protocol_version` — it was only threaded into the two `initialize()` fallback call sites. Traced the exact control flow to confirm before changing anything.

Went with your option 1: since the PR description's own stated intent is "enabling clients to negotiate older or custom protocol versions," an override has to actually win, not depend on whether the server happens to answer `server/discover`. So `negotiate_auto` now short-circuits at the top — if `protocol_version` is set, it skips the probe entirely and goes straight to the legacy handshake at that version. The two now-unreachable inline branches that used to pass `protocol_version` to the fallback `initialize()` calls collapsed back to their original unconditional form, since the override is always `None` by the time the loop runs.

Added:
- A unit test on `negotiate_auto` proving the probe is skipped even when the stub's discover script would otherwise return a valid modern result (`test_a_protocol_version_override_skips_discovery_and_forces_the_legacy_handshake`).
- An e2e test over a real streamable-HTTP server with `mode="auto"` (default) + override, asserting `initialize` is the only method sent and `server/discover` never is (`test_auto_mode_with_a_protocol_version_override_skips_discover_and_initializes`), plus its requirements-manifest entry.

Both fail on the pre-fix code and pass after (verified via `git stash`). Full suite, ruff, and pyright are clean; the only failures are 31 pre-existing ones unrelated to this file (confirmed identical with the fix stashed out).

Separate, smaller point from your review I did not change: the arbitrary-custom-version consistency gap (`ClientSession.initialize()` still rejects any server-echoed version outside `HANDSHAKE_PROTOCOL_VERSIONS`) is real but out of scope for this fix — happy to open a follow-up if that's wanted.

One more thing surfaced while testing this end-to-end: for an in-process `Server`/`MCPServer` target, `Client(server, protocol_version_override=...)` with the default `mode="auto"` currently hits a hard `MCPError: Method not found` rather than silently dropping the override — `_connect_inproc`'s non-legacy connector wires up a `DirectDispatcher` peer (`modern_on_request`) that never implements the legacy `initialize` RPC at all, so `negotiate_auto`'s new short-circuit has nothing to call. That's a separate, pre-existing connector-selection gap (mode picks the connector before `protocol_version_override` is known), unrelated to this bug and outside what I'm comfortable deciding unilaterally here. Flagging in case it's worth its own issue — happy to file one if useful.

STiFLeR7:
Rebased onto main (89 commits) and fixed a genuine lint failure the rebase surfaced: RequestResponder and mcp.shared.session were removed upstream since this test was added. Updated message_handler to the current IncomingMessage type, matching every other handler in this file. Ruff, pyright, and all 92 tests in the file pass.

STiFLeR7:
Fixed cubic's P1 (in-proc auto-mode + override never connected) and P2 (no validation on the override) findings, plus the test-coverage gap. Left the elicitation-capability-gating note unaddressed pending confirmation it applies across handshake-era sub-versions.

PR #2658, opened 2026-05-21
Add list_all_* helpers that drain pagination on the client

### LINKED ISSUE #2556: Add `list_all_*` helpers to drain pagination
### Description

What I hit

Three layers: 

Proposal

```python
```

Caller side would just be:

```python
```

### References

adityasingh2400:
Closes #2556.

## Summary

- Add `list_all_tools` / `list_all_prompts` / `list_all_resources` / `list_all_resource_templates` on `Client`. Each walks `next_cursor` until exhausted and returns the combined list.
- Add `iter_all_tools` / `iter_all_prompts` / `iter_all_resources` / `iter_all_resource_templates` async iterators, for streaming consumers that don't want to materialize every page in memory.
- Update the docstrings on the single-page `list_*` methods to say so plainly and point at the new drains.
- Switch `ClientSessionGroup._aggregate_components` to drain pagination so multi-server aggregators don't silently drop pages past the first (this was specifically called out in the issue as the scenario where high tool counts are most likely).

Existing tests that mocked the session's `list_tools` / `list_resources` / `list_prompts` now set `next_cursor=None` on the mock return values so the new drain loop terminates after a single page.

## Why

The single-page `list_*` methods are easy to misuse, which is the whole point of the issue:

> Connected Claude Code to an AWS Bedrock AgentCore Gateway with 31 tools registered across 7 targets. Only 14 showed up. The gateway returns 14 on page 0 with a `nextCursor`, then 17 on page 1. Claude Code calls `tools/list` once and drops the cursor.

The existing primitives are kept; this just adds a drain helper next to them so downstream client code can get the "give me everything" behavior in one call instead of writing the loop themselves.

## Test plan

- [x] `uv run pytest tests/client/test_list_all_pagination.py tests/client/test_session_group.py tests/client/test_list_methods_cursor.py` (25 passed)
- [x] `uv run pytest tests/client/` (208 passed, 1 xfailed)
- [x] `uv run pytest tests/` (1180 passed, 98 skipped, 1 xfailed)
- [x] `uv run ruff check . && uv run ruff format --check .` (clean)
- [x] `uv run pyright src/mcp/ tests/client/test_list_all_pagination.py tests/client/test_session_group.py` (0 errors)
- [x] 100% coverage on `src/mcp/client/client.py` and `src/mcp/client/session_group.py` via `tests/client/`

New tests cover:
- Multi-page drain across 2 and 3 pages for tools, prompts, resources, and resource templates.
- The cursor sent on subsequent requests matches the `next_cursor` returned by the previous page.
- Single-page and empty-server cases.
- `iter_all_tools` yields one tool at a time and pages lazily.
- `ClientSessionGroup` aggregates across paginated `list_tools` calls and supplies the cursor on the follow-up request.

## Notes

I went with `Client` helpers only and left `ClientSession` as the single-page primitive (matches what the issue suggested as the conservative call). Happy to also add session-level helpers if you'd prefer that.

### DIFF
--- a/src/mcp/client/client.py
+++ b/src/mcp/client/client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import AsyncIterator
 from contextlib import AsyncExitStack
 from dataclasses import KW_ONLY, dataclass, field
 from typing import Any
@@ -26,10 +27,14 @@ from mcp.types import (
     ListToolsResult,
     LoggingLevel,
     PaginatedRequestParams,
+    Prompt,
     PromptReference,
     ReadResourceResult,
     RequestParamsMeta,
+    Resource,
+    ResourceTemplate,
     ResourceTemplateReference,
+    Tool,
 )
 
 
@@ -195,7 +200,11 @@ class Client:
         cursor: str | None = None,
         meta: RequestParamsMeta | None = None,
     ) -> ListResourcesResult:
-        """List available resources from the server."""
+        """List a single page of available resources from the server.
+
+        Returns one page only. The result may include a `next_cursor` if more
+        pages are available. Use `list_all_resources` to drain every page.
+        """
         return await self.session.list_resources(params=PaginatedRequestParams(cursor=cursor, _meta=meta))
 
     async def list_resource_templates(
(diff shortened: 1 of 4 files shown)


CJGjr:
Hi @adityasingh2400, thanks for implementing this.

I opened the original issue (#2556) and wanted to use these helpers, but the PR went stale against the v2 rework on main and now conflicts. Rather than let it sit, I rebased your work onto current main, resolved the conflicts, and added a guard for the non-advancing-cursor case @agaonker raised on the issue, then put it up as #3021 so it is ready to merge. Your authorship is preserved on the feature commit.

PR #2660, opened 2026-05-22
fix(oauth): narrow context.lock scope in async_auth_flow

### LINKED ISSUE #1326: Clients Using Storage Face Deadlock on Token Refresh for SSE
### Initial Checks

### Resulting Behavior - After tokens expire or after a new instance is re-built from storage, the SSE server will not work until full token-clearing and re-auth

As long as a token refresh attempt is not made, the server will work. 

There are two circumstances when a refresh attempt is made, causing the SSE call to hang:
1. Tokens expire during a working SSE session
2. When a Client uses a stale token from storage

### Source of failure for each scenario

**Failure Mode**
1. `OAuthClientProvider` lazy loads and is passed into `aconnect_sse()` without having checked refresh
2. The `OAuthClientProvider.async_auth_flow()` generator enters the refresh logic
3. It yields a refresh request and waits for a response (line ~544 in auth.py)
4. `aconnect_sse()` uses `client.stream()` which is trying to open an SSE connection
5. There's no response mechanism - SSE expects to start streaming events, not handle a token refresh request
6. DEADLOCK: Auth generator waiting for refresh response that will never come

**Temporary Success**
- First request in `aconnect_sse()` uses `client.stream()` yield 401 which sends everything back out of the SSE context
- `OAuthClientProvider` is able to perform full re-auth
- When next sent to `aconnect_sse()` fresh auth headers are availble  and SSE stream succeeds
- SSE is long-lived so works for as long as tokens don't need refresh

### Root cause

In /mcp/client/sse.py, this call with `aconnect_sse` hangs infinitely.

```python
    try:
        logger.debug(f"Connecting to SSE endpoint: {remove_request_params(url)}")
        async with httpx_client_factory(
            headers=headers, auth=auth, timeout=httpx.Timeout(timeout, read=sse_read_timeout)
        ) as client:
            async with aconnect_sse(  #<======================== This hangs indefinitely with a 200 response
                client,
                "GET",
                url,
            ) as event_source:
                event_source.response.raise_for_status()

```

This is because `OAuthClientProvider`, which is attached to the client, on token refresh, tries to yield when in the middle of `client.stream` here:
```python
            if not self.context.is_token_valid() and self.context.can_refresh_token():
                # Try to refresh token
                refresh_request = await self._refresh_token()
                refresh_response = yield refresh_request. #<======================== Causes client.stream() to hang

                if not await self._handle_refresh_response(refresh_response):
                    # Refresh failed, need full re-authentication
                    self._initialized = False
```
Other yields in the broader method occur after the stream got a 401 and so are not in the same context.

### The Fix - Force token refresh check outside of the stream context

Any fix needs to remove the `OAuthClientProvider` from `client` before calling `aconnect_sse`. That way it won't yield the refresh request.

There is probably a more elgant/direct approach but the below worked. I had to create a separate client to avoid some lock conflict.

```python
    try:
        logger.debug(f"Connecting to SSE endpoint: {remove_request_params(url)}")
        async with httpx_client_factory(
            headers=headers, auth=auth, timeout=httpx.Timeout(timeout, read=sse_read_timeout)
        ) as client:
            
            #----- Start of fix -------
            auth_headers = {}
            if auth:
                # Initialize auth to load stored tokens
                if hasattr(auth, '_initialize') and not getattr(auth, '_initialized', False):
                    await auth._initialize()

                # Check if tokens need refresh
                if hasattr(auth, 'context') and auth.context:
                    # Check if token is expired or needs refresh
                    if not auth.context.is_token_valid() and auth.context.can_refresh_token():
                        logger.debug("Token needs refresh before SSE connection")

                        # Use a separate client for refresh to avoid lock issues
                        async with httpx.AsyncClient() as refresh_client:
                            # Create a dummy request to trigger the refresh flow
                            dummy_request = httpx.Request("GET", url)

                            # Run the auth flow to refresh tokens
                            auth_gen = auth.async_auth_flow(dummy_request)
                            try:
                                # Start the generator
                                refresh_request = await auth_gen.asend(None)

                                # Execute the refresh request
                                refresh_response = await refresh_client.request(
                                    refresh_request.method,
                                    refresh_request.url,
                                    data=refresh_request.content,
                                    headers=refresh_request.headers,
                                )

                                # Send response back to generator
                                await auth_gen.asend(refresh_response)
                                logger.debug("Token refreshed successfully before SSE connection")
                            except StopAsyncIteration:
                                # Normal completion of auth flow
                                pass
                            except Exception as e:
                                logger.error(f"Failed to refresh token before SSE: {e}")

                    # Extract the token after potential refresh
                    if auth.context.current_tokens and auth.context.current_tokens.access_token:
                        token = auth.context.current_tokens.access_token
                        auth_headers['Authorization'] = f"Bearer {token}"

            # CRITICAL: Remove auth from client to prevent generator interference
            client.auth = None
            #----- Resume existing code -------
            async with aconnect_sse(
                client,
                "GET",
                url,
                headers=auth_headers,    #<============================== Must add this too
            ) as event_source:
                event_source.response.raise_for_status()
```

Since you removed auth from the client, you have to insert the headers later for the POST call as well:

```python
                    async def post_writer(endpoint_url: str):
                        try:
                            async with write_stream_reader:
                                async for session_message in write_stream_reader:
                                    logger.debug(f"Sending client message: {session_message}")
                                    response = await client.post(
                                        endpoint_url,
                                        json=session_message.message.model_dump(
                                            by_alias=True,
                                            mode="json",
                                            exclude_none=True,
                                        ),
                                        headers=auth_headers,  #<======================== Cause client.stream() to hang
```

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
1.12.4 (but same code as current version)
```

peisuke:
Narrows the lock scope inside `OAuthClientProvider.async_auth_flow` so the actual HTTP request `yield` runs outside any lock. Concurrent requests through the same provider — in particular, a POST issued while a GET SSE long-poll is in flight — no longer serialize on `context.lock`. A new `refresh_lock` preserves single-flight semantics for token refresh so the change does not introduce a refresh race.

## Motivation and Context

`OAuthClientProvider.async_auth_flow` previously held `self.context.lock` for the entire flow body, including the `response = yield request` that hands the request off to httpx. For requests that complete quickly this is fine, but a Streamable HTTP transport opens a long-lived **GET SSE stream** for server-initiated messages alongside the **POST stream** for client-initiated RPCs. Both go through the same `OAuthClientProvider` instance, so the GET SSE long-poll's `yield request` holds `context.lock` for the entire SSE lifetime. Any concurrent `tools/call` POST then blocks on the lock until the SSE returns, producing a multi-second stall in OAuth-authenticated MCP connections.

This commit splits the single coarse lock into purpose-specific scopes:

- **Phase 1** (`context.lock`): initialize state, capture `protocol_version`, and decide whether a refresh is needed. Short-held; no HTTP I/O.
- **Phase 2** (`refresh_lock`, new): single-flight token refresh. The refresh `yield` happens outside any lock. A double-check inside `context.lock` ensures concurrent waiters do not redundantly refresh after another coroutine completed one.
- **Phase 3** (no lock): add the auth header and yield the actual request. GET SSE long-polls and other long-running requests no longer block unrelated traffic.
- **Phase 4** (`context.lock`): 401 / 403 full OAuth re-auth path. Conservatively kept under lock because this path is rare and its yielded sub-requests target the AS, not the resource server. A future change can narrow this further with the same pattern.

`refresh_lock` is added to `OAuthContext` (alongside the existing `lock`) so concurrent expirations trigger at most one refresh request. The double-check pattern inside `refresh_lock` makes the second waiter observe the freshly-updated token and skip its own refresh.

## How Has This Been Tested?

Five new tests in `tests/client/test_auth.py` exercise the refactor:

- `test_concurrent_request_not_blocked_by_pending_long_running_request` — the core regression test. Drives one `async_auth_flow` generator to its yield (simulating a GET SSE long-poll), then opens a second flow on the same provider and asserts it reaches its own yield within `anyio.fail_after(5)`. Pre-fix this hangs indefinitely.
- `test_refresh_lock_double_check_skips_redundant_refresh` — a flow started after another finished its refresh observes the fresh token in Phase 1 and skips Phase 2 entirely; no second refresh request is sent.
- `test_double_check_inside_refresh_lock_skips_second_refresh` — monkeypatches `is_token_valid` to return False in Phase 1 and True in the Phase 2 double-check, exercising the branch where a second coroutine's refresh is correctly elided inside `refresh_lock`.
- `test_refresh_with_failed_status_clears_tokens` — covers the failed-refresh path (`status_code != 200`) including the `_initialized = False` reset.
- `test_refresh_with_invalid_json_clears_tokens` — covers the `ValidationError` branch when the refresh response body is malformed.

`./scripts/test` is clean: 1177 passed / 98 skipped / 1 xfailed, **100.00%** coverage on `src/mcp/client/auth/oauth2.py`, `strict-no-cover` reports no leftover `# pragma: no cover` violations. Four previously-needed pragmas in `_handle_refresh_response` and the Phase 2 yield are now covered by the new tests and removed.

## Breaking Changes

None. Public API is unchanged. Concurrency semantics narrow only — previously-serialized requests now proceed in parallel; token refresh remains serialized via `refresh_lock`.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Targets `main` (v2 development). The same bug exists on `v1.x`; if a backport is wanted, that can land as a follow-up PR.

Fixes #1326

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -533,7 +533,12 @@ class OAuthClientProvider(httpx.Auth):
             # Capture protocol version from request headers
             self.context.protocol_version = request.headers.get(MCP_PROTOCOL_VERSION)
 
-            if not self.context.is_token_valid() and self.context.can_refresh_token():
+            # pragma: no branch — coverage.py on Python 3.10/3.11 (sys.settrace
+            # backend) cannot reliably track both arms of compound boolean
+            # predicates inside an ``async with`` block in an async generator.
+            # Python 3.12+ (sys.monitoring) handles this correctly; the pragmas
+            # below are workarounds for the legacy backend only.
+            if not self.context.is_token_valid() and self.context.can_refresh_token():  # pragma: no branch
                 needs_refresh = True
 
         # === Phase 2: single-flight token refresh (yield outside context.lock) ===
@@ -543,14 +548,14 @@ class OAuthClientProvider(httpx.Auth):
                 # refreshed while we were waiting on refresh_lock.
                 refresh_request: httpx.Request | None = None
                 async with self.context.lock:
-                    if not self.context.is_token_valid() and self.context.can_refresh_token():
+                    if not self.context.is_token_valid() and self.context.can_refresh_token():  # pragma: no branch
                         refresh_request = await self._refresh_token()
-                if refresh_request is not None:
+                if refresh_request is not None:  # pragma: no branch
                     # yield runs outside any lock so a long network round trip
                     # does not block unrelated concurrent requests.
                     refresh_response = yield refresh_request
                     async with self.context.lock:
-                        if not await self._handle_refresh_response(refresh_response):
+                        if not await self._handle_refresh_response(refresh_response):  # pragma: no branch
                             # Refresh failed; fall through to 401 handling below.
                             self._initialized = False
 
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -463,7 +463,7 @@ class OAuthClientProvider(httpx.Auth):
 
         return httpx.Request("POST", token_url, data=refresh_data, headers=headers)
 
-    async def _handle_refresh_response(self, response: httpx.Response) -> bool:  # pragma: no cover
+    async def _handle_refresh_response(self, response: httpx.Response) -> bool:
         """Handle token refresh response. Returns True if successful."""
         if response.status_code != 200:
             logger.warning(f"Token refresh failed: {response.status_code}")
@@ -544,13 +544,13 @@ class OAuthClientProvider(httpx.Auth):
                 refresh_request: httpx.Request | None = None
                 async with self.context.lock:
                     if not self.context.is_token_valid() and self.context.can_refresh_token():
-                        refresh_request = await self._refresh_token()  # pragma: no cover
+                        refresh_request = await self._refresh_token()
                 if refresh_request is not None:
                     # yield runs outside any lock so a long network round trip
                     # does not block unrelated concurrent requests.
-                    refresh_response = yield refresh_request  # pragma: no cover
+                    refresh_response = yield refresh_request
                     async with self.context.lock:
-                        if not await self._handle_refresh_response(refresh_response):  # pragma: no cover
+                        if not await self._handle_refresh_response(refresh_response):
                             # Refresh failed; fall through to 401 handling below.
                             self._initialized = False
 
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -541,14 +541,11 @@ class OAuthClientProvider(httpx.Auth):
             async with self.context.refresh_lock:
                 # Re-check under context.lock: another coroutine may already have
                 # refreshed while we were waiting on refresh_lock.
+                refresh_request: httpx.Request | None = None
                 async with self.context.lock:
-                    still_invalid = (
-                        not self.context.is_token_valid()
-                        and self.context.can_refresh_token()
-                    )
-                    if still_invalid:
+                    if not self.context.is_token_valid() and self.context.can_refresh_token():
                         refresh_request = await self._refresh_token()  # pragma: no cover
-                if still_invalid:
+                if refresh_request is not None:
                     # yield runs outside any lock so a long network round trip
                     # does not block unrelated concurrent requests.
                     refresh_response = yield refresh_request  # pragma: no cover
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -114,7 +114,18 @@ class OAuthContext:
     token_expiry_time: float | None = None
 
     # State
+    #
+    # `lock` guards short-lived reads/writes of provider state (initialization
+    # flag, token cache mutation, protocol_version assignment). It is held only
+    # while mutating state and is released before any HTTP request is yielded
+    # so a long-running request (e.g. GET SSE long-poll) does not block
+    # unrelated concurrent requests.
+    #
+    # `refresh_lock` provides single-flight semantics for token refresh: only
+    # one concurrent refresh fires; other waiters block on this lock, then
+    # re-check the token cache and proceed without re-refreshing.
     lock: anyio.Lock = field(default_factory=anyio.Lock)
+    refresh_lock: anyio.Lock = field(default_factory=anyio.Lock)
 
     def get_authorization_base_url(self, server_url: str) -> str:
         """Extract base URL by removing path component."""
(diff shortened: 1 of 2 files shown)


Bartok9:
Re-confirmation + offer of a conflict-free rebase.

This fix is needed: #2847 just landed (June 12) as a **second, independent production repro** of the exact `RuntimeError: The current task is not holding this lock` this PR addresses — a gateway connecting to multiple OAuth MCP servers concurrently, where `anyio.Lock` (which records task identity at `acquire()` and enforces same-task `release()`) is held across the `yield` in `async_auth_flow`, so httpx driving the generator from a different task throws on release. The reporter explicitly asks for this PR to be reviewed/merged.

The PR currently shows `CONFLICTING` against `main`. The conflict is small and isolated to the Phase-1 refresh block (`main` still has the inline refresh under the held lock; this PR restructures it into Phase-1 decision / Phase-2 single-flight refresh outside the lock). I rebased the four commits onto current `main` (`616476f`-era) — the resolution is purely "take this PR's restructured refresh path" — and the result is green:

```
uv run pytest tests/client/test_auth.py -q
108 passed, 1 xfailed (0.29s)
uv run ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py
All checks passed!
```

Conflict-free rebase (author/commits preserved): https://github.com/Bartok9/python-sdk/tree/salvage/2660-oauth-lock-rebase — happy to open it as a fresh PR if @peisuke is unavailable to push, or @peisuke can cherry-pick. The single-flight `refresh_lock` + re-check-under-`lock` pattern is the right shape; narrowing the lock so no HTTP `yield` (especially the long-poll GET SSE) runs while holding `context.lock` is exactly what unblocks the concurrent case.

Bartok9:
Opened #2858 — a conflict-free rebase of this PR's four commits onto current `main` (your authorship preserved on all four), so the fix can be merged without waiting on this draft to be un-conflicted. **Full credit is yours, @peisuke** — I only resolved the rebase. It's green (108 auth tests pass, ruff clean) and closes #2847. Glad to close #2858 if you'd rather push the rebase here or if maintainers prefer to merge this PR directly — just didn't want the fix to keep sitting now that #2847 is a second production repro.

Bartok9:
Hi @peisuke — wanted to give you a clear heads-up on what I did here and why, so nothing feels like it went around you.

**What I did:** I opened #2858, which is your four commits from this PR rebased cleanly onto current `main`. Your authorship is preserved on every commit (`peisuke.com@gmail.com`), and the PR title and description both state plainly that the fix is yours — I only resolved the rebase, I didn't change the approach or claim the work.

**Why I did it:** This PR has been a `CONFLICTING` draft since May 22 with no maintainer review, and the conflict against `main` has been the thing blocking it. Then #2847 landed on June 12 as a second, independent production repro of the exact `RuntimeError` your fix addresses, with the reporter explicitly asking for this PR to be merged. Rather than let a correct, well-tested fix keep sitting behind a mechanical rebase conflict, I put up a ready-to-merge version so a maintainer can act on it today.

**Your call from here:** If you'd rather push the rebase to this branch yourself, or maintainers prefer to merge this PR directly, I'll close #2858 immediately — it exists only to unblock your fix, not to replace it. The credit is and should remain yours. Just let me know what you prefer.

Thanks for the original fix — the narrow-the-lock-scope + single-flight `refresh_lock` approach is the right shape, and it keeps trio portability (unlike the `asyncio.Lock` swap others floated).

peisuke:
Thanks @Bartok9 — really appreciate you picking this up, and especially the care around authorship and credit.

Let's go with **#2858**. It's already green, rebased cleanly onto `main`, and closes #2847, so it's the fastest path to getting this in front of a maintainer — no need to wait on me to un-conflict this draft.

I'm happy to treat #2858 as the canonical PR from here, and to close this one (#2660) as superseded if maintainers prefer a single thread — just say the word. The narrow-the-lock + single-flight `refresh_lock` shape is exactly what I intended, and keeping trio portability was deliberate, so I'm glad that carried through the rebase.

Thanks again for unblocking it.

PR #2662, opened 2026-05-22
docs(auth): clarify that resource_server_url must include the transport path

adityasingh2400:
Issue #1264 keeps surfacing because RFC 9728 §3.3 requires strict equality between the client's resource identifier and the `resource` value advertised in protected resource metadata, but the public surface of `AuthSettings.resource_server_url` does not say this. PR #1407 established the contract that `resource_server_url` must be the full endpoint URL including the transport path, and the `simple-auth` example was updated, but the matching `examples/snippets/servers/oauth_server.py` snippet (which is rendered into `README.v2.md` by `scripts/update_readme_snippets.py`) still passes `http://localhost:3001` with no path, the `AuthSettings.resource_server_url` description still says only "URL of the MCP server", and `simple-auth/server.py` hardcodes `/mcp` even when the user passes `--transport=sse`.

This change updates the `AuthSettings.resource_server_url` field description to spell out the contract, including the requirement that the URL include the transport path and the reference to RFC 9728 §3.3. The `oauth_server.py` snippet is changed to `http://localhost:3001/mcp` with a comment explaining why, and `README.v2.md` is regenerated so the embedded snippet matches. In `simple-auth/server.py`, the override that built `server_url` picks `/sse` or `/mcp` based on the `--transport` flag so the SSE path no longer produces an RFC 9728 mismatch.

Per the closing comment on #2189, the maintainer asked for a docs and example fix here rather than auto-appending `streamable_http_path`, because `streamable_http_path` is an internal Starlette route and not always the public URL (for example when the app is mounted at `/api` with `streamable_http_path="/"`, or behind a reverse proxy). Verified with `uv run --frozen ruff check`, `uv run --frozen pyright`, `uv run --frozen pytest tests/server/auth/ tests/server/mcpserver/` (388 passed), and `uv run --frozen python scripts/update_readme_snippets.py --check`.

Refs #1264

### DIFF
--- a/README.v2.md
+++ b/README.v2.md
@@ -1024,8 +1024,8 @@ mcp = MCPServer(
     # Auth settings for RFC 9728 Protected Resource Metadata.
     # `resource_server_url` MUST be the full public URL of the MCP endpoint, including
     # the transport path (e.g. `/mcp` for streamable-http, `/sse` for sse). RFC 9728
-    # §3.3 requires strict equality between the client's resource identifier and the
-    # `resource` value advertised in the protected resource metadata.
+    # section 3.3 requires strict equality between the client's resource identifier and
+    # the `resource` value advertised in the protected resource metadata.
     auth=AuthSettings(
         issuer_url=AnyHttpUrl("https://auth.example.com"),  # Authorization Server URL
         resource_server_url=AnyHttpUrl("http://localhost:3001/mcp"),  # Public MCP endpoint URL
--- a/README.v2.md
+++ b/README.v2.md
@@ -1021,10 +1021,14 @@ mcp = MCPServer(
     "Weather Service",
     # Token verifier for authentication
     token_verifier=SimpleTokenVerifier(),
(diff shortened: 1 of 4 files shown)


adityasingh2400:
Pushed d54b216 to fix the 10 Windows CI failures.

Root cause was the `§` section symbol I put in the README code snippet at line 1027. `tests/test_examples.py::test_docs_examples` pipes the snippet through ruff via stdin, and on Windows the pipe defaults to cp1252 so ruff fails to decode the two-byte UTF-8 sequence with `stream did not contain valid UTF-8`. Linux and macOS pipes default to UTF-8 so all 17 non-Windows jobs passed.

Replaced `§` with the word "section" across the four files I introduced it in (README.v2.md, src/mcp/server/auth/settings.py, and the two example servers). Pre-existing `§` references in tests, routes.py, auth.py, and oauth2.py are untouched since they are not run through pytest_examples.

PR #2664, opened 2026-05-23
fix(auth): accept client credentials from Basic auth header in token endpoint (#1315)

### LINKED ISSUE #1315: OAuth TokenHandler should check Authorization header for client credentials

arkh-node:
## Problem

Fixes #1315.

When a client uses HTTP Basic authentication per [RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), its `client_id` and `client_secret` arrive in the `Authorization` header rather than the request body. Two failure paths caused this to break:

1. **`ClientAuthenticator`** read `client_id` from form data first, and raised `AuthenticationError("Missing client_id")` before it even reached the Basic auth parsing branch.

2. **`TokenHandler`** validated `form_data` against `TokenRequest` (which requires `client_id`), so even if authentication had succeeded, the Pydantic model would reject the request.

## Fix

**`client_auth.py`** — before the `Missing client_id` guard, attempt to extract `client_id` from a `Basic` `Authorization` header if it is absent from form data:

```python
if not client_id:
    auth_header = request.headers.get("Authorization", "")
    if auth_header.startswith("Basic "):
        try:
            decoded = base64.b64decode(auth_header[6:]).decode("utf-8")
            if ":" in decoded:
                client_id = unquote(decoded.split(":", 1)[0])
        except (ValueError, UnicodeDecodeError, binascii.Error):
            pass
```

**`token.py`** — after `client_authenticator` successfully authenticates the request (so `client_info` is already verified), populate `client_id` from `client_info` when absent from form data:

```python
form_data = dict(await request.form())
if "client_id" not in form_data:
    form_data["client_id"] = client_info.client_id
```

This is safe because `client_info` is the result of successful authentication — we already know who the client is.

## Tests

Two new tests in `test_auth_integration.py`:

- `test_basic_auth_without_client_id_in_body` — `authorization_code` grant with `client_id` only in `Authorization` header → `200 OK` with access token
- `test_basic_auth_refresh_token_without_client_id_in_body` — `refresh_token` grant with `client_id` only in `Authorization` header → `200 OK` with new access token

All 66 existing auth tests continue to pass.

### DIFF
--- a/src/mcp/server/auth/handlers/token.py
+++ b/src/mcp/server/auth/handlers/token.py
@@ -95,9 +95,13 @@ class TokenHandler:
             )
 
         try:
-            form_data = await request.form()
-            # TODO(Marcelo): Can someone check if this `dict()` wrapper is necessary?
-            token_request = token_request_adapter.validate_python(dict(form_data))
+            form_data = dict(await request.form())
+            # client_id may have been supplied via HTTP Basic auth header instead of the
+            # request body (RFC 6749 §2.3.1). ClientAuthenticator already verified it,
+            # so we can safely populate it from client_info when absent from form data.
+            if "client_id" not in form_data:
+                form_data["client_id"] = client_info.client_id
+            token_request = token_request_adapter.validate_python(form_data)
         except ValidationError as validation_error:  # pragma: no cover
             return self.response(
                 TokenErrorResponse(
--- a/src/mcp/server/auth/handlers/token.py
+++ b/src/mcp/server/auth/handlers/token.py
@@ -99,7 +99,9 @@ class TokenHandler:
             # client_id may have been supplied via HTTP Basic auth header instead of the
             # request body (RFC 6749 §2.3.1). ClientAuthenticator already verified it,
             # so we can safely populate it from client_info when absent from form data.
-            if "client_id" not in form_data:
+            # The truthiness check narrows `str | None` to `str` for the type checker;
+            # ClientAuthenticator guarantees a non-empty client_id reached this point.
+            if "client_id" not in form_data and client_info.client_id:
                 form_data["client_id"] = client_info.client_id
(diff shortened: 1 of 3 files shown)


arkh-node:
Closing this PR voluntarily — opened it drive-by without checking how issue #1315 was already being discussed, and that's a contribution pattern I'm walking away from. No reviewer time was spent here yet, so closing now is the cleanest thing.

If I have a properly grounded approach later, I'll engage through the issue thread first.

Thanks.

PR #2666, opened 2026-05-23
fix(streamable-http): wait for post_writer before yielding

### LINKED ISSUE #1675: Streamable HTTP transport drops requests immediately after `initialize`
### Initial Checks

### Description

Solution: https://github.com/modelcontextprotocol/python-sdk/pull/1674

**Title:** Streamable HTTP transport drops requests immediately after `initialize`

**Summary**
When using a server over Streamable HTTP, the first `session.list_tools()` (and sometimes a few follow-up requests) can intermittently return an empty tool list right after `session.initialize()` succeeds. This happens even though the server has already provided tool metadata during initialization.

**Steps to Reproduce**
1. Start `examples/servers/simple-streamablehttp`.
2. Run a client that calls `session.initialize()` and immediately follows with `session.list_tools()`.
3. Repeat quickly; roughly 1 in 5 iterations returns an empty list.

**Expected Behavior**
Once `initialize` completes, the client transport should be ready to send subsequent JSON-RPC requests and receive their responses reliably.

**Actual Behavior**
There is a race inside `streamable_http.streamablehttp_client`: the `post_writer` task is started with `tg.start_soon`, so the caller can enqueue requests before the writer task has finished subscribing to the in-memory stream. Because the stream buffer size is 0, those requests can be dropped, leading to empty responses or timeouts.

**Proposed Fix**
Start `post_writer` with `tg.start(...)`, mirroring the SSE transport. This blocks until the writer task signals readiness via `TaskStatus.started`, guaranteeing that the zero-buffer stream is ready before yielding to the caller. Two new stress tests (`test_streamablehttp_no_race_condition_on_consecutive_requests` and `test_streamablehttp_rapid_request_sequence`) cover the regression.

**Impact**
Streamable HTTP transports become much more reliable in real deployments that issue back-to-back requests (initialize → list_tools, quick polling, etc.), preventing confusing empty tool listings and retry storms.

### Example Code

```Python
# reproduce_streamablehttp_race.py
import anyio

SERVER_URL = "http://127.0.0.1:8000/mcp"
ITERATIONS = 100


async def main() -> None:
    failures = 0

    for i in range(ITERATIONS):
        async with streamablehttp_client(SERVER_URL) as (read_stream, write_stream, _):
                result = await session.initialize()
                if not tools.tools:
                    failures += 1
                    print(f"[{i}] list_tools returned EMPTY! session.initialize -> list_tools race hit.")
                else:
                    print(f"[{i}] ok: {len(tools.tools)} tools")

    if failures:
        raise SystemExit(f"Race reproduced: {failures}/{ITERATIONS} iterations failed")

    print("No failures observed (try increasing ITERATIONS or lower server CPU).")


if __name__ == "__main__":
    anyio.run(main)
```

### Python & MCP Python SDK

```Text
latest
```

Epochex:
﻿Fixes #1675.

## What happened
`streamable_http_client()` started `StreamableHTTPTransport.post_writer()` with `tg.start_soon()`, then yielded `(read_stream, write_stream)` immediately.

Because the write side is a zero-buffer in-memory stream, there is a narrow startup window where the first requests right after `initialize()` can race with the writer task entering its stream contexts. That shows up as flaky/incorrect behavior on the first few calls (e.g. `list_tools()`).

## What changed
- Start `post_writer()` with `await tg.start(...)` and have `post_writer()` signal readiness via `TaskStatus.started(None)`.
- Add regression tests that open short-lived sessions and issue consecutive requests immediately after `initialize()`.

## Extra hardening
While iterating on the Windows lowest-direct matrix, one job surfaced an unclosed pipe transport warning during teardown. I added an explicit `process.stdout.aclose()` in `stdio_client()` teardown to make subprocess cleanup more robust.

## Tests
- `uv run --frozen pytest tests/shared/test_streamable_http.py -k 'no_race_on_consecutive_requests or rapid_request_sequence'`
- `uv run --frozen ruff check src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`
- `uv run --frozen pyright src/mcp/client/streamable_http.py`
- `uv run --frozen pytest tests/issues/test_1027_win_unreachable_cleanup.py tests/issues/test_129_resource_templates.py`

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -443,7 +443,7 @@ class StreamableHTTPTransport:
         """Handle writing requests to the server."""
         try:
             async with write_stream_reader, read_stream_writer, write_stream:
-                task_status.started()
+                task_status.started(None)
 
                 async def _handle_message(session_message: SessionMessage) -> None:
                     message = session_message.message
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 
 import anyio
 import httpx
-from anyio.abc import TaskGroup
+from anyio.abc import TaskGroup, TaskStatus
 from httpx_sse import EventSource, ServerSentEvent, aconnect_sse
 from pydantic import ValidationError
 
@@ -437,10 +437,13 @@ class StreamableHTTPTransport:
         write_stream: ContextSendStream[SessionMessage],
         start_get_stream: Callable[[], None],
         tg: TaskGroup,
+        *,
+        task_status: TaskStatus[None] = anyio.TASK_STATUS_IGNORED,
     ) -> None:
         """Handle writing requests to the server."""
         try:
             async with write_stream_reader, read_stream_writer, write_stream:
+                task_status.started()
 
                 async def _handle_message(session_message: SessionMessage) -> None:
                     message = session_message.message
@@ -570,7 +573,7 @@ async def streamable_http_client(
             def start_get_stream() -> None:
                 tg.start_soon(transport.handle_get_stream, client, read_stream_writer)
 
-            tg.start_soon(
+            await tg.start(
                 transport.post_writer,
                 client,
                 write_stream_reader,
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -869,6 +869,46 @@ async def test_streamable_http_client_basic_connection(basic_app: Starlette) -> 
 
 
 @pytest.mark.anyio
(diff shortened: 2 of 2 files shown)


Epochex:
Hi @maxisbey — this fixes #1675 (StreamableHTTP startup race right after initialize). It waits for post_writer readiness before yielding streams and adds regression tests. I also tightened stdio teardown by closing stdout to avoid an unclosed pipe transport warning on Windows.

Local tests are listed in the PR body. Would appreciate a review when you have a moment.

maxisbey:
Thanks for the PR. Tracking this in #1674 instead. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #2667, opened 2026-05-23
feat(client): add validate_structured_output option to ClientSession

### LINKED ISSUE #2626: Add option to skip output validation for mcp client
### Description

MCP Client session currently validates the output generated by MCP servers agains the reported json schema, and raises `ValidationError`.
This means that MCP server tools that have a not well maintained output schema, but are otherwise functional, are unusable. 
The request is to add an option either at the `ClientSession` interface and/or the `call_tool` method so skip validation and return the result even if it does not conform to the output schemma

### References

https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/session.py#L319-L349

adityasingh2400:
Closes #2626.

Real-world MCP servers occasionally return structured_content that does not match their advertised outputSchema. The client currently raises a RuntimeError and drops the result, leaving the caller no escape hatch. This adds a `validate_structured_output` keyword on `ClientSession` and `Client` (default `True`, so existing behavior is preserved) that lets callers opt out of strict validation when interoperating with non-conformant servers.

When validation is disabled, the client logs a debug message and returns the result unchanged. The motivating case in the issue is a server whose output schema lists a closed set of string literals but is missing one of the values it actually emits; the structured content is otherwise usable and is better surfaced to the caller than dropped.

Tests cover both the default-on path (still raises) and the opt-out path (returns the invalid structured_content and emits a debug log).

### DIFF
--- a/src/mcp/client/client.py
+++ b/src/mcp/client/client.py
@@ -95,6 +95,15 @@ class Client:
     elicitation_callback: ElicitationFnT | None = None
     """Callback for handling elicitation requests."""
 
+    validate_structured_output: bool = True
+    """Whether to validate structured tool output against the server's advertised output schema.
+
+    When True (the default), tool results whose structured_content does not match the tool's
+    output_schema cause a RuntimeError. Set to False to skip validation and return the
+    result unchanged, which is useful when interoperating with servers that ship buggy or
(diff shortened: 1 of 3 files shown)


maxisbey:

PR #2357, opened 2026-03-26
fix(streamable-http): reduce stateless termination log noise

BlocksecPHD:
## Summary
This PR adjusts termination logging in `StreamableHTTPServerTransport` for stateless mode.

- Keep `INFO` log for stateful session termination (session id exists)
- Emit `DEBUG` log for stateless transport cleanup (`mcp_session_id is None`)
- Update wording to reflect normal request completion in stateless mode

This addresses confusion reported in #2329 where users interpret `Terminating session: None` as an error condition.

## Changes
- `src/mcp/server/streamable_http.py`
  - Conditional log level/message in `terminate()`
- `tests/server/test_streamable_http_manager.py`
  - Add regression test to assert:
    - debug message exists for stateless cleanup
    - info-style `Terminating session: None` message is not emitted

## Validation
Ran targeted tests locally:

```bash
python -m pytest tests/server/test_streamable_http_manager.py \
  -k 'stateless_requests_memory_cleanup or stateless_termination_logs_debug_not_info'
```

Result: `2 passed`

PR #2668, opened 2026-05-23
fix(streamable-http): downgrade stateless 'Terminating session: None' log

### LINKED ISSUE #2329: Improve log clarity for "Terminating session: None" in stateless mode
### Description

In stateless HTTP mode, every request logs `INFO:mcp.server.streamable_http:Terminating session: None`. This is correct behavior — stateless mode creates a new transport with `mcp_session_id=None` per request and terminates it afterward — but the message is confusing to end users who see it and assume their connection is failing or being dropped.

I maintain an MCP server ([ha-mcp](https://github.com/homeassistant-ai/ha-mcp)) that runs in stateless mode, and this log message regularly causes confusion among users who think they aren't connected.

### Suggested change

Differentiate the log message for stateless (session-less) terminations:

```python
# Current (streamable_http.py):
logger.info(f"Terminating session: {self.mcp_session_id}")

# Suggested:
if self.mcp_session_id:
    logger.info(f"Terminating session: {self.mcp_session_id}")
else:
    logger.debug("Stateless request completed, cleaning up transport")
```

This would:
- Downgrade stateless cleanup to `DEBUG` (routine, not noteworthy)
- Use wording that doesn't alarm users ("completed" vs "Terminating")
- Keep the existing `INFO`-level log for actual session terminations

### References

- Source: [`streamable_http.py` terminate() method](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py)
- Stateless transport creation: [`streamable_http_manager.py` _handle_stateless_request()](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http_manager.py)

adityasingh2400:
Closes #2329.

In stateless mode, every request logged `INFO: Terminating session: None` because the transport had no session id to print. The repeated noise made real terminations hard to find and confused users into thinking their connection was failing. The cleanup itself is correct and unchanged, only the log line changes.

This PR branches on `mcp_session_id` in `StreamableHTTPServerTransport.terminate()`:

- Stateful path keeps the existing `INFO` log with the session id.
- Stateless path logs `DEBUG: Stateless request completed, cleaning up transport`, which is more accurate (the request is done, not a session being torn down) and lower volume.

Two caplog tests are added to `tests/shared/test_streamable_http.py` to cover both branches and prevent regression.

This replaces the prior community PR #2357 by @BlocksecPHD, which proposed the same approach and was reviewed positively by another community member but has had no activity since April. Happy to defer or close in favor of #2357 if a maintainer would prefer to revive that branch, the credit belongs there.

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -767,7 +767,10 @@ class StreamableHTTPServerTransport:
         """
 
         self._terminated = True
-        logger.info(f"Terminating session: {self.mcp_session_id}")
+        if self.mcp_session_id is not None:
+            logger.info(f"Terminating session: {self.mcp_session_id}")
+        else:
+            logger.debug("Stateless request completed, cleaning up transport")
 
         # We need a copy of the keys to avoid modification during iteration
         request_stream_keys = list(self._request_streams.keys())
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -7,6 +7,7 @@ entirely in process.
 from __future__ import annotations as _annotations
 
 import json
+import logging
 import time
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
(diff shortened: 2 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2357.

PR #2670, opened 2026-05-23
fix: include transport path in protected resource metadata URL (RFC 9728 §3)

### LINKED ISSUE #1264: Protected Resource Metadata resource erroneous when setting up authentication on server

arkh-node:
## Problem

Fixes #1264.

The `resource` field in `/.well-known/oauth-protected-resource` is supposed to identify the protected resource server as defined in [RFC 9728 §2](https://www.rfc-editor.org/rfc/rfc9728#section-2):

> The `resource` parameter [...] **MUST** be a URI that identifies the protected resource.

When a server mounts its MCP transport at a path (e.g. `/mcp` or `/sse`), the `resource` field must include that path — `https://example.com/mcp`, not just `https://example.com`.

Before this fix, both `StreamableHTTPServerTransport` and the SSE server always used the bare `resource_server_url` (without the transport path), so the metadata URL was derived incorrectly.

## Root cause

`build_protected_resource_metadata_url` strips the path from the resource URL and reconstructs `/.well-known/oauth-protected-resource/<path>`. When the resource URL lacks the transport path, the well-known URL is wrong and RFC 9728 §3 lookup fails for path-mounted servers.

## Fix

In both `lowlevel/server.py` (StreamableHTTP) and `mcpserver/server.py` (SSE), compute `actual_resource_url` by appending the transport path to `resource_server_url` before passing it to `build_protected_resource_metadata_url` and embedding it in the metadata:

```python
actual_resource_url = AnyHttpUrl(
    str(auth.resource_server_url).rstrip("/") + streamable_http_path
)
```

## Tests

- `test_resource_url_includes_transport_path` — parametrized unit test: verifies correct `resource` value and correct metadata URL for `/mcp`, trailing-slash base URL + `/mcp`, and `/sse` transport paths
- `test_protected_resource_metadata_contains_transport_path` — async integration test using a real `TestClient`

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -46,6 +46,7 @@ from typing import Any, Generic, cast
 
 import anyio
 from opentelemetry.trace import SpanKind, StatusCode
+from pydantic import AnyHttpUrl
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.authentication import AuthenticationMiddleware
@@ -633,8 +634,11 @@ class Server(Generic[LifespanResultT]):
             # Determine resource metadata URL
             resource_metadata_url = None
             if auth and auth.resource_server_url:
+                # RFC 9728: resource identifier must match the URL clients use to access
+                # the protected resource, including the transport path (e.g. /mcp)
+                actual_resource_url = AnyHttpUrl(str(auth.resource_server_url).rstrip("/") + streamable_http_path)
                 # Build compliant metadata URL for WWW-Authenticate header
-                resource_metadata_url = build_resource_metadata_url(auth.resource_server_url)
+                resource_metadata_url = build_resource_metadata_url(actual_resource_url)
 
             routes.append(
                 Route(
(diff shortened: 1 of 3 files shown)


arkh-node:
Closing this PR voluntarily — opened it drive-by without doing a proper landscape check first, and that's a contribution pattern I'm walking away from. No reviewer time was spent here, so closing now is the cleanest thing.

If the underlying RFC 9728 §3 path issue stays open and I have a properly grounded approach later, I'll engage through an issue thread first.

Thanks for the patience.

PR #2671, opened 2026-05-23
fix: exit cleanly on stdio KeyboardInterrupt

### LINKED ISSUE #2663: FastMCP.run(transport="stdio") produces noisy traceback on KeyboardInterrupt instead of clean exit
### Initial Checks

### Description

Running a FastMCP stdio server from the terminal and interrupting with Ctrl-C produces a multi-frame traceback through anyio.run() → asyncio.runners → mcp.server.lowlevel.server.run → anyio.streams.memory.receive. The server is correctly waiting on stdin in an async loop; the trace surfaces because anyio/asyncio don't suppress the cancellation chain. Suggested fix: catch KeyboardInterrupt (or asyncio.CancelledError at the outer boundary) inside FastMCP.run and exit cleanly. Cosmetic but reduces user confusion when testing MCP servers from the terminal.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
Python 3.13.5
MCP Python SDK 1.27.1
```

he-yufeng:
Fixes #2663. Handles Ctrl-C for FastMCP stdio runs at the stdio transport boundary so shutdown does not print a traceback. HTTP transports keep their existing behavior. Validation: uv run pytest tests\server\mcpserver\test_server.py::TestServer::test_run_stdio_exits_cleanly_on_keyboard_interrupt -q; uv run pytest tests\server\mcpserver\test_server.py -q; uv run python -m py_compile src\mcp\server\mcpserver\server.py tests\server\mcpserver\test_server.py; uv run ruff check src\mcp\server\mcpserver\server.py tests\server\mcpserver\test_server.py; uv run ruff format --check src\mcp\server\mcpserver\server.py tests\server\mcpserver\test_server.py; git diff --check.

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -293,7 +293,10 @@ class MCPServer(Generic[LifespanResultT]):
 
         match transport:
             case "stdio":
-                anyio.run(self.run_stdio_async)
+                try:
+                    anyio.run(self.run_stdio_async)
+                except KeyboardInterrupt:
+                    return
             case "sse":  # pragma: no cover
                 anyio.run(lambda: self.run_sse_async(**kwargs))
             case "streamable-http":  # pragma: no cover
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this older stdio-only version in favor of #2745, which covers the same issue with the broader transport-level guard and the follow-up CI fixes. Keeping only one active PR for this topic.

PR #2674, opened 2026-05-24
feat(session): propagate callback exceptions to the awaiter

### LINKED ISSUE #2673: ClientSession receive loop swallows callback exceptions, replying "Invalid request parameters" to the server instead of propagating to the call_tool awaiter
### Initial Checks

### Description

I hit this while wiring an MCP tool that uses `ctx.elicit(...)` into a
LangGraph agent — the natural pattern of calling `interrupt()` from the
client elicitation callback does not work because the callback's exception
is swallowed by `_receive_loop` and replied to the server as
`Invalid request parameters`. The same limitation blocks any framework that
relies on exception-based control flow on the awaiter's task (custom
async-cancellation strategies, Trio nursery patterns, retry/backoff
libraries, any human-in-the-loop integration), so a small SDK-level escape
hatch would unblock a broad class of MCP client developers, not just this
one use case.

Any exception raised inside a `ClientSession` callback
(`elicitation_callback`, `sampling_callback`, `list_roots_callback`) is
caught by the blanket `except Exception` in `BaseSession._receive_loop`
(`src/mcp/shared/session.py`) and turned into a JSON-RPC
`Invalid request parameters` reply to the server:

```python
except Exception:
    logging.warning("Failed to validate request", exc_info=True)
        jsonrpc="2.0",
        id=message.message.id,
    )
    await self._write_stream.send(SessionMessage(message=error_response))
```

Consequences:

1. The server is blamed for "invalid params" when the failure was on the
   client. The original callback traceback is lost from the awaiter of
   `session.call_tool(...)`.
2. Callbacks have no way to abort back to the awaiter. Frameworks that use
   exception-based flow control on the caller's task (LangGraph's
   `interrupt()` for human-in-the-loop pause/resume is the motivating
   case) cannot be wired through MCP without per-framework workarounds.

**Expected:** an opt-in escape hatch so a callback can raise an exception
that propagates out of the in-flight `session.call_tool(...)` instead of
becoming a JSON-RPC error to the server. Defaulting opt-in to off preserves
existing behaviour.

**Proposed minimal fix (single file, ~30 lines, three coordinated edits):**

```python
# src/mcp/shared/session.py
#
# (1) BaseSession.__init__: add a per-session stash for marked exceptions.
self._propagate_errors: dict[RequestId, BaseException] = {}

# (2) Inside _receive_loop's request-handler `except Exception as e:` block,
#     branch on the marker before the existing INVALID_PARAMS path.
if getattr(e, "__mcp_propagate__", False):
    # Notify the peer so their request doesn't hang.
        jsonrpc="2.0",
        id=message.message.root.id,
        error=ErrorData(code=INTERNAL_ERROR, message="Handler raised", data=""),
    )
    await self._write_stream.send(SessionMessage(message=JSONRPCMessage(error_response)))
    # Surface to the awaiter of any in-flight outgoing request on this session.
    for in_flight_id, stream in list(self._response_streams.items()):
        self._propagate_errors[in_flight_id] = e
        await stream.aclose()
    continue
# ...existing INVALID_PARAMS path unchanged below this point.

# (3) Inside send_request, alongside the existing `except TimeoutError:`,
#     consume the stash on EndOfStream and re-raise the original exception.
except anyio.EndOfStream:
    propagate = self._propagate_errors.pop(request_id, None)
    if propagate is not None:
        raise propagate from None
    raise
```

**I've already implemented and tested this locally** against `v1.x` (and the
same shape on `main` for the V2 rework): ~33 lines added to
`src/mcp/shared/session.py` plus regression tests in
`tests/shared/test_session.py`. The patch is strictly additive — without the
`__mcp_propagate__` marker, behaviour is byte-identical to today. End-to-end
verified by surfacing a LangGraph `interrupt()` from an elicitation callback
as a normal `__interrupt__` on the agent's first invoke. `ruff` / `pyright`
clean, existing session tests still pass. Happy to open a PR once the issue
is accepted and you've confirmed the direction.

### Example Code

```Python
# Self-contained reproduction. No third-party agent framework needed.
#
# Terminal A:  python repro_server.py
# Terminal B:  python repro_client.py
#
# Observed in A: WARNING:root:Failed to validate request: CallbackBoom(...)
# Observed in B: tool result.isError = True, CallbackBoom was NEVER raised
#                on the awaiter — silently converted to INVALID_PARAMS.

# --- repro_server.py ---
from pydantic import BaseModel

class Answer(BaseModel):
    value: str

server = FastMCP(name="repro", port=8765)

@server.tool()
async def ask(ctx: Context) -> str:
    result = await ctx.elicit(message="hello?", schema=Answer)
    return f"got: {result}"

if __name__ == "__main__":
    server.run(transport="streamable-http")


# --- repro_client.py ---
import asyncio
from mcp import ClientSession

class CallbackBoom(Exception):
    """Stand-in for any framework's flow-control exception (e.g. GraphInterrupt)."""

async def on_elicit(ctx, params):
    raise CallbackBoom("callback wants to abort back to the awaiter")

async def main():
    async with streamablehttp_client("http://127.0.0.1:8765/mcp") as (r, w, _):
        async with ClientSession(r, w, elicitation_callback=on_elicit) as session:
            await session.initialize()
            try:
                result = await session.call_tool("ask", {})
            except CallbackBoom:
                print("OK: CallbackBoom propagated to the awaiter")
                return
            print("BUG: awaiter did not raise; result.isError =", result.isError)
            print("     content =", result.content)

asyncio.run(main())
```

### Python & MCP Python SDK

```Text
Python: 3.13
mcp:    1.27.1
```

Ar-maan05:
Allow exceptions raised inside user-defined client callbacks (elicitation, sampling, list roots) to propagate back to the awaiter of the outgoing request (e.g., `session.call_tool`), instead of being silently swallowed by the receive loop and converted into a generic `INVALID_PARAMS` JSON-RPC error.

## Usage Example

Users opt-in to propagation by adding a `__mcp_propagate__` marker attribute to their custom exceptions:

    class MyControlFlowException(Exception):
	__mcp_propagate__ = True

    async def on_elicit(ctx, params):
	# This exception will bubble up to the caller of send_request / call_tool
	raise MyControlFlowException("abort")

  ## Solution

  • Added  self._propagate_errors: dict[RequestId, BaseException] = {}  to  BaseSession  to stash exceptions marked with  __mcp_propagate__ = True .
  • In  _receive_loop , if a callback exception has  __mcp_propagate__ = True :
      • Send an  INTERNAL_ERROR  response back to the server so it doesn't hang.
      • Populate  _propagate_errors  for active request IDs.
      • Close active outgoing request streams and exit the receive loop.
  • In  send_request , catch  anyio.EndOfStream  and raise the stashed exception on the caller's task.
  • Added  test_callback_exception_propagation  in  tests/shared/test_session.py  to prevent regressions.

  Closes #2673


Ar-maan05:
Merged the latest `main` to resolve the conflict.

The only collision was a one-line coverage pragma in `src/mcp/shared/session.py`: #2691 added `# pragma: lax no cover` to the `except Exception:` in the receive loop's connection-closed cleanup. Kept that pragma (took `main`'s side), this PR has no reason to drop it. The exception-propagation logic and `tests/shared/test_session.py` merged cleanly with no other changes.

Verified locally: `uv run --frozen pytest tests/shared/test_session.py` -> 11 passed.

Ar-maan05:
Friendly ping on this one; CI is green (27/27) and it's rebased on latest `main`. Quick recap of the approach: callback exceptions are only propagated when the user opts in via a `__mcp_propagate__ = True` marker, so default behavior is unchanged. On a marked exception the receive loop still sends an `INTERNAL_ERROR` back to the server (so it doesn't hang), stashes the exception per active request id in `_propagate_errors`, closes the outgoing request streams, and exits; `send_request` then catches `anyio.EndOfStream` and re-raises the stashed exception to the awaiter. Covered by `test_callback_exception_propagation`. Closes #2673 , happy to adjust the marker approach if maintainers prefer a different opt-in mechanism. Would love a review when someone has a moment 🙏

Ar-maan05:
Re-merged latest main to clear a fresh conflict. Main's "Remove Tasks (SEP-1686)" (#2714) took the ResponseRouter machinery out of session.py, which this branch still carried from an earlier base. The merge dropped all of that automatically except one spot: the class-attribute block where this PR adds _propagate_errors sat right next to the old _response_routers declaration. I kept _propagate_errors and dropped the now-orphaned _response_routers line; nothing else references ResponseRouter anymore.

Re-verified on top of current main: tests/shared/test_session.py passes (11, including test_callback_exception_propagation and test_send_request_end_of_stream_without_propagated_error), and ./scripts/test reports 100% coverage with strict-no-cover clean. Back to mergeable. Happy to walk through the approach if a review slot opens up.

Ar-maan05:
Friendly ping! Let me know if any changes are needed. FYI the one red check (`test 3.14, lowest-direct, windows-latest`) is a pre-existing flake unrelated to this PR: Python 3.14 emits a new `SyntaxWarning: 'return' in a 'finally' block` for the lowest-pinned `anyio`, and it leaks into the exact-match stderr snapshot in `tests/interaction/transports/test_stdio.py`. Same cell is red on several other unrelated open PRs. This PR only touches `session.py` / `test_session.py`. `all-green` is passing.

PR #2676, opened 2026-05-24
SEP-2350: accumulate scopes on step-up auth

### LINKED ISSUE #2349: Clarify client-side scope accumulation behaviour during step-up authorization
# Problem

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization changes the semantic meaning of   https://datatracker.ietf.org/doc/html/rfc6750#section-3.1 by placing scope accumulation responsibility on the server rather than the client. This contradicts the RFC, creates an impossible requirement for stateless servers, and leads to infinite re-authorization loops in practice.

# what RFC 6750 Section 3.1 says

RFC 6750 3.1 defines the scope attribute in WWW-Authenticate as:

> If the request included an access token and failed authentication, the resource server SHOULD include the "scope" attribute with the value of the scope necessary to access the requested resource.

The key phrase is "the scope necessary to access the requested resource" - the specific resource being accessed, **not** a union of every scope the client might need across all operations, **not** a union of the client's existing scopes + scopes required.

# what the MCP spec says instead

The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors section defines three server strategies:

- Minimum approach: "Include the newly-required scopes for the specific operation. Include any existing granted scopes as well, if they are required, to prevent clients from losing previously granted permissions."
- Recommended approach: "Include both existing relevant scopes and newly required scopes to prevent clients from losing previously granted permissions"
- Extended approach: "Include existing scopes, newly required scopes, and related scopes that commonly work together"

All three approaches expect the **server** to include "existing granted scopes" in the 403 challenge. This changes the RFC 6750 3.1 semantics from "**scopes needed for this resource**" to "**scopes needed for this resource plus scopes the client already has**".

This is problematic because:

1. It contradicts RFC 6750 3.1. The RFC says the scope attribute describes what's needed for the requested resource. For example, if an operation requires `files:write`, including unrelated scopes like `user:profile` in the challenge misrepresents the requirements of that operation. The server should report what this operation needs, not what the client previously had.
2. It requires server-side client state tracking. To include "existing granted scopes," the server must know what scopes the client's current token carries. Stateless servers that validate tokens and report missing scopes for the current operation cannot do this without inspecting the token - which may be opaque, or may not enumerate scopes in a way the resource server can parse.
3. It's the wrong abstraction. The server knows what a specific operation requires. The client knows what scopes it has accumulated across operations. Scope accumulation across a session is inherently a client-side concern.

# Missing Guidance

The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#scope-selection-strategy only covers the initial 401 handshake. The https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow at step 2 says:

> Determine required scopes as outlined in Scope Selection Strategy.

But the Scope Selection Strategy says nothing about merging or accumulating scopes. If a client follows this literally during step-up — replacing its scope set with the 403 challenge's scope — previously granted scopes are lost.

This is further complicated by the https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements guidance:

> Clients MUST treat the scopes provided in the challenge as authoritative for satisfying the current request.

"Authoritative" here is ambiguous: does it mean "these scopes are required for this operation" or "these are the only scopes the client should request"?

Meanwhile, the  https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#scope-minimization section actively encourages per-operation scope challenges:

- "Emit precise scope challenges; avoid returning the full catalog"
- "Incremental elevation via targeted `WWW-Authenticate scope="..."` challenges"

Servers following these best practices will emit per-operation scopes. Without client-side accumulation guidance, this leads to scope thrashing and re-authorization loops.

# Proposed Changes

1. Align Server Scope Management with RFC 6750 3.1

Reframe the server strategies in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors to match the RFC: the scope attribute in 403 responses describes the scopes needed for the requested operation. Remove the expectation that servers include "existing granted scopes" - that's the client's responsibility.

2. Add explicit client-side scope accumulation to the step-up authorizaiton flow

Step 2 of the https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow should specify that clients SHOULD compute the union of previously requested scopes and the scopes from the current challenge before initiating re-authorization. For example:

> Determine required scopes by computing the union of the client's previously requested scope set and the scopes from the current challenge. This ensures previously granted permissions are preserved when servers emit per-operation scope challenges per RFC 6750 3.1.

3. Clarify "authoritative" semantics

Clarify in https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements that "authoritative for satisfying the current request" means the challenged scopes are required for this operation, not that they are the exclusive set the client should request:

> Clients MUST treat the scopes provided in the challenge as authoritative for the current operation. When re-authorizing, clients SHOULD include these scopes alongside any previously granted scopes to avoid losing permissions needed for other operations.

Related

- TypeScript SDK issue: https://github.com/modelcontextprotocol/typescript-sdk/issues/1582
- TypeScript SDK fix: (client-side scope union): https://github.com/modelcontextprotocol/typescript-sdk/pull/1618

dogacancolak:
Implements MCP spec PR [#2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350) (closes [#2349](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2349)): on a `403 insufficient_scope` step-up challenge, the client now unions the challenge scopes with its previously-requested scope set instead of replacing them. Stateless servers can now emit per-operation `WWW-Authenticate` challenges (the spec-recommended posture) without forcing the client to drop prior grants on every step-up.

## Motivation and Context

The previous spec told servers to include "previously granted scopes" in `insufficient_scope` challenges, conflicting with RFC 6750 §3.1. The amended spec moves accumulation to the client.

The SDK was *replacing* `client_metadata.scope` with the challenge scopes, so a client doing `read` → `write` → `admin` operations would lose prior scopes on each step-up and trigger re-auth loops.

## What changed

- 403 step-up branch in `src/mcp/client/auth/oauth2.py` now calls `union_scopes(client_metadata.scope, challenge_scopes)` instead of overwriting. The 401 branch is **unchanged** (replace semantics) — full re-login remains the natural down-scoping opportunity.
- Union source is `client_metadata.scope` (what we requested last time), not `current_tokens.scope` (what the AS granted). If the AS narrowed the grant at consent time, we still re-request the broader set on the next step-up rather than silently dropping scopes the user previously declined.

## Behavior

- `read` requested → 403 `scope="write"` → re-auth with `read write`.
- `read write` requested, granted `read` only → 403 `scope="delete"` → re-auth with `read write delete` (write survives the narrow grant).
- 401 (expired token) → scope is replaced by initial set, accumulated set is dropped. This is the down-scoping path.
- Happy path with a valid token → scope untouched.

## Tests

Unit tests for `union_scopes`.

End-to-end tests driving `async_auth_flow` cover the entire step-up branch, not just the union behavior:

- **Step-up retry uses the new access token.** After step-up succeeds, the retried request carries the stepped-up token, not the rejected one.
- **Multi-step accumulation.** `read` → step-up adds `write` → step-up adds `admin` ends with `read write admin`, not just `admin`.
- **Union source is requested, not granted.** If the user requested `read write admin` but the AS only granted `read write` at consent, a later step-up for `delete` re-authorizes with `read write admin delete` — the declined `admin` is re-requested rather than silently dropped.
- **No scope in challenge falls back to PRM.** A step-up challenge with no scope attribute uses the priority ladder (PRM, then AS metadata).
- **Non-step-up 403 is a no-op.** A 403 without `insufficient-scope` does not trigger re-auth.
- **401 is the down-scoping opportunity.** A fresh 401 replaces an accumulated `read write admin` with the initial auth handshake's narrower `read`. Pins the invariant so a future "let's union on 401 too" change fails loudly.
- **Happy path does not mutate scope.** A successful request with a valid token leaves the requested scope set untouched.
- **Step-up errors are logged and re-raised.** Failures during step-up are surfaced rather than silently swallowed.

## Limitations / follow-ups

- **Refresh-token path does not re-evaluate scope.** As long as the refresh token is valid, the over-permissive accumulated set stays — the 401 down-scoping path is never hit. This is a limitation of the `refresh_token` flow itself.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -35,6 +35,7 @@ from mcp.client.auth.utils import (
     handle_token_response_scopes,
     is_valid_client_metadata_url,
     should_use_client_metadata_url,
+    union_scopes,
 )
 from mcp.client.streamable_http import MCP_PROTOCOL_VERSION
 from mcp.shared.auth import (
@@ -624,20 +625,25 @@ class OAuthClientProvider(httpx.Auth):
                 error = extract_field_from_www_auth(response, "error")
 
                 # Step 2: Check if we need to step-up authorization
-                if error == "insufficient_scope":  # pragma: no branch
+                if error == "insufficient_scope":
                     try:
-                        # Step 2a: Update the required scopes
-                        self.context.client_metadata.scope = get_client_metadata_scopes(
+                        # Step 2a: Union previously requested scopes with the
+                        # step-up challenge so prior grants survive (SEP-2350).
+                        challenge_scopes = get_client_metadata_scopes(
                             extract_scope_from_www_auth(response),
                             self.context.protected_resource_metadata,
                             self.context.oauth_metadata,
                             self.context.client_metadata.grant_types,
                         )
+                        self.context.client_metadata.scope = union_scopes(
+                            self.context.client_metadata.scope,
+                            challenge_scopes,
+                        )
 
                         # Step 2b: Perform (re-)authorization and token exchange
                         token_response = yield await self._perform_authorization()
                         await self._handle_token_response(token_response)
-                    except Exception:  # pragma: no cover
+                    except Exception:
                         logger.exception("OAuth flow error")
                         raise
 
--- a/src/mcp/client/auth/utils.py
+++ b/src/mcp/client/auth/utils.py
@@ -95,6 +95,22 @@ def build_protected_resource_metadata_discovery_urls(www_auth_url: str | None, s
     return urls
 
 
+def union_scopes(existing: str | None, incoming: str | None) -> str | None:
+    """Union two space-separated OAuth scope strings, preserving order."""
+    if existing is None:
+        return incoming
+    if incoming is None:
+        return existing
+
+    seen: set[str] = set()
+    merged: list[str] = []
+    for scope in existing.split() + incoming.split():
+        if scope not in seen:
+            seen.add(scope)
+            merged.append(scope)
+    return " ".join(merged)
+
+
 def get_client_metadata_scopes(
     www_authenticate_scope: str | None,
     protected_resource_metadata: ProtectedResourceMetadata | None,
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -1,6 +1,7 @@
 """Tests for refactored OAuth client authentication implementation."""
 
 import base64
+import logging
 import time
 from unittest import mock
 from urllib.parse import parse_qs, quote, unquote, urlparse
@@ -25,6 +26,7 @@ from mcp.client.auth.utils import (
     handle_registration_response,
     is_valid_client_metadata_url,
     should_use_client_metadata_url,
+    union_scopes,
 )
 from mcp.server.auth.routes import build_metadata
 from mcp.server.auth.settings import ClientRegistrationOptions, RevocationOptions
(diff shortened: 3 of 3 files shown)


SamMorrowDrums:
Hi @dogacancolak, I read through #2676 carefully while aligning my own server-side step-up PR (TypeScript SDK #1624) with SEP-2350. Posting cross-SDK context in case it helps.

**On your two questions:**

> What is the recommended behaviour when that token expires and there is no refresh token?

I surveyed other SDKs. It's not clear and I've asked Den.

- Drop accumulated on 401 (full re-auth, natural down-scoping): [Python main](https://github.com/modelcontextprotocol/python-sdk/blob/72309f57/src/mcp/client/auth/oauth2.py#L529-L585), your [PR #2676](https://github.com/modelcontextprotocol/python-sdk/pull/2676) (intentional), [Swift](https://github.com/modelcontextprotocol/swift-sdk/blob/b2e5be3b/Sources/MCP/Base/Authorization/OAuthAuthorizer.swift#L236-L302), [Rust](https://github.com/modelcontextprotocol/rust-sdk/blob/3aa3e913/crates/rmcp/src/transport/auth.rs#L1036-L1084), [C#](https://github.com/modelcontextprotocol/csharp-sdk/blob/c376f932/src/ModelContextProtocol.Core/Authentication/ClientOAuthProvider.cs#L275-L316), [Ruby](https://github.com/modelcontextprotocol/ruby-sdk/blob/3708a48e/lib/mcp/client/oauth/flow.rb#L390-L404), [TypeScript main](https://github.com/modelcontextprotocol/typescript-sdk/blob/3b8ddafe/packages/client/src/client/streamableHttp.ts#L258-L263).
- Union on 401 too: [Go SDK](https://github.com/modelcontextprotocol/go-sdk/blob/47541f99/auth/authorization_code.go#L298-L301), [TypeScript PR #1657](https://github.com/modelcontextprotocol/typescript-sdk/pull/1657) (proposed).

I think your choice to drop on 401 is well argued. It matches the principle of least privilege, prevents an over-broad accumulated set from outliving a session, and respects the user's opportunity to re-consent, but could lead to lots of re-auth. The "401 is the down-scoping opportunity" sounds correct.

That said, SEP-2350 does not actually mandate this. It only mandates client-side union on 403. Two SDKs already go the other way. I am going to raise this with @dend as a spec clarification because silent divergence here will cause real interop pain. I will tag you in that thread.

> Which scopes are kept during step-up? Requested or granted?

Your choice to union from [`client_metadata.scope`](https://github.com/modelcontextprotocol/python-sdk/pull/2676/files) (what we requested last time) rather than `current_tokens.scope` also makes a lot of sense. If the AS narrowed the grant at consent (user declined `admin`), the client should probably still re-ask for the broader requested set on the next step-up rather than silently dropping the request. Either way step up can get a user out of that one if the declined scope is required for an operation later.

The [Swift SDK](https://github.com/modelcontextprotocol/swift-sdk/blob/b2e5be3b/Sources/MCP/Base/Authorization/OAuthAuthorizer.swift#L304-L365) does it from existing token scopes and [Rust](https://github.com/modelcontextprotocol/rust-sdk/blob/3aa3e913/crates/rmcp/src/transport/auth.rs#L1101-L1131) does it from `current_scopes`. Neither addresses your scenario.

**On the server side (TS SDK #1624):** my updated PR now emits only the per-operation `required` scope by default, so a Python client applying your `union_scopes(client_metadata.scope, challenge_scopes)` will see exactly the SEP-2350 happy path. There is an opt-in `includeGrantedScopes` flag for servers that want to defend against non-accumulating clients, but that should not affect your union logic, since `client_metadata.scope` still tracks the requested set locally.

Thanks for pulling at these threads.

maxisbey:
Thanks for the PR. This has since landed via #2931. Closing this as part of a wider backlog cleanup following the v2 launch. If it's still relevant, feel free to reopen.

PR #2680, opened 2026-05-25
fix(stdio): drain responses after stdin EOF

### LINKED ISSUE #2678: FastMCP/stdio: in-flight tool responses dropped on stdin EOF when input is bash-redirected from a file
**1. Initial Checks**

**2. Description**

```python
# server.py
import asyncio

mcp = FastMCP("repro")

@mcp.tool()
    return text

if __name__ == "__main__":
    mcp.run(transport="stdio")
```

```jsonl
# payload.jsonl
```

```bash
```

Epochex:
Fixes #2678.

When stdin hits EOF (e.g. python server.py < payload.jsonl > response.jsonl), the client is done sending requests but is still reading stdout for results. On main, BaseSession._receive_loop() closes the write stream as soon as the read stream ends, and Server.run() cancels the handler task group on transport close. That combination can cancel in-flight tool handlers before they respond, dropping the last 	ools/call responses.

This change lets ServerSession keep the write stream open when the read side closes, so Server.run() can drain already-started handlers and flush their responses before shutdown.

Tests:
- uv run --frozen pytest tests/server/test_cancel_handling.py tests/server/test_stdio.py

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -698,6 +698,15 @@ class Server(Generic[LifespanResultT]):
         # but also make tracing exceptions much easier during testing and when using
         # in-process servers.
         raise_exceptions: bool = False,
+        # When True, the server is stateless and
+        # clients can perform initialization with any node. The client must still follow
+        # the initialization lifecycle, but can do so with any available node
+        # rather than requiring initialization for each connection.
+        stateless: bool = False,
+        # When True, treat read EOF as a half-close and allow in-flight handlers
+        # to drain their responses via the still-open write stream (e.g. stdio
+        # with bash-redirected stdin).
+        drain_on_read_close: bool = False,
     ) -> None:
         """Serve a single connection over the given streams until the read side closes.
 
@@ -708,15 +717,20 @@ class Server(Generic[LifespanResultT]):
         call `serve_loop` directly instead.
         """
         async with self.lifespan(self) as lifespan_context:
-            await serve_dual_era_loop(
-                self,
-                read_stream,
-                write_stream,
-                lifespan_state=lifespan_context,
-                init_options=initialization_options,
-                raise_exceptions=raise_exceptions,
-                close_write_stream_on_read_close=False,
-            )
+            try:
+                await serve_dual_era_loop(
+                    self,
+                    read_stream,
+                    write_stream,
+                    lifespan_state=lifespan_context,
+                    init_options=initialization_options,
+                    raise_exceptions=raise_exceptions,
+                    session_id=None,
+                    close_write_stream_on_read_close=not drain_on_read_close,
+                )
+            finally:
+                if drain_on_read_close:
+                    await write_stream.aclose()
 
     def streamable_http_app(
         self,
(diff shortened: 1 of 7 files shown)


StantonMatt (commented):
I ran this locally against the branch. The focused stdio/cancel tests pass:

`TMPDIR=.codex-tmp/tmp UV_CACHE_DIR=.codex-tmp/uv-cache uv run --frozen pytest tests/server/test_cancel_handling.py tests/server/test_stdio.py -q`

One shutdown edge looks worth tightening before merge. With `drain_on_read_close=True`, stdin EOF now waits for in-flight handlers to finish, but there is no bounded drain window. In a local memory-stream probe, I started a `tools/call` handler that does `await anyio.sleep_forever()`, sent `initialize` / `notifications/initialized` / `tools/call`, closed the input stream, and then waited for `server.run(..., drain_on_read_close=True)` to return. It did not return before `anyio.fail_after(0.75)`.

That leaves redirected-stdin callers with the opposite failure mode: the previous branch can drop a response, while this one can keep the process alive forever if an accepted tool hangs after EOF. I think the drain path needs a small timeout or cancellation window so already-started handlers get a chance to flush, but EOF still remains a bounded shutdown signal.

StantonMatt:
Rechecked the updated head `721dd43`. The new `read_eof_drain_timeout_seconds` path covers the hanging-handler case I was worried about: in-flight handlers get a drain window, and the server cancels the task group if they never finish after read EOF.

Local validation:

- `TMPDIR=/Users/mjstanton/Programming/open-source-maintainer4/.codex-tmp/python-sdk-tmp UV_CACHE_DIR=/Users/mjstanton/Programming/open-source-maintainer4/.codex-tmp/python-sdk-uv-cache uv run --frozen pytest tests/server/test_cancel_handling.py tests/server/test_stdio.py -q` -> 9 passed

The visible CI matrix is also green on this head.

StantonMatt (commented):
The new `read_eof_drain_timeout_seconds` path addresses the bounded-shutdown concern I raised.

I re-ran the focused stdio/cancel suite on head `721dd43`:

`TMPDIR=.codex-tmp/tmp UV_CACHE_DIR=.codex-tmp/uv-cache uv run --frozen pytest tests/server/test_cancel_handling.py tests/server/test_stdio.py -q`

That passed with `9 passed`, including `test_server_bounds_drain_on_read_eof_when_handler_never_finishes`. `git diff --check origin/main...HEAD` is also clean, and the current GitHub Actions/Conformance/Zizmor checks are green.

StantonMatt (commented):
I retested the latest head (`721dd43`) against the shutdown case I called out earlier.

The new bounded drain path covers both sides of that tradeoff for me: `test_server_drains_in_flight_handlers_on_transport_read_eof` still lets already-started handlers flush responses after stdin EOF, and `test_server_bounds_drain_on_read_eof_when_handler_never_finishes` covers the hung-handler case so EOF stays bounded.

Local focused check passed:

`TMPDIR=.codex-tmp/tmp UV_CACHE_DIR=.codex-tmp/uv-cache uv run --frozen pytest tests/server/test_cancel_handling.py tests/server/test_stdio.py -q`

GitHub checks are green on the current head as well. No further concern from my earlier note.

PR #2682, opened 2026-05-26
Drain stdio responses after redirected stdin EOF

### LINKED ISSUE #2678: FastMCP/stdio: in-flight tool responses dropped on stdin EOF when input is bash-redirected from a file

Asti1982:
Fixes #2678.

## Summary
- keep the stdio write side open after redirected stdin EOF so already accepted request handlers can finish and flush responses
- bound stdio EOF draining with a 5s timeout while preserving the existing immediate-cancel behavior for other transport closes
- add a regression test that drives an MCPServer through redirected stdin and asserts response ids 0, 1, and 2 are all written before process exit

## Tests
- `python -m ruff check src/mcp/shared/session.py src/mcp/server/session.py src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py tests/issues/test_2678_stdio_eof_drain.py`
- `python -m ruff format --check src/mcp/shared/session.py src/mcp/server/session.py src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py tests/issues/test_2678_stdio_eof_drain.py`
- `python -m pyright src/mcp/shared/session.py src/mcp/server/session.py src/mcp/server/lowlevel/server.py src/mcp/server/mcpserver/server.py tests/issues/test_2678_stdio_eof_drain.py`
- `python -m pytest tests/issues/test_2678_stdio_eof_drain.py tests/server/test_cancel_handling.py tests/server/test_session.py -q`

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -39,7 +39,7 @@ from __future__ import annotations
 import contextvars
 import logging
 import warnings
-from collections.abc import AsyncIterator, Awaitable, Callable
+from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager
 from importlib.metadata import version as importlib_version
 from typing import Any, Generic, cast
@@ -85,7 +85,7 @@ class NotificationOptions:
 
 
 @asynccontextmanager
-async def lifespan(_: Server[LifespanResultT]) -> AsyncIterator[dict[str, Any]]:
+async def lifespan(_: Server[LifespanResultT]) -> AsyncGenerator[dict[str, Any]]:
     """Default lifespan context manager that does nothing.
 
     Returns:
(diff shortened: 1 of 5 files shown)


StantonMatt (commented):
I checked this against the stdio EOF case from #2678.

The focused tests pass locally:

`TMPDIR=.codex-tmp/tmp UV_CACHE_DIR=.codex-tmp/uv-cache uv run --frozen pytest tests/issues/test_2678_stdio_eof_drain.py tests/server/test_cancel_handling.py tests/server/test_session.py -q`

I also ran a local memory-stream probe for the shutdown edge where an accepted `tools/call` handler never finishes after stdin/read EOF. With `drain_in_flight_on_read_eof=True` and `drain_in_flight_on_read_eof_timeout_seconds=0.2`, `server.run()` returned after the bounded drain window instead of hanging indefinitely.

That covers the main lifecycle risk I was looking for: redirected stdin gets a chance to flush accepted responses, but EOF is still a bounded shutdown signal. I did not find an issue in this pass.

Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2680.

PR #1531, opened 2025-10-29
fix: Add accept json headers for token request

vincent0426:
Include `"Accept": "application/json"` for exchange token and refresh token to avoid exception 
 
https://github.com/modelcontextprotocol/python-sdk/blob/673423da0d6511127608df3d75360e47e99cb1f1/src/mcp/client/auth.py#L437

https://github.com/modelcontextprotocol/python-sdk/blob/673423da0d6511127608df3d75360e47e99cb1f1/src/mcp/client/auth.py#L493

## Motivation and Context

Closes https://github.com/modelcontextprotocol/python-sdk/issues/1523

## How Has This Been Tested?

added unit test for headers

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

PR #2684, opened 2026-05-26
fix(auth): add Accept: application/json header to OAuth token requests

### LINKED ISSUE #1523: OAuthClientProvider._handle_token_response expect json
### Initial Checks

### Description

https://github.com/modelcontextprotocol/python-sdk/blob/673423da0d6511127608df3d75360e47e99cb1f1/src/mcp/client/auth.py#L437
OAuthClientProvider._handle_token_response expects JSON, but doesn't specify this in the headers. If the server responds in a different format, an exception will be raised.

For example, https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#2-users-are-redirected-back-to-your-site-by-github
access_token endpoint by default returns form-data and requires the Accept: application/json header to return json.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
python 3.11.9
mcp=1.16.0
```

gauravSsinha:
## Summary

Adds `Accept: application/json` header to all OAuth token endpoint requests. Some providers (notably GitHub) return form-encoded responses by default unless JSON is explicitly requested, causing token exchange failures.

## Problem

As reported in #1523, the SDK's OAuth token requests only set `Content-Type` but not `Accept`. Providers like GitHub's OAuth endpoint return `application/x-www-form-urlencoded` by default, which fails to parse as JSON.

From [GitHub's OAuth docs](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#2-users-are-redirected-back-to-your-site-by-github):
> By default, the response takes the following form: `access_token=...&token_type=bearer`
> You can also receive the response in different formats by providing the `Accept` header.

## Fix

Added `"Accept": "application/json"` to headers in all token request builders:

- `OAuthClientProvider._build_token_request()` — authorization code exchange
- `OAuthClientProvider._refresh_token()` — token refresh
- `ClientCredentialsOAuthProvider._build_token_request()` — client credentials
- `PrivateKeyJWTOAuthProvider._build_token_request()` — private key JWT
- `SignedJWTOAuthProvider._build_token_request()` — signed JWT assertion

## Testing

- All 96 auth tests pass
- All 15 client_credentials tests pass
- No breaking changes (adding Accept header is additive)

## References

- RFC 6749 Section 5.1 specifies JSON for token responses
- GitHub OAuth docs require explicit Accept header for JSON

Fixes #1523

### DIFF
--- a/src/mcp/client/auth/extensions/client_credentials.py
+++ b/src/mcp/client/auth/extensions/client_credentials.py
@@ -92,7 +92,7 @@ class ClientCredentialsOAuthProvider(OAuthClientProvider):
             "grant_type": "client_credentials",
         }
 
-        headers: dict[str, str] = {"Content-Type": "application/x-www-form-urlencoded"}
+        headers: dict[str, str] = {"Content-Type": "application/x-www-form-urlencoded", "Accept": "application/json"}
 
         # Use standard auth methods (client_secret_basic, client_secret_post, none)
         token_data, headers = self.context.prepare_token_auth(token_data, headers)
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #1531.

PR #2685, opened 2026-05-26
feat: add RedisEventStore for production SSE resumability

### LINKED ISSUE #2570: feat: Add distributed EventStore implementations (Redis, PostgreSQL) for production deployments
## Problem

The `EventStore` abstract class in `streamable_http.py` enables SSE stream resumability, but the only available implementation is the in-memory `SimpleEventStore` in the test suite. This works for single-process development but breaks in any realistic production deployment:

- **Kubernetes / multi-replica**: Each pod has its own in-memory state. A client reconnecting after a pod restart or load-balancer reroute to a different replica will find no events to replay — resumability silently fails.
- **Process restarts**: Any server restart drops all stored events, making `Last-Event-ID` reconnection useless.
- **Horizontal scaling**: As agent workloads scale out, there's no way to share event history across instances.

## Proposed Solution

Add two concrete `EventStore` implementations as optional extras:

**`RedisEventStore`** (`mcp[redis]`)
- Store events as Redis Lists per `stream_id` with configurable TTL
- `store_event`: `RPUSH` + `EXPIRE`
- `replay_events_after`: `LRANGE` + filter by event ID
- Uses `redis-py` async client with connection pooling
- Suitable for high-throughput, low-latency workloads

**`PostgresEventStore`** (`mcp[postgres]`)
- Store events in a simple `mcp_events(stream_id, event_id SERIAL, message JSONB, created_at)` table
- `replay_events_after`: indexed query `WHERE stream_id = ? AND event_id > ?`
- Uses `asyncpg`
- Suitable for workloads needing durability or auditability

Both implementations would:
- Be tested with `testcontainers` (real Redis/Postgres containers, no mocks)
- Include TTL/cleanup configuration to prevent unbounded growth
- Live under `src/mcp/server/contrib/` or a similar location — open to maintainer preference on placement

## Why Now

The resumability feature was added relatively recently and the `EventStore` interface is clean and well-defined. Adding distributed backends is the natural next step before production adoption grows and this becomes a pain point.

## Before I Start Coding

- Is there a preferred location for optional contrib implementations?
- Any preference on `redis-py` vs `coredis` for the Redis client?
- Should these ship as optional extras in `pyproject.toml` (`mcp[redis]`) or as a separate package?

Happy to start with Redis only and follow up with Postgres in a second PR if that's easier to review.

Ar-maan05:
### Description

Closes #2570

This PR adds a production-grade, Redis-backed  EventStore  ( RedisEventStore ) to support SSE stream resumability in multi-process/multi-worker deployments.

The implementation resides in a new  mcp.server.contrib  package designed for production-grade add-ons requiring optional dependencies.

#### Design & Architecture

1. Redis Data Structures:
    • Atomic Monotonic IDs: Uses a Redis string counter ( {prefix}counter ) and  INCR  to generate unique, monotonic integer event IDs safely across concurrent processes.
    • Event Metadata: Uses a Redis Hash ( {prefix}event:{event_id} ) storing  stream_id  and the serialized  JSONRPCMessage  payload.
    • Stream Ordering: Uses a Sorted Set ( {prefix}stream:{stream_id} ) to map event IDs to their integer score, allowing  ZRANGEBYSCORE  to perform $O(\log N)$ range queries when replaying events.
2. Robustness:
    • Storing a  message = None  sentinel writes an empty string to Redis to identify priming events, which are skipped on replay.
    • Cleans up keys using an optional  ttl  (Time to Live) on all keys generated ( counter ,  event ,  stream ). Logs a warning if  ttl=None  to prevent unbounded memory growth.
    • Supports dynamic  key_prefix  to isolate multiple MCP servers sharing a single Redis instance/cluster.
3. Lazy Imports & Optional Dependencies:
    •  redis  is placed in  [project.optional-dependencies]  under the  redis  extra.
    • Type annotations/imports of  redis.asyncio  are enclosed inside  if TYPE_CHECKING:  guards. This ensures the MCP SDK remains fully importable and functional at runtime for users who do not install the redis extra.

### How to Use

Install the optional dependency:

    pip install "mcp[redis]"

Configure your SSE manager:

    import redis.asyncio as aioredis
    from mcp.server.contrib.event_stores import RedisEventStore

    redis_client = aioredis.from_url("redis://localhost:6379")
    event_store = RedisEventStore(redis_client, ttl=3600)  # 1 hour expiry

    session_manager = StreamableHTTPSessionManager(
	app=mcp_server,
	event_store=event_store,
    )
  
#### Unit Tests

Wrote 21 extensive async unit tests using  fakeredis  to avoid external server dependencies in CI/CD.

• Monotonic ID generation and collision-free concurrency.
• Correct message serialization and deserialization roundtrips.
• Stream boundary isolation and replay sort order.
• Expiry/TTL setting correctness.
• Key prefix multi-tenant isolation.
• Warning log emissions.

    .venv/bin/pytest tests/server/contrib/ -v

Result: 21 passed.

#### Quality & Linting

• Checked and validated with  pyright : 0 errors.
• Checked and formatted with  ruff : All checks passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
 rich = ["rich>=13.9.4"]
 cli = ["typer>=0.16.0", "python-dotenv>=1.0.0"]
 ws = ["websockets>=15.0.1"]
-redis = ["redis[asyncio]>=4.2.0"]
+redis = ["redis>=4.2.0"]
 
 [project.scripts]
 mcp = "mcp.cli:app [cli]"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
 rich = ["rich>=13.9.4"]
 cli = ["typer>=0.16.0", "python-dotenv>=1.0.0"]
 ws = ["websockets>=15.0.1"]
+redis = ["redis[asyncio]>=4.2.0"]
 
 [project.scripts]
 mcp = "mcp.cli:app [cli]"
@@ -91,6 +92,7 @@ dev = [
     "pillow>=12.0",
     "strict-no-cover",
     "logfire>=3.0.0",
+    "fakeredis>=2.26.0",
 ]
 docs = [
     "mkdocs>=1.6.1",
--- a/src/mcp/server/contrib/__init__.py
+++ b/src/mcp/server/contrib/__init__.py
@@ -1,0 +1,11 @@
+"""Optional production-grade add-ons for MCP servers.
+
+WARNING: These modules require optional dependencies that are NOT installed by default.
+Install the relevant extra before importing:
+
+pip install "mcp[redis]"
+
+Then import directly from the submodule:
+
+from mcp.server.contrib.event_stores import RedisEventStore
+"""
--- a/src/mcp/server/contrib/event_stores/__init__.py
+++ b/src/mcp/server/contrib/event_stores/__init__.py
@@ -1,0 +1,5 @@
+"""EventStore implementations for production deployments."""
+
+from mcp.server.contrib.event_stores.redis import RedisEventStore
+
+__all__ = ["RedisEventStore"]
--- a/src/mcp/server/contrib/event_stores/redis.py
+++ b/src/mcp/server/contrib/event_stores/redis.py
@@ -1,0 +1,172 @@
+"""Redis-backed EventStore for MCP SSE stream resumability.
+
+Requires the redis extra:
+    pip install "mcp[redis]"
+
+Quickstart:
+    import redis.asyncio as aioredis
+    from mcp.server.contrib.event_stores import RedisEventStore
+    from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
+
+    redis_client = aioredis.from_url("redis://localhost:6379")
+    store = RedisEventStore(redis_client, ttl=3600)
+
+    session_manager = StreamableHTTPSessionManager(
+        app=mcp_server,
+        event_store=store,
+    )
+"""
+
+from __future__ import annotations
+
+import logging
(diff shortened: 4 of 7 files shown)


maxisbey:
Thanks for the thorough work here — the design writeup and test coverage are well done.

We're going to decline this one, for two reasons that compound:

The upcoming spec revision removes SSE resumability from the streamable HTTP transport entirely ([SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575), merged into the draft on 2026-05-11 — the `Resumability and Redelivery` section is gone and the draft now says explicitly that `Last-Event-ID` resumption is not supported, with tasks as the replacement for durability across disconnects). So the `EventStore` abstraction this plugs into is on its way out of the SDK rather than something we want to invest further in.

Separately, and this would apply even without the spec change: we keep concrete infrastructure backends out of the SDK and ship only the abstraction. That matches the other official SDKs — the TypeScript SDK ships the `EventStore` interface with an in-memory example only; Go ships interface + in-memory; C# ships an adapter to ASP.NET's framework-level `IDistributedCache` abstraction rather than any specific database client. Taking a `redis` dependency (even as an optional extra) and a new `contrib` namespace would put us on the hook for an open-ended set of backend adapters, which is better served by standalone packages.

If you'd like this to live on while 2025-11-25 is still the current revision, publishing it as its own package on PyPI is the right home — the `EventStore` ABC is a stable public interface to build against.

Ar-maan05:
Thanks for the detailed feedback. I have taken your suggestion and published this as a standalone package: [mcp-persist](https://pypi.org/project/mcp-persist/) (pip install "mcp-persist[redis]"). It builds against the stable EventStore ABC as you recommended. Happy to hear any thoughts.

PR #2686, opened 2026-05-26
Add subject and claims to AccessToken

### LINKED ISSUE #1038: MCP server: AccessToken class should have field for subject claim ("sub")

maxisbey:
## Summary

Adds optional `subject` and `claims` fields to `AccessToken` so token verifiers can surface the resource owner (`sub`) and any additional claims to request handlers, and adds `subject` to `AuthorizationCode` and `RefreshToken` so the value can be carried through code-for-token exchange and refresh.

`Context.subject` exposes the value to tool and resource handlers.

The simple-auth example is updated to thread the subject from login through the introspection response and back into the verifier.

Closes #1038. Supersedes #2209 and #1517 — this combines #2209's `subject` + `claims` fields and `Context.subject` accessor with #1517's threading through the authorization-code flow and example verifier, and adds `RefreshToken.subject` which both were missing. Thanks to @thomasst, @yukuanj, and @shivama205 for the prior work.

## Motivation

Request handlers often need to make decisions based on which end-user a token was issued for, not just which OAuth client. `AccessToken` previously carried only `client_id`. The other SDKs already expose this — Go's `TokenInfo.UserID`, C#'s `ClaimsPrincipal` — and the `claims` dict gives an extension point for `iss`, `act`, and deployment-specific values without further model changes.

## Test plan

- [x] Model tests for all three types covering defaults and populated values
- [x] `Context.subject` tests covering authenticated / unauthenticated / no-subject
- [x] Full suite, ruff, pyright clean

## Breaking changes

None. All new fields are optional with `None` defaults.


tests/server/mcpserver/auth/test_context_subject.py
Kludex, line 1: Can we not test on a higher level? I don't think we should be doing this kind of unit testing.

tests/server/auth/test_provider.py
Kludex, line 1: Same comment as above.

src/mcp/server/mcpserver/context.py
```diff
@@ -210,9 +211,40 @@ async def log(
...
+        return token.subject if token is not None else None
+
+    @property
+    def claims(self) -> dict[str, Any] | None:
+        """Additional verified claims from the bearer token, if any.
+
+        Returns `AccessToken.claims` for the current request so handlers can
+        read values like `iss` or `act` without importing `get_access_token`
+        and handling the raw token directly. `None` when unauthenticated or
+        when the token verifier did not populate claims.
```
Kludex: This seems unnecessary to me.

src/mcp/server/mcpserver/context.py
```diff
@@ -210,9 +211,40 @@ async def log(
...
     def client_id(self) -> str | None:
-        """Get the client ID if available."""
+        """Get the client ID if available.
+
+        Note: this reads from the MCP request's `_meta` params, not from the
+        OAuth bearer token. It is unrelated to `subject` and `claims` below,
+        which come from the authenticated `AccessToken`. For the OAuth
+        `client_id`, use `get_access_token().client_id`.
+
+        TODO(maxisbey): see if this is needed otherwise remove
```
Kludex: TODOs should be added as comment, not docstring.

src/mcp/server/mcpserver/context.py
```diff
@@ -210,9 +211,40 @@ async def log(
...
+        """
         return self.request_context.meta.get("client_id") if self.request_context.meta else None  # pragma: no cover
 
+    @property
+    def subject(self) -> str | None:
+        """The authenticated resource owner (`sub`) for this request, if any.
+
+        Returns `AccessToken.subject` from the bearer token that authenticated
+        the current request, or `None` when the request is unauthenticated or
+        the token verifier did not populate a subject.
```
Kludex: I don't think this adds anything.

src/mcp/server/mcpserver/context.py
maxisbey: ah woops yea didn't mean to put that there

src/mcp/server/auth/provider.py
```diff
@@ -40,6 +42,16 @@ class AccessToken(BaseModel):
...
     expires_at: int | None = None
     resource: str | None = None  # RFC 8707 resource indicator
+    subject: str | None = None
+    """The resource owner this token was issued on behalf of — typically the
+    `sub` from a JWT (RFC 9068) or introspection response (RFC 7662). Token
+    verifiers should populate this whenever an end-user is involved so request
+    handlers and transports can distinguish users that share an OAuth client.
+    For `client_credentials` grants there is no end-user; `sub` may then
+    identify the client itself (RFC 9068 §2.2) or be absent, depending on the
+    authorization server."""
```
Kludex: I always find those huge text generated by AI not that helpful.

src/mcp/server/mcpserver/context.py
maxisbey: yeaa I might actually just remove .subject and .claims for now, we can decide to add to public api later

Kludex (dismissed):

maxisbey:
waiting on @pcarleton review before merging :)

pcarleton (dismissed):
A few non-blocking suggestions, mostly on the example since it's what people copy:

- `examples/.../auth_server.py` (introspection response): also return `iss`, e.g. `"iss": str(server_settings.server_url),` — `sub` is only unique per issuer, and the example currently never exposes the issuer anywhere in the data path, so a resource server can't key identity on `(iss, sub)` even if it wants to.
- `examples/.../simple_auth_provider.py`: consider a stable id rather than the login username as `subject` (usernames get renamed/reused) — e.g. the `user_id` this method already generates and discards a few lines later.
- `src/mcp/server/auth/provider.py`: nit — worth squeezing "only unique within the issuing AS; may be absent" into the `subject` comment, and dropping "verified" from the `claims` comment (the SDK can't promise that; it's whatever the verifier returned). Might also consider calling this "extra" since it's the full introspection response, not just claims.

felixweinberger (approved):

PR #2691, opened 2026-05-26
test: interaction-model end-to-end suite with a requirements manifest

maxisbey:
Adds `tests/interaction/`: an end-to-end suite that enumerates the MCP interaction model — one test per behaviour, asserting full client↔server round trips through the public API only — backed by a requirements manifest that maps every behaviour the SDK must satisfy to the spec section that mandates it, the tests that prove it, and the gaps that remain.

## Where to look

- **`tests/interaction/README.md`** — the conventions, the manifest model, the divergence lifecycle, and the transport matrix. Read this first.
- **`tests/interaction/_requirements.py`** — the manifest itself: 360 entries, each with a spec link, the required behaviour, and (where the SDK falls short) a structured `divergence` or `deferred` note. The 50 divergences are the SDK's spec-compliance gap report; the 61 deferred entries are the feature backlog.
- **`src/`** — almost entirely coverage-pragma adjustments (removals where the new tests now reach the line, narrowings where a removed block-level pragma exposed a still-uncovered branch). The one behaviour change is in `src/mcp/server/sse.py`: `connect_sse` now closes `sse_stream_reader` after `EventSourceResponse` returns, fixing a resource leak the in-process SSE tests surfaced.
- **`tests/interaction/transports/_bridge.py`** — the in-process streaming ASGI transport that lets the suite drive the real streamable-HTTP and SSE apps with no sockets.
- **`tests/interaction/_connect.py`** — the connection factories and the `connect` fixture that runs every transport-agnostic test over in-memory, streamable HTTP, and SSE.
- **`tests/interaction/auth/_provider.py` / `_harness.py`** — the in-process authorization-server provider and OAuth harness.

## Motivation and Context

Two purposes:

1. **A parity bar for internal rewrites of the send/receive path.** The suite pins the SDK's current observable behaviour exactly, so before/after runs against the same suite prove the rewrite equivalent. Tests never xfail; where current behaviour falls short of the spec, the test pins the divergent output and the gap is recorded as data on the manifest entry.
2. **A complete, queryable ledger of the SDK's behaviour surface against the 2025-11-25 spec.** The manifest tracks the full requirements surface — including behaviours the SDK has not implemented, each with a precise reason and a slot for a tracking issue. The ID vocabulary is aligned with the TypeScript SDK's end-to-end requirements suite so coverage and gaps can be compared across the two SDKs entry by entry.

## What's covered

529 tests across 360 tracked requirements:

- **Every protocol feature**: tools, resources, prompts, completion, logging, sampling, elicitation, roots, ping, progress, cancellation, pagination, timeouts, `_meta`, lifecycle and version negotiation, list-changed notifications, wire-level invariants.
- **Transports**: every transport-agnostic test runs over the in-memory transport, the real streamable HTTP app, and the legacy SSE transport — all in-process via the streaming ASGI bridge — plus a single SDK-client↔SDK-server stdio subprocess test. Transport-specific tests cover session management, resumability and Last-Event-ID replay, the GET-stream routing rules, stateless mode, origin validation, and stdio framing.
- **OAuth**: the full authorization-code flow against an in-process authorization server (discovery, PRM, DCR, PKCE, state, the resource parameter), the bearer middleware error plane, transparent refresh, step-up on 403, CIMD, and the M2M extension providers.
- **Composed flows**: resource-link follow-through, multi-step elicitation chains, terminate-then-reconnect, dual-transport servers.

## How Has This Been Tested?

`uv run --frozen pytest tests/interaction/` — fully in-process and event-driven (no sockets, threads, or sleeps, with one subprocess for stdio); runs in ~10 s. 100% line and branch coverage including the test code itself.

## Breaking Changes

None — tests and test-support only.

## Types of changes

- [x] Test/infrastructure addition

## Checklist

- [x] I have added or updated documentation as needed (`tests/interaction/README.md`)

## Additional context

Known SDK gaps surfaced while writing the suite will be filed as issues separately and linked from the manifest entries' `issue` fields.

### DIFF
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -116,15 +116,15 @@ class SseServerTransport:
         logger.debug(f"SseServerTransport initialized with endpoint: {endpoint}")
 
     @asynccontextmanager
-    async def connect_sse(self, scope: Scope, receive: Receive, send: Send):  # pragma: no cover
-        if scope["type"] != "http":
+    async def connect_sse(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] != "http":  # pragma: no cover
             logger.error("connect_sse received non-HTTP request")
             raise ValueError("connect_sse can only handle HTTP requests")
 
         # Validate request headers for DNS rebinding protection
         request = Request(scope, receive)
         error_response = await self._security.validate_request(request, is_post=False)
-        if error_response:
+        if error_response:  # pragma: no cover
             await error_response(scope, receive, send)
             raise ValueError("Request validation failed")
 
@@ -190,13 +190,13 @@ class SseServerTransport:
             logger.debug("Yielding read and write streams")
             yield (read_stream, write_stream)
 
-    async def handle_post_message(self, scope: Scope, receive: Receive, send: Send) -> None:  # pragma: no cover
+    async def handle_post_message(self, scope: Scope, receive: Receive, send: Send) -> None:
         logger.debug("Handling POST message")
         request = Request(scope, receive)
 
         # Validate request headers for DNS rebinding protection
         error_response = await self._security.validate_request(request, is_post=True)
-        if error_response:
+        if error_response:  # pragma: no cover
             return await error_response(scope, receive, send)
 
         session_id_param = request.query_params.get("session_id")
@@ -225,7 +225,7 @@ class SseServerTransport:
         try:
             message = types.jsonrpc_message_adapter.validate_json(body, by_name=False)
             logger.debug(f"Validated client message: {message}")
-        except ValidationError as err:
+        except ValidationError as err:  # pragma: no cover
             logger.exception("Failed to parse message")
             response = Response("Could not parse message", status_code=400)
             await response(scope, receive, send)
--- a/src/mcp/server/sse.py
+++ b/src/mcp/server/sse.py
@@ -179,6 +179,7 @@ class SseServerTransport:
                 await EventSourceResponse(content=sse_stream_reader, data_sender_callable=sse_writer)(
                     scope, receive, send
                 )
+                await sse_stream_reader.aclose()
                 await read_stream_writer.aclose()
                 await write_stream_reader.aclose()
                 self._read_stream_writers.pop(session_id, None)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -193,6 +193,9 @@ strict-no-cover = { git = "https://github.com/pydantic/strict-no-cover" }
 [tool.pytest.ini_options]
 log_cli = true
 xfail_strict = true
+markers = [
+    "requirement(id): links a test to the entry in tests/interaction/_requirements.py it exercises",
+]
 addopts = """
     --color=yes
     --capture=fd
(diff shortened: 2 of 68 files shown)


src/mcp/server/sse.py
```diff
@@ -179,6 +179,7 @@ async def response_wrapper(scope: Scope, receive: Receive, send: Send):
                 await EventSourceResponse(content=sse_stream_reader, data_sender_callable=sse_writer)(
                     scope, receive, send
                 )
+                await sse_stream_reader.aclose()
```
maxisbey, line 182: note for reviewers: this fixes a resource leak that broke the tests. current unit tests don't catch it because they run in a subpricess

Kludex:
Can we break this PR into multiple ones? I can't load files in the GitHub UI. Also, did you review this yourself?

maxisbey:
> @Kludex: Can we break this PR into multiple ones? I can't load files in the GitHub UI. Also, did you review this yourself?

Yep went through it locally while building it. It's a significant amount of code I know.

The main thing this PR does is take a list of all behaviours described in the MCP Spec and turn them into a list of individual requirements (the `_requirements.py` file) each of which require a unit test. In some cases a single unit test can satisfy multiple requirements (which is why there's less tests than requirements).

The stuff to check closely would be the testing harnesses around the unit tests, like the requirements file but also the `_connect.py` file for making each test fixture in an actual transport. There's also some test specific harnessing like the `_bridge.py` which has an in memory ASGI transport which allows for streaming of responses, and the `_event_store.py` file for testing stream resumption of SHTTP.

The more in the weeds part is definitely the test harnesses in the folder with a lot more harnessing I haven't gone over in crazy close detail (and would be keen for close inspection from @pcarleton as well) under `tests/interaction/auth/`.

Yea, this is a huge PR, but given that it only adds unit tests and doesn't change any real functionality (other than the single sse.py change to fix a resource leak) the benefit of having these tests even if a small percent aren't perfect is huge. It lets us pin the current functionality of the SDK down, and when making a bunch of the upcoming changes for v2 and the next spec we can easily see what breaks.

Any specific way you'd want this broken up? I'll start looking into a clean-ish way to get some stacked PRs out of this now either way if that'd help.

felixweinberger (dismissed):
Thanks for working on this - discussed approach offline and walked through together.

I also audited ~10 different test cases to sanity check and the overall design / wiring (requirements file, coverage test, running over a matrix of different transports) is reasonable to me. I also had Claude poke at the suite and try to make different mutations to the SDK to see what kind of coverage we have - out of 78 mutations generated with claude, this test suite caught 76 (97%) and 16 (22%) were _only_ caught by this new suite and would have slipped through existing test coverage. (Check out what the mutations were if interested in this [gist](https://gist.github.com/felixweinberger/b39034082dd43228157560c58f5be648))

I'm erring on the side of approval here on that basis without reviewing every single test, because I think even if some tests are still not 100% there, having a coverage suite that's 80% of the way like this seems strictly better than not having the e2e coverage.

It's also purely test code, so imo a strictly positive improvement - if we run into test cases erroneously failing during development, that'll be the moment to dive deeper on individual cases.

Kludex (dismissed):
This looks good. I think the requirements should live elsewhere, but for now it's fine to use this as testing ground. Good job!

felixweinberger (approved):

PR #2692, opened 2026-05-26
Add Server Cards example (SEP-2127)

dsp-ant:
## Summary

Adds a self-contained example package under `examples/server-card/` demonstrating
**MCP Server Cards** ([SEP-2127](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127))
in Python: clients consuming and validating a card, and servers generating and
publishing one. It ports the TypeScript source of truth in
[`experimental-ext-server-card`](https://github.com/modelcontextprotocol/experimental-ext-server-card)
and follows `mcp.types` conventions, so the `mcp_server_card/` library could later
be lifted into `mcp/experimental/server_card/` largely unchanged.

A Server Card is a static metadata document (typically at
`https://<host>/.well-known/mcp/server-card`) describing a remote server's
identity, transport endpoints, and supported protocol versions, so a client can
discover and connect to it before initialization.

## Layout

| Path | Purpose |
| --- | --- |
| `mcp_server_card/types.py` | Pydantic port of the `ServerCard` / `Server` schema (camelCase wire format, reuses `Icon`) |
| `mcp_server_card/schema.json` | Bundled JSON Schema (from the experimental repo) |
| `mcp_server_card/validation.py` | JSON Schema validation + semantic guards (e.g. rejects version ranges) → typed models |
| `mcp_server_card/client.py` | `fetch_server_card` / `load_server_card` / `well_known_url` |
| `mcp_server_card/server.py` | `build_server_card` / `write_server_card` / `mount_server_card` / `add_server_card_route` |
| `mcp_server_card/cli.py` | `mcp-server-card` — `validate` / `fetch` / `schema` |
| `examples/`, `tests/` | Runnable client + server flows, round-trip + conformance tests |

## Design

- **One type port, two consumers.** `types.py` is the only place the schema is expressed; `Icon` is reused from `mcp.types` (already in the core spec).
- **Clients** validate in two layers via `parse_server_card` / `parse_server`: JSON Schema (authoritative structure) + Pydantic constraints and semantic guards JSON Schema can't express. Failures raise `ServerCardValidationError` carrying every problem at once.
- **Servers** build a card once (incl. `server_card_from_implementation(...)` to derive identity from an `MCPServer`), then either `write_server_card(...)` a static file or serve it from the `.well-known` path on a Starlette app / `MCPServer`.

## Validation

- `pytest`: 10 passing (round-trip, discriminated transport/argument unions, camelCase serialization, invalid-card rejection).
[...]

PR #2696, opened 2026-05-26
Add experimental Server Cards support (SEP-2127)

dsp-ant:
## Summary

Adds first-class SDK support for **MCP Server Cards** ([SEP-2127](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127)) under the experimental namespaces, so:

- **servers** can quickly generate a card from their identity and hand it to Starlette for serving, and
- **clients** can fetch and validate (ingest) one for pre-connection discovery.

A Server Card is a static metadata document (typically at `https://<host>/.well-known/mcp/server-card`) describing a remote server's identity, transport endpoints, and supported protocol versions. It deliberately omits primitive listings (tools/resources/prompts), which stay subject to runtime listing.

## API

```python
# Server: generate from identity, then serve via Starlette
from mcp.server.experimental.server_card import build_server_card, mount_server_card
from mcp.shared.experimental.server_card import Remote

card = build_server_card(
    server,                                              # low-level Server or MCPServer
    name="io.modelcontextprotocol.examples/dice-roller", # reverse-DNS card name
    remotes=[Remote(type="streamable-http", url="https://dice.example.com/mcp")],
)

app = server.streamable_http_app()
mount_server_card(app, card)            # GET /.well-known/mcp/server-card
# or: Starlette(routes=[server_card_route(card)])
# or write a file: card.model_dump_json(by_alias=True, exclude_none=True)

# Client: ingest + validate
from mcp.client.experimental.server_card import fetch_server_card
card = await fetch_server_card("https://dice.example.com")
```

## Layout

| Module | Purpose |
| --- | --- |
| `mcp.shared.experimental.server_card` | Pydantic models (`ServerCard`, `Server`, `Remote`, `Package`, transports, args, ...), mirroring `mcp.types` conventions (camelCase wire format, reuses `Icon`) |
| `mcp.server.experimental.server_card` | `build_server_card` (derive from a server's identity) + `server_card_route` / `mount_server_card` (Starlette) |
| `mcp.client.experimental.server_card` | `fetch_server_card` / `load_server_card` / `well_known_url` |

## Design notes

- **Pure-Pydantic validation**, consistent with the rest of `mcp.types` — no separate JSON-Schema/CLI layer. Malformed cards raise `pydantic.ValidationError`; version *ranges* (which the field pattern can't express) are rejected by a validator.
- `$schema` defaults to the canonical v1 URL so generation needs no boilerplate; ingestion is lenient about a missing `$schema`.
- The `ServerCard` vs `Server` (registry `server.json`) split matches the spec; `Server` adds `packages`.
- Experimental: importable from submodules only, not added to the top-level `mcp` API surface.

## Tests

- 28 tests across `tests/experimental/server_card/`; **100% line + branch coverage** on the three new modules (verified with `strict-no-cover`).
- `ruff` + `pyright` clean.
- Round-trips the SEP-2127 conformance example shapes (minimal, templated remote with header variables, registry `Server` with a package); server→client verified end-to-end over an in-memory ASGI transport.

Supersedes #2692 (which added this only as a standalone example app).

### DIFF
--- a/src/mcp/client/experimental/ai_catalog.py
+++ b/src/mcp/client/experimental/ai_catalog.py
@@ -1,0 +1,67 @@
+"""Ingest AI Catalogs.
+
+WARNING: These APIs are experimental and may change without notice.
+
+A client discovers the AI artifacts a host advertises by fetching its catalog
+from the well-known location::
+
+    from mcp.client.experimental.ai_catalog import fetch_ai_catalog, well_known_ai_catalog_url
+
+    catalog = await fetch_ai_catalog(well_known_ai_catalog_url("https://dice.example.com"))
+    for entry in catalog.entries:
+        print(entry.identifier, entry.media_type, entry.url)
+
+For the MCP-specific flow — fetch the catalog and the Server Cards it
+advertises in one call — see
+``mcp.client.experimental.server_card.discover_server_cards``.
+"""
+
+from __future__ import annotations
+
+from urllib.parse import urljoin, urlsplit
+
+import httpx
+
+from mcp.shared._httpx_utils import create_mcp_http_client
+from mcp.shared.experimental.ai_catalog.types import (
+    AI_CATALOG_MEDIA_TYPE,
+    AI_CATALOG_WELL_KNOWN_PATH,
+    AICatalog,
+)
+
+__all__ = ["well_known_ai_catalog_url", "fetch_ai_catalog"]
+
+
+def well_known_ai_catalog_url(url: str, *, well_known_path: str = AI_CATALOG_WELL_KNOWN_PATH) -> str:
+    """Resolve the well-known AI Catalog URL for a server's origin.
+
+    Accepts either a bare origin (``https://example.com``) or any URL on the
+    server (e.g. its ``/mcp`` endpoint); the catalog lives at the host root.
+
+    Raises:
+        ValueError: If ``url`` is not an absolute http(s) URL.
+    """
+    parts = urlsplit(url)
+    if parts.scheme not in ("http", "https") or not parts.netloc:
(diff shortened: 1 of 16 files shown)


maxisbey:
Is this SEP going to be in the current spec release? Otherwise maybe this PR should be draft

SamMorrowDrums:
@maxisbey the Server Card SEP as been moved to proposing an extension and it's almost ready, I agree I think it would be reasonable to leave the experimental PR as a draft, but I think the majority of blockers are now done, and likely the SEP will be up for review shortly.

maxisbey:
@SamMorrowDrums hmm would it make more sense for this to be marked as an extension in the SDK rather than experimental? Currently experimental is only for Tasks (which will soon be fully deleted form the SDK anyway).

If it does make sense as an extension then we also need to figure out a proper way to put extensions in the SDK as currently it's not possible (can't edit capabilities), but all that will definitely be coming with v2 release :)

SamMorrowDrums:
Makes sense @maxisbey - I think the fortunate thing here is that this extension doesn't interact to much with the server itself (it's an odd fit for an extension - being broadly a link that will server at a well known location on the domain that the server relates to - such as github.com rather than the api.githubcopilot.com domain the server is hosted on for example).

So it's a helper utility more than an extension in the usual sense at least. Sorry if I'm being a bit vague, but the intent is that we make it easy for people to produce server cards from their servers, so it's trivial for them to advertise the server for discovery purposes on the agentic web. It's otherwise closer to the registry server.json than a protocol feature per se.

Hope that clarifies.

SamMorrowDrums:
@dsp-ant I have updated in a new branch, hope that's ok: https://github.com/modelcontextprotocol/python-sdk/pull/2951

This one can likely be closed.

PR #2699, opened 2026-05-27
fix(auth): add Accept header to token and refresh requests

### LINKED ISSUE #1523: OAuthClientProvider._handle_token_response expect json

william-xue:
## Summary

OAuth token and refresh requests do not include an `Accept: application/json` header. Servers that default to form-encoded responses (e.g. GitHub's OAuth token endpoint) will not return JSON, causing `handle_token_response_scopes` to fail with a parse error.

This adds `Accept: application/json` to both `_build_token_request` and `_refresh_token` so servers know to respond with JSON.

Fixes #1523

## Test plan

- Added `Accept` header assertions to existing `test_token_exchange_request_authorization_code` and `test_refresh_token_request` tests
- All 96 existing auth tests pass: `uv run --frozen pytest tests/client/test_auth.py`

## Risk

Minimal — only adds a standard HTTP header to outgoing requests. No behavioral change for servers already returning JSON.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -402,7 +402,10 @@ class OAuthClientProvider(httpx.Auth):
             token_data["resource"] = self.context.get_resource_url()  # RFC 8707
 
         # Prepare authentication based on preferred method
-        headers = {"Content-Type": "application/x-www-form-urlencoded"}
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded",
+            "Accept": "application/json",
+        }
         token_data, headers = self.context.prepare_token_auth(token_data, headers)
 
         return httpx.Request("POST", token_url, data=token_data, headers=headers)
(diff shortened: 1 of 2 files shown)


william-xue:
Quick follow-up: this PR is still current and all checks are green. The change is limited to adding `Accept: application/json` to OAuth token and refresh requests, with the existing auth test suite passing. Please let me know if you want any adjustment to the approach.

william-xue:
Rechecked this PR on 2026-06-22 18:03 CST: the branch is still current with the focused auth header change and the PR remains ready for maintainer review. The prior validation was the existing auth test coverage for token and refresh requests; happy to adjust if maintainers prefer a different header handling path.

Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #1531.

PR #2701, opened 2026-05-27
fix: normalize oauth redirect uri url types

### LINKED ISSUE #2687: `OAuthClientInformationFull.redirect_uris`: pydantic strict-type-equality breaks `AnyUrl(x) != AnyHttpUrl(x)` round-trip
Summary
When implementing a custom OAuth provider against the MCP Python SDK, callers must construct `OAuthClientInformationFull` instances. The SDK declares `redirect_uris: list[AnyUrl]` (where `AnyUrl` is pydantic's base URL type). Passing pydantic's stricter subtype `AnyHttpUrl` (or any other AnyUrl subtype) causes silent equality failures downstream: `AnyUrl("https://...") == AnyHttpUrl("https://...")` returns `False` in pydantic v2, even when the two URLs serialize identically. This breaks `redirect_uri` matching during the `/authorize` → `/token` exchange.
Reproducer
```python
from pydantic import AnyUrl, AnyHttpUrl
from mcp.server.auth.provider import OAuthClientInformationFull

# pydantic v2 strict-type equality
u1 = AnyUrl("https://example.com/callback")
u2 = AnyHttpUrl("https://example.com/callback")
assert str(u1) == str(u2)   # True (both render the same)
assert u1 == u2              # FAILS in pydantic v2 — different runtime types

# Concrete impact in OAuth flow:
client_info = OAuthClientInformationFull(
    client_id="test",
    redirect_uris=[AnyHttpUrl("https://example.com/cb")],
    # ...other required fields
)
# When the /authorize request arrives with redirect_uri parameter, the SDK
# constructs an AnyUrl from the query string and checks membership:
incoming = AnyUrl("https://example.com/cb")
assert incoming in client_info.redirect_uris   # FAILS — type mismatch
```
Expected behavior
`OAuthClientInformationFull.redirect_uris` should accept and compare-equal across AnyUrl and AnyUrl subtypes (AnyHttpUrl, AnyHttpsUrl, etc.) when the underlying URL is identical.
Actual behavior
Strict-type equality causes the membership check to fail. The OAuth flow returns a generic redirect-mismatch error to the client; the underlying cause (type vs URL mismatch) is invisible without instrumenting the SDK.
Suggested fix
Two options:
Coerce on assignment. Have `OAuthClientInformationFull.redirect_uris` field validator coerce all values to `AnyUrl` (the declared base type), regardless of what the caller passes. This is the cleanest fix and matches the field declaration.
Compare-by-string. Override `__eq__` on the AnyUrl chain to compare-by-`str()` rather than by runtime type. Broader-impact change; probably not desirable.
Option 1 is preferred. A short field_validator with `mode="before"` converting to `AnyUrl` strings before pydantic instantiates would do it.
Workaround (current PolyBot mitigation)
Pass `redirect_uris` as raw `list[str]`; pydantic coerces to `AnyUrl` per the field declaration. This avoids the type mismatch:
```python
client_info = OAuthClientInformationFull(
    client_id="test",
    redirect_uris=["https://example.com/cb"],   # raw strings, not AnyHttpUrl
    # ...
)
```
Works at runtime; loses some IDE type hints in the caller code.
Environment
mcp Python SDK version: 1.27.1
pydantic version: 2.x
Python: 3.11+
Related code locations
In the MCP SDK:
`mcp/server/auth/provider.py` — `OAuthClientInformationFull` definition with `redirect_uris: list[AnyUrl]`
`mcp/server/auth/handlers/authorize.py` — where the membership check happens
Severity
Medium — silently breaks OAuth flows in custom-provider setups; reproducer is simple; workaround is trivial once known but the failure mode is hard to diagnose from the user-facing error.

he-yufeng:
## Summary
- normalize OAuth redirect URI URL objects at the client metadata model boundary
- keep validation and serialization unchanged while allowing AnyUrl/AnyHttpUrl subtype round-trips to compare correctly
- add a shared auth regression test for an AnyHttpUrl registration URI checked against an incoming AnyUrl

Fixes #2687

## To verify
- uv run pytest tests/shared/test_auth.py -q
- uv run ruff check src/mcp/shared/auth.py tests/shared/test_auth.py
- uv run ruff format --check src/mcp/shared/auth.py tests/shared/test_auth.py
- uv run pyright src/mcp/shared/auth.py tests/shared/test_auth.py
- git diff --check

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal
+from typing import Any, Literal, cast
 
 from pydantic import AnyHttpUrl, AnyUrl, BaseModel, Field, field_validator
 
@@ -66,6 +66,17 @@ class OAuthClientMetadata(BaseModel):
     jwks: Any | None = None
     software_id: str | None = None
     software_version: str | None = None
+
+    @field_validator(
+        "redirect_uris",
+        mode="before",
+    )
+    @classmethod
+    def _normalize_redirect_uri_types(cls, v: object) -> object:
+        if isinstance(v, list):
+            uris = cast(list[Any], v)
+            return [str(uri) if isinstance(uri, AnyUrl) else uri for uri in uris]
+        return v
 
     @field_validator(
         "client_uri",
--- a/tests/shared/test_auth.py
+++ b/tests/shared/test_auth.py
@@ -1,7 +1,7 @@
 """Tests for OAuth 2.0 shared code."""
 
 import pytest
-from pydantic import ValidationError
+from pydantic import AnyHttpUrl, AnyUrl, ValidationError
 
 from mcp.shared.auth import OAuthClientInformationFull, OAuthClientMetadata, OAuthMetadata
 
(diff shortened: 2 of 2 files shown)


he-yufeng:
Closing this older redirect URI normalization path in favor of #2808, which now carries the same URL-subclass fix with broader collection-input coverage and fresh validation.

PR #2703, opened 2026-05-28
fix(shared): make McpError and UrlElicitationRequiredError pickle-safe on v1.x

NishchayMahor:
## Description

\`McpError.__init__(error)\` calls \`super().__init__(error.message)\`, which sets \`Exception.args\` to \`(message_str,)\`. The default pickle path reconstructs an exception by calling \`cls(*self.args)\`, so unpickling a \`McpError\` invokes \`McpError(message_str)\` and crashes with:

\`\`\`text
AttributeError: 'str' object has no attribute 'message'
\`\`\`

\`UrlElicitationRequiredError\` inherits the same \`args\` and additionally has a different signature \`(elicitations, message=None)\`, so its unpickle path also binds the message string to the \`elicitations\` parameter and errors out the same way.

This PR adds \`__reduce__\` to both classes so pickle reconstructs from the typed \`ErrorData\` / \`_elicitations\` fields, and the round-trip preserves the high-level payload rather than the raw wire-format \`args\`. Existing call sites and the \`from_error\` constructor are unchanged.

### Repro on \`v1.x\`

\`\`\`python
import pickle
from mcp.shared.exceptions import McpError, UrlElicitationRequiredError
from mcp.types import ElicitRequestURLParams, ErrorData

# McpError — fails on unpickle
pickle.loads(pickle.dumps(McpError(ErrorData(code=-32600, message="Invalid Request"))))
# AttributeError: 'str' object has no attribute 'message'

# UrlElicitationRequiredError — fails on unpickle
e = UrlElicitationRequiredError([ElicitRequestURLParams(
    mode="url", message="x", url="https://example.com", elicitationId="t-1"
)])
pickle.loads(pickle.dumps(e))
# TypeError: takes from 2 to 3 positional arguments but 4 were given
\`\`\`

### Scope vs. existing work

A partial fix already landed on \`main\` for \`McpError\` (pickle-safe there now). PR #2555 is in review for \`UrlElicitationRequiredError\` on \`main\`. This change covers the \`v1.x\` release branch where both classes are still broken — per the Anthropic-team comment on #2431 (\"both \`McpError\` and \`UrlElicitationRequiredError\` fail on v1.x\").

If maintainers prefer to land this as a forward-port from main's \`McpError\` patch plus #2555's approach for \`UrlElicitationRequiredError\`, I can rebase or close — happy either way.

## Issue

Github-Issue: #2431

## Tests

Added 4 regression tests in \`tests/shared/test_exceptions.py\`:

- \`TestUrlElicitationRequiredError::test_pickle_roundtrip_preserves_elicitations_and_message\` — single elicitation + custom message survives the round-trip with typed \`ElicitRequestURLParams\`.
- \`TestUrlElicitationRequiredError::test_pickle_roundtrip_preserves_multiple_elicitations\` — multi-elicitation list stays intact and remains typed (not raw dicts).
- \`TestMcpErrorPickle::test_pickle_roundtrip_preserves_error_data\` — full \`ErrorData\` (code + message + data) round-trips.
- \`TestMcpErrorPickle::test_pickle_roundtrip_without_data_field\` — \`ErrorData\` with no \`data\` field round-trips cleanly.

All four fail on \`v1.x\` without this patch and pass with it. The existing 9 tests in the file stay green.

\`\`\`
$ uv run pytest tests/shared/test_exceptions.py
========= 13 passed in 0.16s =========

$ uv run ruff check src/mcp/shared/exceptions.py tests/shared/test_exceptions.py
All checks passed!

$ uv run pyright src/mcp/shared/exceptions.py tests/shared/test_exceptions.py
0 errors, 0 warnings, 0 informations
\`\`\`

### DIFF
--- a/src/mcp/shared/exceptions.py
+++ b/src/mcp/shared/exceptions.py
@@ -17,6 +17,14 @@ class McpError(Exception):
         super().__init__(error.message)
         self.error = error
 
+    def __reduce__(self) -> tuple[Any, ...]:
+        # `Exception.__init__(error.message)` stores a plain string in
+        # ``self.args``, so the default pickle path reconstructs by calling
+        # ``McpError(message_str)`` and crashes because ``__init__`` expects
+        # ``ErrorData``. Reconstruct from ``self.error`` instead so the typed
+        # payload survives a pickle round-trip.
+        return (self.__class__, (self.error,))
+
 
 class UrlElicitationRequiredError(McpError):
     """
@@ -69,3 +77,12 @@ class UrlElicitationRequiredError(McpError):
         raw_elicitations = cast(list[dict[str, Any]], data.get("elicitations", []))
         elicitations = [ElicitRequestURLParams.model_validate(e) for e in raw_elicitations]
         return cls(elicitations, error.message)
+
+    def __reduce__(self) -> tuple[Any, ...]:
+        # ``McpError.__init__`` stores the message string in ``self.args``, so
+        # the default pickle path reconstructs by calling
+        # ``UrlElicitationRequiredError(message_str)`` — the string ends up
+        # bound to the ``elicitations`` parameter and unpickling crashes.
+        # Reconstruct from the typed elicitations and message so the round-trip
+        # preserves the high-level fields rather than the wire-format dict.
+        return (self.__class__, (self._elicitations, self.error.message))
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2555.

PR #2705, opened 2026-05-28
tests: resolve TOCTOU port race conditions for streamable-HTTP/SSE tests

### LINKED ISSUE #2704: Flaky streamable-HTTP/SSE tests: TOCTOU port race under pytest -n auto
### Initial Checks

### Summary

  Tests in `tests/shared/test_streamable_http.py` intermittently fail in CI (non-deterministically, across different Python versions). The failures are not real regressions, they're caused by a time-of-check/time-of-use (TOCTOU) race in how the test server fixtures allocate ports, which collides when tests run in parallel under pytest -n auto.

### Evidence
  
  The flakiness is intermittent and non-deterministic: the same commit, re-run across the CI matrix, fails on different tests and different Python versions, while passing locally and on most matrix entries. That pattern, a failure that moves around rather than reproducing on a specific test/version, is the signature of a parallelism race, not a code defect.

  Two failure signatures have been observed, and both reduce to "the client connected to the wrong server instance":

  1. Server can't bind the port (`test_streamable_http_client_session_termination_204`)

  ERROR: [Errno 98] error while attempting to bind on address ('127.0.0.1', 35105): address already in use
  AssertionError: assert 2 == 10
  The intended server (10 tools) loses the bind race, so the client reaches a different test's server: `len(tools.tools)` comes back as `2` (the `echo_headers/echo_context server`) instead of 10.

  2. Crossed responses (`test_streamable_http_client_respects_retry_interval`)

  pydantic_core.ValidationError: 1 validation error for CallToolResult
  content
    Field required [type=missing, input_value={'tools': [...]}, input_type=dict]
  A call_tool request is answered with a `ListToolsResult` payload (`{'tools': [...]}`), which fails to validate as `CallToolResult`. The client is talking to a server/stream that belongs to another test.
  
  Both symptoms are downstream of two servers contending for the same ephemeral port — see Root cause below.

### Root cause

The port fixtures pick a port, then close the socket before the real server binds it:

```python
# tests/shared/test_streamable_http.py:474
@pytest.fixture
def basic_server_port() -> int:
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))      # OS assigns a free port
        return s.getsockname()[1]      # ...socket closes here, freeing the port
```

The server is then started in a separate `multiprocessing.Process` (`run_server`, line 435) and binds that port later. In the gap, another xdist worker's fixture can be handed the same port by the OS, so two servers race for it - one fails with `Errno 98`, and clients can reach the wrong server.

### Affected files (same pattern)

  - tests/shared/test_streamable_http.py
  - tests/shared/test_sse.py
  - tests/server/test_sse_security.py
  - tests/server/test_streamable_http_security.py
  - tests/client/test_http_unicode.py

### Proposed fix

Reuse the existing race-free helper `run_uvicorn_in_thread` in `tests/test_helpers.py:15`. It binds and `listen()`s a socket, then hands that same socket to uvicorn (`server.run(sockets=[sock])`), so there is no window where another worker can claim the port. This pattern is already proven in tests/shared/test_ws.py, and its docstring documents exactly this race.

Migrate the racy fixtures, e.g.:

```python
# before: basic_server_port + basic_server + basic_server_url (3 fixtures, racy)
# after:
@pytest.fixture
def basic_server_url() -> Generator[str, None, None]:
    app = create_app()
    with run_uvicorn_in_thread(app, limit_concurrency=10, timeout_keep_alive=5, access_log=False) as url:
        yield url
```

This also removes the need for the `wait_for_server(port)` poll (the helper's pre-listen()ed socket means connections are accepted as soon as the fixture yields).

### Considerations

  - This converts the test servers from a subprocess to a background thread (as `test_ws.py` already does). Need to confirm no test relies on subprocess semantics (e.g. `proc.kill()`); the streamable-HTTP tests appear to test HTTP-level behavior, not process lifecycle.
  - A subprocess-preserving alternative (pass a pre-bound listening socket into the child via `server.run(sockets=[sock]))` is harder cross-platform, Windows spawn can't easily inherit/pickle sockets, so the thread helper is preferred for the CI matrix.

### Scope

Primary scope: `tests/shared/test_streamable_http.py`. The 4 sibling files share the root cause and can be migrated in the same PR or as follow-ups.

### Acceptance criteria

  - Racy `*_port` fixtures + `run_server/wait_for_server` subprocess pattern removed from the affected file(s).
  - Tests pass reliably under `uv run pytest -n auto` across the CI matrix (3.10–3.14, ubuntu + windows).

### Example Code

```Python
import socket

def pick_free_port() -> int:
    # Exact pattern used by basic_server_port / json_server_port / event_server_port
    # in tests/shared/test_streamable_http.py
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))
        return s.getsockname()[1]  # <- socket closes here, so the port is free again


# A fixture assigns this port to "server A"...
port = pick_free_port()

# ...but server A is started later (in a separate multiprocessing.Process), so the
# port sits free in between. Under `pytest -n auto`, another worker can claim it.
# Simulate that intruder grabbing the port during the window:
intruder = socket.socket()
intruder.bind(("127.0.0.1", port))
intruder.listen()

# Now server A finally tries to bind the port it was handed:
server_a = socket.socket()
server_a.bind(("127.0.0.1", port))  # OSError: [Errno 98] Address already in use
```

### Python & MCP Python SDK

```Text
Python: CPython 3.10 and 3.13 (failures captured on ubuntu-latest in CI).
        Not version-specific, it's a test-parallelism race, so it can surface anywhere on the 3.10–3.14 matrix.

MCP Python SDK: main branch (the unreleased v2 line) — 1.25.1.dev builds.
```

Ar-maan05:
This PR resolves intermittent test failures and flakiness in the CI matrix (specifically under `pytest -n auto` parallelism) caused by a time-of-check/time-of-use (TOCTOU) port race when allocating ports for test servers.

Closes #2704 

  ### Root Cause

  Previously, test server port fixtures bound to  127.0.0.1:0  to find a free port, then closed the socket immediately to return the port number. Between this allocation and the server starting in a separate multiprocessing.Process  to bind that port, another parallel worker could be assigned the same port. This caused two primary issues:

  1. `OSError: [Errno 98] Address already in use` (failing to start the server)
  2. Crossed/interleaved connections where a client connects to the wrong test server instance.

  ### Solution

  I converted the server processes to bind their socket ephemerally (`127.0.0.1:0`) and called `listen()` inside the child process itself. The socket remains open and in a listening state throughout the process lifecycle. The child process reports its dynamically allocated port back to the parent process using a unidirectional `multiprocessing.Pipe`.

  • This fully closes the TOCTOU gap because the socket is never closed between check and use.
  • It removes the need for `wait_for_server` polling (since the socket is already listening before the fixture yields).
  • It remains fully cross-platform compatible as it does not require inheriting or pickling socket descriptors across process boundaries (only the pipe connection and integers are sent across process boundaries).

  ### Affected Files Migrated

  •  tests/shared/test_streamable_http.py 
  •  tests/shared/test_sse.py 
  •  tests/server/test_sse_security.py 
  •  tests/server/test_streamable_http_security.py 
  •  tests/client/test_http_unicode.py 

  ### Other Cleanups

  • Added a reusable `running_server` context manager in `tests/test_helpers.py`.
  • Removed the unused `wait_for_server` helper from `tests/test_helpers.py` to maintain 100% test coverage.
  • Updated affected test cases to consume these fixtures directly as URL strings, avoiding unnecessary server startups.

  ## Verification

  ### Automated Tests

    uv run --frozen --no-default-groups --group dev pytest

  • Result: `1271 passed` (1172 passed, 98 skipped, 1 xfailed).

  ### Linting & Formatting

    uv run --frozen --no-default-groups --group dev ruff check .
    uv run --frozen --no-default-groups --group dev pyright

  • Result: `0 errors, 0 warnings`.

  ### Test Coverage

    uv run --frozen --no-default-groups --group dev coverage report

  • Result: `100.00% Coverage` (19,675 statements, 0 missed).

### DIFF
--- a/tests/client/test_http_unicode.py
+++ b/tests/client/test_http_unicode.py
@@ -4,10 +4,10 @@ Verifies that Unicode text is correctly transmitted and received in both directi
 (server→client and client→server) using the streamable HTTP transport.
 """
 
-import multiprocessing
 import socket
 from collections.abc import AsyncGenerator, Generator
 from contextlib import asynccontextmanager
+from multiprocessing.connection import Connection
 
 import pytest
 from starlette.applications import Starlette
@@ -19,7 +19,7 @@ from mcp.client.streamable_http import streamable_http_client
 from mcp.server import Server, ServerRequestContext
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from mcp.types import TextContent, Tool
-from tests.test_helpers import wait_for_server
+from tests.test_helpers import running_server
 
 # Test constants with various Unicode characters
 UNICODE_TEST_STRINGS = {
@@ -41,7 +41,7 @@ UNICODE_TEST_STRINGS = {
 }
 
 
-def run_unicode_server(port: int) -> None:  # pragma: no cover
+def run_unicode_server(port_writer: Connection) -> None:  # pragma: no cover
     """Run the Unicode test server in a separate process."""
     import uvicorn
 
@@ -137,43 +137,28 @@ def run_unicode_server(port: int) -> None:  # pragma: no cover
         lifespan=lifespan,
     )
 
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    sock.bind(("127.0.0.1", 0))
+    sock.listen()
+    port = sock.getsockname()[1]
+    port_writer.send(port)
+    port_writer.close()
+
     # Run the server
     config = uvicorn.Config(
         app=app,
-        host="127.0.0.1",
-        port=port,
         log_level="error",
     )
     uvicorn_server = uvicorn.Server(config)
-    uvicorn_server.run()
+    uvicorn_server.run(sockets=[sock])
 
 
 @pytest.fixture
-def unicode_server_port() -> int:
-    """Find an available port for the Unicode test server."""
-    with socket.socket() as s:
-        s.bind(("127.0.0.1", 0))
-        return s.getsockname()[1]
-
-
-@pytest.fixture
-def running_unicode_server(unicode_server_port: int) -> Generator[str, None, None]:
+def running_unicode_server() -> Generator[str, None, None]:
     """Start a Unicode test server in a separate process."""
-    proc = multiprocessing.Process(target=run_unicode_server, kwargs={"port": unicode_server_port}, daemon=True)
-    proc.start()
-
-    # Wait for server to be ready
-    wait_for_server(unicode_server_port)
-
-    try:
-        yield f"http://127.0.0.1:{unicode_server_port}"
-    finally:
-        # Clean up - try graceful termination first
-        proc.terminate()
-        proc.join(timeout=2)
-        if proc.is_alive():  # pragma: no cover
-            proc.kill()
-            proc.join(timeout=1)
+    with running_server(run_unicode_server) as url:
+        yield url
 
 
 @pytest.mark.anyio
--- a/tests/server/test_sse_security.py
+++ b/tests/server/test_sse_security.py
@@ -4,6 +4,7 @@ import logging
 import multiprocessing
 import re
 import socket
+from multiprocessing.connection import Connection
 
 import anyio
 import httpx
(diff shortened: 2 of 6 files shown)


Ar-maan05:
Superseded by #2764/#2765.

PR #2708, opened 2026-05-28
Fix #348: Allow ToolError to carry custom content and is_error

### LINKED ISSUE #348: No way to set isError=True for arbitrary tool result content

Stoobyy:
## Why

Closes #348. There is currently no way to set `isError=True` when returning non-text content from tool calls (e.g., images, mixed content). The only way to get `isError=True` is to let an unhandled exception propagate, which always produces a single `TextContent` block.

## What changed

- `ToolError` now accepts `content` and `is_error` keyword arguments
- `Tool.run()` re-raises `ToolError` instead of wrapping it
- `_handle_call_tool` catches `ToolError` before the generic exception handler, constructing a `CallToolResult` with the caller's custom content and `is_error` value

## Usage

```python
from mcp.server.mcpserver.exceptions import ToolError
from mcp.types import TextContent, ImageContent

@server.tool()
def my_tool():
    raise ToolError(
        "Processing failed",
        content=[
            TextContent(type="text", text="Detailed error message"),
            ImageContent(type="image", data="base64...", mimeType="image/png"),
        ],
    )
```

## Validation

- All 91 existing tests pass
- 2 new tests added: ToolError with custom content, ToolError with default content

### DIFF
--- a/src/mcp/server/mcpserver/exceptions.py
+++ b/src/mcp/server/mcpserver/exceptions.py
@@ -1,4 +1,5 @@
 """Custom exceptions for MCPServer."""
+from __future__ import annotations
 
 
 class MCPServerError(Exception):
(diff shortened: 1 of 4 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #1824.

PR #2710, opened 2026-05-28
[v2] Dispatcher/ServerRunner receive-path swap — replaces BaseSession

maxisbey:
V2 server receive path: `Transport → JSONRPCDispatcher → ServerRunner → Server`, replacing the `BaseSession`/`ServerSession` message loop. Consolidates the stack under #2562 into one PR against `main`.

## What changed

- `JSONRPCDispatcher` (`mcp/shared/`): MCP-agnostic JSON-RPC engine — ID correlation, per-request task isolation, cancellation/progress, exception→wire boundary
- `ServerRunner`: per-connection MCP layer — initialize handshake, pre-init request gate, handler lookup/validation/dispatch
- `Connection`: per-connection state (peer info, scratch `state`, `exit_stack`) and the standalone outbound channel; `ServerSession` is now a thin facade over it
- `Server` is a plain registry: public `add_request_handler`/`add_notification_handler`, constructor `on_*` handlers
- `Server.middleware` wraps every inbound request and notification (raw params, runs before validation)
- All server transports route through the new path; the v1 `BaseSession` server loop is gone

## Parity bar

The 529 transport-parametrized tests in `tests/interaction/` (#2691), written against the v1 path, pass byte-unchanged — one deliberate exception below.

## Intentional behavior changes

- Handlers never receive `None` params: a message with no `params` member validates `{}` against the registered model — all-optional models arrive with their defaults; models with required fields reject as `INVALID_PARAMS` before the handler runs (Go SDK parity). One interaction snapshot updated (`test_roots.py`: `[None]` → `[NotificationParams()]`).

## Migration

Breaking changes documented in `docs/migration.md`.

Supersedes #2562 and the four PRs stacked below it.


src/mcp/shared/direct_dispatcher.py
```diff
@@ -0,0 +1,183 @@
...
+    Returns:
+        A ``(left, right)`` pair. Conventionally ``left`` is the client side
+        and ``right`` is the server side, but the wiring is symmetric.
+    """
+    ctx = TransportContext(kind=DIRECT_TRANSPORT_KIND, can_send_request=can_send_request, headers=headers)
+    left = DirectDispatcher(ctx)
+    right = DirectDispatcher(ctx)
+    left.connect_to(right)
+    right.connect_to(left)
+    return left, right
```
maxisbey: make client and server instead of left and right

src/mcp/shared/dispatcher.py
```diff
@@ -0,0 +1,157 @@
...
+    All keys are optional. Dispatchers ignore keys they do not understand.
+    """
+
+    timeout: float
+    """Seconds to wait for a result before raising and sending ``notifications/cancelled``."""
+
+    on_progress: ProgressFnT
+    """Receive ``notifications/progress`` updates for this request."""
+
+    resumption_token: str
```
maxisbey, line 61: investigate how this will work on new spec

docs/migration.md
```diff
@@ -442,7 +442,7 @@ async def handle_set_logging_level(level: str) -> None:
 mcp._mcp_server.subscribe_resource()(handle_subscribe)  # pyright: ignore[reportPrivateUsage]
 ```
 
-In v2, the lowlevel `Server` no longer has decorator methods (handlers are constructor-only), so the equivalent workaround is `_add_request_handler`:
+In v2, the lowlevel `Server` no longer has decorator methods (handlers are constructor-only), so the equivalent workaround is `add_request_handler`:
```
maxisbey: probably wouldn't say workaround here. Instead would say "this is now supported by ___" or something

docs/migration.md
```diff
@@ -1039,6 +1090,39 @@ from mcp.server import ServerRequestContext
...
+# where `dispatcher` is a JSONRPCDispatcher and `connection` is a Connection
+```
+
+In practice, replace direct `ServerSession` use with `Server.run(read_stream, write_stream, init_options)` and let the framework wire it up.
+
+**Removed from `mcp.server.session`:**
+
+- `InitializationState` enum and `ServerSession._initialization_state` — initialization tracking is now on `Connection` (`connection.initialized` is an `anyio.Event`, `connection.client_params` holds the init params).
+- `ServerRequestResponder` type alias.
+- `ServerSession.incoming_messages` stream — there is no longer a public stream of inbound messages to iterate. Register handlers via the `on_*` constructor params (or `add_request_handler`) and use `Server.middleware` to observe every request.
```
maxisbey: need to make sure middleware works even if a method doesn't exist I think

src/mcp/server/lowlevel/server.py
```diff
@@ -58,22 +56,45 @@ async def main():
...
+    Stored in `Server._request_handlers` / `_notification_handlers` and consumed
+    by `ServerRunner` to validate, build `Context`, and invoke. The handler's
+    second-argument type is erased to `Any` in storage (each entry has a
+    different concrete params type and `Callable` parameters are contravariant);
+    the precise type is recoverable via `params_type`. The correlation is
+    enforced at registration time by `Server.add_request_handler`.
+    """
+
+    params_type: type[BaseModel]
+    handler: RequestHandler[LifespanResultT, Any]
```
maxisbey, line 96: guessing we can't make params_type generic here with it being set the same as the handelr any?

src/mcp/server/lowlevel/server.py
```diff
@@ -191,53 +212,75 @@ def __init__(
...
         self.lifespan = lifespan
-        self._request_handlers: dict[str, Callable[[ServerRequestContext[LifespanResultT], Any], Awaitable[Any]]] = {}
-        self._notification_handlers: dict[
-            str, Callable[[ServerRequestContext[LifespanResultT], Any], Awaitable[None]]
-        ] = {}
+        self._request_handlers: dict[str, HandlerEntry[LifespanResultT]] = {}
+        self._notification_handlers: dict[str, HandlerEntry[LifespanResultT]] = {}
         self._session_manager: StreamableHTTPSessionManager | None = None
+        # Context-tier middleware consumed by `ServerRunner`. Additive; the
+        # existing `run()` path ignores it.
```
maxisbey: what does this mean that run() ignores it?

src/mcp/server/lowlevel/server.py
```diff
@@ -191,53 +212,75 @@ def __init__(
...
+        self.middleware: list[ServerMiddleware[LifespanResultT]] = []
```
maxisbey, line 224: worth thinking about or putting a todo that this will change with further middleware redesigns before v2 final release

src/mcp/server/lowlevel/server.py
```diff
@@ -191,53 +212,75 @@ def __init__(
...
+            {m: HandlerEntry(pt, h) for m, pt, h in _spec_notifications if h is not None}
         )
 
-    def _add_request_handler(
+    def add_request_handler(
         self,
         method: str,
-        handler: Callable[[ServerRequestContext[LifespanResultT], Any], Awaitable[Any]],
+        params_type: type[_ParamsT],
+        handler: RequestHandler[LifespanResultT, _ParamsT],
```
maxisbey, line 255: ah I see we do type enforcement here? why does it have to be any on the handler then?

src/mcp/server/lowlevel/server.py
```diff
@@ -191,53 +212,75 @@ def __init__(
...
+        handler: NotificationHandler[LifespanResultT, _ParamsT],
+    ) -> None:
+        """Register a notification handler for `method`.
+
+        `params_type` should subclass `NotificationParams` so `_meta`
+        parses uniformly. Replaces any existing handler.
+        """
+        self._notification_handlers[method] = HandlerEntry(params_type, handler)
+
+    # --- ServerRegistry protocol (consumed by ServerRunner) ------------------
```
maxisbey: remove this commet

src/mcp/server/connection.py
```diff
@@ -0,0 +1,173 @@
+"""`Connection` - per-client connection state and the standalone outbound channel.
```
maxisbey, line 1: note to reviewers: this will change with fast follow Context redesign

src/mcp/server/connection.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @property
+    def client_info(self) -> Implementation | None:
+        """The client's `Implementation` from `initialize`; `None` before initialization."""
+        return self.client_params.client_info if self.client_params is not None else None
+
+    @property
+    def client_capabilities(self) -> ClientCapabilities | None:
+        """The client's `ClientCapabilities` from `initialize`; `None` before initialization."""
+        return self.client_params.capabilities if self.client_params is not None else None
```
maxisbey: are these needed? Can't we just expose client_params instead?

src/mcp/server/connection.py
```diff
@@ -0,0 +1,173 @@
...
+
+    async def send_raw_request(
+        self,
+        method: str,
+        params: Mapping[str, Any] | None,
+        opts: CallOptions | None = None,
```
maxisbey, line 124: maybe worth putting in the docstring what CallOptions is? unelss that has been typed, but I can't easily browse to that while reviewing.

src/mcp/server/connection.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Raises:
+            MCPError: The peer responded with an error.
+            NoBackChannelError: `has_standalone_channel` is `False`.
+        """
+        if not self.has_standalone_channel:
+            raise NoBackChannelError(method)
+        return await self._outbound.send_raw_request(method, params, opts)
+
+    async def notify(self, method: str, params: Mapping[str, Any] | None) -> None:
```
maxisbey, line 177: probably better to be called send_raw_notification, no?

src/mcp/server/connection.py
```diff
@@ -0,0 +1,173 @@
...
+            logger.debug("dropped %s: standalone stream closed", method)
+
+    async def ping(self, *, meta: Meta | None = None, opts: CallOptions | None = None) -> None:
+        """Send a `ping` request on the standalone stream.
+
+        Raises:
+            MCPError: The peer responded with an error.
+            NoBackChannelError: `has_standalone_channel` is `False`.
+        """
+        await self.send_raw_request("ping", dump_params(None, meta), opts)
```
maxisbey, line 198: is dump_params public and supposed to be public? otherwise maybe module/package private. would be good to cut down on public interface where we can

src/mcp/server/context.py
```diff
@@ -1,21 +1,147 @@
 from __future__ import annotations
```
maxisbey: is this needed? please review this on all files we touch as to whether it's still needed otherwise remove

src/mcp/server/context.py
```diff
@@ -1,21 +1,147 @@
...
+    handler.
+    """
+
+    def __init__(
+        self,
+        dctx: DispatchContext[TransportContext],
+        *,
+        lifespan: LifespanT,
+        connection: Connection,
+        meta: RequestParamsMeta | None = None,
```
maxisbey, line 61: why is this here? I think it'd make more sense to have a like raw request object somewhere availble. I know we pass the params into method handlers, don't those params contain the meta field?

we need to think through how to do something like ctx.???.jsonrpc_message or something. that'd be later though I think

src/mcp/server/context.py
```diff
@@ -1,21 +1,147 @@
...
+    ) -> None:
+        super().__init__(dctx, meta=meta)
+        self._lifespan = lifespan
+        self._connection = connection
+
+    @property
+    def lifespan(self) -> LifespanT:
```
maxisbey: when doing backwards compat stuff need to make sure to remember to add per-session lifespan shims

src/mcp/server/runner.py
```diff
@@ -0,0 +1,304 @@
...
+    init_options: InitializationOptions | None = None
+    """`InitializeResult` payload. Defaults to `server.create_initialization_options()`."""
+    session_id: str | None = None
+    stateless: bool = False
+    dispatch_middleware: list[DispatchMiddleware] = field(default_factory=list[DispatchMiddleware])
+
+    connection: Connection = field(init=False)
+    session: ServerSession = field(init=False)
+    """Connection-scoped: the same instance reaches every request as `ctx.session`."""
+    _initialized: bool = field(init=False)
```
maxisbey: need to remember to remove and restructure later. or potentially now we should move sessions/initialization to be middleware instead somehow...

src/mcp/server/session.py
```diff
@@ -1,188 +1,97 @@
...
+`ctx.session` and use the typed helpers (`create_message`, `elicit_form`,
+`send_log_message`, ...) to call back to the client.
+
+The receive-loop, initialize handling, and per-request task isolation that
+used to live here are now owned by `JSONRPCDispatcher` and `ServerRunner`.
 """
 
-from enum import Enum
-from typing import Any, TypeVar, overload
+from __future__ import annotations
```
maxisbey: again, is this needed?

src/mcp/shared/dispatcher.py
```diff
@@ -0,0 +1,182 @@
...
+    timeout: float
+    """Seconds to wait for a result before raising and sending `notifications/cancelled`."""
+
+    on_progress: ProgressFnT
+    """Receive `notifications/progress` updates for this request."""
+
+    resumption_token: str
+    """Opaque token to resume a previously interrupted request (transport-dependent)."""
+
+    on_resumption_token: Callable[[str], Awaitable[None]]
```
maxisbey, line 71: who is this used by? client or server?

src/mcp/shared/dispatcher.py
```diff
@@ -0,0 +1,182 @@
...
+        transports attach `ServerMessageMetadata` (the HTTP request, SSE
+        stream-close callbacks); stdio and in-memory dispatch attach nothing.
+        Tied to the `SessionMessage` wire format - goes away when transports
+        stop delivering messages that way.
+        """
+        # TODO(maxisbey): remove for context rework
+        ...
+
+    @property
+    def cancel_requested(self) -> anyio.Event:
```
maxisbey, line 163: need to add usability and a high level test for the different ways that cancellation works

src/mcp/shared/jsonrpc_dispatcher.py
```diff
@@ -0,0 +1,649 @@
...
+    ) -> None: ...
+    def __init__(
+        self,
+        read_stream: ReadStream[SessionMessage | Exception],
+        write_stream: WriteStream[SessionMessage],
+        *,
+        transport_builder: Callable[[MessageMetadata], TransportT] | None = None,
+        peer_cancel_mode: PeerCancelMode = "interrupt",
+        raise_handler_exceptions: bool = False,
+        inline_methods: frozenset[str] = frozenset(),
```
maxisbey, line 249: we need to think about these. They're not in the Dispatcher protocol, so I don't really like them a lot.

Just to confirm my mental model: the idea is that the SHTTP transport classes would build this JSONRPCDispatcher, and then hand the dispatcher to the ServerRunner which accepts the Protocol class right?

If so, these seem like potentially useful stuff to formalize somehow.. or really think through the usability of. Especially the peer_cancel_mode and raise_handler_exceptions. Also the inline_methods here wouldn't work in the model I described given that the init would happen before ServerRunner even exists.

src/mcp/shared/jsonrpc_dispatcher.py
```diff
@@ -0,0 +1,649 @@
...
+
+        metadata = _outbound_metadata(_related_request_id, opts)
+        target = out_params.get("name")
+        span_name = f"MCP send {method}{f' {target}' if isinstance(target, str) else ''}"
+        # TODO(maxisbey): the otel span + inject below mirror
+        # BaseSession.send_request for parity. They belong in an outbound
+        # middleware (symmetric with otel_middleware on the inbound side) once
+        # that seam exists; the dispatcher should not own otel.
+        try:
+            with otel_span(
```
maxisbey, line 334: will need to rethink this as middleware or something, but maybe not this PR

src/mcp/shared/jsonrpc_dispatcher.py
```diff
@@ -0,0 +1,649 @@
...
+            self._spawn(self._handle_request, req, dctx, scope, on_request, sender_ctx=sender_ctx)
+
+    def _dispatch_notification(
+        self,
+        msg: JSONRPCNotification,
+        metadata: MessageMetadata,
+        on_notify: OnNotify,
+        sender_ctx: contextvars.Context | None,
+    ) -> None:
+        if msg.method == "notifications/cancelled":
```
maxisbey, line 555: worth adding a rationalisation comment here, explaining why we need this in the Dispatcher when it's mcp specific (basically cause the cancellation is tied to a specific json rpc request id, basically by spec definition coupled)

src/mcp/shared/jsonrpc_dispatcher.py
```diff
@@ -0,0 +1,649 @@
...
+        transport_ctx = self._transport_builder(metadata)
+        dctx = _JSONRPCDispatchContext(
+            transport=transport_ctx,
+            _dispatcher=self,
+            _request_id=req.id,
+            message_metadata=metadata,
+            _progress_token=progress_token,
+        )
+        scope = anyio.CancelScope()
+        self._in_flight[req.id] = _InFlight(scope=scope, dctx=dctx)
```
maxisbey: need to handle the case of if a client sends a request id that already exists in session. check spec language

src/mcp/shared/peer.py
```diff
@@ -0,0 +1,216 @@
...
+    ToolChoice,
+)
+
+__all__ = ["Meta", "Peer", "PeerMixin", "dump_params"]
+
+Meta = dict[str, Any]
+"""Type alias for the `_meta` field carried on request/notification params."""
+
+
+def dump_params(model: BaseModel | None, meta: Meta | None = None) -> dict[str, Any] | None:
```
maxisbey, line 42: ah I see this, yea probably worth putting somewhere module private I'd think? unless you can see a reason for an sdk consumer to use this?

src/mcp/shared/peer.py
```diff
@@ -0,0 +1,216 @@
...
+    precedence over any `_meta` already present on the model.
+    """
+    out = model.model_dump(by_alias=True, mode="json", exclude_none=True) if model is not None else None
+    if meta:
+        out = dict(out or {})
+        out["_meta"] = {**out.get("_meta", {}), **meta}
+    return out
+
+
+class PeerMixin:
```
maxisbey: yea probably wroth renaming to ClientPeer or something? idk have a think

tests/issues/test_malformed_input.py
```diff
@@ -1,151 +0,0 @@
...
-import pytest
-
-from mcp.server.models import InitializationOptions
-from mcp.server.session import ServerSession
-from mcp.shared.message import SessionMessage
-from mcp.types import INVALID_PARAMS, JSONRPCError, JSONRPCMessage, JSONRPCRequest, ServerCapabilities
-
-
-@pytest.mark.anyio
-async def test_malformed_initialize_request_does_not_crash_server():
```
maxisbey, line 14: sanity check we don't need to test this as well

tests/shared/test_progress_notifications.py
```diff
@@ -1,264 +0,0 @@
-from typing import Any
```
maxisbey, line 1: do we still have or need progress notofication tests? or do we just rely on the interaction/ tests now?

src/mcp/shared/direct_dispatcher.py
maxisbey: Done — locals + docstring renamed.

docs/migration.md
maxisbey: Reworded — heading drops "(workaround)", text now says supported via `add_request_handler`.

src/mcp/server/lowlevel/server.py
maxisbey: Correct — making it generic just moves the `Any` to the dict value type (heterogeneous storage + contravariant param). The correlation is enforced at `add_request_handler`; docstring at L88-92 covers this.

src/mcp/server/lowlevel/server.py
maxisbey: Stale post-swap — `run()` does consume it now via `ServerRunner._on_request`. Rewrote the comment.

src/mcp/server/lowlevel/server.py
maxisbey: Added the TODO.

src/mcp/server/lowlevel/server.py
maxisbey: Yep — this is the typed correlation point, then it's erased for heterogeneous storage (see L88-92). No tighter bound exists.

src/mcp/server/lowlevel/server.py
maxisbey: Deleted.

src/mcp/server/connection.py
maxisbey: Dropped both — only our own test was reading them; v1 only ever exposed `client_params`. `check_capability` reads through `client_params` now.

src/mcp/server/connection.py
maxisbey: Added a one-line `opts` pointer here and on `Outbound.send_raw_request`.

src/mcp/server/connection.py
maxisbey: Dropped from `peer.__all__`. Didn't underscore-rename — cross-module callers would need `reportPrivateUsage` ignores.

src/mcp/shared/dispatcher.py
maxisbey: Annotated both as client-side / streamable-HTTP only.

src/mcp/shared/jsonrpc_dispatcher.py
maxisbey: Docstring added pointing at the module-level rationale + the `_in_flight`/`_pending` correlation reason.

src/mcp/shared/peer.py
maxisbey: Dropped from `__all__` (see the `connection.py` thread). `Meta` kept since it appears in user-facing kwarg signatures.

src/mcp/shared/dispatcher.py
maxisbey: Checked 2025-11-25 + draft: `on_progress`/`timeout` hold up unchanged. Resumption is removed in draft — annotated both keys as 2025-11-25-and-earlier.

docs/migration.md
maxisbey: Restructured: `Server.middleware` now wraps the entire `_on_request` and `_on_notify` body (raw `Mapping` params), so it sees `initialize`, `METHOD_NOT_FOUND`, validation failures, and notifications. Doc claim is now true.

src/mcp/server/connection.py
maxisbey: Ack — tracking the cluster for the Context rework.

src/mcp/server/connection.py
maxisbey: Keeping `notify` — the asymmetry exists because `send_raw_request` has a typed sibling to differ from; notifications have many named conveniences so bare `notify` is unambiguous. The pair is consistent across all 8 layer classes.

src/mcp/server/context.py
maxisbey: Removed (no forward refs). Also dropped from `session.py`.

src/mcp/server/context.py
maxisbey: Tracking for Context rework — `ServerRequestContext.meta` is load-bearing for now; plan is to expose the raw inbound message on `DispatchContext` instead of cherry-picking.

src/mcp/server/context.py
maxisbey: Tracking for Context rework — `session_lifespan` hook vs documenting `connection.state`/`exit_stack`.

src/mcp/server/runner.py
maxisbey: Partially done by the t04 restructure: middleware now wraps `initialize` + `notifications/initialized`. `_handle_initialize` is still a special-case branch (mutates runner state); making it a registered handler is rework territory.

src/mcp/server/session.py
maxisbey: Removed — no actual cycle, promoted `Connection` to a plain import.

src/mcp/shared/dispatcher.py
maxisbey: Tracking for Context rework — `cancel_requested`/signal mode are dormant until plumbed to `Server.run()`.

src/mcp/shared/jsonrpc_dispatcher.py
maxisbey: Tracking for Context rework — move `inline_methods`/`raise_handler_exceptions` to `Dispatcher.run()` so the protocol-aware caller sets them. `peer_cancel_mode` could go now if you want to shrink the surface.

src/mcp/shared/jsonrpc_dispatcher.py
maxisbey: Already TODO'd + tracked.

src/mcp/shared/jsonrpc_dispatcher.py
maxisbey: Spec puts uniqueness on the sender; v1 and TS both blind-overwrite same as us. Left a TODO for now (parity); can add the `INVALID_REQUEST` guard later.

src/mcp/shared/peer.py
maxisbey: Renamed to `ClientPeerMixin`/`ClientPeer`. Fixed the stale docstring (Connection doesn't mix it in).

tests/issues/test_malformed_input.py
maxisbey: Added `test_runner_initialize_with_absent_params_returns_invalid_params_and_stays_alive` — re-covers the deleted regression on the new path.

tests/shared/test_progress_notifications.py
maxisbey: Covered — interaction `test_progress.py` (7 tests vs requirements manifest) + the dispatcher unit tests for callback-exception.

src/mcp/server/runner.py
```diff
@@ -0,0 +1,396 @@
...
+    async def _on_request(
+        self,
+        dctx: DispatchContext[TransportContext],
+        method: str,
+        params: Mapping[str, Any] | None,
+    ) -> dict[str, Any]:
+        ctx = self._make_context(dctx, _extract_meta(params))
+
+        async def _inner() -> HandlerResult:
+            # TODO(maxisbey): pinned compat. `BaseSession._receive_loop`
```
maxisbey: for all massive comments like this, reduce to one or two sentences

src/mcp/client/_memory.py
Kludex, line 1: Are those changes necessary? Seems unrelated.

src/mcp/server/lowlevel/server.py
```diff
@@ -58,22 +56,45 @@ async def main():
...
+
+    Stored in `Server._request_handlers` / `_notification_handlers` and consumed
+    by `ServerRunner` to validate, build `Context`, and invoke. The handler's
+    second-argument type is erased to `Any` in storage (each entry has a
+    different concrete params type and `Callable` parameters are contravariant);
+    the precise type is recoverable via `params_type`. The correlation is
+    enforced at registration time by `Server.add_request_handler`.
+    """
+
+    params_type: type[BaseModel]
```
Kludex, line 95: This doesn't seem correct.

src/mcp/server/_typed_request.py
```diff
@@ -0,0 +1,86 @@
...
+    ) -> BaseModel:
+        """Send a typed server-to-client request and return its typed result.
+
+        For spec request types the result type is inferred. For custom requests
+        pass `result_type=` explicitly.
+
+        Raises:
+            MCPError: The peer responded with an error.
+            NoBackChannelError: No back-channel for server-initiated requests.
+            KeyError: `result_type` omitted for a non-spec request type.
```
Kludex: I guess validationError as well?

src/mcp/server/connection.py
```diff
@@ -0,0 +1,168 @@
...
+    peer-info remains `None` (no handshake reaches a stateless connection).
+    """
+
+    def __init__(self, outbound: Outbound, *, has_standalone_channel: bool, session_id: str | None = None) -> None:
+        self._outbound = outbound
+        self.has_standalone_channel = has_standalone_channel
+        self.session_id: str | None = session_id
+
+        self.client_params: InitializeRequestParams | None = None
+        """The full `initialize` request params; `None` before initialization."""
```
Kludex: This is fully wrong. There's no such a thing as docstring in the middle of a constructor.

src/mcp/server/context.py
```diff
@@ -1,21 +1,160 @@
...
+    session: ServerSession
     lifespan_context: LifespanContextT
+    request_id: RequestId | None = None
+    meta: RequestParamsMeta | None = None
     request: RequestT | None = None
     close_sse_stream: CloseSSEStreamCallback | None = None
     close_standalone_sse_stream: CloseSSEStreamCallback | None = None
+
+
+LifespanT = TypeVar("LifespanT", default=Any, covariant=True)
```
Kludex: What's the difference between LifespanT and LifespanContextT above?

src/mcp/server/context.py
```diff
@@ -1,21 +1,160 @@
...
+
+
+class Context(BaseContext[TransportContext], ClientPeerMixin, TypedServerRequestMixin, Generic[LifespanT]):
```
Kludex: @maxisbey I don't really think mixins are a good idea in general. Makes the code very hard to follow.

src/mcp/server/runner.py
```diff
@@ -0,0 +1,396 @@
...
+        `connection.exit_stack` is unwound (shielded) so any per-connection
+        cleanup registered by handlers or middleware runs to completion.
+        """
+        try:
+            await self.dispatcher.run(self._compose_on_request(), self._on_notify, task_status=task_status)
+        finally:
+            with anyio.CancelScope(shield=True):
+                try:
+                    await self.connection.exit_stack.aclose()
+                except Exception:
```
Kludex, line 204: What's the right exception?

src/mcp/client/_memory.py
maxisbey: These are a consequence of the swap rather than a drive-by. The dispatcher exits on read-stream EOF, so the transport can now shut the server down gracefully by closing the streams instead of hard-cancelling the task group — and the cancel actually has to go, because it throws into the test's own task, which on CPython 3.11 hits python/cpython#106749 (`coro.throw()` drops `'call'` trace events) and desyncs coverage's CTracer; the old receive loop masked that by suspending a second time on join, the dispatcher's EOF fast path doesn't, so the 3.11 coverage job fails without this. The grace-period fallback exists because EOF (unlike cancel) lets user lifespan teardown run, and a hung teardown would otherwise block forever. Full mechanism is written up in 9f63603's commit message.

src/mcp/server/runner.py
maxisbey: `Exception` is deliberate here. The exit stack runs arbitrary user cleanup (handlers and middleware push callbacks onto `connection.exit_stack`), so there's no narrower set to name — `aclose()` re-raises whatever the callback raised. Letting it propagate would crash stdio servers on a normal disconnect and, via raise-in-finally, mask `dispatcher.run()`'s real outcome — including the `CancelledError` that the streamable HTTP idle-timeout teardown checks via `idle_scope.cancelled_caught`. `BaseException` would be too wide: it would eat the `move_on_after` deadline's own cancellation (breaking the abandon-after-grace warning) as well as KeyboardInterrupt/SystemExit. The behavior is pinned by `test_runner_exit_stack_cleanup_exception_is_logged_not_propagated` and documented on `Connection.exit_stack` ("exceptions are logged and swallowed").

src/mcp/server/lowlevel/server.py
maxisbey: The erasure here is deliberate: each entry stores a different concrete params model, and `Callable` parameters are contravariant, so the entry dict has no useful common supertype — the `params_type`/handler correlation is enforced at `add_request_handler` registration instead (a mismatch is a type error at the call site). Chasing this did surface a real wart nearby though: handlers could receive `None` when a message omitted `params`, contradicting the registered handler type. Fixed in 33130a9 — absent params now validate `{}` against the model, so handlers get the model with its defaults, never `None` (required-field models reject as `INVALID_PARAMS` first; same guarantee as the Go SDK). If you meant something else here, happy to dig further.

src/mcp/server/connection.py
maxisbey: Fair — moved them out in fda5736. The public attributes are now declared in the class body with their docstrings under the declarations, and `__init__` just assigns; pyright/Pylance hover and the mkdocstrings rendering pick the docs up exactly as before (verified both). The private `_inline_methods` one in the dispatcher is now a regular comment above the assignment, since the docs site never rendered it anyway.

src/mcp/server/context.py
maxisbey: They're different on purpose, though the naming wasn't helping — renamed in 53e531c to make it visible. `LifespanContextT` (invariant, `dict[str, Any]` default) parameterizes `ServerRequestContext`, where the lifespan is a writable dataclass field, so it can't be covariant. The new context class exposes `lifespan` read-only, so its var is covariant (`Context[AppState]` passes where `Context[object]` is expected, which middleware composition relies on) — it's now spelled `LifespanT_co` to make that distinction explicit. Full consolidation of the remaining invariant vars (`LifespanResultT` etc.) is planned for the upcoming Context rework, where these seams get rewritten anyway.

src/mcp/server/context.py
maxisbey: Agreed — they're gone as of 186bf6e. `send_request` now lives directly on `Connection` (its only live host), the peer methods live on `ClientPeer` as a plain class, and the unwired `Context` lost that surface entirely (it comes back as composition, not inheritance, in the upcoming Context rework). Net −98 lines and no mixins left in the codebase.

src/mcp/server/_typed_request.py
maxisbey: Added in dd2d770 — also to the peer methods and `ServerSession.send_request`, which had the same gap (that one was missing `MCPError` too). The docstrings now live on `Connection`, since `_typed_request.py` was folded into it in 186bf6e.

felixweinberger (approved):

PR #2729, opened 2026-05-30
fix(client): send same-origin Origin header from streamable HTTP client

### LINKED ISSUE #2727: streamable_http client does not send `Origin` header → rejected with 403 by spec-compliant servers (e.g. go-sdk `CrossOriginProtection`)
## Summary

The Python SDK's `streamable_http_client` opens its POST handshake without an `Origin` header (and without `Sec-Fetch-Site`). The official Go SDK (`modelcontextprotocol/go-sdk` v1.4.x) wraps every streamable-HTTP handler with Go 1.25's stdlib `http.CrossOriginProtection`, which enforces the spec's anti-DNS-rebinding rule and **denies any state-changing request that cannot prove same-origin** via one of:

- `Sec-Fetch-Site: same-origin | same-site | none` (browser-only), or
- An `Origin` header whose host matches the server's `Host`, or
- An origin explicitly listed via `CrossOriginProtection.AddTrustedOrigin`.

Since the Python client sends none of those, a perfectly legitimate server-to-server connection from the official Python client to the official Go server is indistinguishable from a CSRF attempt → **HTTP 403 Forbidden** on the very first POST.

So the two reference SDKs from the same org are out of sync by one spec revision: the Go server enforces the new rule; the Python client doesn't yet send the headers that satisfy it.

## Reproduction

**Server** — a Go MCP server built with `modelcontextprotocol/go-sdk@v1.4.1` and the standard handler:

```go
handler := mcp.NewStreamableHTTPHandler(
    func(_ *http.Request) *mcp.Server { return srv },
    nil, // default CrossOriginProtection: deny non-same-origin
)
http.Handle("/mcp", handler)
```

**Client** — Python `mcp` SDK:

```python

async with streamablehttp_client("http://my-go-server:8081/mcp") as (read, write, _):
        await session.initialize()   # never completes
```

**Observed:**

1. `httpx` POSTs to `/mcp` with no `Origin` header, no `Sec-Fetch-*` headers.
2. Go server returns `HTTP/1.1 403 Forbidden` immediately.
3. Python client `post_writer` swallows the non-2xx (see #2110), and `session.initialize()` hangs forever on the read stream.
4. Eventually the caller (e.g. a FastAPI startup hook with a `wait_for` timeout) cancels, which surfaces as `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in` because the `streamable_http_client` task group was entered in one task and is being torn down in another.

**Expected:** the Python client should send an `Origin` header derived from the target URL by default, so a spec-compliant server accepts the handshake.

## Workarounds (today)

- **Server side, Go**: pass `&mcp.StreamableHTTPHandlerOptions{CrossOriginProtection: cop}` with `cop.AddTrustedOrigin(...)`, or set `GODEBUG=disablecrossoriginprotection=1`. Both require code/env changes on every server deployment.
- **Client side, Python**: monkey-patch the httpx client to inject `Origin`. Brittle; depends on internals of `streamable_http`.

Neither is satisfactory if both reference SDKs are supposed to interoperate out of the box.

## Suggested fix

In `mcp.client.streamable_http`, when opening the `httpx.AsyncClient`, derive a default `Origin` header from the target URL's scheme + netloc and add it to every outgoing request:

```python
parsed = urlparse(self.url)
default_origin = f"{parsed.scheme}://{parsed.netloc}"
headers.setdefault("Origin", default_origin)
```

This makes the Python client's traffic indistinguishable from a same-origin browser request as far as `CrossOriginProtection.Check` is concerned, without weakening any server's CSRF posture. Callers who *want* a different `Origin` (e.g. multi-tenant proxies) can still override via the existing custom-headers path.

Optionally, also set `Sec-Fetch-Site: same-origin` so the Go middleware short-circuits on the cheaper check.

## Related

- #2110 — explains why the resulting 403 manifests as a client hang instead of a clean exception.
- #1798 / #861 — the Python *server* added equivalent DNS-rebinding protection; the *client* never picked up the matching header behavior.

## Environment

- `mcp` (Python SDK): latest installed via `mcp >= 1.x` (tested under Locus's `locus.integrations.fastmcp.MCPClient` wrapper, which is a thin pass-through to `streamablehttp_client`).
- `modelcontextprotocol/go-sdk@v1.4.1` (server).
- Go 1.25 (stdlib `http.CrossOriginProtection`).
- Python 3.13, anyio 4.x.

Bartok9:
## Summary

- The streamable HTTP **client** now sends a same-origin `Origin` header by default.
- This lets the official Python client interoperate with spec-compliant servers (e.g. the Go SDK's `http.CrossOriginProtection`) that otherwise reject the handshake with `403`.

## Motivation

Closes #2727.

The Python `streamablehttp_client` opened its POST handshake **without** an `Origin` header. The official Go SDK (`modelcontextprotocol/go-sdk` v1.4.x) wraps every streamable-HTTP handler with Go 1.25's `http.CrossOriginProtection`, which denies any state-changing request that cannot prove same-origin via `Sec-Fetch-Site`, a matching `Origin`, or an allow-listed origin. A legitimate server-to-server connection from the Python client therefore looks like a CSRF attempt → `HTTP 403 Forbidden` on the first POST, and the client (per #2110) swallows the non-2xx and hangs forever on `session.initialize()`.

The two reference SDKs from the same org were out of sync by one spec revision: the Go server enforces the rule; the Python client never sent the header that satisfies it.

## Fix

`StreamableHTTPTransport._prepare_headers()` now derives a same-origin value (`scheme://host[:port]`) from the target URL and sends it as the `Origin` header on every request. The derivation:

- Uses `urllib.parse.urlsplit` on the configured URL.
- Returns `None` (adds no header) when the URL has no scheme or host, so malformed/relative URLs are unaffected.

Callers needing a different `Origin` (e.g. multi-tenant proxies) can still set one on the underlying `httpx.AsyncClient` default headers.

## Verification

- `uv run pytest tests/shared/test_streamable_http.py` — **63 passed** (includes the existing `test_streamable_http_client_mcp_headers_override_defaults` / custom-header tests, unchanged).
- New tests:
  - `test_prepare_headers_includes_same_origin` — `http://my-go-server:8081/mcp` → `Origin: http://my-go-server:8081`; `https://…/path?x=1` → `https://example.com`.
  - `test_prepare_headers_omits_origin_for_invalid_url` — no `Origin` added for a URL lacking scheme/host.
- `uv run ruff check` / `ruff format --check` — clean.

Diff: 2 files, +41/-0.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -202,7 +202,7 @@ Invalid Host header
 The fix is the same `transport_security=TransportSecuritySettings(allowed_hosts=[...], allowed_origins=[...])` shown under `Server returned an error response`. Two of its edges are worth naming:
 
 * An `allowed_hosts` entry is an exact string. `"mcp.example.com"` matches a bare `Host` header and `"mcp.example.com:*"` matches any explicit port. List both.
-* A `403` with the body `Invalid Origin header` is the sibling check on the `Origin` header. It only fires for browsers (nothing else sends `Origin`), and `allowed_origins=` is its allowlist.
+* A `403` with the body `Invalid Origin header` is the sibling check on the `Origin` header, and `allowed_origins=` is its allowlist. Browsers send `Origin`, and so does the python `Client`: it stamps a same-origin value (`scheme://host[:port]`) derived from the URL you connect to, so that spec-compliant servers enforcing CSRF / DNS-rebinding protection accept the handshake. That is why the allowlist above names `http://mcp.example.com` alongside the host — the client's own `Origin` has to be on it.
 
 **[Deploy & scale](run/deploy.md)** has the full treatment, including when switching the check off is the honest configuration.
 
--- a/docs_src/troubleshooting/tutorial004.py
+++ b/docs_src/troubleshooting/tutorial004.py
@@ -13,6 +13,6 @@ def forecast(city: str) -> str:
 app = mcp.streamable_http_app(
     transport_security=TransportSecuritySettings(
         allowed_hosts=["mcp.example.com", "mcp.example.com:*"],
-        allowed_origins=["https://app.example.com"],
+        allowed_origins=["http://mcp.example.com", "http://mcp.example.com:*"],
     )
 )
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -7,6 +7,7 @@ import logging
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
+from urllib.parse import urlsplit
 
 import anyio
 import httpx
@@ -108,6 +109,19 @@ class StreamableHTTPTransport:
         # `notifications/cancelled` at 2026 can abort it; see
         # `_consume_modern_cancellation`. Keys are verbatim-typed ("1" is not 1).
         self._in_flight_posts: dict[RequestId, _InFlightPost] = {}
+        self._default_origin = self._derive_origin(url)
+
+    @staticmethod
+    def _derive_origin(url: str) -> str | None:
+        """Derive a same-origin ``Origin`` value (scheme://host[:port]) from a URL.
+
+        Returns ``None`` when the URL has no scheme or host, in which case no
+        ``Origin`` header is added.
+        """
+        parsed = urlsplit(url)
+        if not parsed.scheme or not parsed.netloc:
+            return None
+        return f"{parsed.scheme}://{parsed.netloc}"
 
     def _prepare_headers(self) -> dict[str, str]:
         """Build MCP-specific request headers for any outbound HTTP request.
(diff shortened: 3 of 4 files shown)


Bartok9:
**CI note:** the single red cell (`test (3.12, lowest-direct, ubuntu-latest)`) is an infra/flake failure unrelated to this change, not a regression from this PR.

- Failing test: `tests/shared/test_streamable_http.py::test_streamable_http_client_session_termination_204` (a pre-existing async server test, not added by this PR).
- Root cause in the log: `[Errno 98] error while attempting to bind on address ('127.0.0.1', 50093): address already in use` → `httpx.ConnectError: All connection attempts failed`. This is the TOCTOU test-port race tracked in #2704.
- The two tests this PR adds (`test_prepare_headers_includes_same_origin`, `test_prepare_headers_omits_origin_for_invalid_url`) are synchronous unit tests that instantiate `StreamableHTTPTransport`, call `_prepare_headers()`, and assert on the returned dict. They never start a server, bind a port, or open a socket, so they cannot produce a bind/connect error.
- All other 23 test cells pass; `all-green` aggregate is green; PR is `MERGEABLE`.

A rerun of the failed job should clear it (I don't have rerun permission on the fork PR). Happy to rebase if main has moved.

Bartok9:
**CI note** — rebased onto current `main`. The only remaining red cell (`test (3.14, lowest-direct, …)`) is a **pre-existing, main-wide failure** unrelated to this PR: `tests/interaction/transports/test_stdio.py::test_tool_call_and_notification_round_trip…` asserts byte-exact child stderr, but `anyio==4.9` (the `lowest-direct` floor) has `return` inside a `finally:` in `from_thread.py:119`, which Python 3.14 emits as `SyntaxWarning` to stderr and pollutes the snapshot. This PR does not touch stdio or that test. Full root-cause + the 6 parallel fixes already in flight are consolidated in https://github.com/modelcontextprotocol/python-sdk/pull/2734#issuecomment-4628718448. All other cells (incl. the previously-flaky termination test, now run in-process via #2767) are green.

PR #2730, opened 2026-05-30
Send SEP-2243 Mcp-Method/Mcp-Name headers from the StreamableHTTP client

### LINKED ISSUE #2715: SEP-2243: Send required Mcp-Method/Mcp-Name HTTP headers
## Description

Per [SEP-2243 (HTTP Standard Headers)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2243-http-standard-headers.md), MCP clients MUST send the following HTTP headers on POST requests:

- `Mcp-Method`: Set to the JSON-RPC `method` value from the request body (REQUIRED on every POST)
- `Mcp-Name`: Set from `params.name` (for tools/call, prompts/get) or `params.uri` (for resources/read, resources/subscribe, resources/unsubscribe) when present

## Context

The cxxmcp (C++ SDK) conformance testing revealed that the official conformance test suite's reference client (TypeScript SDK) does not send these headers, making the conformance suite self-contradictory. See:
- [typescript-sdk#2176](https://github.com/modelcontextprotocol/typescript-sdk/issues/2176) — TypeScript SDK bug report
- [conformance#323](https://github.com/modelcontextprotocol/conformance/issues/323) — Conformance suite self-contradiction
- [typescript-sdk#2178](https://github.com/modelcontextprotocol/typescript-sdk/pull/2178) — Fix PR for TypeScript SDK

## Impact

Servers that enforce SEP-2243 (reject missing `Mcp-Method` with HTTP 400) are incompatible with clients that don't send these headers. This affects cross-SDK interoperability.

## Request

Please implement `Mcp-Method` and `Mcp-Name` header support per SEP-2243 to ensure cross-SDK compatibility.

vidigoat:
## Summary

Fixes #2715. Per [SEP-2243](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2243-http-standardization.md) (accepted — see the [draft changelog](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/draft/changelog.mdx)), the Streamable HTTP transport requires `Mcp-Method` and `Mcp-Name` headers on POST requests so servers/intermediaries can route without parsing the JSON-RPC body. The Python client did not send them.

## Changes

- `_set_mcp_request_headers()` sets, for outgoing POST messages:
  - `Mcp-Method`: the JSON-RPC method — for requests **and** notifications
  - `Mcp-Name`: `params.name` (tools/prompts) or, when absent, `params.uri` (resources)
  - responses/errors (no method) get neither header
- `_encode_mcp_header_value()` applies the SEP-2243 value-encoding rules to `Mcp-Name`: safe printable-ASCII values pass through; non-ASCII, control characters, or significant leading/trailing whitespace are wrapped as `=?base64?<b64>?=` (matching the spec's examples).

The encoding is not just spec-compliance — without it, a tool name/URI containing non-ASCII raises `UnicodeEncodeError` and breaks the request, and a value containing CR/LF is a header-injection vector. Verified both: `café` now sends as `=?base64?Y2Fmw6k=?=` and is accepted by httpx; `"a\r\nEvil: 1"` is base64-wrapped.

The broader `x-mcp-header` custom-header part of SEP-2243 is intentionally out of scope here (separate feature); this PR covers the `Mcp-Method`/`Mcp-Name` requirement from #2715.

## Tests

- `test_set_mcp_request_headers` — request with `name`, with `uri`, no params, non-string `name`, notification, response, error, and a non-ASCII name (end-to-end encoding).
- `test_encode_mcp_header_value` — passthrough + each base64 trigger (non-ASCII, whitespace, control char, sentinel collision), using SEP-2243's own example values.

Full `tests/shared/test_streamable_http.py` passes (76 tests); `ruff` and `pyright` clean.

```
uv run --frozen pytest tests/shared/test_streamable_http.py
```

> Authored with the assistance of an AI coding assistant; reviewed and verified by me.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations as _annotations
 
+import base64
 import contextlib
 import logging
 from collections.abc import AsyncGenerator, Awaitable, Callable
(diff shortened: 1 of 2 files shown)


vidigoat:
Updated to also apply SEP-2243's value-encoding rules to `Mcp-Name`. Without it, a tool name or URI containing non-ASCII characters raises `UnicodeEncodeError` and breaks the request, and CR/LF in a value is a header-injection vector. Now safe printable-ASCII values pass through unchanged and everything else is wrapped as `=?base64?<b64>?=`, matching the spec's examples. Added `test_encode_mcp_header_value` covering passthrough plus each base64 trigger (non-ASCII, leading/trailing whitespace, control char, sentinel collision).

vidigoat:
Closing as superseded — the SEP-2243 header support has since landed upstream via #2950, #2974, and #2990, which together are a strict superset of this change (same base64 sentinel encoding and name/uri mapping), and test_legacy_wire now pins the 2025-negotiated behavior this PR would have conflicted with. Nothing here that main doesn't already do better. Thanks!

PR #2731, opened 2026-05-30
fix: send Origin header for streamable HTTP

### LINKED ISSUE #2727: streamable_http client does not send `Origin` header → rejected with 403 by spec-compliant servers (e.g. go-sdk `CrossOriginProtection`)

he-yufeng:
﻿Fixes #2727.

## Summary
- derive an `Origin` header from the streamable HTTP endpoint for client requests
- keep a caller-provided `Origin` on custom `httpx.AsyncClient` headers
- cover the default and custom-Origin paths in streamable HTTP client tests

## To verify
- `uv run pytest tests/shared/test_streamable_http.py -q`
- `uv run ruff check src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`
- `uv run ruff format --check src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`
- `uv run pyright src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`
- `python -m py_compile src\mcp\client\streamable_http.py tests\shared\test_streamable_http.py`

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -79,10 +79,12 @@ class StreamableHTTPTransport:
             url: The endpoint URL.
         """
         self.url = url
+        parsed_url = httpx.URL(url)
+        self.origin = f"{parsed_url.scheme}://{parsed_url.netloc.decode()}" if parsed_url.netloc else None
         self.session_id: str | None = None
         self.protocol_version: str | None = None
 
-    def _prepare_headers(self) -> dict[str, str]:
+    def _prepare_headers(self, client: httpx.AsyncClient | None = None) -> dict[str, str]:
         """Build MCP-specific request headers.
 
         These headers will be merged with the httpx.AsyncClient's default headers,
@@ -92,6 +94,8 @@ class StreamableHTTPTransport:
             "accept": "application/json, text/event-stream",
             "content-type": "application/json",
         }
+        if self.origin and (client is None or "origin" not in client.headers):
+            headers["origin"] = self.origin
         # Add session headers if available
         if self.session_id:
             headers[MCP_SESSION_ID] = self.session_id
@@ -189,7 +193,7 @@ class StreamableHTTPTransport:
                 if not self.session_id:
                     return
 
-                headers = self._prepare_headers()
+                headers = self._prepare_headers(client)
                 if last_event_id:
                     headers[LAST_EVENT_ID] = last_event_id
 
@@ -225,7 +229,7 @@ class StreamableHTTPTransport:
 
     async def _handle_resumption_request(self, ctx: RequestContext) -> None:
         """Handle a resumption request using GET with SSE."""
-        headers = self._prepare_headers()
+        headers = self._prepare_headers(ctx.client)
         if ctx.metadata and ctx.metadata.resumption_token:
             headers[LAST_EVENT_ID] = ctx.metadata.resumption_token
         else:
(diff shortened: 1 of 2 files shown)


he-yufeng:

PR #2742, opened 2026-05-31
fix: surface stateful HTTP session crash cause

### LINKED ISSUE #2741: Stateful Streamable HTTP: in-session exceptions are masked as an empty -32603; real cause only in a separate "Session crashed" log
### Initial Checks

## Summary

In **stateful** Streamable HTTP (`stateless_http=False`), when a session's `app.run()` raises, the client receives JSON-RPC `-32603` with the empty message `"Error handling POST request: "`, and the only prominent server log is `anyio.ClosedResourceError`. The real exception is logged separately as `Session <id> crashed` and is never tied to the request — so the actual cause is nearly invisible from the client or a quick log scan.

This is **distinct from the DoS crash fixed in 1.10.0** (GHSA-j975-95f5-7wqh): the process stays up. This is the remaining **error-propagation / diagnosability** gap.

## Observed

Repro server is in *Example Code* below; served via `uvicorn repro:app --port 8765` and hit with one `initialize` POST.

Client response — the real `RuntimeError` is absent:

```
HTTP/1.1 500 Internal Server Error
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32603,"message":"Error handling POST request: "}}
```

Server log — the truth, in a separate task:

```
Session <id> crashed
  RuntimeError: BOOM-distinctive-root-cause
Error handling POST request
  anyio.ClosedResourceError
Exception in ASGI application
  anyio.ClosedResourceError
```

(My real-world trigger was `OSError: [Errno 24] inotify instance limit reached` from a per-session resource; every request then 500'd with an opaque `ClosedResourceError`, hiding the `OSError` entirely.)

## Root cause

`_handle_stateful_request` (`streamable_http_manager.py`) starts `run_server` (runs `app.run()` on the session's memory streams), then concurrently calls `http_transport.handle_request()`. If `app.run()` raises, `run_server` logs `Session … crashed` and `connect()` tears down the streams. The concurrent `_handle_post_request` (`streamable_http.py`) then:

1. raises `ClosedResourceError` at `writer.send(session_message)` (session read-stream already closed) — it never sees the real exception, only "stream closed";
2. builds the 500 via `_create_error_response(f"Error handling POST request: {err}")`, but `err` is now that `ClosedResourceError`, whose `str()` is empty → empty client message;
3. calls `writer.send(Exception(err))` into the already-closed stream, raising `ClosedResourceError` again.

The real exception is decoupled from, and invisible to, the request and the client.

## Expected

The client error and/or request-path log should surface the actual exception that crashed the session, not an opaque empty `-32603`.

## Suggested fix

- Capture `run_server`'s exception on the transport and surface it from `_handle_post_request`.
- Guard the trailing `writer.send(Exception(err))` with `try/except (ClosedResourceError, BrokenResourceError)`.
- When `str(err)` is empty, fall back to `repr(err)` / the exception type.

## Related

- GHSA-j975-95f5-7wqh / #967 (1.10.0) fixed the *crash*, not propagation.
- #1219, #1658 — *stateless* `ClosedResourceError`-on-connect.
- #1304 — *stateless* lifespan-per-request.

This is the **stateful error-propagation** gap that remains.

### Example Code

```Python
# Minimal repro — confirmed on mcp 1.27.2 (latest), Python 3.14
# Serve with:  uvicorn repro:app --port 8765

@asynccontextmanager
async def lifespan(server):
    # FastMCP's server lifespan re-enters PER SESSION in stateful mode.
    # Simulate any per-session startup failure:
    raise RuntimeError("BOOM-distinctive-root-cause")
    yield {}

mcp = FastMCP(
    "repro",
    stateless_http=False,
    json_response=True,
    lifespan=lifespan,
    transport_security=TransportSecuritySettings(enable_dns_rebinding_protection=False),
)

@mcp.tool()
def ping() -> str:
    return "pong"

# Then send one initialize POST (single line):
# curl -s -i -X POST http://127.0.0.1:8765/mcp -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"c","version":"1"}}}'
```

### Python & MCP Python SDK

```Text
Python 3.14 · mcp 1.27.2 (latest) · anyio 4.13 · starlette 1.2.1. (Originally observed on 1.27.1; confirmed identical on 1.27.2.)
```

he-yufeng:
## Summary
- surface the original stateful session `app.run()` exception when a POST fails after the session stream closes
- clean up the per-request memory stream on that failure path
- avoid a secondary `ClosedResourceError` when reporting the failure back to an already closed writer

Fixes #2741.

## To verify
- `python -m pytest tests\server\test_streamable_http_manager.py -q`
- `python -m ruff check src\mcp\server\streamable_http.py src\mcp\server\streamable_http_manager.py tests\server\test_streamable_http_manager.py`
- `python -m ruff format --check src\mcp\server\streamable_http.py src\mcp\server\streamable_http_manager.py tests\server\test_streamable_http_manager.py`
- `git diff --check -- src\mcp\server\streamable_http.py src\mcp\server\streamable_http_manager.py tests\server\test_streamable_http_manager.py`

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -10,7 +10,7 @@ import logging
 import re
 from abc import ABC, abstractmethod
 from collections.abc import AsyncGenerator, Awaitable, Callable
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, suppress
 from dataclasses import dataclass
 from http import HTTPStatus
 from typing import Any
@@ -171,6 +171,7 @@ class StreamableHTTPServerTransport:
         ] = {}
         self._sse_stream_writers: dict[RequestId, MemoryObjectSendStream[dict[str, str]]] = {}
         self._terminated = False
+        self._session_run_error: BaseException | None = None
         # Idle timeout cancel scope; managed by the session manager.
         self.idle_scope: anyio.CancelScope | None = None
 
@@ -178,6 +179,16 @@ class StreamableHTTPServerTransport:
     def is_terminated(self) -> bool:
         """Check if this transport has been explicitly terminated."""
         return self._terminated
+
+    def note_session_run_error(self, exc: BaseException) -> None:
+        self._session_run_error = exc
+
+    def _post_error_message(self, err: BaseException) -> str:
+        display_error = self._session_run_error or err
+        display_error_text = str(display_error) or type(display_error).__name__
+        if display_error is not err and str(display_error):
+            display_error_text = f"{type(display_error).__name__}: {display_error_text}"
+        return f"Error handling POST request: {display_error_text}"
 
     def close_sse_stream(self, request_id: RequestId) -> None:
         """Close SSE connection for a specific request without terminating the stream.
@@ -361,6 +372,10 @@ class StreamableHTTPServerTransport:
                 # Remove the request stream from the mapping
                 self._request_streams.pop(request_id, None)
 
+    async def _clean_up_post_request_stream(self, request_id: RequestId | None) -> None:
+        if request_id is not None:
+            await self._clean_up_memory_streams(request_id)
+
     async def handle_request(self, scope: Scope, receive: Receive, send: Send) -> None:
         """Application entry point that handles all HTTP requests."""
         request = Request(scope, receive)
@@ -441,6 +456,7 @@ class StreamableHTTPServerTransport:
         writer = self._read_stream_writer
         if writer is None:  # pragma: no cover
             raise ValueError("No read stream writer available. Ensure connect() is called first.")
+        request_id: RequestId | None = None
         try:
             # Validate Accept header
             if not await self._validate_accept_header(request, scope, send):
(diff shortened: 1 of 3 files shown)


he-yufeng:
Rebased this on current `main` and carried the crash-cause reporting over the updated streamable HTTP code.

Validation run locally on Windows:

```text
python -m py_compile src\mcp\server\streamable_http.py src\mcp\server\streamable_http_manager.py tests\server\test_streamable_http_manager.py
uv run pytest tests/server/test_streamable_http_manager.py -q -k "session_crash_cause or post_error_message" --basetemp .tmp\pytest-2742-20260610 -p no:cacheprovider
# 2 passed, 20 deselected

git diff --check upstream/main..HEAD
```

The error log lines in that targeted pytest run are expected: the new regression test deliberately crashes the session runner and asserts that the JSON response reports the original session exception instead of only the downstream closed-stream error.

he-yufeng:
Closing this. Since I opened it, `main` has reworked the streamable-HTTP POST error path in the opposite direction — it logs the exception server-side and returns a generic "Error handling POST request" message rather than surfacing the underlying error to the client. This PR's whole point was to surface the session crash cause, which now cuts against that deliberate choice, so it's cleaner to drop it than to force a conflicting reconcile. Thanks for the refactor.

PR #2743, opened 2026-05-31
fix: prevent CRLF translation in stdio_server on Windows

### LINKED ISSUE #2433: Windows: TextIOWrapper in stdio_server() emits CRLF instead of LF, corrupting newline-delimited JSON messages

Epochex:
Fixes #2433.

On Windows, TextIOWrapper defaults to newline=None and translates \n -> \r\n on output. stdio_server wraps sys.stdout.buffer with TextIOWrapper, so every JSON-RPC message was emitted with CRLF line endings.

This sets newline= on the stdio stdin/stdout wrappers to disable platform newline translation and keep LF-delimited NDJSON.

Tests:
- uv run --frozen pytest tests/server/test_stdio.py

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -39,9 +39,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     # python is platform-dependent (Windows is particularly problematic), so we
     # re-wrap the underlying binary stream to ensure UTF-8.
     if not stdin:
-        stdin = anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8", errors="replace"))
+        stdin = anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8", errors="replace", newline=""))
     if not stdout:
-        stdout = anyio.wrap_file(TextIOWrapper(sys.stdout.buffer, encoding="utf-8"))
+        stdout = anyio.wrap_file(TextIOWrapper(sys.stdout.buffer, encoding="utf-8", newline=""))
 
     read_stream_writer, read_stream = create_context_streams[SessionMessage | Exception](0)
     write_stream, write_stream_reader = create_context_streams[SessionMessage](0)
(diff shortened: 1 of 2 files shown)


yyzquwu:
I ran an independent Windows-focused repro packet from Agent Windows Lab and it supports the direction of this PR.

Environment/evidence:
- Windows 11 `10.0.26200`, Python `3.13.13`
- `stdio_newline_framing`: Python text-mode stdout emitted CRLF (`0d0a`) while binary stdout emitted LF (`0a`)
- `mcp_stdio_jsonrpc_probe`: binary stdio round-tripped an MCP-style JSON-RPC `initialize` message with LF framing
- The generated report was redacted and passed the redaction verifier

Command used:

```powershell
python .\scripts\run_agent_windows_lab.py --out .\artifacts\issue-packet --case mcp --issue-target modelcontextprotocol-python-sdk
python .\scripts\verify_redacted_report.py .\artifacts\issue-packet\agent-windows-lab-report.json
```

Public harness commit: https://github.com/yyzquwu/agent-windows-lab/commit/ab5e7149006df3638e0e2c8711af629b18725d43
Windows proof run: https://github.com/yyzquwu/agent-windows-lab/actions/runs/26735979548

This gives independent Windows evidence for treating stdio framing as byte-level transport behavior and preventing text-mode newline translation from leaking into JSON-RPC framing.

Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2552.

PR #2746, opened 2026-06-01
fix: gracefully terminate active Streamable HTTP sessions during shutdown (#2150)

### LINKED ISSUE #2150: Active Streamable HTTP sessions are not terminated during shutdown

kaushik701:
Fixes #2150

## Problem

When shutting down a FastMCP/MCP server (CTRL+C) while clients have active streaming connections, Uvicorn logs:

```
ERROR: ASGI callable returned without completing response.
```

There were two compounding issues:

1. `StreamableHTTPSessionManager.run()` never called `terminate()` on active transports before cancelling the task group — it just cleared the dict, leaving `EventSourceResponse` coroutines killed mid-stream.
2. `terminate()` itself didn't close `_sse_stream_writers`. It only cleaned up `_request_streams`, but the SSE stream writers (which keep `EventSourceResponse` alive) were stored separately and never touched, so `EventSourceResponse` kept waiting on its reader.
## Fix

**`src/mcp/server/streamable_http_manager.py`** — In the `run()` method's `finally` block, iterate over all active transports and call `terminate()` on each before cancelling the task group:

```python
finally:
    logger.info("StreamableHTTP session manager shutting down")
    # Gracefully terminate all active sessions before cancelling tasks
    for transport in list(self._server_instances.values()):
        try:
            await transport.terminate()
        except Exception:
            logger.debug("Error terminating transport during shutdown", exc_info=True)
    # Cancel task group to stop all spawned tasks
    tg.cancel_scope.cancel()
    ...
```

**`src/mcp/server/streamable_http.py`** — In `terminate()`, close all `_sse_stream_writers` before cleaning up request streams, so active `EventSourceResponse` instances can complete gracefully:

```python
# Close all SSE stream writers to allow EventSourceResponse to complete gracefully
for writer in list(self._sse_stream_writers.values()):
    writer.close()
self._sse_stream_writers.clear()
```

## Notes

- Stateless mode is not affected (transports are already terminated per-request).
- - The `_terminated` flag ensures that any concurrent requests arriving during shutdown get a 404, not a partial response.
- - Using sync `.close()` on `MemoryObjectSendStream` (rather than `await aclose()`) is safe and avoids awaiting inside the shutdown path.

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -771,6 +771,11 @@ class StreamableHTTPServerTransport:
         self._terminated = True
         logger.info(f"Terminating session: {self.mcp_session_id}")
 
+        # Close all SSE stream writers to allow EventSourceResponse to complete gracefully
+        for writer in list(self._sse_stream_writers.values()):
+            writer.close()
+        self._sse_stream_writers.clear()
+
         # We need a copy of the keys to avoid modification during iteration
         request_stream_keys = list(self._request_streams.keys())
 
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -133,6 +133,12 @@ class StreamableHTTPSessionManager:
                 yield  # Let the application run
             finally:
                 logger.info("StreamableHTTP session manager shutting down")
+                # Gracefully terminate all active sessions before cancelling tasks
+                for transport in list(self._server_instances.values()):
+                    try:
+                        await transport.terminate()
+                    except Exception:
+                        logger.debug("Error terminating transport during shutdown", exc_info=True)
                 # Cancel task group to stop all spawned tasks
                 tg.cancel_scope.cancel()
                 self._task_group = None


StantonMatt:
I think this still needs regression coverage before merge.

The current-head matrix is red in every `checks / test (...)` job. A representative full-suite job, `checks / test (3.10, locked, windows-latest)`, ran `1741 passed, 99 skipped, 1 xfailed`, then failed coverage at `99.99 < 100.00` with these missing lines:

- `src/mcp/server/streamable_http.py:1050` (`connect()`'s `await self._clean_up_memory_streams(stream_id)` cleanup path)
- `src/mcp/server/streamable_http_manager.py:140-141` (the defensive `except Exception` branch around `transport.terminate()` during manager shutdown)

That first missed line looks tied to the new shutdown ordering: manager shutdown now calls `transport.terminate()` before cancelling the task group, so request streams can already be cleared before `connect()` reaches its own cleanup loop. The behavior may be fine, but with this repo's 100% coverage gate this change needs tests for the new shutdown/cleanup paths.

Local checks on head `e37d8aaaa3fb9f8fc894d921240207929bf03b1f`:

- `uv run --frozen pytest tests/server/test_streamable_http_manager.py -q` -> 20 passed
- `uv run --frozen pytest tests/shared/test_streamable_http.py::test_server_close_sse_stream_via_context tests/shared/test_streamable_http.py::test_streamable_http_client_auto_reconnects tests/shared/test_streamable_http.py::test_streamable_http_client_receives_priming_event -q` -> 3 passed
- `uv run --frozen ruff check src/mcp/server/streamable_http.py src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py tests/shared/test_streamable_http.py` -> passed
- `uv run --frozen pyright src/mcp/server/streamable_http.py src/mcp/server/streamable_http_manager.py tests/server/test_streamable_http_manager.py tests/shared/test_streamable_http.py` -> 0 errors
- `git diff --check origin/main...HEAD` -> clean

I also ran two small runtime probes: `terminate()` with a registered `_sse_stream_writers['req-1']` cleared the writer map and the paired reader observed `EndOfStream`; `StreamableHTTPSessionManager.run()` shutdown called all three fake active transports' `terminate()` methods and continued after one raised. So I don't see a behavior blocker from those probes, just the missing regression coverage that CI is already flagging.

One Ubuntu shard also failed `test_streamable_http_client_receives_priming_event` with `httpx.ConnectError: All connection attempts failed`; I could not reproduce that focused test locally, so I would treat the repeatable blocker as the coverage gap above.

Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2253.

PR #2038, opened 2026-02-11
test: fix progress notification assertions for related_request_id

akshan-main:
## Summary

Fixes failing CI after #2002 added `related_request_id` to progress notifications.

The `report_progress()` method now passes `related_request_id=self.request_id` to `send_progress_notification()`, but the test in `test_176_progress_token.py` was not updated to expect this new argument. This caused all 20 test matrix jobs to fail.

Updated the three `assert_any_call` assertions to include `related_request_id="test-request"`. Added unit test

PR #2747, opened 2026-06-01
test: add regression test for issue #2001 (progress related_request_id)

### LINKED ISSUE #2001: Progress notifications not delivered via SSE in stateless HTTP mode
### Initial Checks

## Summary

`Context.report_progress()` notifications are not delivered to clients when using the streamable HTTP transport, while `Context.info()` and other logging methods work correctly.

## Root Cause

In `src/mcp/server/fastmcp/server.py`, the `report_progress` method does not pass `related_request_id` to `send_progress_notification()`:

```python
async def report_progress(self, progress: float, total: float | None = None, message: str | None = None) -> None:
    progress_token = self.request_context.meta.progressToken if self.request_context.meta else None

    if progress_token is None:
        return

    await self.request_context.session.send_progress_notification(
        progress_token=progress_token,
        progress=progress,
        total=total,
        message=message,
        # related_request_id is missing here
    )
```

The SSE transport uses `related_request_id` in the message metadata to route notifications to the correct client. Without it, progress notifications are silently dropped.

In contrast, `send_log_message` (used by `info()`, `warning()`, etc.) correctly passes `related_request_id`:

## Fix

Add `related_request_id=self.request_id` to the `send_progress_notification()` call:
```python
await self.request_context.session.send_progress_notification(
    progress_token=progress_token,
    progress=progress,
    total=total,
    message=message,
    related_request_id=self.request_id,
)
```

## Reproduction

1. Create a FastMCP server with `stateless_http=True`
2. In a tool, call both `await ctx.info("test")` and `await ctx.report_progress(0.5, total=1.0, message="test")`
3. Observe that `info()` messages are delivered via SSE but `report_progress()` messages are not

## Environment

- MCP SDK version: 1.26.0
- Python version: 3.11
- Transport: Streamable HTTP (`stateless_http=True`)

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
- MCP SDK version: 1.26.0
- Python version: 3.11
```

kaushik701:
Adds a dedicated regression test file for issue #2001.

Context.report_progress() was silently dropping progress notifications in stateless HTTP / SSE transports because send_progress_notification() was called without related_request_id. The fix (already in main via mcpserver/context.py) passes related_request_id=self.request_id, consistent with send_log_message().

This test file covers:
- The primary regression: related_request_id is forwarded on every call
- Edge case: no progress_token -> notification is skipped (no-op)
- Edge case: integer progress_token (e.g. 0) works correctly

Closes #2001

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/tests/issues/test_2001_progress_related_request_id.py
+++ b/tests/issues/test_2001_progress_related_request_id.py
@@ -13,75 +13,75 @@ from mcp.server.mcpserver import Context
 pytestmark = pytest.mark.anyio
 
 async def test_report_progress_passes_related_request_id() -> None:
-      """report_progress must forward request_id as related_request_id."""
-      mock_session = AsyncMock()
-      mock_session.send_progress_notification = AsyncMock()
-      request_context = ServerRequestContext(
-      request_id="req-2001",
-      session=mock_session,
-      meta={"progress_token": "tok-progress"},
-      lifespan_context=None,
-      experimental=Experimental(),
-      )
-      ctx = Context(request_context=request_context, mcp_server=MagicMock())
-      await ctx.report_progress(25, 100, message="quarter done")
-      await ctx.report_progress(50, 100)
-      await ctx.report_progress(100, 100, message="complete")
-      assert mock_session.send_progress_notification.call_count == 3
-      mock_session.send_progress_notification.assert_any_call(
-      progress_token="tok-progress",
-      progress=25.0,
-      total=100.0,
-      message="quarter done",
-      related_request_id="req-2001",
-      )
-      mock_session.send_progress_notification.assert_any_call(
-      progress_token="tok-progress",
-      progress=50.0,
-      total=100.0,
-      message=None,
-      related_request_id="req-2001",
-      )
-      mock_session.send_progress_notification.assert_any_call(
-      progress_token="tok-progress",
-      progress=100.0,
-      total=100.0,
-      message="complete",
-      related_request_id="req-2001",
-      )
+    """report_progress must forward request_id as related_request_id."""
+    mock_session = AsyncMock()
+    mock_session.send_progress_notification = AsyncMock()
+    request_context = ServerRequestContext(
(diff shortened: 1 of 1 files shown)


StantonMatt (commented):
I checked this against the current test tree and I think this PR mostly duplicates coverage that is already on `main`.

The main regression for #2001 is already covered by `tests/server/mcpserver/test_server.py::test_report_progress_passes_related_request_id`, including the `related_request_id` assertion. The integer `progress_token=0` case is already covered by `tests/issues/test_176_progress_token.py::test_progress_token_zero_first_call`, also with `related_request_id`. The no-token no-op behavior is covered at the interaction level by `tests/interaction/mcpserver/test_context.py::test_report_progress_without_a_progress_token_sends_nothing`.

I verified this branch and the overlapping tests locally on `0d3e9ef`:

- `uv run --frozen pytest tests/issues/test_2001_progress_related_request_id.py tests/server/mcpserver/test_server.py::test_report_progress_passes_related_request_id tests/issues/test_176_progress_token.py -q` -> 5 passed
- `uv run --frozen ruff check tests/issues/test_2001_progress_related_request_id.py tests/server/mcpserver/test_server.py tests/issues/test_176_progress_token.py`
- `uv run --frozen ruff format --check tests/issues/test_2001_progress_related_request_id.py tests/server/mcpserver/test_server.py tests/issues/test_176_progress_token.py`
- `uv run --frozen pyright tests/issues/test_2001_progress_related_request_id.py`

So the tests pass, but I do not see much new coverage from adding this extra issue file as-is. If maintainers want an issue-specific file for #2001, it may be worth trimming it to only a missing case, otherwise this can probably be closed in favor of the existing regression and interaction tests.

maxisbey:
Thanks for the PR. This has since landed via #2038. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2749, opened 2026-06-01
fix(resources): escape literal regex metacharacters in ResourceTemplate.matches

vidigoat:
## Problem

`ResourceTemplate.matches()` (`src/mcp/server/mcpserver/resources/templates.py`) builds its matching regex by naive string substitution:

```python
pattern = self.uri_template.replace("{", "(?P<").replace("}", ">[^/]+)")
match = re.match(f"^{pattern}$", uri)
```

The **literal** portions of the template are never `re.escape`-d, so any regex metacharacter in the template text is interpreted as a regex operator. This causes both:

- **False positives** — `api://v1.0/{version}` treats `.` as "any character" and wrongly matches `api://v1X0/abc`, routing a URI to a template it shouldn't match (returning `{'version': 'abc'}` instead of `None`).
- **False negatives** — a template with `+`, `(`, `[`, etc. in a literal segment (e.g. `res://a+b/{x}`) fails to match its own valid URI.

## Fix

Tokenize the template into literal/placeholder parts, `re.escape` the literal text, and turn each `{param}` into a named capture group:

```python
parts: list[str] = []
for literal, param in re.findall(r"([^{]*)(?:\{(\w+)\})?", self.uri_template):
    parts.append(re.escape(literal))
    if param:
        parts.append(f"(?P<{param}>[^/]+)")
pattern = "".join(parts)
```

## Verification

```
before:  pytest tests/server/mcpserver/resources/test_resource_template.py -k escapes_literal_regex
         1 failed   (matches('api://v1X0/abc') returned {'version': 'abc'}, expected None)
after :  16 passed  (full test_resource_template.py)
no regressions: tests/server/mcpserver/resources/ -> 60 passed;
                tests/issues/test_129/test_141 resource templates -> green
```

`ruff check` and `ruff format --check` pass on both changed files. Adds a regression test asserting metacharacters in literal segments are matched literally.

### DIFF
--- a/tests/server/mcpserver/resources/test_resource_template.py
+++ b/tests/server/mcpserver/resources/test_resource_template.py
@@ -157,6 +157,47 @@ def test_matches_escapes_template_literals():
     assert t.matches("data://v1X0/42") is None
 
 
+def test_matches_literal_dot_does_not_route_to_wrong_template():
+    """A "." in a template's literal text matches only a literal dot (#2961).
+
+    The old regex-based matcher let "api://v1.0/{v}" match "api://v1X0/abc",
+    so an earlier registered template could capture URIs meant for a later one.
+    """
+    t = _make("api://v1.0/{version}")
+    assert t.matches("api://v1.0/abc") == {"version": "abc"}
+    assert t.matches("api://v1X0/abc") is None
+
+
+def test_matches_literal_regex_metacharacters_carry_no_regex_meaning():
+    """Quantifiers and grouping characters in literal text match only themselves (#2961).
+
+    Under the old regex-based matcher, "+" / "*" / "?" acted as quantifiers and
+    "(" ")" created groups, producing false negatives on the template's own URIs.
+    """
+    plus = _make("res://a+b/{x}")
+    assert plus.matches("res://a+b/v") == {"x": "v"}
+    assert plus.matches("res://aaab/v") is None  # "+" is not one-or-more
+
+    star = _make("res://a*b/{x}")
+    assert star.matches("res://a*b/v") == {"x": "v"}
+    assert star.matches("res://b/v") is None  # "*" is not zero-or-more
+
+    optional = _make("res://a?b/{x}")
+    assert optional.matches("res://a?b/v") == {"x": "v"}
+    assert optional.matches("res://ab/v") is None  # "?" is not optionality
+
+    parens = _make("res://a(b)c/{x}")
+    assert parens.matches("res://a(b)c/v") == {"x": "v"}
+    assert parens.matches("res://abc/v") is None  # "(" ")" is not a group
+
+
+def test_matches_template_of_every_regex_metacharacter_matches_itself():
+    """A literal run of all the regex metacharacters matches only itself (#2961)."""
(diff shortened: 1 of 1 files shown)


StantonMatt (commented):
I checked this against the resource-manager routing surface, since the bad `.` behavior can let an earlier template catch a URI intended for a later one. With `api://v1.0/{version}` registered before `api://v1X0/{version}`, `ResourceManager.get_resource(api://v1X0/abc)` now returns the `v1X0` template result (`x:abc`), while `api://v1.0/abc` still returns the dotted template result (`dot:abc`).

Local checks:
- `uv run --frozen pytest tests/server/mcpserver/resources/test_resource_template.py tests/issues/test_973_url_decoding.py -q`
- `uv run --frozen ruff check src/mcp/server/mcpserver/resources/templates.py tests/server/mcpserver/resources/test_resource_template.py`

vidigoat:
Gentle nudge on this one whenever a maintainer has a moment. CI is green, and @StantonMatt independently reproduced the template-routing issue and verified the fix against the resource-manager surface. It's a small, self-contained change (escaping literal regex metacharacters in `ResourceTemplate.matches`) with a regression test. Happy to adjust anything — thanks for your time!

Robin1987China (commented):
Fix logic is correct: `re.findall` splits the template into literal/param pairs, `re.escape()` on literals avoids metacharacter interpretation. The `(\w+)` for parameter names properly constrains to valid Python identifiers (matching MCPServer's function parameter constraint).

The test covers `.` and `+` — good minimal coverage. Suggestion: consider adding a few more boundary cases for completeness, since the fix changes regex behavior for ALL special characters:

- `*` (asterisk) — could match zero-or-more incorrectly  
- `?` (question mark) — could act as optional quantifier
- `(` `)` (parentheses) — could create false capture groups
- A full-special-char regression test (e.g., `data:.+*?^$|()[]{name}`) to validate all metacharacters are escaped

These would prevent regressions if the regex construction is refactored later. Happy to share test cases if helpful.

AI assistance disclosure: Reviewed with AI assistance (opencode). I have verified the fix logic against the original bug report (#2961).

vidigoat:
@StantonMatt Rebased onto latest main, and there's a wrinkle worth flagging for the routing scenario you verified: main's RFC 6570 rewrite (mcp.shared.uri_template) replaced the regex construction in ResourceTemplate.matches entirely, and the new matcher compares literal text literally — so api://v1.0/{version} no longer captures api://v1X0/abc even without my code change. I've reduced this PR to the regression tests (as top-level functions, per current test conventions) so that behavior stays pinned if matching is ever refactored again.

vidigoat:
@Robin1987China Added the boundary cases you listed: `*`, `?`, and `()` each get a literal-match plus a no-regex-meaning assertion, and there's a full run of every metacharacter (`data:.+*?^$|()[]{name}`) asserting it matches only itself. After rebasing onto main, the RFC 6570 UriTemplate rewrite has replaced the regex construction and handles literals without regex at all, so the code change is subsumed — the PR is now regression tests only, which pins exactly the refactor-safety concern you raised. All pass against the new matcher.

PR #2753, opened 2026-06-01
fix(clientsessiongroup): only query negotiated capabilities

### LINKED ISSUE #2689: `ClientSessionGroup` attempts to use capabilities that were not advertised
### Initial Checks

### Description

`ClientSessionGroup` calls `session.list_tools`, `session.list_prompts()`, and `session.list_resources()` unconditionally on every connect. It never consults the `ServerCapabilities` returned by `initialize`.

For a server that doesn't implement tools, prompts, or resources, the corresponding calls return a JSON-RPC `Method not found` error. The `MCPError` is squashed and a `WARNING` is emitted, e.g.,

```
WARNING:root:Could not fetch prompts: Method not found
```

This seems to contradict the MCP lifecycle spec ([2025-11-25, *Capability Negotiation* / *Operation*](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#operation)):

> Both parties **MUST**: ... Only use capabilities that were successfully negotiated.

`ClientSessionGroup` should inspect `session.initialize_result.capabilities` and only interrogate capabilities that the server advertised, e.g.,

```python
caps = session.initialize_result.capabilities if session.initialize_result else None

if caps and caps.prompts is not None:
    prompts = (await session.list_prompts()).prompts
    ...
if caps and caps.resources is not None:
    resources = (await session.list_resources()).resources
    ...
if caps and caps.tools is not None:
    tools = (await session.list_tools()).tools
    ...
```

### Example Code

```Python
"""Repro: tools-only server + ClientSessionGroup emits WARNINGs."""

import anyio

from mcp import Client, types
from mcp.client.session_group import ClientSessionGroup
from mcp.server import Server, ServerRequestContext


async def handle_list_tools(
    ctx: ServerRequestContext, params: types.PaginatedRequestParams | None
) -> types.ListToolsResult:
    return types.ListToolsResult(
        tools=[
            types.Tool(
                name="ping",
                description="Ping",
                input_schema={"type": "object", "properties": {}},
            )
        ]
    )


async def handle_call_tool(
    ctx: ServerRequestContext, params: types.CallToolRequestParams
) -> types.CallToolResult:
    return types.CallToolResult(content=[types.TextContent(type="text", text="pong")])


async def main() -> None:
    server = Server(
        "tools-only-server",
        on_list_tools=handle_list_tools,
        on_call_tool=handle_call_tool,
    )

        # Server omits prompts/resources capabilities:
        assert client.initialize_result.capabilities.prompts is None
        assert client.initialize_result.capabilities.resources is None

        group = ClientSessionGroup()
        await group.connect_with_session(client.initialize_result.server_info, client.session)


if __name__ == "__main__":
    anyio.run(main)
```

### Python & MCP Python SDK

```Text
Python: 3.14.4
MCP Python SDK: main @ da4c118 (also reproducible on v1.27.1)
OS: Linux 6.6.87.2-microsoft-standard-WSL2 x86_64
```

mr-brobot:
## Summary

Fixes #2689

`ClientSessionGroup` queried prompts, resources, and tools unconditionally on every connect, ignoring the `ServerCapabilities` returned by `initialize()`. A server advertising only some capabilities (e.g. tools) returned JSON-RPC `Method not found` for the rest, which was swallowed into spurious `WARNING` logs — contrary to the [lifecycle spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#operation): clients MUST "Only use capabilities that were successfully negotiated."

## Changes

- `_aggregate_components` now gates each `list_*` call on the matching capability from `session.initialize_result.capabilities`.
- The `MCPError` fallback is preserved for servers that advertise a capability but still fail the method; a missing `initialize_result` falls back to the prior unconditional behavior.
- Adds `test_client_session_group_skips_unsupported_capabilities`, parametrized over tools/prompts/resources-only servers.

The same code path exists on `v1.x` and likely warrants a `[v1.x]` backport.

## Note

I followed the [suggestion](https://github.com/modelcontextprotocol/python-sdk/issues/2689#issuecomment-4592013818) strictly, but I think the fallback behavior for "misbehaving servers" has questionable value. Sharing as-is for feedback.

### DIFF
--- a/src/mcp/client/session_group.py
+++ b/src/mcp/client/session_group.py
@@ -344,36 +344,45 @@ class ClientSessionGroup:
         tools_temp: dict[str, types.Tool] = {}
         tool_to_session_temp: dict[str, mcp.ClientSession] = {}
 
+        # Per the lifecycle spec, only invoke methods for capabilities the
+        # server advertised during initialization. If the initialize result
+        # is missing, fall back to the prior unconditional behavior so the
+        # existing MCPError handler can still cope with servers that misbehave.
+        capabilities = session.initialize_result.capabilities if session.initialize_result is not None else None
+
         # Query the server for its prompts and aggregate to list.
-        try:
-            prompts = (await session.list_prompts()).prompts
-            for prompt in prompts:
-                name = self._component_name(prompt.name, server_info)
-                prompts_temp[name] = prompt
-                component_names.prompts.add(name)
-        except MCPError as err:  # pragma: no cover
-            logging.warning(f"Could not fetch prompts: {err}")
+        if capabilities is None or capabilities.prompts is not None:
+            try:
+                prompts = (await session.list_prompts()).prompts
+                for prompt in prompts:
+                    name = self._component_name(prompt.name, server_info)
+                    prompts_temp[name] = prompt
+                    component_names.prompts.add(name)
+            except MCPError as err:  # pragma: no cover
+                logging.warning(f"Could not fetch prompts: {err}")
 
         # Query the server for its resources and aggregate to list.
-        try:
-            resources = (await session.list_resources()).resources
-            for resource in resources:
-                name = self._component_name(resource.name, server_info)
-                resources_temp[name] = resource
-                component_names.resources.add(name)
-        except MCPError as err:  # pragma: no cover
-            logging.warning(f"Could not fetch resources: {err}")
+        if capabilities is None or capabilities.resources is not None:
+            try:
+                resources = (await session.list_resources()).resources
+                for resource in resources:
+                    name = self._component_name(resource.name, server_info)
+                    resources_temp[name] = resource
+                    component_names.resources.add(name)
+            except MCPError as err:  # pragma: no cover
+                logging.warning(f"Could not fetch resources: {err}")
 
         # Query the server for its tools and aggregate to list.
-        try:
-            tools = (await session.list_tools()).tools
-            for tool in tools:
-                name = self._component_name(tool.name, server_info)
-                tools_temp[name] = tool
-                tool_to_session_temp[name] = session
-                component_names.tools.add(name)
-        except MCPError as err:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


StantonMatt (commented):
I found one edge in the new capability-gated path. When an already-initialized session advertises none of `tools`, `prompts`, or `resources`, `connect_with_session()` now skips all three list calls and then hits the existing cleanup block:

```python
if not any((prompts_temp, resources_temp, tools_temp)):
    del self._session_exit_stacks[session]
```

For `connect_with_session()` there is no session exit stack registered, so this raises `KeyError` instead of connecting an empty session.

Local probe on this head:

```text
capabilities=ServerCapabilities()
await ClientSessionGroup().connect_with_session(..., session)
=> KeyError <AsyncMock spec='ClientSession' ...>
list_tools awaited 0
list_prompts awaited 0
list_resources awaited 0
```

The new parametrized test covers exactly one advertised capability at a time, but not the zero-capability case that now takes this skip-all path. A small guard around the cleanup, or a zero-capability regression test, should close it.

Focused checks I ran locally:
- `uv run --frozen pytest tests/client/test_session_group.py -q` -> 14 passed
- `uv run --frozen ruff check src/mcp/client/session_group.py tests/client/test_session_group.py`
- `uv run --frozen pyright src/mcp/client/session_group.py tests/client/test_session_group.py`

I also checked the red public `checks / test (3.12, lowest-direct, windows-latest)` job: the tests themselves completed (`1744 passed`, `99 skipped`, `1 xfailed`), then the job failed coverage at `99.96 < 100.00` on `tests\interaction\transports\_stdio_server.py`, which this branch does not touch.

maxisbey:
Thanks for the PR. Tracking this in #2698 instead. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

Issue #2064, opened 2026-02-14
ClientDisconnect during _handle_post_request crashes stateless session with ClosedResourceError


einarfd:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

  When a client disconnects while a stateless streamable-HTTP server is reading the request body, `_handle_post_request` catches the
  `ClientDisconnect` but the error handler in `_handle_message` (lowlevel/server.py:694) then tries to `send_log_message()` back to the
  client. Since the session was already terminated and the write stream closed, this raises `ClosedResourceError`, which is unhandled and
  crashes the stateless session with an `ExceptionGroup`.

  This is a **different code path** from what PR #1384 fixed. That PR addressed `ClosedResourceError` in the message router loop. This bug
  is in the error recovery path: catch exception → try to log it to client → write stream already closed → crash.

  ## Versions

  - **mcp**: 1.26.0 (also reproduced on 1.25.0; believed to affect >= 1.12.0)
  - **Python**: 3.14.2 (also reproducible on 3.12+)
  - **starlette**: 0.48.0
  - **uvicorn**: 0.34.3

  ## Steps to reproduce
  Run the attached repro script to reproduce the problem.

  ## Expected behavior

  The server should log a warning about the client disconnect and cleanly discard the failed request, without crashing the stateless session.

  ## Root cause

  In `lowlevel/server.py`, `_handle_message` has a catch-all exception handler (line ~690) that calls `session.send_log_message()` to notify
   the client about the error. When the error *is* a client disconnect, the write stream is already closed, so `send_log_message` →
  `send_notification` → `_write_stream.send()` raises `ClosedResourceError`. This is unhandled in the TaskGroup and crashes the session.

  A possible fix would be to catch `ClosedResourceError` (and/or `BrokenResourceError`) in the error handler at `_handle_message`, since
  failing to notify a disconnected client is expected and harmless.

  ## Related issues

  - #1190 — closed, partially fixed by PR #1384 (message router path only)
  - #1219 — closed as duplicate of #1190
  - #1658 — closed as duplicate of #1190

  None of these cover the `_handle_post_request` → `_handle_message` → `send_log_message` path.

### Example Code

```Python
"""Minimal reproduction: MCP SDK crashes with ClosedResourceError on client disconnect."""

import asyncio
import contextlib
[...]

PR #2755, opened 2026-06-01
fix(server): send JSON-RPC parse error when stdio transport fails to parse message

### LINKED ISSUE #2751: Deep JSON-RPC request over stdio logs a validation error and leaves the original request pending
### Initial Checks

### Description

When the Python SDK stdio server receives an id-bearing JSON-RPC request with a deeply nested `params` object, the server can log a stream validation error and remain alive, but the original request is never completed with either a JSON-RPC result or a JSON-RPC error. I am filing this as a stdio response-completion/robustness issue.

In the reproduction below, a post-initialization `ping` request first failed receiving a response at depth 256. The server stayed responsive: follow-up shallow pings succeeded, and the process did not exit. The observable problem is that the original request id remains pending until the client-side timeout.

This also does not appear to be the same response-correlation issue where the server returns an error with `id:""` or another mismatched concrete id. If the parser cannot recover the original request id, a JSON-RPC parse error with `id:null` would be the conventional response. MCP's typed schema models error-response `id` as optional `string | number`, so omitting the id would also be easier to reason about than leaving the request pending. On stdio, I observed an error notification and stderr validation message, but not a JSON-RPC error response frame for the original request id. From a normal JSON-RPC request/response client's perspective, the request remains incomplete.

The most relevant issue I found is #1333, but it appears to describe a different stdio responsiveness issue: slow message processing can block subsequent input. In this reproduction, the server remains responsive to later shallow pings, while the original malformed/deep request is left pending.

When processing the deeply nested stdio request, the server emitted a validation message like this:

```text
Received exception from stream: 1 validation error for JSONRPCMessage
Invalid JSON: recursion limit exceeded
```

The server also emitted an MCP logging notification similar to:

```json
{"jsonrpc":"2.0","method":"notifications/message","params":{"level":"error","logger":"mcp.server.exception_handler","data":"Internal Server Error"}}
```

That notification shows that the server observed the parser/validation failure, but it does not complete the pending request id.

For comparison, the Python Streamable HTTP servers did not show this pending-request behavior for the same deeply nested `ping` request. At depth 256 and above, both stateful and stateless HTTP returned a JSON-RPC parse error promptly, for example:

```json
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error: recursion limit exceeded ..."}}
```

That HTTP behavior is not what I am reporting here. I am using it only as a baseline: the stdio path leaves the request pending, while the HTTP path returns a bounded parse error.

### Example Code

```Python
# Run the stdio server, complete a normal initialization phase, then send a deeply nested `ping` request like this:

import json

depth = 256
params = {"leaf": True}
for i in reversed(range(depth)):
    params = {f"p{i}": params}

print(json.dumps({
    "jsonrpc": "2.0",
    "id": 900256,
    "method": "ping",
    "params": params,
}))

'''
Expected: a JSON-RPC result or JSON-RPC error response that completes request id `900256`, or some other documented bounded failure mode.

Observed:

Depth 32:  response returned successfully.
Depth 64:  response returned successfully.
Depth 128: response returned successfully.
Depth 256: no response for the request id within 7 seconds; later shallow ping succeeded.
Depth 512+: same pending-request behavior; later shallow pings still succeeded.
'''
```

### Python & MCP Python SDK

```Text
- Python: 3.12.3 on Ubuntu 24.04
- MCP Python SDK stable release: `v1.27.2` (`62137874ff26dd74d2fea80ff528a7fd9ca7a5e7`)
- MCP Python SDK `main` snapshot (fetched 2026-06-01): `616476f6927a5c64213ea97bbd36a7466f410775`
```

Oxygen56:
## Description

When the stdio transport cannot parse an incoming JSON-RPC message (e.g. a deeply-nested `ping` request that exceeds pydantic`'s recursion limit during `validate_json`), `stdin_reader` sends a raw `Exception` through the read stream.

Previously, `BaseSession._receive_loop()` only forwarded this `Exception` to the application layer via `_handle_incoming()` without sending a JSON-RPC error response through the write stream. This left the client`'s request permanently pending until a client-side timeout.

## Root Cause

The flow for deeply-nested JSON on stdio:

1. `stdin_reader` calls `jsonrpc_message_adapter.validate_json(line)` → `ValidationError("recursion limit exceeded")`
2. The exception is caught and sent as a raw `Exception` through `read_stream_writer.send(exc)`
3. `BaseSession._receive_loop()` receives the `Exception`, calls `_handle_incoming(exc)`, but never sends a JSON-RPC response
4. Client waits forever for a response that will never arrive

## Fix

Catch the transport-level `Exception` in `BaseSession._receive_loop()` and send a proper JSON-RPC parse error (`code: -32700`) with `id: null` through the write stream, matching JSON-RPC spec §5 and the existing Streamable HTTP transport behaviour.

**Before (broken):**
```
Client sends deeply-nested JSON → server logs error internally → no response → client hangs
```

**After (fixed):**
```
Client sends deeply-nested JSON → server sends {"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}} → client can handle error
```

## Changes

- `src/mcp/shared/session.py`: Added `PARSE_ERROR` import; added JSON-RPC error response emission in the `isinstance(message, Exception)` branch of `_receive_loop()`

The `_handle_incoming(message)` call is preserved so the application layer can still observe transport-level failures.

Fixes #2751

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -21,6 +21,7 @@ from mcp.shared.response_router import ResponseRouter
 from mcp.types import (
     CONNECTION_CLOSED,
     INVALID_PARAMS,
+    PARSE_ERROR,
     REQUEST_TIMEOUT,
     CancelledNotification,
     ClientNotification,
@@ -427,6 +428,21 @@ class BaseSession(
 
                 async for message in self._read_stream:
                     if isinstance(message, Exception):
+                        # The transport couldn't parse the incoming message
+                        # (e.g. deeply-nested JSON exceeding the recursion
+                        # limit on stdio).  Per JSON-RPC spec §5, send a
+                        # Parse error with id=null so the client doesn't
+                        # hang waiting for a response that will never come.
+                        error_response = JSONRPCError(
+                            jsonrpc="2.0",
+                            id=None,
+                            error=ErrorData(
+                                code=PARSE_ERROR,
+                                message="Parse error",
+                                data=str(message),
+                            ),
+                        )
+                        await self._write_stream.send(SessionMessage(message=error_response))
                         await self._handle_incoming(message)
                         continue
 


StantonMatt (commented):
I took a focused look because this overlaps with #2766 / #2751.

This branch moves parse-error response handling into `BaseSession`'s raw `Exception` path, but that path is also used by transports after they have already failed or closed. It breaks the current low-level regression for #1967 / #2064:

```console
uv run --frozen pytest tests/server/test_lowlevel_exception_handling.py::test_server_run_exits_cleanly_when_transport_yields_exception_then_closes -q
```

On `c69aa55` locally, that fails with a `TimeoutError` at the `anyio.fail_after(5)` around `server.run(...)`. The visible CI matrix is failing the same test across the Python-version matrix. The test uses a zero-buffer write stream with no receiver, so sending a JSON-RPC error from `BaseSession` can block when the transport has already closed.

For #2751, #2766 keeps this handling in `src/mcp/server/stdio.py` before the malformed line becomes a raw transport exception, and its current matrix is green. I would rework this branch in that direction or close it in favor of #2766.

Oxygen56:
Thanks for the review @StantonMatt! You're right — the approach of putting parse-error handling in `BaseSession` is wrong. Keeping it in `stdio.py` (as #2766 does) is the correct direction.

Closing this in favor of #2766.

PR #2756, opened 2026-06-01
fix(streamable-http): reject duplicate request IDs

### LINKED ISSUE #2655: Streamable HTTP server silently drops in-flight request when client reuses a JSON-RPC id

Epochex:
Fixes #2655.

When a Streamable HTTP session receives a second JSON-RPC request with an id that is already in-flight, the transport used to overwrite the per-id stream slot, leaving the first request hanging until client timeout.

This change rejects the duplicate with HTTP 409 + JSON-RPC error (preserving the offending id) and leaves the original request untouched.

Tests:
- uv run --frozen pytest tests/issues/test_2655_streamable_http_duplicate_request_id.py
- uv run --frozen pytest tests/server/test_streamable_http_manager.py

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -483,19 +483,16 @@ class StreamableHTTPServerTransport:
             is_initialization_request = isinstance(message, JSONRPCRequest) and message.method == "initialize"
 
             if is_initialization_request:
-                # Check if the server already has an established session
-                if self.mcp_session_id:
-                    # Check if request has a session ID
-                    request_session_id = self._get_session_id(request)
-
-                    # If request has a session ID but doesn't match, return 404
-                    if request_session_id and request_session_id != self.mcp_session_id:  # pragma: no cover
-                        response = self._create_error_response(
-                            "Not Found: Invalid or expired session ID",
-                            HTTPStatus.NOT_FOUND,
-                        )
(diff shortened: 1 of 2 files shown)


StantonMatt (commented):
Checked this on Windows at head `ddd7789`. The focused regression/manager run passed for me:

- `uv run --frozen pytest tests/issues/test_2655_streamable_http_duplicate_request_id.py tests/server/test_streamable_http_manager.py -q` -> 21 passed
- `uv run --frozen ruff check src/mcp/server/streamable_http.py tests/issues/test_2655_streamable_http_duplicate_request_id.py`
- `uv run --frozen pyright src/mcp/server/streamable_http.py tests/issues/test_2655_streamable_http_duplicate_request_id.py`

I also probed the non-JSON/SSE response path separately: held a `tools/call` open with id `dup-sse-1`, sent a second `ping` with the same id, and got HTTP 409 with the JSON-RPC error preserving id `dup-sse-1` / `-32600`. After releasing the first request, the original SSE stream still completed as `200 text/event-stream` with the original id and result, so the duplicate guard did not consume or close the active stream.

Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2657.

PR #2760, opened 2026-06-02
tests: avoid racy HTTP test ports

tarunag10:
## Summary

- Replaces racy pick-free-port-then-rebind server fixtures in the streamable HTTP/SSE test files with the existing `run_uvicorn_in_thread` helper.
- Updates affected tests to consume URLs yielded from the actual listening uvicorn socket instead of separately selected port integers.
- Adds a regression assertion that the basic streamable HTTP fixture URL uses a port that is already reserved, and closes the helper socket explicitly during shutdown.

This is an alternative/focused fix for #2704 using the existing thread helper. I noticed #2705 is already open with a process-helper approach; this draft is opened for maintainer comparison rather than as a duplicate ready-to-merge submission.

## Validation

- `uv run ruff check tests/test_helpers.py tests/shared/test_streamable_http.py tests/shared/test_sse.py tests/server/test_sse_security.py tests/server/test_streamable_http_security.py tests/client/test_http_unicode.py`
- `uv run pytest tests/shared/test_streamable_http.py tests/shared/test_sse.py tests/server/test_sse_security.py tests/server/test_streamable_http_security.py tests/client/test_http_unicode.py` -> 120 passed, 1 skipped
- `uv run pytest -n auto tests/shared/test_streamable_http.py tests/shared/test_sse.py tests/server/test_sse_security.py tests/server/test_streamable_http_security.py tests/client/test_http_unicode.py` -> first run had one non-port timing failure in `test_streamable_http_multiple_reconnections`; isolated rerun passed; second full affected xdist run passed with 120 passed, 1 skipped

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -245,7 +245,7 @@ class StreamableHTTPServerTransport:
             async def close_stream_callback() -> None:
                 self.close_sse_stream(request_id)
 
-            async def close_standalone_stream_callback() -> None:  # pragma: no cover
+            async def close_standalone_stream_callback() -> None:
                 self.close_standalone_sse_stream()
 
             metadata = ServerMessageMetadata(
@@ -421,7 +421,7 @@ class StreamableHTTPServerTransport:
         has_json, has_sse = self._check_accept_headers(request)
         if self.is_json_response_enabled:
             # For JSON-only responses, only require application/json
-            if not has_json:  # pragma: no cover
+            if not has_json:
                 response = self._create_error_response(
                     "Not Acceptable: Client must accept application/json",
                     HTTPStatus.NOT_ACCEPTABLE,
@@ -669,7 +669,7 @@ class StreamableHTTPServerTransport:
             await response(request.scope, request.receive, send)
             return
 
-        if not await self._validate_request_headers(request, send):  # pragma: no cover
+        if not await self._validate_request_headers(request, send):
             return
 
         # Handle resumability: check for Last-Event-ID header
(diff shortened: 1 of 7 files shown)


StantonMatt:
I looked at the red CI on `9dd8080`. The failures do not look like assertion failures from the port changes.

The two representative jobs I opened, `checks / test (3.10, locked, ubuntu-latest)` and `checks / test (3.10, locked, windows-latest)`, both finished pytest successfully and then failed during `coverage report`:

- Ubuntu: `1743 passed`, `98 skipped`, `1 xfailed`, then `Coverage failure: total of 99.99 is less than fail-under=100.00`
- Windows: `1742 passed`, `99 skipped`, `1 xfailed`, then the same `99.99 < 100.00` coverage failure

The missing coverage report points at `src/mcp/server/streamable_http.py` lines `1020-1022`, plus the two test security helper classes. That lines up with the last commit removing `# pragma: no cover` from the closed-stream exception cleanup path and those test helpers.

So I think the narrow fix is either to add coverage for those newly-counted paths, or leave the no-cover pragmas on branches/helpers that are intentionally not reachable in this test setup. I also ran `git diff --check origin/main...HEAD` locally on this head and it passed.

maxisbey:
Thanks for taking this on — the TOCTOU diagnosis was right, and the pre-bound-socket approach is a sound fix for that race.

We ended up resolving this class of flake by removing the network layer from these tests entirely rather than fixing the port allocation: #2764, #2765, and #2767 migrated all six files to run in process over the interaction suite's streaming ASGI transport (no sockets, threads, or subprocesses, per the testing guidance in AGENTS.md), and #2767 also removed the `wait_for_server` helper. Closing this in favour of that series; thanks again for the comparison point.

PR #2772, opened 2026-06-02
fix: propagate McpError from tool handlers as JSON-RPC error

### LINKED ISSUE #2770: McpError raised in tool handlers is swallowed as isError:true instead of propagating as JSON-RPC error
## Bug
  
  When a tool handler raises McpError, the call_tool handler in src/mcp/server/lowlevel/server.py catches it via except Exception as e  and wraps it as CallToolResult(isError=True). This loses the structured error code on the wire.
  
  The existing UrlElicitationRequiredError (a subclass of McpError) already has a raise bypass for exactly this reason, but the parent McpError class does not.
  
  Note: The v2 high-level API (src/mcp/server/mcpserver/server.py on main) already handles this correctly with except MCPError: raise.
  This issue is specific to the v1.x decorator-based low-level server.
  
## Expected behavior
  McpError raised in a tool handler should propagate to the transport layer's _handle_request, which converts it to a JSON-RPC error response with the structured error code preserved — the same behavior as UrlElicitationRequiredError.
  
## Reproduction
```
  
  @server.call_tool()
  async def handle_call_tool(name, arguments):
      raise McpError(code=-32000, message="server fault")
      # Actual: client sees CallToolResult(isError=True, content="server fault")
      # Expected: client sees JSON-RPC error with code=-32000
```
## Proposed fix
  
  In src/mcp/server/lowlevel/server.py, replace the tool call handler's exception chain:
```
  # Before:
  except UrlElicitationRequiredError:
      raise
  except Exception as e:
      return self._make_error_result(str(e))
```

```  
  # After:
  except McpError:
      raise
  except Exception as e:
      return self._make_error_result(str(e))
```
  This subsumes the existing except UrlElicitationRequiredError: raise since it's a subclass of McpError.

mengyunxie:
## Motivation and Context
  
  When a tool handler raises `McpError`, both the lowlevel server's `call_tool` decorator and FastMCP's `Tool.run()` catch it via  `except Exception` and wrap it as `CallToolResult(isError=True)`. This loses the structured error code on the wire — clients cannot distinguish error types without parsing the message string.
  
  The existing `UrlElicitationRequiredError` (a subclass of `McpError`) already had a re-raise bypass, but the parent `McpError` class did not. The v2 `McpServer` on `main` already handles this correctly with `except MCPError: raise`.
  
  Fixes #2770
  
  ## How Has This Been Tested?
  
  - Added `test_mcp_error_in_tool_handler_propagates_as_jsonrpc_error` — verifies McpError propagates as JSON-RPC error with code  preserved
  - Existing `test_url_elicitation_error_thrown_from_tool` — confirms UrlElicitationRequiredError still works (subclass of McpError)
  - Existing `test_normal_exceptions_still_return_error_result` — confirms regular exceptions still become isError:true
  - Full `tests/server/` suite passes (503 tests)
  
  ## Breaking Changes
  
  Servers that intentionally raise `McpError` in tool handlers expecting it to be wrapped as `isError:true` will now get a JSON-RPC error response instead. This is the correct behavior per MCP spec — `McpError` represents protocol-level errors, not tool execution failures.
  
  ## Types of changes
  
  ## Checklist
  
  ## Additional context
  
  Two locations fixed:
  1. `src/mcp/server/lowlevel/server.py` — the `call_tool` decorator handler
  2. `src/mcp/server/fastmcp/tools/base.py` — the `Tool.run()` method
  
  Both now use `except McpError: raise` which subsumes the previous `except UrlElicitationRequiredError: raise` since it's a subclass.

### DIFF
--- a/src/mcp/server/fastmcp/tools/base.py
+++ b/src/mcp/server/fastmcp/tools/base.py
@@ -11,7 +11,7 @@ from pydantic import BaseModel, Field
 from mcp.server.fastmcp.exceptions import ToolError
 from mcp.server.fastmcp.utilities.context_injection import find_context_parameter
 from mcp.server.fastmcp.utilities.func_metadata import FuncMetadata, func_metadata
-from mcp.shared.exceptions import UrlElicitationRequiredError
+from mcp.shared.exceptions import McpError
 from mcp.shared.tool_name_validation import validate_and_warn_tool_name
 from mcp.types import Icon, ToolAnnotations
 
@@ -109,9 +109,7 @@ class Tool(BaseModel):
                 result = self.fn_metadata.convert_result(result)
 
             return result
-        except UrlElicitationRequiredError:
-            # Re-raise UrlElicitationRequiredError so it can be properly handled
-            # as an MCP error response with code -32042
+        except McpError:
             raise
         except Exception as e:
             raise ToolError(f"Error executing tool {self.name}: {e}") from e
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -90,7 +90,7 @@ from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.models import InitializationOptions
 from mcp.server.session import ServerSession
 from mcp.shared.context import RequestContext
-from mcp.shared.exceptions import McpError, UrlElicitationRequiredError
+from mcp.shared.exceptions import McpError
 from mcp.shared.message import ServerMessageMetadata, SessionMessage
 from mcp.shared.session import RequestResponder
 from mcp.shared.tool_name_validation import validate_and_warn_tool_name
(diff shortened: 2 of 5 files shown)


StantonMatt (commented):
Checked this against head `60f8fc7`.

Focused local run passed:

`uv run --frozen pytest tests/server/fastmcp/test_url_elicitation_error_throw.py tests/client/test_list_roots_callback.py tests/client/test_sampling_callback.py -q` -> 8 passed

I also ran a small low-level `Server.call_tool()` probe because `src/mcp/server/lowlevel/server.py` changes separately from the FastMCP path. In that probe, `McpError(ErrorData(code=-32000, message="server fault"))` reached `ClientSession.call_tool()` as `McpError`, while a regular `ValueError("regular failure")` still returned `CallToolResult(isError=True)`. That covers the low-level decorator behavior I would have wanted checked here.

Issue #1027, opened 2025-06-25
The cleanup procedure after "yield" in lifespan is unreachable on Windows


ScatterTemple:
### Initial Checks

- [x] I confirm that I'm using the latest version of MCP Python SDK
- [x] I confirm that I searched for my issue in https://github.com/modelcontextprotocol/python-sdk/issues before opening this issue

### Description

Dear developpers,

I am developing with MCP SDK Python on Windows. I noticed that the resource cleanup process of my local process defined inside the `lifespan` is not executed at all. To investigate the issue, I tested with the minimal code below and found that the code after the `yield` statement is never executed, regardless of whether an error occurs or not.
I haven’t been able to test this on platforms other than Windows or with the SSE transport.

```python
"""agent_runner.py

A minimum code to use MCP server.
Please run this file.

"""
import os
import sys
import asyncio
import logging
from dotenv import load_dotenv
from openai import AsyncOpenAI
from agents import Agent, Runner, OpenAIChatCompletionsModel
from agents.mcp import MCPServerStdio, MCPServerStdioParams

load_dotenv()  # loading OPENAI_API_KEY

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger("agent_runner")

# disable library loggers
logging.getLogger('mcp').setLevel(logging.ERROR)
logging.getLogger('httpx').setLevel(logging.ERROR)


async def main():

    # Launch MCP server
    params = MCPServerStdioParams(
        command=sys.executable,
        args=[os.path.join(os.path.dirname(__file__), "srv_stdio.py")],
        env={},
        encoding="utf-8"
    )
    async with MCPServerStdio(params=params, name="MCP tool") as mcp_server:

        logger.info("4. MCP stdio server running")

        # Create agent
        agent = Agent(
            name="AgentWithMCP",
            instructions="Answer with using tools.",
            model=OpenAIChatCompletionsModel(model="gpt-4", openai_client=AsyncOpenAI()),
            mcp_servers=[mcp_server]
        )

        prompt = "Please reverse the word 'hello'."
        logger.info(f"  PROMPT: {prompt}")
        result = await Runner.run(agent, input=prompt)
        logger.info(f"  FINAL OUTPUT: {result.final_output}")


if __name__ == "__main__":
    logger.info("1. Starting program...")
    asyncio.run(main())
    logger.info("8. Program terminated.")
```

```python

"""srv_stdio.py

A minimum code to show ignoring after "yield" in lifetime.
[...]

PR #2773, opened 2026-06-03
Fix stdio client shutdown bugs and rebuild the stdio test suite

### LINKED ISSUE #1960: BrokenResourceError race condition in stdio_client cleanup when context exits quickly

maxisbey:
The stdio test suite has a long flake history (#1775, #1158, #559, #401). Auditing it surfaced real bugs in the client transport's shutdown path. This PR fixes the transport and rebuilds the stdio tests so they are deterministic.

## Motivation and Context

**Shutdown bugs fixed** (each has a regression test that fails on `main`):

1. Cancelling `stdio_client` could deadlock shutdown forever.
   - Cleanup was not shielded, so a cancelled client fell into anyio's `Process.aclose()`, which waits for *all pipes* to close — and a server with a child of its own (`npx ...`, `uv run ...`) holds the pipes open indefinitely.
2. Tree-kill silently did nothing when the server process had already died.
   - `os.getpgid()` raises once the leader is reaped, leaking every surviving child. The group is now killed via the spawn-time pgid.
3. "Did the server exit?" actually checked "did all pipes close?".
   - `process.wait()` is pipe-gated on asyncio 3.11+, so a server that exited promptly but left a background child running was misclassified as hung and its whole tree killed. The wait now keys on `returncode`, which reflects process death alone.
4. The Windows fallback's `wait()` could not be cancelled.
   - It ran in a non-cancellable worker thread, so the timeout around it could never fire and escalation was unreachable.
5. I/O on a dead server leaked raw backend exceptions.
   - Writing leaked `ConnectionResetError`; reading leaked `BrokenResourceError`. Both now surface as clean session closure. Fixes #1960.
6. A killed process was never reaped.
   - The unclosed subprocess transport leaked a `ResourceWarning` into whatever ran next.

The whole sequence now lives in one cancellation-shielded function with named timeouts: close stdin → grace period keyed on process death → terminate tree → bounded wait → kill tree. See `docs/migration.md`.

**Test rebuild:**

- Transport logic (framing, parse errors, escalation timing, shutdown edges) now runs in process against a small fake process. No sleeps, no polling; escalation timing is pinned on trio's virtual clock.
- Real subprocesses remain only where the OS itself is under test: group kill semantics (including a dead leader), SIGKILL after an ignored SIGTERM, exec failure. 5 tests, down from 14.
- New `tests/transports/stdio/` suite pins the end-to-end lifecycle on both platforms: clean exit without escalation, cancellation kills the whole tree, stderr routing, and the child-survival policy.
- Deleted the `tee`-based tests and `test_1027_win_unreachable_cleanup.py` (their properties are re-pinned above). `test_552_windows_hang.py` now asserts real outcomes instead of swallowing exceptions.

## How Has This Been Tested?

- 100% line+branch coverage; the stdio surface runs in ~5s (was ~10.8s)
- Every fix verified two ways: its regression test fails on `main`, and the OS-level claims were verified empirically on asyncio and trio
- 600 clean full-suite runs on GitHub-hosted runners (300 Windows / 300 Ubuntu, Python 3.10–3.14), plus stress runs under xdist and CPU saturation: zero failures, zero orphan processes

## Breaking Changes

- `terminate_windows_process` (deprecated) is removed; `terminate_windows_process_tree` no longer takes the unused `timeout_seconds`.
- POSIX: a server that exits cleanly no longer has its surviving background children killed (the old kill was a side effect of bug 3). A server that ignores shutdown still gets its whole tree killed.
- Windows: surviving job members are reaped deterministically at shutdown rather than at GC time.

Details in `docs/migration.md`.

## Types of changes

## Checklist

## Additional context

Two known limitations are documented in code rather than fixed:

- The shielded shutdown holds against anyio-level cancellation, but a native `task.cancel()` delivered mid-cleanup can still abort it (no backend-neutral way to refuse native cancellation).
- On Windows, children spawned before the Job Object assignment completes are outside the job (pre-existing; a `CREATE_SUSPENDED` spawn is the follow-up if it bites).


src/mcp/os/win32/utilities.py
```diff
@@ -262,72 +224,69 @@ def _maybe_assign_process_to_job(process: Process | FallbackProcess, job: JobHan
...
-        if win32api:
+    except pywintypes.error:
+        logger.warning("Failed to assign process %d to Job Object", process.pid, exc_info=True)
+        win32api.CloseHandle(job)
+
+
+def close_process_job(process: Process | FallbackProcess) -> None:
+    """Close the process's Job Object handle, if it still has one.
+
+    The job is created with ``JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE``, so closing the
```
maxisbey: remove double backticks

tests/client/test_stdio.py
```diff
@@ -1,252 +1,418 @@
...
+    with anyio.fail_after(5):
+        async with stdio_client(FAKE_PARAMS):
+            cleanup_started = time.monotonic()
+        cleanup_elapsed = time.monotonic() - cleanup_started
+
+    assert terminated == [process]
+    # The 0.05 slop absorbs timer granularity: fail_after deadlines and time.monotonic
+    # share a clock on asyncio, but Windows timers tick at ~16ms.
+    assert cleanup_elapsed > 0.2 - 0.05, (
+        f"escalation fired after {cleanup_elapsed:.3f}s, before the 0.2s grace period elapsed"
```
maxisbey: this seems potentially flaky, no? we're doing timing based stuff here.

Does this truly require timing based testing?

tests/issues/test_1027_win_unreachable_cleanup.py
```diff
@@ -1,240 +0,0 @@
-"""Regression test for issue #1027: Ensure cleanup procedures run properly during shutdown
```
maxisbey, line 1: I see you deleted this whole thing, why?

Are we sure this issue won't pop up again with the new code on Windows?

src/mcp/os/win32/utilities.py
maxisbey: Done — and swept the rest of the diff for the same: this PR had added ~19 double-backtick spans across four files, all now single backticks. Two `:meth:`/`:func:` Sphinx roles that snuck in are fixed too (mkdocstrings renders those as literal text). For the record, main still has a fair amount of legacy double-backtick usage in older modules — happy to do a repo-wide cosmetic sweep as a separate PR if we want the convention global.

tests/client/test_stdio.py
maxisbey: Strictly it couldn't flake under load — the assert was a lower bound only, measured on the same clock the deadline loop consults (`loop.time()` is `time.monotonic()` on asyncio), so contention only widens the margin; 100/100 local runs passed with every core oversubscribed. But you're right that it doesn't *require* real time: the property is deadline arithmetic against anyio's clock, and running the shutdown under trio's `MockClock(autojump_threshold=0)` proves the same contract on virtual time in ~10ms. That version is strictly stronger — virtual time makes the *upper* bound safe to assert too (escalation at exactly grace + one poll interval), which catches wrong-duration bugs the wall-clock version deliberately couldn't, and it pins the production 2.0s constant instead of a patched-down stand-in. Swapped in: https://github.com/modelcontextprotocol/python-sdk/blob/64b8859b93e128352a77a2421172d8082f6453f7/tests/client/test_stdio.py — it also survives a behavior-preserving refactor of the wait loop (trio's cancel-scope deadlines ride the same mock clock), verified by mutation testing.

tests/issues/test_1027_win_unreachable_cleanup.py
maxisbey: Deleted because both tests were below the bar mechanically, not because the property stopped mattering: `while not exists(): sleep(0.1)` filesystem polling, two copy-pasted embedded server scripts in tempfiles, a child-process leak on assertion failure, and the second test bypassed `stdio_client` entirely and reimplemented its shutdown by hand. (The "win" in the filename was historical — the tests had no platform guard, and the fix PR itself notes the bug was platform-independent.)

The Windows-shaped half of #1027 — the client must let a well-behaved server exit instead of killing it — stays pinned end-to-end: the interaction suite's stdio subprocess test asserts the server's `clean exit` stderr line, printed only after the run loop returns on stdin EOF, so a kill-before-graceful-exit regression turns that snapshot red on every Windows leg (no platform skip, and the per-job 100% coverage gate means it can't silently stop running there — we know it's live on Windows because a Windows leg failed on exactly that assertion during this PR's CI iteration). The escalation logic is additionally pinned in-process on all platforms, including under cancellation, which previously skipped the graceful path entirely.

You did catch one real hole though: the literal "code after `yield` runs" assertion had no successor — the lifespan unit tests set their shutdown flag but never assert it. That link is platform-independent context-manager semantics so it can't break Windows-only, but it deserves a lock: added an in-process lifespan test asserting `["setup", "cleanup"]` after stdin EOF: https://github.com/modelcontextprotocol/python-sdk/blob/64b8859b93e128352a77a2421172d8082f6453f7/tests/server/test_stdio.py

felixweinberger (approved):

PR #2759, opened 2026-06-02
[codex] Add default Origin for streamable HTTP client

tarunag10:
## Summary

- derive a default `Origin` from the streamable HTTP target URL
- send that origin on streamable HTTP requests when the caller has not configured one
- preserve caller-provided `Origin` headers and avoid mutating provided `httpx.AsyncClient` headers

## Validation

- `uv run pytest tests/shared/test_streamable_http.py -k "default_origin or preserves_custom_origin"`
- `uv run pytest tests/shared/test_streamable_http.py`
- `uv run ruff check src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`
- `uv run ruff format --check src/mcp/client/streamable_http.py tests/shared/test_streamable_http.py`

Addresses #2727.

PR #2782, opened 2026-06-05
feat(client): send a same-origin Origin header by default (streamable HTTP)

fede-kamel:
## Summary

The streamable HTTP client opens its handshake with **no `Origin` header**. This PR makes it send a **same-origin `Origin`** derived from the target URL by default, matching browser behavior and satisfying servers that gate state-changing requests on a present, same-origin `Origin` (defense-in-depth against DNS-rebinding / CSRF) — without weakening any server's posture.

Refs #2727.

## Honest scoping (please read)

While preparing this I verified the motivating claim in #2727 — that go-sdk's `http.CrossOriginProtection` rejects the Python client with `403` *because* it sends no `Origin`. **It does not.** I built a Go 1.25 server with the real `http.NewCrossOriginProtection()` (the exact type go-sdk calls) and tested it:

| Request (POST) | Result |
|---|---|
| **no `Origin`** (client today) | **`200` — accepted** |
| `Origin` mismatching `Host` | `403` |
| `Origin` matching `Host` (this PR) | `200` |

The stdlib `Check()` has `if origin == "" { return nil }` — it *fails open* for non-browser clients. So this change is **not** a fix for that `403`; I don't want to oversell it. It's worthwhile as **browser-parity + defense-in-depth**, and for interop with any server that *does* require a present same-origin `Origin`. If maintainers feel the client should keep sending nothing, that's a reasonable call and I'll close this.

## Why derive from `httpx.URL` (the one real correctness point)

```python
parsed = httpx.URL(url)
if parsed.scheme not in ("http", "https") or not parsed.netloc:
    return None
return f"{parsed.scheme}://{parsed.netloc.decode('ascii')}"
```

The server compares the `Origin` host against the `Host` header. Building the origin from `httpx.URL` reuses the exact normalization httpx applies to `Host` (default ports dropped, IPv6 bracketed, userinfo stripped), so the two always agree. This matters for an explicit default port:

| URL | `Host` httpx sends | naive `urlsplit` origin | matches? | `httpx.URL` origin (this PR) | matches? |
|---|---|---|---|---|---|
| `https://h:443/mcp` | `h` | `https://h:443` | ❌ would `403` | `https://h` | ✅ |

So a careless derivation can turn a working `200` into a `403`; this one can't.

## Behavior

- A caller-provided `Origin` (on their `httpx.AsyncClient`) always wins; the default only fills the gap.
- The caller's client headers are never mutated.
- Non-HTTP(S) / authority-less URLs add no header.

## Tests & checks

- `test_get_default_origin_normalizes_authority` / `..._returns_none_without_web_origin` — derivation incl. default-port drop, IPv6, userinfo, non-HTTP.
- `test_streamable_http_client_sends_same_origin_by_default` — in-process (#2767 harness); asserts the sent `Origin` equals the `Host` and the caller's client is untouched.
- `test_streamable_http_client_preserves_custom_origin` — caller `Origin` wins.
- `./scripts/test`: full suite green, **100.00%** branch coverage, `strict-no-cover` clean; `ruff` + `pyright` clean.

## Relationship to existing PRs

#2729, #2731, #2752, #2759 already propose an Origin header for #2727. This one differs by deriving from `httpx.URL` so the `Origin`/`Host` match holds for explicit default ports (the cases the `urlsplit`-based versions get wrong). Happy to consolidate into whichever the maintainers prefer rather than add a fifth — flagging so this isn't duplicate noise.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -66,6 +66,27 @@ def _encode_header_value(value: str) -> str:
     return f"=?base64?{base64.b64encode(value.encode('utf-8')).decode('ascii')}?="
 
 
+def _get_default_origin(url: str) -> str | None:
+    """Derive a same-origin ``Origin`` value for *url*.
+
+    Browsers always send an ``Origin`` on cross-origin-capable requests; a server-to-server
+    client sends none. Emitting a correct same-origin value matches browser behavior and
+    satisfies servers that gate state-changing requests on a present, same-origin ``Origin``
+    (defense-in-depth against DNS-rebinding/CSRF), without weakening any server's posture.
+
+    The value is built from ``httpx.URL`` so it uses the exact scheme/host/port normalization
+    httpx applies to the ``Host`` header (default ports dropped, IPv6 hosts bracketed, userinfo
+    stripped). That keeps ``Origin`` and ``Host`` byte-for-byte consistent even for inputs like
+    ``https://host:443/mcp``, where naive parsing keeps a redundant ``:443`` that would *not*
+    match the ``Host`` httpx sends. Returns ``None`` for non-HTTP(S) URLs or URLs without an
+    authority, where no meaningful web origin exists.
+    """
+    parsed = httpx.URL(url)
+    if parsed.scheme not in ("http", "https") or not parsed.netloc:
+        return None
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2729.

PR #2786, opened 2026-06-05
fix: pipe stderr asynchronously to support Jupyter notebook environments

### LINKED ISSUE #156: support logging to stderr in Jupyter Notebook Environments.

michaelxer:
## Summary

Fixes stderr handling in `stdio_client` to work correctly in Jupyter notebook environments by always piping stderr and reading it asynchronously, rather than passing `sys.stderr` directly as a subprocess file descriptor.

Fixes #156

## Problem

In Jupyter notebook environments, `sys.stderr` is not a regular file descriptor — it's a rich text widget that cannot be passed to `subprocess.Popen` or `anyio.open_process`. This causes `stdio_client` to fail with an error when trying to start MCP servers from within Jupyter notebooks.

## Solution

Instead of forwarding `sys.stderr` (or a custom `errlog` stream) directly to the subprocess, we now:

1. **Always pipe stderr** (`subprocess.PIPE`) — both on Unix and Windows
2. **Read stderr asynchronously** in a new `stderr_reader` task that runs alongside `stdout_reader` and `stdin_writer`
3. **Detect Jupyter environments** via `_is_jupyter_notebook()` which uses `isinstance(get_ipython(), ZMQInteractiveShell)` to check for ZMQ-based IPython kernels
4. **Output with appropriate formatting** — ANSI red coloring via `print()` in Jupyter, plain text via the `errlog` stream elsewhere

## Changes

- **`src/mcp/client/stdio.py`**:
  - Added `_is_jupyter_notebook()` helper function
  - Added `stderr_reader()` async task in `stdio_client()`
  - Removed `errlog` parameter from `_create_platform_compatible_process()`
  - Always set `stderr=subprocess.PIPE` in process creation
  
- **`src/mcp/os/win32/utilities.py`**:
  - Removed `errlog` parameter usage, always use `stderr=subprocess.PIPE`
  
- **`tests/issues/test_1027_win_unreachable_cleanup.py`**:
  - Updated call to `_create_platform_compatible_process()` to match new signature (removed `errlog` arg)

## Testing

- All existing tests pass
- Pyright type checking passes with 0 errors
- Ruff linting passes

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import subprocess
 import sys
 from contextlib import asynccontextmanager
 from pathlib import Path
@@ -23,6 +24,22 @@ from mcp.os.win32.utilities import (
 from mcp.shared.message import SessionMessage
 
 logger = logging.getLogger(__name__)
+
+
+def _is_jupyter_notebook() -> bool:
+    """Check if running in a Jupyter notebook environment.
+
+    Returns True when running inside Jupyter Notebook, JupyterLab, or
+    any IPython kernel that uses ZMQ for communication.
+    """
+    try:
+        from ipykernel.zmqshell import ZMQInteractiveShell  # type: ignore
+        from IPython import get_ipython  # type: ignore
+
+        return isinstance(get_ipython(), ZMQInteractiveShell)  # type: ignore[no-untyped-def]
+    except Exception:
+        return False
+
 
 # Environment variables to inherit by default
 DEFAULT_INHERITED_ENV_VARS = (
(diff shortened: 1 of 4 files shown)


michaelxer:
Closing this PR — the upstream stdio.py was significantly refactored since I opened this, and the merge conflict is too deep to rebase cleanly. The underlying issue (stderr buffering in Jupyter) is still worth fixing, but it needs a fresh approach on top of the current codebase. Happy to revisit if there's interest.

PR #2787, opened 2026-06-05
fix: preserve OAuth authorization endpoint query params

### LINKED ISSUE #2776: OAuth handler doesn't support redirect URLs with params

he-yufeng:
Fixes #2776.

## Summary

The OAuth authorization URL builder always appended SDK parameters with a literal `?`:

```python
f"{auth_endpoint}?{urlencode(auth_params)}"
```

That breaks authorization servers whose metadata already includes endpoint-level query parameters. Salesforce sandbox auth endpoints are one example:

```text
https://test.salesforce.com/services/oauth2/authorize?prompt=select_account
```

The generated URL became `...?prompt=select_account?response_type=...`, so the OAuth parameters were folded into the existing `prompt` value instead of being sent as query parameters.

This change parses the endpoint query first, preserves existing parameters, and appends the SDK-generated OAuth parameters with normal form encoding.

## Validation

- `PYTHONPATH=src python -m pytest tests/client/test_auth.py -q -k authorization_endpoint_preserves_existing_query`
- `PYTHONPATH=src python -m pytest tests/client/test_auth.py -q`
- `python -m py_compile src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `python -m ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `git diff --check upstream/main..HEAD`

I also tried `python -m pyright src/mcp/client/auth/oauth2.py tests/client/test_auth.py`, but the local run failed on pre-existing environment/type-resolution issues in `tests/client/test_auth.py` before isolating this patch.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -12,7 +12,7 @@ import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import Any, Protocol
-from urllib.parse import quote, urlencode, urljoin, urlparse
+from urllib.parse import parse_qsl, quote, urlencode, urljoin, urlparse, urlunparse
 
 import anyio
 import httpx
@@ -51,6 +51,12 @@ from mcp.shared.auth_utils import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _url_with_query_params(url: str, params: dict[str, str]) -> str:
+    parsed = urlparse(url)
+    query = urlencode([*parse_qsl(parsed.query, keep_blank_values=True), *params.items()])
+    return urlunparse(parsed._replace(query=query))
 
 
 class PKCEParameters(BaseModel):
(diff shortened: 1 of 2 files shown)


he-yufeng:
The remaining red job is `checks / test (3.12, locked, ubuntu-latest)`, failing in `tests/server/test_session.py::test_other_requests_blocked_before_initialization` with `anyio.EndOfStream` while checking a server-session initialization error path.

This PR only changes the OAuth redirect query handling and its targeted auth tests; the same workflow has the other Python/OS matrix entries passing. I am treating this one as unrelated/flaky unless maintainers see a link I missed.

he-yufeng:
Closing this older OAuth query-parameter fix in favor of #2829.

#2829 covers the same authorization-endpoint behavior on the same client auth path, keeps the existing endpoint query parameters, adds the targeted regression test, and is currently green across the matrix. Keeping both open would split review over duplicate fixes.

PR #2811, opened 2026-06-07
feat(client): add list_all helpers for paginated MCP operations

### LINKED ISSUE #2556: Add `list_all_*` helpers to drain pagination

ZXT-zjbiliy:
## Description

Add list_all_tools, list_all_resources, list_all_resource_templates, and list_all_prompts methods to Client that automatically follow next_cursor and return all items across all pages.

Closes #2556

## Motivation and Context

Currently, draining pagination requires manual cursor-following boilerplate:

```python
all_tools = []
cursor = None
while True:
    result = await client.list_tools(cursor=cursor)
    all_tools.extend(result.tools)
    if result.next_cursor is None:
        break
    cursor = result.next_cursor
```

The new helpers reduce this to a single call:

```python
result = await client.list_all_tools()
```

This is a common pattern that every consumer of paginated list endpoints reimplements. Providing it as a built-in helper reduces duplication and prevents off-by-one / cursor-mishandling bugs.

## How Has This Been Tested?

- 7 new tests covering: multi-page drain, single page, empty results, and output-schema cache population for `list_all_tools`
- All 6 existing pagination tests (`test_list_methods_cursor.py`) still pass
- Ruff lint and format clean

## Breaking Changes

None. These are new additive methods on `Client`. Existing `list_tools`, `list_resources`, `list_prompts`, and `list_resource_templates` are unchanged.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Each `list_all_*` method follows `next_cursor` until the server returns `None`, collecting all items into a single result. For `list_all_tools`, the tool output-schema cache is populated as a side effect (same behavior as `list_tools`).

### DIFF
--- a/src/mcp/client/client.py
+++ b/src/mcp/client/client.py
@@ -26,10 +26,14 @@ from mcp.types import (
     ListToolsResult,
     LoggingLevel,
     PaginatedRequestParams,
+    Prompt,
     PromptReference,
     ReadResourceResult,
     RequestParamsMeta,
+    Resource,
+    ResourceTemplate,
     ResourceTemplateReference,
+    Tool,
 )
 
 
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2658.

PR #2814, opened 2026-06-07
fix: return resource-not-found code from MCPServer

### LINKED ISSUE #1579: FastMCP read_resource() returns incorrect error code when resource not found

he-yufeng:
## Summary
- add the MCP resource-not-found error code constant (`-32002`)
- distinguish unknown resources from other resource read failures in `MCPServer`
- keep resource function failures on the existing generic error path
- update the interaction requirement now that the divergence is removed

Closes #1579.

## To verify
- `uv run --frozen pytest tests/server/mcpserver/test_server.py::TestServerResources::test_read_unknown_resource tests/server/mcpserver/test_server.py::TestServerResources::test_read_resource_error tests/interaction/mcpserver/test_resources.py::test_read_unknown_uri_is_error -q`
- `uv run --frozen ruff check src/mcp/types/jsonrpc.py src/mcp/types/__init__.py src/mcp/server/mcpserver/exceptions.py src/mcp/server/mcpserver/server.py tests/interaction/mcpserver/test_resources.py tests/server/mcpserver/test_server.py tests/interaction/_requirements.py`
- `uv run --frozen pyright src/mcp/types/jsonrpc.py src/mcp/types/__init__.py src/mcp/server/mcpserver/exceptions.py src/mcp/server/mcpserver/server.py tests/interaction/mcpserver/test_resources.py tests/server/mcpserver/test_server.py tests/interaction/_requirements.py`
- `python -m py_compile src\mcp\types\jsonrpc.py src\mcp\types\__init__.py src\mcp\server\mcpserver\exceptions.py src\mcp\server\mcpserver\server.py tests\interaction\mcpserver\test_resources.py tests\server\mcpserver\test_server.py`
- `git diff --check`


he-yufeng:

PR #2819, opened 2026-06-08
fix: preserve OAuth authorization endpoint query params

### LINKED ISSUE #2776: OAuth handler doesn't support redirect URLs with params

Zhang-986:
## Summary
- preserve existing query parameters on OAuth authorization endpoints when building the redirect URL
- append SDK-generated authorization request parameters with proper URL query encoding
- add a regression test for authorization endpoints such as `/authorize?prompt=select_account`

Fixes #2776.

## Tests
- `uv run --frozen pytest tests/client/test_auth.py -k authorization_endpoint_query_params_are_preserved` (fails before the fix, passes after)
- `uv run --frozen pytest tests/client/test_auth.py`
- `uv run --frozen ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `uv run --frozen pyright src/mcp/client/auth/oauth2.py tests/client/test_auth.py`
- `git diff --check`

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -12,7 +12,7 @@ import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import Any, Protocol
-from urllib.parse import quote, urlencode, urljoin, urlparse
+from urllib.parse import parse_qsl, quote, urlencode, urljoin, urlparse, urlsplit, urlunsplit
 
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2779.

PR #2836, opened 2026-06-11
Fix unknown-method error code and add a protocol version registry

### LINKED ISSUE #1561: Invalid method names return error code -32602 instead of -32601
### Initial Checks

### Description

**How to reproduce:**
```bash
echo '{"jsonrpc":"2.0","method":"invalid/method","id":1,"params":{}}' | your-mcp-server
```

**Expected result:**
```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"Method not found"}}
```

**Actual result:**
```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Invalid request parameters","data":""}}
```

The issue appears to occur because request validation in `mcp/shared/session.py` (lines 360-375) catches the validation exception and returns `-32602` before the proper method-not-found handler in `mcp/server/lowlevel/server.py` (lines 716-718) can be reached.

### Example Code

```Python

```

### Python & MCP Python SDK

```Text
**Environment:**
- Python SDK version: `1.2.0`
- Python version: `3.12`
- Platform: macOS
```

maxisbey:
Bundled fixes around the same version/dispatch seams: unknown methods now get the JSON-RPC-correct error code on both sides, and protocol version comparisons go through a small registry instead of raw string compares.

## Motivation and Context

**Unknown methods answered `-32602` instead of `-32601`.** A request with an unknown method failed inside the request-union validation and fell into the generic invalid-params handler, so peers received `-32602` (Invalid params) instead of `-32601` (Method not found). This affected the server runner for pre-initialization requests and the client-side receive loop. JSON-RPC 2.0 reserves `-32601` for exactly this case, and clients that probe for optional/newer methods need to distinguish "method doesn't exist" from "bad params" to fall back correctly. Fixes #1561 (its repro — an unknown method sent before any handshake — is the exact pre-initialization path fixed here).

**Lexicographic version compares on wire input.** Three call sites compared protocol versions with raw string ordering (e.g. `>= "2025-11-25"`), including against the client-supplied `MCP-Protocol-Version` header. Any non-date string sorts *somewhere* (letters sort after digits, so e.g. `"draft"` compares newer than every release) and can silently flip version-gated behavior. Comparisons now go through `mcp.shared.version`: an ordered registry of released protocol versions plus `is_version_at_least()`, which treats unknown wire strings conservatively (no version-gated behavior enabled) and raises on an unknown *threshold*, since that's a typo in our own code.

Also included:
- `ServerSession.protocol_version`: read access to the negotiated version (the field was previously write-only internal state).
- `docs/migration.md` correction: it claimed extra fields on MCP types raise a `ValidationError`; they are silently ignored. The doc now describes actual behavior.

## How Has This Been Tested?

Regression tests for both error-code fixes were written first and verified to fail on `main` with the old behavior (`-32602` with empty data) before the fix. Full suite green locally at 100% line+branch coverage; pyright and ruff clean. The registry refactor of the three compare sites is behavior-preserving for all released versions (covered by equivalence tests); only unknown-string inputs change outcome, conservatively.

## Breaking Changes

No API changes. One wire-visible behavior change: requests with unknown methods now receive `-32601` (Method not found) instead of `-32602` — the JSON-RPC-specified code. Anything matching on the old, incorrect code for this case would need updating.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Commits are stacked for review (registry first, then the two error-code fixes, then the property and doc fix) and are independently revertable. A v1.x backport of the `-32601` fix is prepared separately.

### DIFF
--- a/src/mcp/shared/version.py
+++ b/src/mcp/shared/version.py
@@ -1,4 +1,4 @@
-"""Protocol-version registry and comparison/classification helpers.
+"""Protocol-version registry and comparison helpers.
 
 Date-string protocol revisions happen to sort lexicographically, but versions
 are an enumerated set, not an ordered scalar: future identifiers are not
@@ -16,33 +16,11 @@ KNOWN_PROTOCOL_VERSIONS: Final[tuple[str, ...]] = (
     "2025-03-26",
     "2025-06-18",
     "2025-11-25",
-    "2026-07-28",  # draft: recognized for ordering/classification, NOT negotiable
 )
-"""Every protocol revision this SDK knows about, oldest to newest."""
-
-DRAFT_PROTOCOL_VERSION: Final[str] = "2026-07-28"
-"""The in-progress spec revision.
-
-Recognized by the helpers in this module but absent from
-SUPPORTED_PROTOCOL_VERSIONS until the SDK actually implements it.
-"""
+"""Every released protocol revision, oldest to newest."""
 
 SUPPORTED_PROTOCOL_VERSIONS: list[str] = ["2024-11-05", "2025-03-26", "2025-06-18", LATEST_PROTOCOL_VERSION]
 """Protocol revisions this SDK can negotiate."""
-
-STATEFUL_PROTOCOL_VERSIONS: Final[frozenset[str]] = frozenset({"2024-11-05", "2025-03-26", "2025-06-18", "2025-11-25"})
-"""Revisions that negotiate via the initialize handshake.
-
-Closed by design: every revision after 2025-11-25 is stateless and negotiates
-per-request, never via initialize. Hardcoded - do not derive from
-SUPPORTED_PROTOCOL_VERSIONS. (Matches typescript-sdk's
-STATEFUL_PROTOCOL_VERSIONS / isStatefulProtocolVersion.)
-"""
-
-
-def is_stateful_protocol_version(version: str) -> bool:
-    """Return True if `version` negotiates via the initialize handshake."""
-    return version in STATEFUL_PROTOCOL_VERSIONS
 
 
 def is_version_at_least(version: str, minimum: str) -> bool:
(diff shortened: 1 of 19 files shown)


src/mcp/shared/version.py
```diff
@@ -1,3 +1,37 @@
...
+from typing import Final
+
 from mcp.types import LATEST_PROTOCOL_VERSION
 
+KNOWN_PROTOCOL_VERSIONS: Final[tuple[str, ...]] = (
+    "2024-11-05",
+    "2025-03-26",
+    "2025-06-18",
+    "2025-11-25",
+)
```
felixweinberger, line 19: not adding 2026-07-28 as part of this yet?

src/mcp/shared/version.py
```diff
@@ -1,3 +1,37 @@
...
+
+    Unknown `version` strings return False (treat unrecognized peers
+    conservatively). `minimum` must be a member of KNOWN_PROTOCOL_VERSIONS;
+    passing anything else is programmer error and raises ValueError.
+    """
+    if minimum not in KNOWN_PROTOCOL_VERSIONS:
+        raise ValueError(f"minimum must be a known protocol version, got {minimum!r}")
+    if version not in KNOWN_PROTOCOL_VERSIONS:
+        return False
+    return KNOWN_PROTOCOL_VERSIONS.index(version) >= KNOWN_PROTOCOL_VERSIONS.index(minimum)
```
felixweinberger, line 37: looks correct but feels brittle to tie the logic specifically to a random array constant being in the right order, but seems fine 🤷‍♂️ we won't change this that often I guess.

felixweinberger (approved):

src/mcp/shared/version.py
maxisbey: yep not yet

PR #2852, opened 2026-06-12
fix: correlate invalid JSON-RPC envelope errors with the original request id

### LINKED ISSUE #2848: Invalid JSON-RPC envelope errors are not correlated with the original request id
### Initial Checks

### Description

After a normal initialization flow, several id-bearing JSON-RPC messages that are syntactically valid JSON but invalid JSON-RPC request envelopes are not correlated back to the original request id.

This is not meant to require implementations to recover an id after every low-level parse/deserialization failure. The narrower concern is that these payloads are parsed far enough to produce request-specific validation errors, while the original top-level id is still present in the payload. Preserving that id where feasible would make the error response easier for clients to correlate.

The tested inputs were:

| Case                    | Request                                                | Validation detail                                       |
| ----------------------- | ------------------------------------------------------ | ------------------------------------------------------- |
| wrong `jsonrpc` version | `{"jsonrpc":"1.0","id":3,"method":"ping","params":{}}` | `JSONRPCRequest.jsonrpc`: value should be `"2.0"`       |
| missing `jsonrpc` field | `{"id":4,"method":"ping","params":{}}`                 | `JSONRPCRequest.jsonrpc`: field required                |
| `method` as number      | `{"jsonrpc":"2.0","id":8,"method":12345,"params":{}}`  | `JSONRPCRequest.method`: input should be a valid string |

For all three envelope-invalid inputs, I would expect `-32600 Invalid Request` or `-32602` error response that can be correlated to the original request where possible.

Observed behavior was consistent by transport:

- **stdio:** The server emits a `notifications/message` log notification with `level:"error"` and `data:"Internal Server Error"`, but no JSON-RPC error response is sent for the original request `id`. A follow-up `ping` succeeds.
- **Streamable HTTP:** The server returns HTTP 400 with a JSON-RPC error response using `id:"server-error"` and `code:-32602`, rather than the original request `id`. A follow-up `ping` succeeds over SSE.

### Example Code

```Python
"""
With a Python SDK MCP server running over stdio or Streamable HTTP,
complete a normal initialization flow, then send these requests in isolation.
"""

requests = [
    # Wrong JSON-RPC version: valid JSON, invalid JSON-RPC envelope.
    {"jsonrpc": "1.0", "id": 3, "method": "ping", "params": {}},

    # Missing jsonrpc field: valid JSON, invalid JSON-RPC envelope.
    {"id": 4, "method": "ping", "params": {}},

    # method is not a string: valid JSON, invalid JSON-RPC envelope.
    {"jsonrpc": "2.0", "id": 8, "method": 12345, "params": {}},
]
```

### Python & MCP Python SDK

```Text
* Python: `3.12.3`
* MCP Python SDK stable release: `v1.27.2` (`62137874ff26dd74d2fea80ff528a7fd9ca7a5e7`)
* Transports: stdio and Streamable HTTP
```

anneheartrecord:
Fixes #2848

When an incoming message is valid JSON but fails JSON-RPC envelope validation (e.g. `"jsonrpc": "1.0"`, a missing `jsonrpc` field, or a non-string `method`), the error response could not be correlated with the originating request:

- **stdio**: the server transport forwarded the exception in-stream and never answered, so the client got no response for its request id.
- **Streamable HTTP**: the server replied 400 with a JSON-RPC error using a null id (`"server-error"` on v1.x), breaking client-side request/response correlation.

This PR extracts the request id best-effort from the raw parsed payload (shared helper `extract_raw_request_id`; only spec-valid id types are accepted — strings and non-bool integers) and preserves it in the error response on both transports, falling back to a null id per the JSON-RPC 2.0 spec when no valid id can be extracted. The stdio transport now also answers unparseable lines with `-32700 Parse error` (null id), matching the HTTP transport's existing parse-error behavior. Envelope-invalid messages are now reported as `-32600 Invalid Request` instead of `-32602 Invalid params`, matching JSON-RPC 2.0 error code semantics (the issue's three repro cases are envelope violations, not parameter errors).

Tests cover all three repro cases from the issue on both transports, plus the null-id fallback (unextractable id and unparseable body). Full suite, `ruff`, and `pyright` pass.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -24,9 +24,36 @@ from io import TextIOWrapper
 import anyio
 import anyio.lowlevel
 import mcp_types as types
+import pydantic_core
 
 from mcp.shared._context_streams import create_context_streams
-from mcp.shared.message import SessionMessage
+from mcp.shared.message import SessionMessage, extract_raw_request_id
+
+
+def _error_response_for_invalid_line(line: str) -> SessionMessage:
+    """Build the JSON-RPC error response for a stdin line that failed message validation.
+
+    Correlates the error with the originating request where possible: for lines that
+    are valid JSON but an invalid JSON-RPC envelope, the request id is extracted
+    best-effort from the raw payload (Invalid Request, -32600); for lines that are
+    not valid JSON, a null id is used (Parse error, -32700), per the JSON-RPC 2.0
+    specification.
+
+    Args:
+        line: The raw stdin line that failed to validate as a JSON-RPC message.
+
+    Returns:
+        A `SessionMessage` wrapping the `JSONRPCError` to write back to the client.
+    """
+    try:
+        raw_message = pydantic_core.from_json(line)
+    except ValueError:
+        request_id = None
+        error = types.ErrorData(code=types.PARSE_ERROR, message="Parse error")
+    else:
+        request_id = extract_raw_request_id(raw_message)
+        error = types.ErrorData(code=types.INVALID_REQUEST, message="Invalid Request")
+    return SessionMessage(types.JSONRPCError(jsonrpc="2.0", id=request_id, error=error))
 
 
 @asynccontextmanager
@@ -53,6 +80,13 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
                     try:
                         message = types.jsonrpc_message_adapter.validate_json(line, by_name=False)
                     except Exception as exc:
+                        try:
+                            await write_stream.send(_error_response_for_invalid_line(line))
+                        except anyio.ClosedResourceError:
+                            # The server side already closed the write stream; the
+                            # error response cannot be delivered, but the exception
+                            # below still surfaces the bad line in-stream.
+                            await anyio.lowlevel.checkpoint()
                         await read_stream_writer.send(exc)
                         continue
 
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -23,10 +23,38 @@ from typing import BinaryIO, Literal, TextIO
 import anyio
 import anyio.lowlevel
 import mcp_types as types
+import pydantic_core
 
 from mcp.os.win32.utilities import rebind_std_handle_to_fd
 from mcp.shared._context_streams import create_context_streams
-from mcp.shared.message import SessionMessage
+from mcp.shared.message import SessionMessage, extract_raw_request_id
+
+
+def _error_response_for_invalid_line(line: str) -> SessionMessage:
+    """Build the JSON-RPC error response for a stdin line that failed message validation.
+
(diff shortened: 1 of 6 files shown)


anneheartrecord:
For reviewers — summary to ease review:

**Bug**: when an incoming message is valid JSON but fails JSON-RPC envelope validation (e.g. `"jsonrpc":"1.0"`, missing `jsonrpc`, non-string `method`), the error response couldn't be correlated to the request — stdio never answered at all; Streamable HTTP replied with a null id. (Issue #2848 has 3 repro cases.)

**Fix**: a shared `extract_raw_request_id()` helper pulls the id best-effort from the raw payload (only spec-valid id types — str / non-bool int), preserved in the error response on both transports, with null-id fallback per JSON-RPC 2.0 when no valid id exists.

**Two deliberate calls worth a look**:
1. stdio now also answers unparseable lines with `-32700 Parse error` (matching HTTP's existing behavior) instead of silently dropping.
2. Envelope-invalid messages now report `-32600 Invalid Request` instead of `-32602 Invalid params` (the repro cases are envelope violations, not param errors) — aligns with the codes touched in the recently-merged #2836.

Both repro cases covered on both transports + null-id fallbacks. Full suite (1785), ruff, pyright all pass. If you'd rather land just the id-correlation core and split the stdio parse-error / error-code changes, I'm glad to narrow it.

anneheartrecord:
Rebased onto `57394b0` — the branch had drifted into conflicts. Same six files, no substantive change to the diff.

On the `needs confirmation` label: all three envelope-invalid inputs from #2848 still behave as reported on current main. Running this PR's tests against unpatched `57394b0`:

- **stdio** — no JSON-RPC error response is written at all: `assert [] == [3, 4, 8, None]`, the ids that should have come back.
- **streamable HTTP** — 400 with `id: null` and `-32602` for all three, e.g. `{"jsonrpc":"2.0","id":null,"error":{"code":-32602,"message":"Validation error: 6 validation errors for union[...]"}}`.

Same numbers @Parker-Fawcett got on the issue.

One thing worth deciding before this lands. This PR covers stdio and the 2025-era streamable HTTP entry; the 2026-07-28 entry still answers an envelope-invalid body with `_INVALID_BODY`, i.e. `id: null`. I left it alone deliberately: that shape is currently asserted as SDK-defined in `test_handle_modern_request_rejects_a_body_that_is_neither_request_nor_notification`, and correlating there flips two of its params — `posted-response` and `posted-error` both carry a valid `id`, and echoing an id back on a body that isn't a request looks like the wrong answer. Restricting correlation to request-shaped bodies is a semantics call, not a bug fix, so I'd rather you make it: happy to extend this PR, or leave it as a follow-up.

Local on `f52691d`: 5748 passed / 9 skipped / 1 xfailed, `coverage report` 100.00%, `ruff check` and `ruff format --check` clean, `pyright` 0 errors.

PR #2853, opened 2026-06-12
fix: retry token refresh without RFC 8707 resource param when the AS rejects it

### LINKED ISSUE #2578: OAuth token refresh sends RFC 8707 resource parameter that Entra ID v2.0 rejects (AADSTS9010010)
## Problem

The MCP Python SDK sends an RFC 8707 `resource` parameter on **all** token requests — including `refresh_token` grants. Microsoft Entra ID v2.0 rejects this with **AADSTS9010010** (`The resource parameter provided in the request doesn't match with the requested scopes`).

This causes MCP servers using Entra ID OAuth to **lose authentication after ~1 hour** when the access token expires and the SDK attempts a silent refresh.

## Root Cause

Two compounding issues:

### 1. Entra v2.0 does not support `resource` on refresh

Entra's v2.0 token endpoint expects `scope`, not `resource`. The `resource` parameter is a v1.0 concept. Since March 2026, Entra strictly validates and rejects `resource` on token refresh (previously it was silently ignored).

### 2. Pydantic v2 `AnyHttpUrl` trailing-slash normalization

`ProtectedResourceMetadata.resource` is typed as `AnyHttpUrl` (`shared/auth.py:143`). When `str()` is called on a bare-domain URL, Pydantic v2 adds a trailing slash:

```python
>>> str(AnyHttpUrl("https://mcp-server.example.com"))
'https://mcp-server.example.com/'   # trailing slash added
```

In `get_resource_url()` (`client/auth/oauth2.py:155`), this trailing-slash version is used:
```python
prm_resource = str(self.protected_resource_metadata.resource)  # adds trailing slash
```

But the Entra app registration has the audience as `https://mcp-server.example.com` (no slash), so the `resource` and `scope` audience don't match.

## Affected Code

`src/mcp/client/auth/oauth2.py`:

```python
async def _refresh_token(self) -> httpx.Request:
    refresh_data = {
        "grant_type": "refresh_token",
        "refresh_token": self.context.current_tokens.refresh_token,
        "client_id": self.context.client_info.client_id,
    }
    # This sends 'resource' on refresh — Entra v2.0 rejects it
    if self.context.should_include_resource_param(self.context.protocol_version):
        refresh_data["resource"] = self.context.get_resource_url()  # RFC 8707
```

The same issue exists in the TypeScript SDK (`packages/client/src/client/auth.ts`), where WHATWG `URL` also normalizes bare-domain URLs with a trailing slash.

### Option A: Strip trailing slash in `get_resource_url()`
```python
def get_resource_url(self) -> str:
    resource = resource_url_from_server_url(self.server_url)
    if self.protected_resource_metadata and self.protected_resource_metadata.resource:
        prm_resource = str(self.protected_resource_metadata.resource).rstrip('/')
        if check_resource_allowed(requested_resource=resource, configured_resource=prm_resource):
            resource = prm_resource
    return resource
```

### Option B: Include `scope` alongside `resource` on refresh
Entra v2.0 tolerates `resource` if `scope` is also present and consistent:
```python
refresh_data["scope"] = " ".join(self.context.scopes)
```

### Option C: Make `resource` on refresh configurable
Allow servers to signal whether the `resource` parameter should be included on refresh grants, since not all authorization servers support RFC 8707.

## Related Issues

- https://github.com/anthropics/claude-code/issues/52871 — same trailing-slash + AADSTS9010010 bug
- https://github.com/microsoft/powerbi-modeling-mcp/issues/68 — same Entra v2.0 incompatibility

## Environment

- MCP Python SDK: v1.27.0
- Authorization server: Microsoft Entra ID v2.0
- MCP server: Azure Container Apps with custom EntraTokenVerifier
- MCP spec version: 2025-06-18 (mandates RFC 8707 `resource`)

## Current Workaround

Server-side: set `resource_server_url=None` in `AuthSettings` and do NOT serve `/.well-known/oauth-protected-resource` metadata. Without PRM, `should_include_resource_param()` returns `False` and `resource` is omitted from refresh requests. Initial auth still works via the `WWW-Authenticate` header fallback.

fede-kamel:
## Motivation and Context

Fixes #2578.

Microsoft Entra ID v2.0 rejects the RFC 8707 `resource` parameter on `refresh_token` grants (`AADSTS9010010`, strictly enforced since March 2026). Because the SDK sends it on every refresh, MCP clients using Entra OAuth lose authentication after ~1 hour: the silent refresh fails, tokens are cleared, and the user is forced through interactive re-auth every hour.

At the same time, the MCP authorization spec (2025-06-18 and later, "Resource Parameter Implementation") requires the `resource` parameter in token requests, and RFC 8707 §2.2 defines it for all grant types including refresh. So unconditionally omitting it on refresh (this PR's original approach) would trade the Entra bug for a spec violation — thanks to the review feedback below for pushing on this.

**This PR now keeps the spec-required behavior and adds a targeted fallback:**

- The `resource` parameter is still sent on refresh requests exactly as before.
- If the refresh fails with HTTP 400 and the error is **not** `invalid_grant`, the refresh is retried **once** without the `resource` parameter before falling back to full re-authentication.
- `invalid_grant` (a dead refresh token) never triggers the retry, since resending without `resource` cannot help.

For conformant authorization servers nothing changes. For Entra-style servers, the cost is one extra request on an already-failing path, instead of an hourly interactive re-auth.

## How Has This Been Tested?

- `tests/client/test_auth.py` + `tests/interaction/auth/test_lifecycle.py`: 155 passed, 1 xfailed (`uv run --frozen pytest`)
- New `TestRefreshResourceParamFallback` covers: the Entra-style 400 → retry without `resource` → success path; retry failure → full re-auth; `invalid_grant` → no retry; old protocol version (no `resource` sent) → no retry
- Existing spec-conformance tests (resource included on refresh for 2025-06-18+) pass unchanged
- `ruff check` and `ruff format` clean

## Breaking Changes

None. Default behavior is unchanged for all conformant authorization servers; the retry only engages after a 400 rejection that is not `invalid_grant`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -54,6 +54,18 @@ from mcp.shared.version import is_version_at_least
 logger = logging.getLogger(__name__)
 
 
+def _normalize_resource_url(resource: str) -> str:
+    """Undo the trailing slash URL parsers add to bare-domain URLs (e.g. pydantic's AnyHttpUrl).
+
+    RFC 9728 requires exact-string identity on the resource identifier, so only a root path
+    with no query or fragment is stripped; trailing slashes on deeper paths are preserved.
+    """
+    parsed = urlparse(resource)
+    if parsed.path == "/" and not parsed.params and not parsed.query and not parsed.fragment:
+        return f"{parsed.scheme}://{parsed.netloc}"
+    return resource
+
+
 class PKCEParameters(BaseModel):
     """PKCE (Proof Key for Code Exchange) parameters."""
 
@@ -152,7 +164,7 @@ class OAuthContext:
 
         # If PRM provides a resource that's a valid parent, use it
         if self.protected_resource_metadata and self.protected_resource_metadata.resource:
-            prm_resource = str(self.protected_resource_metadata.resource)
+            prm_resource = _normalize_resource_url(str(self.protected_resource_metadata.resource))
             if check_resource_allowed(requested_resource=resource, configured_resource=prm_resource):
                 resource = prm_resource
 
@@ -441,9 +453,8 @@ class OAuthClientProvider(httpx.Auth):
             "client_id": self.context.client_info.client_id,
         }
 
-        # Only include resource param if conditions are met
-        if self.context.should_include_resource_param(self.context.protocol_version):
-            refresh_data["resource"] = self.context.get_resource_url()  # RFC 8707
+        # The RFC 8707 resource param is deliberately omitted: some providers
+        # (e.g. Entra ID v2.0) reject it on refresh_token grants.
 
         # Prepare authentication based on preferred method
         headers = {"Content-Type": "application/x-www-form-urlencoded"}
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -5,6 +5,7 @@ Implements authorization code flow with PKCE and automatic token refresh.
 
 import base64
 import hashlib
+import json
 import logging
 import secrets
 import string
@@ -79,6 +80,15 @@ _SECRET_TOKEN_ENDPOINT_AUTH_METHODS = ("client_secret_post", "client_secret_basi
 _REGISTRATION_USABLE_TOKEN_ENDPOINT_AUTH_METHODS: tuple[str | None, ...] = tuple(
     method for method in _KNOWN_TOKEN_ENDPOINT_AUTH_METHODS if method != "private_key_jwt"
 )
+
+
+def _refresh_error_code(body: bytes) -> str | None:
+    """Extract the RFC 6749 ``error`` code from a failed token response, if any."""
+    try:
+        error = json.loads(body).get("error")
+    except (json.JSONDecodeError, UnicodeDecodeError, AttributeError):
+        return None
+    return error if isinstance(error, str) else None
 
 
 def check_registration_usable(client_info: OAuthClientInformationFull) -> None:
@@ -486,8 +496,15 @@ class OAuthClientProvider(httpx2.Auth):
         self.context.update_token_expiry(token_response)
         await self.context.storage.set_tokens(token_response)
 
-    async def _refresh_token(self) -> httpx2.Request:
-        """Build token refresh request."""
+    async def _refresh_token(self, *, include_resource: bool = True) -> httpx2.Request:
+        """Build token refresh request.
+
+        Args:
+            include_resource: Whether to attach the RFC 8707 ``resource`` parameter
+                (when the protocol version calls for it). The retry path passes False
+                for authorization servers that reject the parameter on
+                ``refresh_token`` grants (e.g. Microsoft Entra ID v2.0, AADSTS9010010).
+        """
         if not self.context.current_tokens or not self.context.current_tokens.refresh_token:
             raise OAuthTokenError("No refresh token available")  # pragma: no cover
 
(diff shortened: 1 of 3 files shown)


fede-kamel:
@maxisbey @Kludex — final maintainer-selection request for #2578. I have just rebased this branch onto current main (e8afa1a5), then rerun the focused OAuth surfaces: 153 passed, 1 expected xfail. I also reran the repository test gate after the rebase.\n\nThe issue is still open and there have been multiple overlapping fixes. This PR keeps the change deliberately small: omit resource from refresh_token grants, and normalize only the artificial root trailing slash while preserving meaningful path slashes.\n\nCould you please make a one-line decision on whether this is the implementation you want to take, whether one of the other proposals should be selected instead, or whether the behavior needs a different design? If it is not going to be taken, I will close this PR in 10 days rather than leave duplicate stale work open. Thank you.

fede-kamel:
@maxisbey could this get a look? It has been open since June 12 and #2578 still reproduces on today's main: with PRM present and protocol version 2025-06-18, the refresh request body is ['client_id', 'grant_type', 'refresh_token', 'resource'], and Entra ID v2.0 rejects the resource param on refresh_token grants (AADSTS9010010), so Entra-backed MCP clients lose auth every hour.

I've just rebased onto main and slimmed the PR: the trailing-slash half was superseded by #2925 and is dropped, leaving only the refresh-grant fix. 151 auth tests pass, ruff clean.

keeltrace:

I think this change fixes the Entra-specific failure by violating the MCP authorization profile for the protocol versions where `resource` is required.

There are two separate standards layers here:

1. **RFC 8707 itself is permissive.** A generic OAuth client MAY use `resource`, but §2.2 explicitly defines it for token-endpoint requests **for all grant types**, and specifically discusses `refresh_token` grants. So `resource` is not out of place on a refresh request; it can be used to select/audience-restrict the newly issued access token.

2. **MCP is stricter than bare RFC 8707.** The MCP 2025-06-18 and 2025-11-25 Authorization specs' “Resource Parameter Implementation” require the `resource` parameter in both authorization requests **and token requests**, and say clients MUST send it regardless of whether the authorization server supports it. The current draft keeps the same requirement.

Before this PR, `_refresh_token()` follows that MCP profile via:

```python
if self.context.should_include_resource_param(self.context.protocol_version):
    refresh_data["resource"] = self.context.get_resource_url()
```

This PR removes the check entirely and the updated test now asserts that a 2025-06-18 refresh omits `resource`. That means the test is pinning behavior contrary to that protocol revision's stated requirement.

So I don't think the global statement in the PR body — “Authorization servers that accept it on refresh ignore its absence” — is safe. RFC 8707 allows an AS to use `resource` on a refresh-token access-token request to determine the target resource, and MCP intentionally requires clients to send it.

The Entra interoperability problem looks real, but could it be isolated as an explicit provider-compatibility escape hatch rather than changing the default MCP behavior for every AS? At minimum I think the generic path should preserve `resource` for MCP revisions that require it, with a regression test proving that, and any Entra workaround should be opt-in/provider-specific.

Checked against PR head `d32f7c2`. This is a standards/code-path review; I did not run the repository suite locally.

fede-kamel:
@keeltrace you're right, and thanks for the careful read. I checked RFC 8707.

Reworked as of 69bfdbb: the resource param stays on refresh per spec, and on a 400 whose error is not invalid_grant the refresh is retried once without it before falling back to full re-auth. That keeps conformant servers untouched and turns the Entra failure into one extra request instead of an hourly interactive re-auth. Rather than a provider-specific opt-in flag, the fallback is generic — Entra is detected by behavior, not by name — which also covers other non-conformant ASs. New tests cover the retry, the invalid_grant exclusion, and the no-resource-sent case; the pre-existing spec-conformance tests pass unchanged.

Issue #2644, opened 2026-05-19
Bug: `anyio.Lock` in `oauth2.py` raises "current task is not holding this lock" under cross-task generator driving


gILisH:
# Bug: `anyio.Lock` in `oauth2.py` raises "current task is not holding this lock" under cross-task generator driving

## Summary

`mcp/client/auth/oauth2.py` declares its OAuth state lock as `anyio.Lock` (via a `field(default_factory=anyio.Lock)` on the dataclass). When `httpx` drives the `async_auth_flow` async generator from a task different from the one that originally acquired the lock, the `release()` call raises:

```
RuntimeError: The current task is not holding this lock
```

This happens because `anyio.Lock` records task identity at `acquire()` time and verifies it at `release()` — but `httpx`'s generator-driving pattern does not guarantee the driving task is the same task across `__anext__` calls. The lock object's task-identity invariant is incompatible with how async generators are consumed downstream.

## Reproduction

Install the MCP SDK in any context where an `httpx.AsyncClient` drives `async_auth_flow` across multiple `httpx` request retries. Under sufficient concurrency the task-identity check eventually fails. We see this reliably with Hermes Agent's MCP client when handling token refresh under load.

Stack trace shape:

```
File ".../mcp/client/auth/oauth2.py", line <N>, in async_auth_flow
    async with self.lock:
File ".../anyio/_core/_synchronization.py", line <N>, in __aexit__
    self.release()
File ".../anyio/_core/_synchronization.py", line <N>, in release
    raise RuntimeError("The current task is not holding this lock")
```

## Root cause

`anyio.Lock` is designed for the "one task acquires, same task releases" pattern. Async generators driven by `httpx` violate that contract because `httpx`'s retry/redirect machinery may resume the generator from a different task than the one that suspended it. `asyncio.Lock` does NOT check task identity on release — it's a simple FIFO synchronization primitive, suitable for this pattern.

## Proposed fix

```diff
- import anyio
+ import anyio
+ import asyncio
  ...
- lock: anyio.Lock = field(default_factory=anyio.Lock)
+ lock: asyncio.Lock = field(default_factory=asyncio.Lock)
```

(`asyncio.Lock` is also reentrant-safe for the use pattern in `async_auth_flow` — the only thing being given up is the anyio-cross-runtime portability, which `async_auth_flow` doesn't rely on since it's already inside an asyncio context driven by httpx.)

## Workaround

[...]

PR #2858, opened 2026-06-13
fix(oauth): narrow context.lock scope in async_auth_flow (rebase of #2660 by @peisuke, closes #2847)

### LINKED ISSUE #2847: Bug: anyio.Lock in async_auth_flow causes RuntimeError under concurrent OAuth MCP connections
## Summary

Confirmed production occurrence of the bug described in #2644 (self-closed by the original reporter before they could follow through). Filing to provide a second validated repro and draw attention to the existing draft fix in #2660.

## Environment

- `mcp==1.26.0`
- `anyio==4.13.0`
- `httpx==0.28.1`
- Python 3.11.14, macOS

## What happens

When a gateway session starts and connects to multiple OAuth-authenticated MCP servers concurrently (Notion + TinyFish, both OAuth 2.1 PKCE), Notion fails intermittently:

```
ERROR asyncio: Task exception was never retrieved
RuntimeError: The current task is not holding this lock
  File ".../mcp/client/auth/oauth2.py", line 503, in async_auth_flow
  File ".../mcp/client/auth/oauth2.py", line 484, in async_auth_flow
    raise RuntimeError("The current task is not holding this lock")

WARNING tools.mcp_tool: MCP server 'notion' connection lost (attempt 1/5), reconnecting in 1s
WARNING tools.mcp_tool: Failed to connect to MCP server 'notion': CancelledError
INFO tools.mcp_tool: MCP: registered 114 tool(s) from 4 server(s) (1 failed)
```

Notion is affected more often than TinyFish because its OAuth token refreshes frequently (~every 15–60 min), consistently triggering the refresh `yield` path in `async_auth_flow`. TinyFish tokens expire less often and typically take the happy path (add header, yield once).

## Root cause

`OAuthContext.lock` is `anyio.Lock`, which records task identity at `acquire()` and enforces same-task `release()`. `async_auth_flow` is an async generator that holds this lock across `yield` points. When httpx drives the generator from a different task during concurrent connections, `anyio.Lock.release()` throws.

## Existing draft fix

PR #2660 addresses this correctly by narrowing lock scope so no lock is held across yields — GET SSE long-polls and token refresh yields both run outside any lock. It has full test coverage (100% on `oauth2.py`, 1177 passed) and no breaking changes, but has been sitting as a draft without maintainer review since May 22.

## Workaround applied locally

Replacing `anyio.Lock` with `asyncio.Lock` in `OAuthContext` stops the error since `asyncio.Lock` does not enforce task identity on release. This is a bandaid — it loses trio portability — but unblocks asyncio deployments. Note: this is a mechanical fix and has not been load-tested to exhaustion; the intermittent nature of the bug means full verification requires sustained concurrent load.

## Request

Could a maintainer review and merge PR #2660? The fix is principled and well-tested.

Bartok9:
## Summary

Conflict-free **salvage of #2660 by @peisuke** onto current `main`, opened as a fresh, mergeable PR so the fix can land without waiting on the original draft to be un-conflicted. **All credit for the fix goes to @peisuke** — the commit is authored by them; I only resolved the rebase against `main` and squashed to a single commit.

Closes #2847. Addresses the same `RuntimeError: The current task is not holding this lock` reported in #2644 and #2847.

## Why a new PR

#2660 has been a `CONFLICTING` draft since May 22 with no maintainer review. Issue #2847 (June 12) is a **second, independent production repro** of the exact bug, and the reporter explicitly asks for #2660 to be reviewed/merged. Rather than let the fix keep sitting, this PR provides a ready-to-merge version. If maintainers prefer to merge #2660 directly, or @peisuke wants to push the rebase themselves, this PR can be closed — the goal is just to unblock the fix.

## The bug

`OAuthContext.lock` is an `anyio.Lock`, which records task identity at `acquire()` and enforces same-task `release()`. `async_auth_flow` holds this lock across `yield` points. When httpx drives the generator from a *different* task during concurrent OAuth connections (e.g. Notion's frequent token refresh behind a gateway), `release()` raises `RuntimeError: The current task is not holding this lock`.

## The fix (@peisuke's approach)

Narrow the lock scope so no HTTP `yield` (including the long-poll GET SSE and the token-refresh round trips) runs while holding `context.lock`, plus a single-flight `refresh_lock` with a re-check under the lock. Keeps trio portability (unlike swapping to `asyncio.Lock`).

## Verification

Rebased onto current `main`. Single commit (authored by @peisuke), touches only `src/mcp/client/auth/oauth2.py` and `tests/client/test_auth.py` (+348/-18).

```
$ uv run pytest tests/client/test_auth.py -q
131 passed, 1 xfailed

$ uv run ruff check src/mcp/client/auth/oauth2.py tests/client/test_auth.py
All checks passed!
```

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -1,4 +1,4 @@
-"""OAuth2 Authentication implementation for httpx2.
+"""OAuth2 Authentication implementation for HTTPX.
 
 Implements authorization code flow with PKCE and automatic token refresh.
 """
@@ -11,15 +11,14 @@ import string
 import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from dataclasses import dataclass, field
-from typing import Any, Protocol, get_args
+from typing import Any, Protocol
 from urllib.parse import quote, urlencode, urljoin, urlparse
 
 import anyio
-import httpx2
-from mcp_types.version import is_version_at_least
+import httpx
 from pydantic import BaseModel, Field, ValidationError
 
-from mcp.client.auth.exceptions import OAuthFlowError, OAuthRegistrationError, OAuthTokenError
+from mcp.client.auth.exceptions import OAuthFlowError, OAuthTokenError
 from mcp.client.auth.utils import (
     build_oauth_authorization_server_metadata_discovery_urls,
     build_protected_resource_metadata_discovery_urls,
@@ -41,6 +40,7 @@ from mcp.client.auth.utils import (
     validate_authorization_response_iss,
     validate_metadata_issuer,
 )
+from mcp.client.streamable_http import MCP_PROTOCOL_VERSION
 from mcp.shared.auth import (
     AuthorizationCodeResult,
     OAuthClientInformationFull,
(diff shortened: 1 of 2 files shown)


ctaylor86:
I reproduced this downstream with MCP SDK 1.28.1 against a production Streamable HTTP OAuth server: OAuth and `initialize` succeeded, but a session-long GET held `OAuthContext.lock`, so the subsequent `tools/list` POST never reached the network and timed out. Supplying the same stored bearer token directly listed 171 tools immediately.

I opened the `v1.x` backport in #3243 and validated it against that server plus two existing MCP connections. While testing, I found two details that may also be relevant to this `main` PR:

1. The final resource retries after the 401/403 branches must also yield outside the provider-state critical section; they can themselves be session-long GET requests.
2. `anyio.Lock` is task-owned. HTTPX can resume or close an auth async generator from a different task, reproducing `RuntimeError: The current task is not holding this lock` even for bounded OAuth requests. #3243 uses task-agnostic bounded semaphores for critical sections and includes a cross-task generator-close regression test.

The backport also serializes refresh and authorization transitions under one semaphore, re-checks tokens after acquiring it, and restores each request's MCP protocol version before RFC 8707 request construction.

Bartok9:
Thanks @ctaylor86 — really helpful production repro and the v1.x work in #3243.

Agree on both points for `main` as well:

1. **Final resource retries after 401/403** must yield outside the provider-state critical section. Session-long GETs on those paths would reintroduce the same lock-hold / starvation you hit (`initialize` OK, then `tools/list` blocked behind the long GET).
2. **Task-owned `anyio.Lock`** is still a footgun even with narrowed scope if HTTPX resumes or closes the auth generator from another task. Task-agnostic bounded semaphores + a cross-task generator-close regression (as in #3243) look like the right hardening beyond the original #2660 narrow-scope fix.

This PR is a conflict-free salvage of @peisuke's #2660 onto `main` (credit stays with them). Your backport goes further on semaphore choice, single-flight refresh/auth transitions, protocol-version restore before RFC 8707 construction, and live validation — maintainers may want to land the fuller design on `main` via #3243's approach (or cherry-picks) rather than only the narrower lock-scope change here.

Happy to fold (1)/(2) into this branch if that's the preferred vehicle for `main`; otherwise I'll defer to whatever landing path maintainers pick between #2858 / #3243 / #2660. Appreciate the downstream proof (171 tools once the lock wasn't held across the session GET).

PR #2859, opened 2026-06-13
feat(auth): proactive OAuth token refresh with jitter to reduce concurrent refresh spikes

Bartok9:
## Summary

Refresh OAuth tokens **proactively** at ~80% of their lifetime with a small random jitter, instead of only **reactively** once they've already expired. This reduces the "thundering herd" of simultaneous token refreshes that occurs when a fleet of OAuth-backed MCP connectors is provisioned around the same time.

## The production problem

When many MCP clients each hold an OAuth connection and were provisioned (or last refreshed) in roughly the same window, their access tokens all expire inside the same narrow window too. Today refresh only fires **after** `is_token_valid()` returns `False` (i.e. after hard expiry), so all of those clients try to refresh at nearly the same moment.

For a large fleet that produces a synchronized burst of `grant_type=refresh_token` requests against the authorization server — contention, rate-limit (429) responses, and spurious auth failures, all clustered into the same ~60s window. The herd then re-synchronizes on the new tokens and the spike repeats on the next cycle.

## The design

Add a per-connection **proactive refresh point** that sits before hard expiry and is individually jittered, so a fleet desynchronizes naturally:

```
refresh_at = now + expires_in * refresh_fraction - jitter
```

- **`refresh_fraction` = 0.8** by default → refresh once 80% of the lifetime has elapsed, leaving headroom before hard expiry.
- **`jitter` ∈ [0, 30s]** by default, always **subtracted** so it can only pull the refresh point *earlier* — it can never push past hard expiry. Each connector draws its own jitter, so refreshes spread out across the window rather than bunching up.

New pieces:

1. **`calculate_token_refresh_time(expires_in, *, refresh_fraction=0.8, max_jitter_seconds=30.0, jitter=None)`** in `src/mcp/shared/auth_utils.py` — pure, deterministic-testable (inject `jitter` to bypass the RNG), returns `None` when `expires_in` is `None`.
2. **`OAuthContext.token_refresh_time`** field, set alongside `token_expiry_time` in `update_token_expiry` and cleared in `clear_tokens`.
3. **`OAuthContext.should_refresh_token()`** — `True` when we hold refreshable tokens *and* we're past the jittered proactive-refresh point, even if the token is still technically valid.
4. **`async_auth_flow` Phase 1** now refreshes when the token is hard-invalid **OR** `should_refresh_token()` is `True` (while `can_refresh_token()`), keeping the existing re-check / lock structure intact.

`is_token_valid()` is deliberately unchanged — it still gates whether a token is *usable at all* (hard validity). Proactive refresh is layered on top.

## Edge cases handled

- **`expires_in is None`** → `token_refresh_time` is `None`; `should_refresh_token()` returns `False` and behavior degrades to the existing reactive path.
- **Tiny TTLs** (e.g. `expires_in` smaller than `max_jitter_seconds`): jitter is clamped to the available `(refresh_at - now)` window so the result never goes negative or before `now`.
- **Never past hard expiry**: the result is always clamped into `(now, hard_expiry]`.
- **String `expires_in`** (some servers return it as a string): handled via `int()` like `calculate_token_expiry`.

## Backward compatibility

Fully backward compatible. No public signatures change; defaults preserve the current behavior shape (proactive refresh is strictly an improvement, not a breaking change). Clients that never got an `expires_in` keep the old reactive behavior exactly.

## Test coverage

- `tests/shared/test_auth_utils.py` — 9 new tests for `calculate_token_refresh_time`: normal TTL within the jitter window and strictly before hard expiry, `None` → `None`, string `expires_in`, deterministic injected jitter, jitter ordering (more jitter → earlier), never-past-hard-expiry across many TTLs, tiny-TTL no-negative, zero-TTL collapse, custom fraction.
- `tests/client/test_auth.py` — `should_refresh_token()` predicate (hard-valid-but-past-window → True; fresh → False; no refresh time → False; no refresh token → False), plus two `async_auth_flow` integration tests: one proving a **proactive** refresh request is yielded while the token is still hard-valid, and one proving a fresh token is used directly with no refresh.

All `tests/client/test_auth.py` + `tests/shared/test_auth_utils.py` pass (128 passed, 1 xfailed). `uv run ruff check`, `uv run ruff format --check`, and `uv run pyright` are clean on all touched files.

## Relationship to #2858

This is **complementary to and independent of** #2858. That PR addresses *concurrency/locking* of refresh (narrowing the `anyio.Lock` scope + single-flight `refresh_lock` to fix a `RuntimeError`). This PR is purely about **when** a refresh fires (proactive + jittered), not how it's locked. They touch different concerns and compose cleanly; if #2858 lands first this will need only a trivial rebase.

## Credit

Motivated by production feedback from @Ben-Home (CorpusIQ) on #2847. Refs #2847, #2858.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -46,6 +46,7 @@ from mcp.shared.auth import (
 )
 from mcp.shared.auth_utils import (
     calculate_token_expiry,
+    calculate_token_refresh_time,
     check_resource_allowed,
     resource_url_from_server_url,
 )
@@ -113,6 +114,9 @@ class OAuthContext:
     # Token management
     current_tokens: OAuthToken | None = None
     token_expiry_time: float | None = None
+    # Jittered point (before hard expiry) at which to proactively refresh, so a fleet
+    # of connectors does not all refresh in the same window. See should_refresh_token.
+    token_refresh_time: float | None = None
 
     # State
     lock: anyio.Lock = field(default_factory=anyio.Lock)
@@ -123,11 +127,12 @@ class OAuthContext:
         return f"{parsed.scheme}://{parsed.netloc}"
 
     def update_token_expiry(self, token: OAuthToken) -> None:
-        """Update token expiry time using shared util function."""
+        """Update token expiry and proactive-refresh times using shared util functions."""
         self.token_expiry_time = calculate_token_expiry(token.expires_in)
+        self.token_refresh_time = calculate_token_refresh_time(token.expires_in)
 
     def is_token_valid(self) -> bool:
-        """Check if current token is valid."""
+        """Check if current token is valid (i.e. usable, not past hard expiry)."""
         return bool(
             self.current_tokens
             and self.current_tokens.access_token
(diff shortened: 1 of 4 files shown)


maxisbey:

Bartok9:
Totally understood — thanks for the context on the v2 backlog cleanup and the issue-first process. I'll open a triaged issue first if this stays relevant post-v2. Appreciate the quick note!

PR #2866, opened 2026-06-14
fix(server): duplicate stdio fds to prevent closing real stdio

superShen0916:
## Summary
Fixes #1933. When using `transport="stdio"`, the async wrappers around `sys.stdin`/`sys.stdout` would close the underlying file descriptors on server exit, causing subsequent operations like `print()` to fail with `ValueError: I/O operation on closed file`.

## Changes
- `src/mcp/server/stdio.py`: Use `os.dup()` to duplicate the underlying file descriptors before wrapping them in async files. Falls back to the original behavior when the stream lacks a real fd (e.g. `BytesIO` in tests).

## Test plan
- Existing `tests/server/test_stdio.py` tests exercise both paths (explicit streams with `BytesIO` and system streams via `monkeypatch`)

## Risk
- Low. The fix only affects the default path (when stdin/stdout are not provided explicitly). The `os.dup()`/`os.fdopen()` pattern is standard Unix practice for isolating fd lifecycle.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -17,9 +17,10 @@ Example:
     ```
 """
 
+import os
 import sys
 from contextlib import asynccontextmanager
-from io import TextIOWrapper
+from io import TextIOWrapper, UnsupportedOperation
 
 import anyio
 import anyio.lowlevel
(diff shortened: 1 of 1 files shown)


maxisbey:

PR #2022, opened 2026-02-09
Add idle session timeout to StreamableHTTPSessionManager

felixweinberger:
Port of #1994 (v1.x) to main.

## Motivation and Context

Sessions created via `StreamableHTTPSessionManager` persist indefinitely in `_server_instances` even after the client disconnects, leaking memory over time (issue #1283).

The `session_idle_timeout` parameter automatically terminates and removes sessions that receive no HTTP requests for the configured duration. Each session manages its own lifetime via an anyio `CancelScope` deadline — no background reaper task needed. Incoming requests push the deadline forward to keep active sessions alive.

## How Has This Been Tested?

Tests added to `tests/server/test_streamable_http_manager.py`: idle session reaping (deterministic, no sleeps), terminate idempotency, and parameterized input validation. All existing tests pass unchanged.

## Breaking Changes

None — new optional parameter with default `None`.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #2874, opened 2026-06-15
fix(server): expose session_idle_timeout via streamable_http_app() and pause reaping during active requests

### LINKED ISSUE #2455: session_idle_timeout is not exposed via streamable_http_app() and can cancel active requests

Bartok9:
## Summary

- Expose `session_idle_timeout` through the high-level `Server.streamable_http_app()` / `MCPServer.streamable_http_app()` API so users no longer have to wire up `StreamableHTTPSessionManager` by hand to configure idle reaping.
- Stop counting a request that is actively being processed as an idle session — idle reaping is suspended for the lifetime of every in-flight request.

## Motivation

Closes #2455.

The `session_idle_timeout` feature (from #1994 / #2022) had two linked problems:

1. **Not exposed by `streamable_http_app()`** — the parameter existed on `StreamableHTTPSessionManager` but not on the canonical app-builder API, forcing users to drop down to manual session-manager wiring.
2. **Could cancel active requests** — the idle deadline was only pushed forward when a request *arrived*. A request still being processed when the deadline elapsed could have its session reaped mid-flight, since "in-flight" was never distinguished from "idle."

## Fix

- Thread `session_idle_timeout` through both `streamable_http_app()` methods into the session manager.
- Add `mark_request_started()` / `mark_request_finished()` to `StreamableHTTPServerTransport`:
  - `started` increments an active-request counter and pushes the idle deadline to `+inf` (suspends reaping).
  - `finished` decrements the counter and re-arms the deadline only when the **last** concurrent request completes; it is a no-op when no timeout is configured, and the counter never underflows.
- Wrap both stateful `handle_request()` call sites in `try/finally` start/finished pairs, so an active request is never counted as idle.
- Move `task_status.started()` to **after** `idle_scope` is attached, so the very first request can't race ahead of the scope it needs to suspend.

## Verification

- `uv run pytest tests/server/test_streamable_http_manager.py tests/server/test_streamable_http_security.py -q` — 32 passed
- `uv run pytest tests/ -q -k "streamable or idle or session_manager"` — 249 passed
- Existing `test_idle_session_is_reaped` still green (reaping of genuinely idle sessions unchanged).
- 7 new tests: app forwarding + default `None`; transport suspend/resume across overlapping requests; no-timeout no-op; counter-underflow guard.
- `ruff format` + `ruff check` + `pyright` clean on all changed files.

## Notes

Rebuilds the intent of the stale, now-conflicting #2457 by @shaun0927 (same root cause and approach) cleanly against current `main` after the server modules were refactored. Crediting the original approach on that PR.


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2457.

PR #113, opened 2024-12-26
fix mcp.types.Tool is not following python's naming convention snake_case 

azdolinski:
Renamed the `inputSchema` field to `input_schema` in the `Tool` class to align with Python's naming conventions (snake_case). Added an alias `inputSchema` for backward compatibility.

## Motivation and Context

The name `inputSchema` does not follow Python's naming convention (PEP 8). This change improves code readability and consistency. Fix related: #97 

## How Has This Been Tested?

Tested locally by running existing tests.

## Breaking Changes

- [X] Yes. Users will need to update references to the `inputSchema` field to `input_schema`. Adding an alias is intended to ease the migration.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

## Additional context

The addition of the `inputSchema` alias is intended to provide backward compatibility and facilitate the migration process for users.

PR #2877, opened 2026-06-15
ci(conformance): pin harness to 0.2.0-alpha.3 with expected-failures baseline

maxisbey:
Modernizes the conformance CI to match the [typescript-sdk pattern](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/.github/workflows/conformance.yml) so the job actually gates and the failure set burns down per SEP. Supersedes #1921 (the composite-action approach) — pinning the npm package directly is what typescript-sdk settled on.

## Motivation and Context

The conformance jobs were pinned to `0.1.10`/`0.1.13` (drifted between server and client), ran with `continue-on-error: true` so they never gated, and had no baseline file — failures were silently ignored. The conformance harness has since gained `--expected-failures` ([#113](https://github.com/modelcontextprotocol/conformance/pull/113)) and the `0.2.0-alpha` line with version-aware scenario selection.

## Changes

- Pin the harness via a single workflow-level `CONFORMANCE_VERSION: "0.2.0-alpha.3"` env var (one place to bump).
- Add `.github/actions/conformance/expected-failures.yml` — the path the conformance repo's [`known-sdks.ts`](https://github.com/modelcontextprotocol/conformance/blob/main/src/sdk-runner/known-sdks.ts) already expects. Client baseline is 16 scenarios grouped by SEP (same set as [typescript-sdk's baseline](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/test/conformance/expected-failures.yaml)); server `active` suite is fully green so `server: []`.
- Drop `continue-on-error: true` from both jobs. The runner now exits 0 only when failures match the baseline, 1 on regressions or stale entries.
- Harden `run-server.sh` with a port-conflict guard, dead-process check in the readiness loop, and `curl --max-time`, mirroring typescript-sdk's wrapper.

The server `--suite draft` step (2026-07-28 scenarios) is a follow-up PR; the baseline file already has a `server:` placeholder for it.

## How Has This Been Tested?

Run locally against the conformance harness at `0.2.0-alpha.3`:

- `server --suite active --expected-failures ...` → 30/30 scenarios pass, 42 assertions, exit 0.
- `client --suite all --expected-failures ...` → 40 scenarios: 24 pass, 16 expected-fail, 0 unexpected, 0 stale, exit 0.
- Stale-entry guard: appending a passing scenario (`initialize`) to the baseline → exit 1 with `Stale baseline entries (now passing - remove from baseline): initialize`.
- `bash -n`, YAML parse, and pre-commit all clean.

## Breaking Changes

None for SDK users. For contributors: the conformance jobs now fail on regressions instead of silently passing. `run-server.sh` requires `CONFORMANCE_VERSION` to be set when run locally (the error message points at the workflow pin).

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- The conformance jobs are not branch-protection-required checks, so an npm-registry blip turns them red but does not block merges. Same posture as typescript-sdk; can revisit once it has run cleanly for a while.
- Backporting `run-server.sh` to `v1.x` will need the `env: CONFORMANCE_VERSION` block backported in the same change (the script now reads it via `${CONFORMANCE_VERSION:?}`).
- Follow-up on the conformance repo: add a `python-sdk` (main) entry to `known-sdks.ts` alongside the existing `python-sdk-v1`.

### DIFF
--- a/.github/actions/conformance/expected-failures.yml
+++ b/.github/actions/conformance/expected-failures.yml
@@ -1,0 +1,52 @@
+# Conformance scenarios not yet passing against the Python SDK on main.
+# CI exits 0 if only these fail, exits 1 on unexpected failures or stale entries.
+#
+# Baseline established against @modelcontextprotocol/conformance pinned in
+# .github/workflows/conformance.yml (CONFORMANCE_VERSION = 0.2.0-alpha.3).
+# New conformance releases are adopted by deliberately bumping that pin and
+# reconciling this file in the same change.
+#
+# Entries are grouped by SEP. As each SEP lands in the SDK the corresponding
+# scenarios start passing and MUST be removed from this list (the runner fails
+# on stale entries), so the baseline burns down per milestone.
+
+client:
+  # --- Draft-spec scenarios (in `--suite draft`, also part of `--suite all`) ---
+  # SEP-2575 (request metadata / _meta envelope): client does not populate the
+  # _meta envelope or the MCP-Protocol-Version header semantics yet.
+  - request-metadata
+  # SEP-2322 (multi-round-trip requests): client does not echo requestState /
+  # handle IncompleteResult yet.
+  - sep-2322-client-request-state
+  # SEP-2243 (HTTP standardization): no fixture handler / client header support yet.
+  - http-custom-headers
+  - http-invalid-tool-headers
+  # SEP-2106 (JSON Schema $ref handling): client still dereferences network $refs.
+  - json-schema-ref-no-deref
+  # SEP-2468 (authorization response iss parameter): not implemented in the client.
+  - auth/iss-supported
+  - auth/iss-not-advertised
+  - auth/iss-supported-missing
+  - auth/iss-wrong-issuer
+  - auth/iss-unexpected
+  - auth/iss-normalized
+  - auth/metadata-issuer-mismatch
+  # SEP-2352 (authorization server migration): client does not re-register when
+  # PRM authorization_servers changes.
+  - auth/authorization-server-migration
+  # SEP-837 (application_type during DCR): the check only fires on draft-version
+  # runs; this draft scenario is the one place the client still hits it.
+  - auth/offline-access-not-supported
+
+  # --- Pre-existing scenarios that fail on checks added after conformance 0.1.15 ---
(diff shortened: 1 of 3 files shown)


.github/workflows/conformance.yml
```diff
@@ -46,4 +53,9 @@ jobs:
         with:
           node-version: 24
       - run: uv sync --frozen --all-extras --package mcp
-      - run: npx @modelcontextprotocol/conformance@0.1.13 client --command 'uv run --frozen python .github/actions/conformance/client.py' --suite all
+      - name: Run client conformance (all suite)
+        run: >-
+          npx --yes @modelcontextprotocol/conformance@"$CONFORMANCE_VERSION" client
+          --command 'uv run --frozen python .github/actions/conformance/client.py'
+          --suite all
```
felixweinberger, line 60: I'm actually not sure `all` is the correct one, counterintuitively - I believe it has 2 scenarios that aren't actually load bearing (I think something related to old client auth cases from a pre-2025-11-25 spec version)

.github/actions/conformance/expected-failures.yml
```diff
@@ -0,0 +1,52 @@
...
+  # runs; this draft scenario is the one place the client still hits it.
+  - auth/offline-access-not-supported
+
+  # --- Pre-existing scenarios that fail on checks added after conformance 0.1.15 ---
+  # SEP-2350 (scope step-up): WARNING-only; the expected-failures evaluator
+  # counts WARNINGs as failures.
+  - auth/scope-step-up
+  # SEP-990 (enterprise-managed authorization extension): no fixture handler /
+  # client support for the token-exchange + JWT bearer flow.
+  - auth/enterprise-managed-authorization
```
felixweinberger, line 47: Do we not have EMA in Python at all? Do we need to build this into Python potentially @pcarleton?

felixweinberger (approved):

.github/workflows/conformance.yml
felixweinberger: Land it for now though, we can fix if we end up with remaining failures that aren't covered by any SEPs

.github/actions/conformance/expected-failures.yml
maxisbey: yea it does not apparently

PR #2888, opened 2026-06-16
fix(auth): keep root PRM resource URL canonical

### LINKED ISSUE #2883: AnyHttpUrl normalization adds trailing slash → breaks RFC 9728 canonical resource URL in ProtectedResourceMetadata
## Repro

```python
str(AnyHttpUrl("https://example.com"))
# -> 'https://example.com/'
```

## Context

We use `mcp.shared.auth.ProtectedResourceMetadata.resource` for canonical resource URLs exposed at `/.well-known/oauth-protected-resource`. When `AnyHttpUrl` normalizes a URL whose original path is `/`, it serializes the value with a trailing slash. This breaks RFC 9728 §3.1 expectations and causes OAuth `protected-resource` validation to fail in strict clients — tokens are rejected and the client falls back into an infinite OAuth `authorize`/`token` loop.

## Symptom

Any MCP server exposing a resource with `streamable_http_path="/"` ends up producing a `resource` value with a trailing slash in the `protected-resource` metadata. Clients that validate strict against the canonical form (e.g. Claude Desktop) silently reject the token exchange and reinitiate the OAuth flow indefinitely.

Observed server logs:

```
POST /         → 401 (no token, expected)
GET  /.well-known/oauth-protected-resource → 200  (resource = "https://host/")
GET  /authorize → 302
POST /token    → 200  (token issued)
POST /         → never sent with Authorization: Bearer
GET  /authorize → 302  (new flow)
POST /token    → 200
... loops forever
```

## Suggested fixes (options)

1. Add a custom validator on `ProtectedResourceMetadata.resource` that strips the trailing slash when the original path is exactly `/`.
2. Change the field type from `AnyHttpUrl` to `str` with explicit canonical-form validation.
3. Use a stricter URL normalization option in pydantic if one exists for this case.

## Workaround in consumer repo

We patched it on our side by setting `streamable_http_path="/mcp"` so the URL has a non-trivial path, which `AnyHttpUrl` preserves without appending a slash. The consumer commit is [27c8e0c](https://github.com/viejunoacien/WirMensch-Phoenix/commit/27c8e0c0300fc867e944563abc00bf86f83fa1aa).

But that is a workaround for the symptom — the underlying contract violation in the `protected-resource` metadata is still present for any consumer that uses `streamable_http_path="/"`.

## Happy to help

Happy to open a follow-up PR if you prefer the custom-validator route — let me know which option you would prefer to merge.

Epochex:
## Summary
- serialize `ProtectedResourceMetadata.resource` without a synthetic trailing slash when the resource is the origin root
- keep the internal field typed as `AnyHttpUrl`, so this only changes the served PRM wire shape
- add regression coverage for the root metadata endpoint and direct JSON serialization

## Testing
- uv run pytest tests/server/auth/test_protected_resource.py
- uv run pytest tests/interaction/auth/test_discovery.py -k prm_endpoint_serves_the_resource_url

Fixes #2883.

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,6 +1,7 @@
 from typing import Any, Literal
+from urllib.parse import urlsplit, urlunsplit
 
-from pydantic import AnyHttpUrl, AnyUrl, BaseModel, ConfigDict, Field, field_validator
+from pydantic import AnyHttpUrl, AnyUrl, BaseModel, ConfigDict, Field, field_serializer, field_validator
 
 
 class OAuthToken(BaseModel):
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. This has since landed via #2925. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2908, opened 2026-06-18
fix: stdio_server emits LF only on Windows stdout (fixes #2433)

### LINKED ISSUE #2433: Windows: TextIOWrapper in stdio_server() emits CRLF instead of LF, corrupting newline-delimited JSON messages

kaXianc2-gom:
## Summary

Fix #2433: Windows `TextIOWrapper` in `stdio_server()` emits `\r\n` instead of `\n`, corrupting newline-delimited JSON messages.

## Change

Add `newline=""` to `TextIOWrapper(sys.stdout.buffer, ...)` so the MCP protocol line delimiter (`\n`) is preserved on Windows.

- **`src/mcp/server/stdio.py`**: +3 lines (1 line of code + 2 comment lines), −1 line
- **`tests/server/test_stdio.py`**: +18 lines (new test `test_mcpserver_run_stdio_emits_lf_not_crlf`)

`stdin` is intentionally left with the default `newline=None` so universal-newline behaviour normalises `\r\n` → `\n` on read.

## AI Assistance Disclosure

This contribution was developed with AI assistance (Claude / Anthropic Claude Code). The process was:

1. **Issue triage**: AI identified #2433 as a Windows-specific bug matching our environment
2. **Local sandbox verification**: Forked → cloned to local sandbox → created venv → ran full existing test suite (629 passed, 1 skipped)
3. **Fix applied**: One-line change to `stdio.py` (add `newline=""` parameter)
4. **Cross-validation performed**:
   - ✅ Existing tests: 5/5 passed (including 4 legacy tests)
   - ✅ New CRLF-specific test: passes on Windows
   - ✅ Full server test suite: 629/629 passed
   - ✅ Ruff format: no changes needed
   - ✅ Ruff lint: all checks passed
5. **Human review**: The change was reviewed and understood before submission

## Verification Environment

- **OS**: Windows 11 (the exact platform affected by #2433)
- **Python**: 3.11.15
- **Test runner**: pytest 8.4.2 with anyio
- **Dependencies**: uv sync --no-group docs

## Why this fix is correct

| Scenario | Before (buggy) | After (fixed) |
|----------|---------------|---------------|
| Write `"json\n"` on Windows stdout | Emits `"json\r\n"` ❌ | Emits `"json\n"` ✅ |
| Read `"json\r\n"` from Windows stdin | Normalises to `"json\n"` ✅ | Unchanged ✅ |
| Linux/macOS | No impact (newline=None is no-op) | No impact (newline="" is no-op) |

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -41,7 +41,9 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
     if not stdin:
         stdin = anyio.wrap_file(TextIOWrapper(sys.stdin.buffer, encoding="utf-8", errors="replace"))
     if not stdout:
-        stdout = anyio.wrap_file(TextIOWrapper(sys.stdout.buffer, encoding="utf-8"))
+        # newline="" prevents \n → \r\n translation on Windows, which would
+        # corrupt newline-delimited JSON messages (the MCP protocol uses \n).
+        stdout = anyio.wrap_file(TextIOWrapper(sys.stdout.buffer, encoding="utf-8", newline=""))
 
     read_stream_writer, read_stream = create_context_streams[SessionMessage | Exception](0)
     write_stream, write_stream_reader = create_context_streams[SessionMessage](0)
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2552.

Issue #2894, opened 2026-06-17
New Client (all-versions, low-level + high-level)


maxisbey:
The reworked Client that speaks every version: a connect sequence that probes `server/discover` and falls back to `initialize`, stamps the `_meta` envelope on modern requests, and exposes both a low-level interface (raw requests, handle MRTR yourself) and a high-level one (negotiation/MRTR/etc. handled). Relationship to `ClientSession` is part of this design.

## What's in it

- Low-level + high-level interface; relationship to `ClientSession` (replaces? wraps? grows 2026 under the hood?).
- Connect sequence: probe `server/discover` → fall back to legacy `initialize`; cache era per origin.
- Stamp the `_meta` envelope on every modern request.
- Emit `Mcp-Method` / `Mcp-Name` / `Mcp-Param-*` on modern HTTP (consumes the helpers from #2900).
- Own the client-side wiring of the MRTR loop (_MRTR (multi-round-trip requests)_), listen stream (_subscriptions/listen_), and caching parse (_Caching hints (ttlMs / cacheScope)_) — those issues supply the helpers.
- Never send session-id / open GET / DELETE against modern servers.

## Conformance

- `request-metadata`
- `http-standard-headers` (runs and passes, not skips)
- `http-custom-headers`, `http-invalid-tool-headers` (with #2900)
- `sep-2322-client-request-state` (with #2898)

## Dependencies

- Depends on: #2892

## References

- [Versioning](https://modelcontextprotocol.io/specification/draft/basic/versioning)
- [server/discover](https://modelcontextprotocol.io/specification/draft/server/discover)
- [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575)
- [client/request-metadata.ts](https://github.com/modelcontextprotocol/conformance/blob/main/src/scenarios/client/request-metadata.ts)

maxisbey:
Done via #2950 (`ClientSession` mechanics + `Client(mode=)` policy, default `mode='auto'`) and #2998 (auto-loop driver for `InputRequiredResult`).

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

Issue #2893, opened 2026-06-17
New HTTP serving path (modern + dual-era)


maxisbey:
Tracks: #2803

A single HTTP entry that branches: legacy traffic goes to the existing manager untouched; modern traffic goes to a new sessionless path. This is where the sessionless wire mode and the inbound header validation live.

## What's in it

- Single entry point that branches legacy → existing `StreamableHTTPSessionManager` (untouched, including `stateless_http=True`), modern → a new direct path.
- Sessionless wire mode for modern traffic: no session-id minting, GET/DELETE → 405, disconnect = cancel.
- The inbound validation ladder lives here (header↔body cross-check, error → HTTP status mapping).
- Where the `subscriptions/listen` long-lived stream plugs in.
- What `stateless_http=True` means going forward (current behaviour kept; era is a separate axis).

## Conformance

- `http-header-validation`
- `http-custom-header-server-validation` (with the helpers from #2900)
- Together with _Backwards compatibility & version negotiation_, _Capabilities API + server/discover handler_, _subscriptions/listen_, _Context object (layered, per-request)_: `server-stateless`
- Once the upstream `server-sessionless` scenario lands in the conformance suite and the conformance version pinned in CI is bumped: `server-sessionless`

## Dependencies

- Depends on: #2892

## References

- [Streamable HTTP](https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http)
- [SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567)
- [SEP-2243](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243)
- [server/stateless.ts](https://github.com/modelcontextprotocol/conformance/blob/main/src/scenarios/server/stateless.ts)

maxisbey:
Shipped in [v2.0.0b1](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.0.0b1): the branching entry (legacy manager untouched), sessionless wire mode, the full inbound validation ladder including server-side header/body cross-checks (#3033), and the listen stream plug-in (#3035). `http-header-validation`, `http-custom-header-server-validation`, and `server-stateless` all pass. The one remaining line — the `server-sessionless` scenario — is waiting on that scenario landing upstream in the conformance suite, not on SDK work.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

Issue #2892, opened 2026-06-17
Backwards compatibility & version negotiation


maxisbey:
Tracks: #2803

Make 2026-07-28 a real version the SDK can speak alongside every prior one: per-instance version knobs, the era classifier, and per-request `_meta` validation on the server. This is where the stateless server-side obligations land.

## What's in it

- Per-instance "what versions do I serve/speak" knobs (server + client); add 2026-07-28 to the negotiable set.
- Era classifier + dispatch: server reads the `_meta` envelope (version/clientInfo/capabilities), validates it, emits the unsupported-version / missing-capability errors with typed data, and skips the legacy init gate for modern traffic.
- Era-gate everything removed from core for modern peers: `ping`, `logging/setLevel`, `roots/list_changed`, the tasks surface, the URL-elicitation-required error.
- `resultType` stamped on every modern result (the strip-for-legacy half is already in #2849).

## Conformance

- (no scenario passes on this alone — it's the foundation the others stand on)
- Together with _New HTTP serving path (modern + dual-era)_, _Capabilities API + server/discover handler_, _subscriptions/listen_, _Context object (layered, per-request)_: `server-stateless`
- Together with #2894: `request-metadata`

## Dependencies

- Depends on: none

## References

- [Versioning](https://modelcontextprotocol.io/specification/draft/basic/versioning)
- [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575)
- [SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567)
- [sep-2575.yaml](https://github.com/modelcontextprotocol/conformance/blob/main/src/seps/sep-2575.yaml)

maxisbey:
Shipped in [v2.0.0b1](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.0.0b1): per-instance version configuration, the era classifier with typed unsupported-version errors, era-gated core methods, and resultType stamping on modern results. The `server-stateless` and `request-metadata` conformance scenarios pass.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

Issue #2891, opened 2026-06-17
2026-07-28 spec support — tracking


maxisbey:
Umbrella for the 2026-07-28 release work. Bar: the conformance suite passes with no expected-failures waivers attributable to this release (the [SEP-990](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/990) enterprise-auth waiver is unrelated and remains). Each sub-issue lists the scenarios it's responsible for; design is owned by whoever takes the issue.

This sits alongside the broader v2.0 SDK rework tracked in #1253 — that umbrella covers SDK-ergonomics breaking changes; this one covers spec compliance for the 2026-07-28 protocol revision.

## What's in it

- Sub-issues:
  - #2892
  - #2893
  - #2894
  - #2895
  - #2896
  - #2897
  - #2898
  - #2899
  - #2900
  - #2901
  - #2902
  - #2903
  - #2904
  - #2905
- **Architecture (direction, not law):** lowlevel `Server` = protocol interface, minimal state, no magic. The 2026 features ship as **public helpers/classes** usable directly against lowlevel `Server`. `MCPServer` is a **thin composition** of those same public helpers — a pre-built convenience, not a monolith with private machinery.
- **Capability checkpoints** (shared targets, not tickets):
  - **M1** — a modern client makes a stateless `tools/call` against a modern server over HTTP, no initialize (minimum slice of _Backwards compatibility & version negotiation_ + _New HTTP serving path (modern + dual-era)_ + _New Client (all-versions, low-level + high-level)_)
  - **M2** — the `server-stateless` conformance scenario passes end-to-end (_Backwards compatibility & version negotiation_ + _New HTTP serving path (modern + dual-era)_ + _Context object (layered, per-request)_ + _Capabilities API + server/discover handler_ + _subscriptions/listen_ all in)
  - **M3** — all release-attributable waivers gone
- **Landing model:** PRs land to `main`. 2026-07-28 stays modeled-but-not-negotiable until the version-negotiation work flips it on, so partial work doesn't change behaviour for existing users.
- Standing maintenance (not a sub-issue): when bumping the conformance pin, reconcile `expected-failures.yml` in the same PR; when in-flight spec PRs merge, re-vendor `schema/2026-07-28.json` + regenerate.
- Stretch (optional): contribute missing scenarios upstream — discover caching-hints check, dedicated `subscriptions/listen` server scenario, client-direction sessionless scenario.

## Conformance

[...]

PR #2910, opened 2026-06-19
First end-to-end 2026-07-28 stateless tools/call (experimental entry + ClientSession pin)

maxisbey:
First end-to-end slice of the 2026-07-28 stateless protocol, plus the interaction-suite era axis turned on so the existing transport-agnostic tests run at the new revision.

Part of #2891, #2892, #2893, #2894 (closes none).

## Server

- `mcp.server._streamable_http_modern` — `handle_modern_request` direct-invocation path (private)
- `StreamableHTTPSessionManager.handle_request` branches on `MCP-Protocol-Version` ∈ `MODERN_PROTOCOL_VERSIONS`
- per-request `ServerRunner` over a `SingleExchangeDispatcher` (no back-channel; notifications no-op pending the SSE response mode)

## Client

- `Client`, `ClientSession`, and `streamable_http_client()` gain `protocol_version: str | None`
- pinned to `"2026-07-28"`: born initialized (synthesized `InitializeResult`, placeholder `server_info` pending `server/discover`); `initialize()` is idempotent; stamps the `io.modelcontextprotocol/{protocolVersion,clientInfo,clientCapabilities}` envelope on every request; `cancel_on_abandon=False`
- pinned to a stateful version: `initialize()` requests that version instead of `LATEST_PROTOCOL_VERSION`
- transport pins `MCP-Protocol-Version` from the constructor; `Mcp-Method`/`Mcp-Name` derived per message; `_encode_header_value` Base64-sentinel-wraps RFC-7230-unsafe names
- `MODERN_PROTOCOL_VERSIONS` constant in `mcp.shared.version`

## Types

- `CacheableResult` defaults to `ttl_ms=0, cache_scope="private"` so list/read results constructed without explicit hints validate against the 2026-07-28 surface and never accidentally enable shared caching

## Interaction suite

- `SPEC_VERSIONS = ("2025-11-25", "2026-07-28")` — node ids gain a `-version` suffix
- the `connect` fixture forwards `protocol_version` through `Client`; `in-memory` and `streamable-http-stateless` era-locked to 2025-11-25
- `[streamable-http-2026-07-28]`: **60 pass, 8 xfail** (xfails are progress/logging — handler-emitted notifications not yet streamed on the single-exchange path)
- 13 `lifecycle:*` requirements get `removed_in="2026-07-28"`; 6 error-shape requirements excluded (`modern-error-surface` — tests pin the legacy `code=0` divergence; the modern arm returns the spec-correct `-32603`)
- the wire contract (envelope, headers, no handshake, no session-id) stays pinned by `transports/test_hosting_http_modern.py`

## Conformance

- 23 scenarios now pass at the 2026-07-28 wire across the draft and carried-forward legs
- `tools-call-with-progress` stays expected-fail (progress delivery on the single-exchange path)

## Behaviour changes

- `ClientSession.initialize()` is now idempotent: a second call returns the cached result (or the synthesized one for a stateless pin) instead of re-sending the handshake. All other 2026 behaviour is opt-in via `protocol_version=`. `SUPPORTED_PROTOCOL_VERSIONS` and `LATEST_PROTOCOL_VERSION` are unchanged.

## Flagged for review

- 2× `# type: ignore[call-arg]` in `server/session.py` for `_related_request_id` (TODO at :94 — fix is the `ServerSession`/`Context` rework)
- 2× `# type: ignore[reportPrivateUsage]` in the modern entry (`_compose_on_request`, `_EXIT_STACK_CLOSE_TIMEOUT`) — both fall out of the `ServerRunner` driver-split

## Out of scope

`server/discover`, `_meta` validation ladder, version negotiation, `subscriptions/listen`, MRTR, SSE response mode for the modern entry, public design of the new HTTP path, in-memory modern entry.


src/mcp/server/session.py
```diff
@@ -92,8 +91,16 @@ async def send_request(
             # Fail fast instead of parking forever on a response that cannot
             # arrive; matches `Connection.send_raw_request`.
             raise NoBackChannelError(data["method"])
-        result = await self._dispatcher.send_raw_request(
-            data["method"], data.get("params"), opts or None, _related_request_id=related
+        # TODO: _related_request_id is not on the Dispatcher Protocol; either
+        # add it there or refactor ServerSession once the legacy path is compat-only.
+        result = cast(
```
maxisbey, line 99: seems like we should use the dispatcher context object here then? that's the whole point of it, is it know it's own secret params

src/mcp/server/session.py
```diff
@@ -110,7 +117,7 @@ async def send_notification(
     ) -> None:
         """Send a typed server-to-client notification."""
         data = notification.model_dump(by_alias=True, mode="json", exclude_none=True)
-        await self._dispatcher.notify(data["method"], data.get("params"), _related_request_id=related_request_id)
+        await self._dispatcher.notify(data["method"], data.get("params"), _related_request_id=related_request_id)  # type: ignore[call-arg]
```
maxisbey, line 123: this is bad

src/mcp/client/session.py
```diff
@@ -273,17 +297,17 @@ async def initialize(self) -> types.InitializeResult:
             if self._list_roots_callback is not _default_list_roots_callback
             else None
         )
+        return types.ClientCapabilities(sampling=sampling, elicitation=elicitation, experimental=None, roots=roots)
 
+    async def initialize(self) -> types.InitializeResult:
+        if self._pinned_version is not None:
```
maxisbey: why?

tests/interaction/lowlevel/test_lifecycle_stateless.py
```diff
@@ -0,0 +1,238 @@
...
+)
+from tests.interaction._helpers import RecordingTransport
+from tests.interaction._modern_vocab import MODERN_BODY_TOKENS
+from tests.interaction._requirements import requirement
+
+pytestmark = pytest.mark.anyio
+
+
+@requirement("lifecycle:stateless:request-envelope")
+async def test_pinned_session_stamps_the_envelope_meta_on_every_request_and_never_initializes() -> None:
```
maxisbey: this doesn't seem very high level? why is it being ran like this?

src/mcp/server/session.py
maxisbey: Agreed it's the wrong shape. Reworded the TODO to point at the actual fix (a per-request `Outbound` — the `DispatchContext` — alongside the connection-level one, with `related_request_id` as the channel selector). That belongs with the `ServerSession`/`Context` rework; widening the Protocol is the wrong direction since `related_request_id` is transport-specific.

src/mcp/server/session.py
maxisbey: Same as above — see the reworded TODO at :94.

src/mcp/client/session.py
maxisbey: Pinning to a stateful version is the contradiction, not calling `initialize()` afterward. Tightened the param to `Literal["2026-07-28"]` (`StatelessProtocolVersion`) so a 2025-era pin is a type error; the raise stays as the fail-fast for misuse (a no-op would have to synthesize an `InitializeResult` we don't have).

tests/interaction/lowlevel/test_lifecycle_stateless.py
maxisbey: Dropped this file (and the MockTransport one) in 194f225. The capstone in `transports/test_hosting_http_modern.py` proves everything they did at higher altitude — envelope, headers, no-initialize, and now caller-`_meta`-preserved via a server-side `ctx.meta` echo. The `initialize()`-raises and session-id-ignore checks moved to `tests/client/`.

PR #2915, opened 2026-06-20
fix: forward stdio server stderr through errlog

### LINKED ISSUE #156: support logging to stderr in Jupyter Notebook Environments.

jordansilly77-stack:
## Summary
- pipe stdio server stderr on POSIX instead of handing the current sys.stderr object directly to the subprocess
- forward stderr chunks through the existing errlog stream
- close the stderr pipe during shutdown and add a regression test for errlog forwarding

Fixes #156.

## Testing
- python3 -m py_compile src/mcp/client/stdio.py tests/client/test_stdio.py
- uv run ruff check src/mcp/client/stdio.py tests/client/test_stdio.py
- uv run ruff format src/mcp/client/stdio.py tests/client/test_stdio.py --check

I could not complete the local pytest run on this Mac because the repository virtualenv hits local plugin/import issues, including macOS blocking black's compiled extension and slow logfire/pydantic plugin imports. The focused code and formatting checks above pass; CI should provide the full pytest signal.

## AI disclosure
AI assistance was used to research the issue context, draft the patch, and run local checks. The final changes were reviewed before submission.

### DIFF
--- a/src/mcp/client/stdio.py
+++ b/src/mcp/client/stdio.py
@@ -10,6 +10,7 @@ process nor hang on one.
 
 import logging
 import os
+import subprocess
 import sys
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager, suppress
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. Tracking this in #3268 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

Issue #2902, opened 2026-06-17
Auth — 2026 client-side requirements


maxisbey:
Tracks: #2801 #2800 #2783 #2798 #2795

Spec-compliant client-side OAuth changes for the release. Independent of the protocol-version work; design choices are the implementer's.

## What's in it

- Validate `iss` on the OAuth callback (RFC 9207 / SEP-2468).
- Bind stored credentials to the issuer; re-register on AS change (SEP-2352).
- Send `application_type` on DCR (SEP-837).
- Union scopes on step-up re-authorization (SEP-2350).
- Confirm and close out offline_access (SEP-2207 — already implemented).
- Server-side SHOULDs (PRM offline_access, single-challenge scopes): docs note / out of scope for this release.

## Conformance

- `auth/iss-*` (×6)
- `auth/metadata-issuer-mismatch`
- `auth/authorization-server-migration`
- `auth/offline-access-not-supported`
- `auth/scope-step-up`

## Dependencies

- Depends on: none

## References

- [Authorization](https://modelcontextprotocol.io/specification/draft/basic/authorization)
- [SEP-2468](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2468)
- [SEP-2352](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2352)
- [SEP-837](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/837)
- [SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350)
- [SEP-2207](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2207)
- [python-sdk #2801](https://github.com/modelcontextprotocol/python-sdk/issues/2801)
- [python-sdk #2800](https://github.com/modelcontextprotocol/python-sdk/issues/2800)

maxisbey:
Done via #2921 / #2930 / #2931 / #2933 (the auth bundle) and #2988 (SEP-990 identity assertion). Client auth conformance is green.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

tarunag10:
Opened a targeted regression-fix PR: https://github.com/modelcontextprotocol/python-sdk/pull/3012

It tightens WWW-Authenticate auth-param parsing so `scope` / `resource_metadata` only match complete parameter names, with tests for prefixed names like `error_scope`.

PR #2921, opened 2026-06-20
Validate the `iss` authorization-response parameter (RFC 9207 / SEP-2468)

Kludex:
Implements the first item of #2902: validate the RFC 9207 `iss` authorization-response parameter (SEP-2468).

## What changed

- The OAuth client validates the `iss` query parameter returned on the authorization-response redirect, and the `issuer` of discovered authorization server metadata. Comparison is simple string comparison per RFC 3986 §6.2.1 (no normalization): a mismatched `iss` is rejected, and a missing `iss` is rejected when the AS advertised `authorization_response_iss_parameter_supported`.
- `callback_handler` now returns `AuthorizationCodeResult` (`code`/`state`/`iss`) instead of `tuple[str, str | None]`, so the redirect's `iss` reaches the validation. **Breaking change** — documented in `docs/migration.md`.
- Added `authorization_response_iss_parameter_supported` to `OAuthMetadata`.

`AnyHttpUrl` appends a trailing slash to a bare authority (`https://as` → `https://as/`), which would defeat the byte-exact comparison the spec requires (`iss-supported` must accept `https://host` while `iss-normalized` must reject `https://host/`), so the recorded issuer is compared with that lone authority slash stripped.

## Conformance

Flips `auth/metadata-issuer-mismatch` green (removed from both expected-failures baselines). The `auth/iss-*` scenarios validate `iss` correctly but stay baselined because they reach Dynamic Client Registration and trip the unimplemented SEP-837 `application_type` check — re-grouped under SEP-837 with accurate comments. Both conformance legs pass with no stale entries.

## Out of scope (separate PRs, tracked by #2902)

SEP-837 (`application_type`), SEP-2350 (scope-union step-up), SEP-2352 (AS migration / credential binding). SEP-2207 (offline_access) is already done.

## AI Disclaimer


src/mcp/client/auth/utils.py
```diff
@@ -211,6 +211,63 @@ async def handle_auth_metadata_response(response: Response) -> tuple[bool, OAuth
...
+        return url[:-1]
+    return url
+
+
+def raw_issuer(oauth_metadata: OAuthMetadata) -> str:
+    """Return the issuer as the authorization server transmitted it, before URL normalization."""
+    return _strip_authority_trailing_slash(str(oauth_metadata.issuer))
+
+
+def validate_authorization_response_iss(iss: str | None, oauth_metadata: OAuthMetadata | None) -> None:
```
Kludex, line 214: Core RFC 9207 check. Two reject cases: `iss` present but != issuer (simple string compare), or `iss` absent when the AS advertised `authorization_response_iss_parameter_supported`. Everything else accepts.

Kludex (commented):
Pointers to the substantive logic — the rest of the diff is mostly the mechanical `tuple[str, str | None]` → `AuthorizationCodeResult` rename across tests/examples.

src/mcp/client/auth/utils.py
```diff
@@ -211,6 +211,63 @@ async def handle_auth_metadata_response(response: Response) -> tuple[bool, OAuth
     return True, None
 
 
+def _strip_authority_trailing_slash(url: str) -> str:
```
Kludex: The non-obvious bit. `AnyHttpUrl` rewrites a bare authority `https://as` → `https://as/`, which would break the byte-exact comparison the spec mandates (`iss-supported` must accept `https://host`, `iss-normalized` must reject `https://host/`). We strip *only* that lone authority slash; a real path keeps its slash.

src/mcp/client/auth/utils.py
```diff
@@ -211,6 +211,63 @@ async def handle_auth_metadata_response(response: Response) -> tuple[bool, OAuth
...
+    if iss is not None:
+        if iss != expected:
+            raise OAuthFlowError(f"Authorization response iss mismatch: {iss} != {expected}")
+        return
+
+    if oauth_metadata is not None and oauth_metadata.authorization_response_iss_parameter_supported:
+        raise OAuthFlowError("Authorization response missing iss parameter advertised by the authorization server")
+
+
+def validate_metadata_issuer(oauth_metadata: OAuthMetadata, expected_issuer: str) -> None:
```
Kludex, line 238: RFC 8414 §3.3: the metadata `issuer` must match the issuer the well-known URL was built from. Compared the same slash-stripped way so a bare-authority issuer isn't a false mismatch.

src/mcp/client/auth/oauth2.py
```diff
@@ -356,16 +359,19 @@ async def _perform_authorization_code_grant(self) -> tuple[str, str]:
...
+        result = await self.context.callback_handler()
 
-        if returned_state is None or not secrets.compare_digest(returned_state, state):
-            raise OAuthFlowError(f"State parameter mismatch: {returned_state} != {state}")
+        if result.state is None or not secrets.compare_digest(result.state, state):
+            raise OAuthFlowError(f"State parameter mismatch: {result.state} != {state}")
 
-        if not auth_code:
+        # RFC 9207: validate the authorization-response issuer
+        validate_authorization_response_iss(result.iss, self.context.oauth_metadata)
```
Kludex, line 368: Call site 1: validate the redirect `iss` after the state check, before the code is used for token exchange.

src/mcp/client/auth/oauth2.py
```diff
@@ -570,6 +576,9 @@ async def async_auth_flow(self, request: httpx.Request) -> AsyncGenerator[httpx.
                         if not ok:
                             break
                         if ok and asm:
+                            # SEP-2468: metadata issuer must match the discovery issuer
+                            if self.context.auth_server_url is not None:
+                                validate_metadata_issuer(asm, self.context.auth_server_url)
```
Kludex, line 581: Call site 2: validate the AS metadata issuer against `auth_server_url` (the PRM-advertised AS the well-known URL was derived from). Guarded on `auth_server_url is not None` so the legacy no-PRM path is unaffected.

src/mcp/client/auth/oauth2.py
```diff
@@ -97,7 +100,7 @@ class OAuthContext:
     client_metadata: OAuthClientMetadata
```
Kludex, line 100: The breaking change: `callback_handler` returns `AuthorizationCodeResult` so the redirect's `iss` reaches the validation. Documented in `docs/migration.md`.

tests/interaction/auth/_provider.py
```diff
@@ -53,9 +54,13 @@ def __init__(
...
         fail_next_refresh: bool = False,
         reject_all_tokens: bool = False,
+        issuer: str | None = None,
     ) -> None:
         self._default_scopes = list(default_scopes) if default_scopes is not None else ["mcp"]
-        self._issuer = "http://127.0.0.1:8000"
+        # The authorization-response iss must match the AS issuer the client recorded (RFC 9207
+        # simple string comparison); the client strips the trailing slash AnyHttpUrl adds to a
+        # path-less issuer, so the redirect carries the bare origin. Path-issuer tests pass it.
+        self._issuer = issuer if issuer is not None else BASE_URL
```
Kludex: Test-harness consistency: a compliant AS's redirect `iss` must equal its metadata `issuer` byte-for-byte. We emit the bare origin (slash-less) to match what the client records, mirroring the real conformance harness.

tests/interaction/auth/test_discovery.py
```diff
@@ -311,23 +315,26 @@ async def test_as_metadata_advertises_authorize_token_registration_and_s256() ->
 
 
 @requirement("client-auth:as-metadata-discovery:issuer-validation")
-async def test_as_metadata_with_a_mismatched_issuer_is_accepted_and_the_flow_proceeds() -> None:
-    """Authorization-server metadata whose `issuer` does not match the discovery URL is accepted.
+async def test_as_metadata_with_a_mismatched_issuer_aborts_the_flow() -> None:
```
Kludex, line 317: Inverted from the old test (which asserted a mismatched issuer was *accepted* — a documented divergence). It now asserts the flow aborts with `OAuthFlowError` before any authorize/token request, and the corresponding `divergence` was removed from `_requirements.py`.

.github/actions/conformance/client.py
```diff
@@ -140,6 +141,8 @@ async def handle_redirect(self, authorization_url: str) -> None:
                         self._auth_code = query_params["code"][0]
                         state_values = query_params.get("state")
                         self._state = state_values[0] if state_values else None
+                        iss_values = query_params.get("iss")
+                        self._iss = iss_values[0] if iss_values else None
```
Kludex, line 145: Instead of doing this, can't we use a typeddict with a type adapter, or a basemodel? str | list[str]?

examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
```diff
@@ -160,6 +161,10 @@ def get_state(self):
         """Get the received state parameter."""
         return self.callback_data["state"]
 
+    def get_iss(self):
+        """Get the received iss parameter."""
+        return self.callback_data["iss"]
```
Kludex: can't it be a @property?

docs/migration.md
```diff
@@ -51,6 +51,35 @@ async with http_client:
...
+    params = parse_qs(urlparse(await wait_for_redirect()).query)
+    return AuthorizationCodeResult(
+        code=params["code"][0],
+        state=params.get("state", [None])[0],
+        iss=params.get("iss", [None])[0],
+    )
+```
+
+Forward the `iss` query parameter from the redirect so the validation can run; omitting it disables the issuer check for servers that send `iss`.
+
```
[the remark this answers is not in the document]
Kludex: Good catch - reworded to spell out that omitting `iss` hard-fails (raises `OAuthFlowError`) against servers that advertise `authorization_response_iss_parameter_supported`, and only silently skips the check otherwise.

examples/clients/simple-auth-client/mcp_simple_auth_client/main.py
Kludex: Done - converted both `get_state` and `get_iss` to `@property` (`state`/`iss`) and updated the call site.

.github/actions/conformance/client.py
Kludex: Left as-is: this is the conformance test-driver, and the per-field `vals[0] if vals else None` extraction mirrors the existing `state` parsing right above it. Happy to introduce a helper/model if you prefer, but it felt like churn for driver code.

Kludex:
Prerequisite #2925 opened: it sets `url_preserve_empty_path=True` on the OAuth metadata models so issuers parsed from the wire keep their empty path. Once #2925 merges, I'll rebase this PR and drop the `_strip_authority_trailing_slash` workaround in favor of a direct comparison.

src/mcp/client/auth/utils.py
```diff
@@ -211,6 +211,63 @@ async def handle_auth_metadata_response(response: Response) -> tuple[bool, OAuth
...
+    """
+    if url.endswith("/") and urlparse(url).path == "/":
+        return url[:-1]
+    return url
+
+
+def raw_issuer(oauth_metadata: OAuthMetadata) -> str:
+    """Return the issuer as the authorization server transmitted it, before URL normalization."""
+    return _strip_authority_trailing_slash(str(oauth_metadata.issuer))
+
```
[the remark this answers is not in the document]
Kludex: Fixed. Merged #2925 (sets `url_preserve_empty_path=True` on the OAuth metadata models), so the issuer parsed from the wire now keeps its empty path - `str(oauth_metadata.issuer)` is the byte-exact value the AS transmitted. Removed `_strip_authority_trailing_slash`/`raw_issuer` and compare directly, which is the parse-time capture you suggested. A genuinely slash-terminated issuer (e.g. Auth0 `https://tenant.auth0.com/`) now matches its own `iss` instead of being falsely rejected - added a `trailing-slash-issuer` accept case to lock that in. Conformance `iss-supported`/`iss-normalized` still pass.

Kludex:
## Conformance scenarios this PR addresses (SEP-2468)

Run against `@modelcontextprotocol/conformance@0.2.0-alpha.4`. Every SEP-2468 check now passes:

| Scenario | SEP-2468 check | Result | Scenario overall |
|---|---|---|---|
| `auth/metadata-issuer-mismatch` | `sep-2468-client-validate-metadata-issuer` | ✅ SUCCESS | ✅ **PASSES** (removed from baseline) |
| `auth/iss-supported` | `sep-2468-client-compare-iss-supported` | ✅ SUCCESS | ⛔ blocked by SEP-837 |
| `auth/iss-not-advertised` | `sep-2468-client-proceed-no-iss` | ✅ SUCCESS | ⛔ blocked by SEP-837 |
| `auth/iss-supported-missing` | `sep-2468-client-reject-missing-iss` | ✅ SUCCESS | ⛔ blocked by SEP-837 |
| `auth/iss-wrong-issuer` | `sep-2468-client-compare-iss-supported` | ✅ SUCCESS | ⛔ blocked by SEP-837 |
| `auth/iss-unexpected` | `sep-2468-client-compare-iss-unadvertised` | ✅ SUCCESS | ⛔ blocked by SEP-837 |
| `auth/iss-normalized` | `sep-2468-client-no-normalization` | ✅ SUCCESS | ⛔ blocked by SEP-837 |

**`auth/metadata-issuer-mismatch` is the only one that flips fully green** here (it rejects before DCR), so it's removed from both `expected-failures.yml` and `expected-failures.2026-07-28.yml`.

The six `auth/iss-*` scenarios all run at spec version `2026-07-28` and reach Dynamic Client Registration, where the harness's `sep-837-application-type-present` check fires. Since the client doesn't yet send `application_type` (SEP-837, a separate item of #2902), each scenario fails overall on *that* check despite its SEP-2468 check passing. They stay baselined under SEP-837 with a comment noting iss validation is done; they'll flip green when SEP-837 lands.

Both conformance legs pass with no unexpected or stale baseline entries.

maxisbey (approved):

PR #2849, opened 2026-06-12
Protocol types for 2026-07-28: superset monolith, committed per-version packages, and wire-method maps

maxisbey:
### Overview

Three type sets:
- `mcp.types.v2025_11_25` — generated from the 2025-11-25 schema; serves every version through 2025-11-25
- `mcp.types.v2026_07_28` — generated from the 2026-07-28 (draft) schema
- `mcp.types` (the monolith) — hand-maintained superset of both; this is the public API

`mcp.types.methods` holds `(method, version)` → surface-type maps and the parse/validate/serialize helpers the runner and sessions call.

### Generation

`scripts/gen_surface_types.py` runs `datamodel-codegen` over vendored `schema/*.json`, with a few mappings overridden (`format: byte/uri/uri-template` → `str`; `number` → `int | float`) and `extra="ignore"` everywhere except a per-version allow-list of spec-declared open types (`_meta`, `Result`, `GetTaskPayloadResult`, tool input/output schemas, etc.). CI regenerates with `--check` and fails on drift.

### Wiring

`ServerRunner` and `ClientSession` consult the maps:

- **Inbound requests/notifications** are validated against the negotiated version's surface row. A spec method that doesn't exist at that version is `METHOD_NOT_FOUND` even if a handler is registered; a malformed payload is `INVALID_PARAMS` (or dropped+logged for notifications). Non-spec methods registered on the server skip the gate.
- **Inbound results** (`send_request` return path, both directions) are surface-validated before the monolith parse; mismatches raise at the call site.
- **Outbound server results** are dumped from the monolith, then re-serialized through the surface row. Fields not in that version's schema are dropped, so 2026-only keys (`resultType`, `ttlMs`, `cacheScope`) never reach a pre-2026 peer.

`SUPPORTED_PROTOCOL_VERSIONS` is unchanged — 2026-07-28 is modeled, not yet negotiable.

### Breaking changes

- Server handlers must return spec-valid results; nonconforming output (e.g. `Tool(inputSchema={})` — missing `"type": "object"`) now produces `INTERNAL_ERROR`.
- Clients now surface-validate inbound traffic; spec-invalid server output may raise where it was previously tolerated.
- Elicitation schemas are now validated against `PrimitiveSchemaDefinition`; `Optional[T]` is unwrapped, bare `list[str]` (no enum) is rejected.

Documented in `docs/migration.md`.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
[...]

PR #2924, opened 2026-06-20
Pass json-schema-ref-no-deref conformance scenario (SEP-2106)

### LINKED ISSUE #2903: SEP-2106 — JSON Schema 2020-12 + structuredContent
Tracks: #2792

Tool schemas conform to JSON Schema 2020-12 and `structuredContent` accepts any JSON value (the type widening landed in #2849).

## What's in it

- Client never dereferences network `$ref` anywhere it processes tool schemas; server rejects unresolvable external `$ref` at registration.
- Audit downstream code that assumed `structuredContent` is an object / truthy.

## Conformance

- `json-schema-ref-no-deref`

## Dependencies

- Depends on: none

## References

- [Tools — schemas](https://modelcontextprotocol.io/specification/draft/server/tools)
- [SEP-2106](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106)

Kludex:
Closes #2903.

SEP-2106 permits the full JSON Schema 2020-12 vocabulary in tool schemas, including `$ref`. A `$ref` resolving to a network URI is an SSRF / fetch-DoS vector. The SEP requires:

- Implementations **MUST NOT** automatically dereference a `$ref` that is not a same-document reference.
- A schema that fails to validate **due to an unresolved external `$ref`** **SHOULD** be rejected rather than treated as permissive.

The SDK already satisfies the MUST NOT: there is no `$ref`-fetching code anywhere, and `jsonschema`'s `referencing` backend raises `Unresolvable` rather than performing an HTTP GET. So no `$ref`-resolution change is needed.

## What changed

- Added a `json-schema-ref-no-deref` driver handler to the conformance client. Its absence (not actual dereferencing) was why the scenario failed; the SDK was already compliant.
- Removed the scenario from both `expected-failures` baselines.
- Added a client test confirming a network `$ref` in an output schema, under a property the structured content does not set, is validated without being dereferenced.

## Scope notes

- I deliberately did **not** add eager registration-time rejection of external `$ref`s. The SEP only forbids *dereferencing* them; an external `$ref` that is never fetched is legitimate 2020-12 (the SEP lists `$ref` among the keywords it enables). Eager rejection would forbid valid schemas and go beyond the spec. The issue text says "server rejects unresolvable external `$ref` at registration"; this PR follows the SEP's narrower, conditional SHOULD instead. Happy to revisit if maintainers prefer the stricter reading.
- The `structuredContent` widening to any JSON value already landed in #2849; I audited the consumers and found no object/truthiness assumptions.
- Composition-keyword resource bounds (SEP SHOULD: schema depth / validation-time limits) are not implemented here.

## Verification

`json-schema-ref-no-deref` passes on both the 2025 and 2026-07-28 client legs locally, with "Baseline check passed: all failures are expected" (no stale-entry error). Full suite green, 100% coverage, `strict-no-cover`/pyright/ruff clean.

## AI Disclaimer


Kludex:
Why the conformance wiring is the only change needed:

SEP-2106's security requirement is a **MUST NOT** *dereference* network `$ref`s, not a MUST reject them. The SDK already satisfies it without any code change:

- There is no `$ref`-fetching machinery anywhere in `src/` - no `RefResolver`, no custom `referencing.Registry`, no retrieval hook.
- The only place a tool schema is validated against data is the client's `jsonschema.validate(structured_content, output_schema)` call ([`session.py`](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/session.py)). It uses jsonschema 4.18+'s default `referencing` backend, which - unlike the old `RefResolver` - does **not** auto-fetch: an unresolved network `$ref` raises `Unresolvable` instead of issuing an HTTP GET.
- `ClientSession.list_tools()` only deserializes the tool list; it never walks `inputSchema` or resolves `$ref`s. So the client is structurally incapable of dereferencing during listing - exactly what the conformance scenario checks.

So the scenario was already passing behaviourally; it failed in CI only because our conformance driver (`client.py`) had no handler registered for `json-schema-ref-no-deref`, which made the driver exit 1 ("Client never requested tools/list"). Adding the handler + burning the scenario down in both `expected-failures` baselines is all that's required.

I deliberately did **not** add eager registration-time rejection of external `$ref`s. The SEP only forbids *dereferencing* them; an external `$ref` that is never fetched is valid JSON Schema 2020-12 (the SEP lists `$ref` among the keywords it enables), so rejecting at registration would forbid legitimate schemas and go beyond the spec. The narrower SHOULD ("reject a schema that fails to validate due to an unresolved external `$ref`") is already met: validation raises rather than treating the ref as permissive.

The `structuredContent` widening to any JSON value landed separately in #2849; I audited the consumers for object/truthiness assumptions and found none.

maxisbey (approved):

PR #2930, opened 2026-06-20
Send `application_type` during Dynamic Client Registration (SEP-837)

Kludex:
Implements SEP-837 (the third client-side item of #2902): send `application_type` during Dynamic Client Registration.

## What changed

- Added `application_type: Literal["web", "native"] = "native"` to `OAuthClientMetadata`. It serializes into the DCR `/register` body via the existing `model_dump`.
- Default is `"native"`, matching the Rust SDK and the common MCP case (CLI/desktop clients with loopback redirect URIs - all our examples). Remote browser-based clients override to `"web"`.

Per the spec: omitting `application_type` defaults to `"web"` under OIDC, which an authorization server can reject for the `localhost` redirect URIs native clients use; non-OIDC servers ignore the parameter.

## Conformance

Unblocks and removes from the expected-failures baselines:

- **Default leg** (`expected-failures.yml`): the 6 `auth/iss-*` and `auth/offline-access-not-supported` (their iss/scope checks already passed under #2921; they only failed `sep-837-application-type-present` at DCR).
- **2026-07-28 leg** (`expected-failures.2026-07-28.yml`): the above plus `auth/metadata-default`, `auth/metadata-var{1,2,3}`, `auth/scope-from-*`, `auth/scope-omitted-when-undefined`, and `auth/token-endpoint-auth-{basic,post,none}`.

The 2026-07-28 leg passes cleanly with no stale or unexpected entries.

> **Note - pre-existing default-leg failure, not from this PR:** the default leg still reports `auth/scope-step-up` as an unexpected failure. This fails identically on clean `main` (verified): #2927 removed it from `expected-failures.yml` for the alpha.5 harness (which gates the SEP-2350 scope-union check to 2026-07-28+), but CI is still pinned to alpha.4, where the check fires on the default leg. It resolves when the harness pin moves to alpha.5 (already a TODO in `conformance.yml`) or when SEP-2350 lands. Out of scope here.

## AI Disclaimer

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -67,6 +67,9 @@ class OAuthClientMetadata(BaseModel):
     # servers may also return additional types they support
     response_types: list[str] = ["code"]
     scope: str | None = None
+    # SEP-837: OIDC application_type. Defaults to "native" since MCP clients typically use
+    # loopback redirect URIs; set "web" for remote browser-based clients on a non-local host.
+    application_type: Literal["web", "native"] = "native"
 
     # these fields are currently unused, but we support & store them for potential
     # future use
--- a/.github/actions/conformance/expected-failures.2026-07-28.yml
+++ b/.github/actions/conformance/expected-failures.2026-07-28.yml
@@ -28,32 +28,6 @@ client:
   # branch takes the per-request _meta path.
   - tools_call
 
-  # --- SEP-837 (application_type during DCR) ---
-  # The sep-837-application-type-present check only fires on 2026-version
-  # runs; the client omits application_type during Dynamic Client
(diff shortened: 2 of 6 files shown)


src/mcp/shared/auth.py
```diff
@@ -67,6 +67,9 @@ class OAuthClientMetadata(BaseModel):
     # servers may also return additional types they support
     response_types: list[str] = ["code"]
     scope: str | None = None
+    # SEP-837: OIDC application_type. Defaults to "native" since MCP clients typically use
+    # loopback redirect URIs; set "web" for remote browser-based clients on a non-local host.
```
Kludex, line 71: No need for comment

maxisbey (approved):

PR #2931, opened 2026-06-20
Union previously requested scopes on step-up re-authorization (SEP-2350)

Kludex:
Implements SEP-2350 (the fourth client-side item of #2902): union previously requested scopes on step-up re-authorization.

## What changed

When a `403 insufficient_scope` challenge triggers step-up re-authorization, the OAuth client previously **replaced** `client_metadata.scope` with only the newly challenged scopes, dropping the scopes granted for earlier operations. It now requests the **union** of the previously requested scopes and the challenged scopes.

Per the spec: "Scope accumulation across operations is a client-side responsibility. Clients SHOULD compute the union of previously requested scopes and newly challenged scopes when initiating re-authorization." This lets servers stay stateless about client scope sets while clients keep their permissions.

- New `union_scopes(previous, new)` helper in `utils.py` — order-preserving, deduped.
- Unions at the 403 step-up call site in `oauth2.py`.

## Conformance

Flips **`auth/scope-step-up`** green on the default leg: `sep-2350-scope-union-on-reauth` now reports SUCCESS (21/21). No baseline edits were needed:

- **Default leg**: `scope-step-up` was already removed from `expected-failures.yml` by #2927 (for the alpha.5 harness), so it was an *unexpected* failure on the pinned alpha.4 harness. It now passes, so the default leg is green again - this also resolves the pre-existing failure I flagged in #2930.
- **2026-07-28 leg**: `scope-step-up` stays baselined there - it still fails for an unrelated reason ("client did not make an initial authorization request", the documented 2026 connection-lifecycle gap in `client.py`), not on the SEP-2350 check.

Both legs pass with no stale or unexpected entries.

## AI Disclaimer

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1303,6 +1303,10 @@ If you relied on extra fields round-tripping through MCP types, move that data i
 
 ## New Features
 
+### Step-up authorization unions previously requested scopes (SEP-2350)
+
+When a `403 insufficient_scope` challenge triggers step-up re-authorization, the OAuth client now requests the union of the previously requested scopes and the newly challenged scopes, instead of replacing the scope with only the challenged ones. This keeps permissions granted for earlier operations from being dropped when a later operation escalates. No API change; the wider scope is sent automatically on the re-authorization request.
+
 ### OAuth Dynamic Client Registration sends `application_type` (SEP-837)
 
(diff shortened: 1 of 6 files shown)


Kludex:
Ran a `codex review` pass, which flagged a valid edge case: on restart only the persisted token is reloaded (not `client_metadata.scope`), so the step-up union would re-authorize for less than was previously granted. Fixed in 33bcb51d - the union now also folds in the stored token's scope, and a new test (`test_403_step_up_preserves_scope_from_stored_token`) covers the restart path. Conformance `auth/scope-step-up` still passes 21/21.

maxisbey (approved):

PR #2933, opened 2026-06-20
Bind client credentials to their authorization server (SEP-2352)

Kludex:
Implements SEP-2352 (the last client-side item of #2902): bind OAuth client credentials to their authorization server and re-register when it changes.

## What changed

Persisted client credentials are now bound to the issuer that registered them. `OAuthClientInformationFull` gains an SDK-owned `issuer` field, set after DCR/CIMD. Before reusing stored credentials, the provider compares their recorded issuer to the authorization server discovered from protected resource metadata; on a mismatch it discards the credentials (and the old tokens) and re-registers, instead of presenting one server's `client_id` to another.

- **CIMD** (URL-based client IDs): portable across servers, always match - no re-registration.
- **No recorded issuer** (pre-registered, or stored before this change): no binding to enforce, left as-is.
- **No `TokenStorage` protocol change** - the `issuer` round-trips through the existing `get_client_info`/`set_client_info`, so the 8 in-repo implementations and downstream users are unaffected.

Follows the **Go SDK's** approach (issuer field on the client-info, proactive comparison). The TS SDK takes a reactive invalidate-on-error path that doesn't fit our httpx auth-generator cleanly.

## Conformance

The `auth/authorization-server-migration` scenario's `no-reuse-on-as-change` and `no-cross-as-credential-reuse` checks pass (the client never presents the old AS's `client_id` to the new AS). The `reregister-on-as-change` check is **not yet reached**: the scenario runs only at `2026-07-28`, where `client.py`'s 2025 stateful lifecycle is rejected (400 on `initialize`) before the migration 401 fires. The scenario therefore stays baselined in both expected-failures files, re-annotated to point at the stateless-lifecycle gap (the same gap blocking `tools_call`, `scope-step-up`, etc. on the 2026 leg) rather than SEP-2352. An in-process interaction test covers the migration end to end.

Both conformance legs pass with no stale or unexpected entries.

## #2902 status

This completes the **implementation** of all five client-side items (SEP-2207, SEP-2468, SEP-837, SEP-2350, SEP-2352). The AS-migration conformance scenario will go green once the 2026 stateless client lifecycle lands.

## AI Disclaimer

### DIFF
--- a/.github/actions/conformance/expected-failures.2026-07-28.yml
+++ b/.github/actions/conformance/expected-failures.2026-07-28.yml
@@ -46,8 +46,10 @@ client:
   # SEP-2243 (HTTP standardization): no fixture handler / client header support yet.
   - http-custom-headers
   - http-invalid-tool-headers
-  # SEP-2352 (authorization server migration): client does not re-register when
-  # PRM authorization_servers changes.
+  # SEP-2352 (authorization server migration): the client re-registers and does not reuse the old
+  # AS credentials, but the 2026-mode mock rejects the MCP POST before the migration 401 fires
+  # (client.py drives the 2025 stateful lifecycle), so the re-register check is never reached.
+  # Unblocks with the 2026 stateless client lifecycle.
   - auth/authorization-server-migration
   # auth/enterprise-managed-authorization (SEP-990) is in the 2025 baseline but
   # NOT here: the harness skips it as inapplicable at --spec-version 2026-07-28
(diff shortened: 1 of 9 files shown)


Kludex:
Ran a `codex review` pass, which raised two valid P2 edge cases — both fixed in 523b32b0:

1. **Legacy no-PRM DCR persisted `issuer=None`.** When PRM discovery fails but AS-metadata discovery succeeds, `auth_server_url` is None, so DCR creds were stamped with no issuer — a later PRM pointing elsewhere would reuse them. Now the bound issuer falls back to the discovered `oauth_metadata.issuer`.
2. **CIMD inferred from URL shape.** A registration server returning a URL-shaped `client_id` was misclassified as portable CIMD. Portability is now keyed on the `client_id` equaling the configured `client_metadata_url`, not URL syntax. Added a regression test (`test_credentials_match_issuer_url_shaped_dcr_id_is_not_portable`).

Full suite + 100% coverage + both conformance legs still green.

maxisbey (approved):

PR #2934, opened 2026-06-20
Buffer per-request StreamableHTTP streams to avoid serial-router head-of-line block

### LINKED ISSUE #1764: Race condition in StreamableHTTP: zero-buffer memory streams cause deadlock with concurrent SSE responses

maxisbey:
Two related fixes to the stateful `StreamableHTTPServerTransport` so concurrent requests can't wedge each other and event-store ordering doesn't depend on scheduler timing.

**Buffer the per-request `_request_streams` (head-of-line block).** Each POST registers a buffer-0 `_request_streams[id]` channel, `start_soon`s the `EventSourceResponse`, and pushes the request to the server. The single serial `message_router` then forwards each response with `await _request_streams[id][0].send(...)` — which on a buffer-0 stream parks until that request's `sse_writer` (started lazily, two `start_soon` hops deep) has reached its first `receive()`. While the router is parked on one request, every other in-flight response on the session is head-of-line blocked behind it. This gives the three `_request_streams[EventMessage]` sites a small bounded buffer (`REQUEST_STREAM_BUFFER_SIZE = 16`) so the router can deposit and move on. The downstream `sse_stream` dict streams stay at 0 — they're not on the router's send target.

**Store the priming event before request dispatch (event-store ordering).** With an `event_store` configured, `message_router` calls `store_event(msg)` before its `send()`, while the priming event was stored inside the lazily-started `sse_writer` — so its event-store position raced the first routed message. Splitting the helper into `_mint_priming_event` (store + return wire dict) and `_run_sse_writer` (forward onto the wire) lets the POST handler `await _mint_priming_event(...)` before `writer.send(session_message)`. The server can't emit anything for a request it hasn't received, so the priming row now precedes every router store for that stream by data dependency rather than scheduler timing — the same shape the TypeScript SDK uses.

## Motivation and Context

Fixes #1764. The natural race is hard to reproduce standalone (multiple parties report 0 hits across thousands of loopback iterations) but is observed consistently in production by several reporters on the issue. The unit-level head-of-line block is straightforward to demonstrate though: register two request streams, give B a consumer, leave A consumer-less, write responses A then B — on `main` B never arrives until A is closed; with this change B arrives immediately and A's response is buffered.

The priming-order fix addresses a review finding that the buffer change widened a pre-existing window where a routed message could be stored before the priming event.

## How Has This Been Tested?

- New `tests/server/test_streamable_http_router.py` drives the routing layer directly: one test for the head-of-line block, one asserting `[priming, m₁, …, m₅]` event-store order with no `sse_writer` ever scheduled.
- `tests/shared/test_streamable_http.py` priming-helper unit tests updated for `_mint_priming_event`.
- Existing reconnection / sse-polling / replay tests pass unchanged.

## Breaking Changes

None. `EventStore` ABC is unchanged.

**Intentional behaviour changes:**

- The standalone GET stream now buffers up to 16 server-initiated messages before backpressure engages on a slow GET client (was 0).
- A request whose consumer stalls past 16 messages (e.g. a long-running tool emitting many progress notifications to a stalled SSE client) can still wedge the router — this moves the threshold from 0 to 16, it doesn't remove the serial router. The routerless design in `_streamable_http_modern.py` is the structural fix on `main`.
- `_maybe_send_priming_event` is replaced by `_mint_priming_event` (returns the wire dict instead of sending) and `_run_sse_writer`. Both are private; no public surface change.
- The replay path's priming/live-tail ordering window is pre-existing and unchanged here; tracked separately.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The buffer-size approach matches what reporters on #1764 have field-validated (sed-patching to 10). 16 covers the reported single-response list-request case by construction (one message per id) plus headroom for small notification workloads, while staying bounded for the session-lifetime GET stream.

A separate change is needed for the stateless task-leak half of #1764 (#2145's request-scoped task-group approach); that's not bundled here.

### DIFF
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -1636,80 +1636,30 @@ async def test_handle_sse_event_skips_empty_data() -> None:
 
 @pytest.mark.anyio
 async def test_priming_event_not_sent_for_old_protocol_version() -> None:
-    """_maybe_send_priming_event skips for old protocol versions (backwards compat)."""
-    # Create a transport with an event store
-    transport = StreamableHTTPServerTransport(
-        "/mcp",
-        event_store=SimpleEventStore(),
-    )
-
-    # Create a mock stream writer
-    write_stream, read_stream = anyio.create_memory_object_stream[dict[str, Any]](1)
-
-    try:
-        # Call _maybe_send_priming_event with OLD protocol version - should NOT send
-        await transport._maybe_send_priming_event("test-request-id", write_stream, "2025-06-18")
-
-        # Nothing should have been written to the stream
-        assert write_stream.statistics().current_buffer_used == 0
-
-        # Now test with NEW protocol version - should send
-        await transport._maybe_send_priming_event("test-request-id-2", write_stream, "2025-11-25")
-
-        # Should have written a priming event
-        assert write_stream.statistics().current_buffer_used == 1
-    finally:
-        await write_stream.aclose()
-        await read_stream.aclose()
+    """`_mint_priming_event` skips for old protocol versions (backwards compat)."""
+    transport = StreamableHTTPServerTransport("/mcp", event_store=SimpleEventStore())
+    assert await transport._mint_priming_event("test-request-id", "2025-06-18") is None
+    assert await transport._mint_priming_event("test-request-id-2", "2025-11-25") is not None
 
 
 @pytest.mark.anyio
 async def test_priming_event_not_sent_without_event_store() -> None:
-    """_maybe_send_priming_event returns early when no event_store is configured."""
-    # Create a transport WITHOUT an event store
+    """`_mint_priming_event` returns `None` when no event_store is configured."""
     transport = StreamableHTTPServerTransport("/mcp")
-
-    # Create a mock stream writer
-    write_stream, read_stream = anyio.create_memory_object_stream[dict[str, Any]](1)
-
-    try:
-        # Call _maybe_send_priming_event - should return early without sending
-        await transport._maybe_send_priming_event("test-request-id", write_stream, "2025-11-25")
-
-        # Nothing should have been written to the stream
-        assert write_stream.statistics().current_buffer_used == 0
-    finally:
-        await write_stream.aclose()
-        await read_stream.aclose()
+    assert await transport._mint_priming_event("test-request-id", "2025-11-25") is None
 
 
 @pytest.mark.anyio
 async def test_priming_event_includes_retry_interval() -> None:
-    """_maybe_send_priming_event includes the retry field when retry_interval is set."""
-    # Create a transport with an event store AND retry_interval
+    """`_mint_priming_event` includes the retry field when `retry_interval` is set."""
     transport = StreamableHTTPServerTransport(
         "/mcp",
         event_store=SimpleEventStore(),
         retry_interval=5000,
     )
-
-    # Create a mock stream writer
-    write_stream, read_stream = anyio.create_memory_object_stream[dict[str, Any]](1)
-
-    try:
-        # Call _maybe_send_priming_event with new protocol version
-        await transport._maybe_send_priming_event("test-request-id", write_stream, "2025-11-25")
-
-        # Should have written a priming event with retry field
-        assert write_stream.statistics().current_buffer_used == 1
-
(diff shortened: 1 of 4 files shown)


tests/shared/test_streamable_http.py
```diff
@@ -1636,80 +1636,30 @@ async def test_handle_sse_event_skips_empty_data() -> None:
...
-
-        # Now test with NEW protocol version - should send
-        await transport._maybe_send_priming_event("test-request-id-2", write_stream, "2025-11-25")
-
-        # Should have written a priming event
-        assert write_stream.statistics().current_buffer_used == 1
-    finally:
-        await write_stream.aclose()
-        await read_stream.aclose()
+    """`_mint_priming_event` skips for old protocol versions (backwards compat)."""
```
maxisbey: can we do a more high level end to end test here? if it's too hard to drive the client via only high level interfaces, then whatever you want for client is fine. but really I think we should have a high level test showing it works in memory with the transport and a real server.

felixweinberger (approved):

PR #2939, opened 2026-06-21
fix(mcpserver): preserve Annotated/Field metadata for dict[str, T] return types

### LINKED ISSUE #2935: dict[str, T] tool return types lose Annotated/Field metadata in output schema
### Initial Checks

### Description

Tools that return a string-keyed dict (dict[str, T]) lose any Annotated/Field metadata on the return type when the output schema is generated. A Field(description=...) on the return annotation shows up in the output schema for every other return type, but for dict[str, T] it's dropped.

I'd expect the description to be preserved, the same as other return types. Looks like the dict[str, T] branch in _try_create_model_and_schema builds the model from the Annotated-stripped type instead of the original annotation (there's an existing TODO on that line).

Have a simple, tested fix ready to go. Love the project and would love to contribute if possible! 

### Example Code

```Python
from typing import Annotated
from pydantic import Field
from mcp.server.mcpserver.utilities.func_metadata import func_metadata

def get_config() -> Annotated[dict[str, int], Field(description="Configuration values")]:
    return {"timeout": 30}

print(func_metadata(get_config).output_schema)
# {'type': 'object', 'additionalProperties': {'type': 'integer'}, 'title': 'get_configDictOutput'}
# expected to also include: 'description': 'Configuration values'
```

### Python & MCP Python SDK

```Text
Python 3.14.5
mcp: main branch (mcpserver / V2)
```

anneheartrecord:
## Summary

When a tool declares a `dict[str, T]` return type wrapped in `Annotated` with
Pydantic `Field` metadata (e.g. a description), that metadata was silently dropped
from the generated output schema.

```python
def get_config() -> Annotated[dict[str, int], Field(description="Configuration values")]:
    return {"timeout": 30}

print(func_metadata(get_config).output_schema)
# Before: {'type': 'object', 'additionalProperties': {'type': 'integer'}, 'title': 'get_configDictOutput'}
# After:  {'type': 'object', 'additionalProperties': {'type': 'integer'}, 'title': 'get_configDictOutput', 'description': 'Configuration values'}
```

## Root cause

In `_try_create_model_and_schema`, the `dict[str, T]` branch called
`_create_dict_model(func_name, type_expr)` where `type_expr` is the
`Annotated`-stripped type. The existing TODO comment on that line even flagged
the issue. The fix is to pass `original_annotation` instead, so
`RootModel[Annotated[dict[str, T], Field(...)]]` picks up the metadata.

## Changes

- `src/mcp/server/mcpserver/utilities/func_metadata.py`: pass `original_annotation`
  instead of `type_expr` to `_create_dict_model`; replace TODO comment with a
  clarifying one
- `tests/server/mcpserver/test_func_metadata.py`: add
  `test_structured_output_dict_str_preserves_annotated_metadata` regression test
  covering `Field(description=...)` and `Field(description=..., title=...)`

Fixes #2935

### DIFF
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -349,9 +349,9 @@ def _try_create_model_and_schema(
         if origin is dict:
             args = get_args(type_expr)
             if len(args) == 2 and args[0] is str:
-                # TODO: should we use the original annotation? We are losing any potential `Annotated`
-                # metadata for Pydantic here:
-                model = _create_dict_model(func_name, type_expr)
+                # Use the original annotation (which may be `Annotated[dict[str, T], Field(...)]`)
+                # so that any Annotated metadata (e.g. Field description) is preserved in the schema.
+                model = _create_dict_model(func_name, original_annotation)
             else:
                 # dict with non-str keys needs wrapping
                 model = _create_wrapped_model(func_name, original_annotation)
--- a/tests/server/mcpserver/test_func_metadata.py
+++ b/tests/server/mcpserver/test_func_metadata.py
@@ -261,7 +261,9 @@ def test_structured_output_dict_str_preserves_annotated_metadata():
     )
 
     # Additional metadata (title, max_length, etc.) should also be preserved
-    def get_headers() -> Annotated[dict[str, str], Field(description="HTTP headers", title="Headers")]:  # pragma: no cover
+    def get_headers() -> Annotated[
+        dict[str, str], Field(description="HTTP headers", title="Headers")
+    ]:  # pragma: no cover
         return {"Content-Type": "application/json"}
 
     meta2 = func_metadata(get_headers)
--- a/tests/server/mcpserver/test_func_metadata.py
+++ b/tests/server/mcpserver/test_func_metadata.py
@@ -245,6 +245,31 @@ def test_structured_output_dict_str_types():
     assert "result" in meta.output_schema["properties"]
 
 
+def test_structured_output_dict_str_preserves_annotated_metadata():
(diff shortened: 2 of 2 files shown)


anneheartrecord:
Fixed the pre-commit failure — ruff-format wrapped a long return-type annotation in the test file. All other checks were already green.

itxaiohanglover:
Nice catch! Using `original_annotation` instead of `type_expr` preserves the Annotated metadata for Pydantic. Simple one-line fix with the right regression test covering both description and title metadata.

anneheartrecord:
@maxisbey — this has been open since Jun 21 with no maintainer pass, so surfacing it. It should be a fast one:

**It answers a TODO this repo already wrote.** `func_metadata.py` currently has, at line 510 on `main` today:

```python
# TODO: should we use the original annotation? We are losing any potential `Annotated`
# metadata for Pydantic here:
model = _create_dict_model(func_name, type_expr)
```

The answer is yes, and that is the whole change — passing `original_annotation` instead of `type_expr`, so `Annotated[dict[str, T], Field(description=...)]` keeps its metadata in the output schema. One line of source, plus a regression test covering both `description` and `title`. Fixes #2935.

**State:** +30/-3 across 2 files, all 27 checks green, still `MERGEABLE`. It also survived #3331 untouched — that PR reworked other parts of the same file, but this `origin is dict` branch and its TODO are unchanged on `main`, so the fix still applies as written. Happy to rebase for a fresh run if you want one.

**Independent read:** @itxaiohanglover looked at it on Jun 26 and reached the same conclusion about `original_annotation` vs `type_expr`.

No objection if you would rather resolve the TODO differently — if you prefer the metadata to be dropped deliberately and the comment removed instead, say so and I will close this. I mostly want the TODO to stop being ambiguous either way.

PR #2941, opened 2026-06-21
Slim ServerMiddleware to `(ctx, call_next)` and add `OpenTelemetryMiddleware`

Kludex:
## What

Reshapes the context-tier `ServerMiddleware` interface and adds a context-tier `OpenTelemetryMiddleware`.

- `ServerMiddleware.__call__` goes from `(ctx, method, params, call_next)` to `(ctx, call_next)`. `method` and `params` now live on `ServerRequestContext` as `ctx.method` and raw `ctx.params`.
- `CallNext` takes the context: `Callable[[ServerRequestContext], Awaitable[HandlerResult]]`. Middleware passes it through with `call_next(ctx)`, and can rewrite the inbound message via `call_next(replace(ctx, params=...))`.
- New `OpenTelemetryMiddleware` (context-tier) in `mcp/server/_otel.py`, mirroring the span shape of the existing dispatch-tier `otel_middleware`, which is left intact. It spans both requests and notifications, setting `jsonrpc.request.id` only when present.

## Why

The 4-arg signature duplicated `method`/`params` that already belong on the context, and `call_next()` could only observe, not rewrite. The new shape matches the pure-ASGI middleware model (single context in, pass it through, mutate before next) and lets middleware redirect params before the handler runs.

## Notes

- Raw `ctx.params` is deliberate: middleware runs before method lookup and validation (it wraps the failure path too), so no single validated type exists at that tier. Authors validate on demand with `Model.model_validate(ctx.params or {})`.
- Documented in `docs/migration.md` (no compat shim - the 4-arg form never shipped in a release).
- Reviewed with Codex; the notification `request_id is None` path was confirmed correct.

## AI Disclaimer

### DIFF
--- a/src/mcp/server/_otel.py
+++ b/src/mcp/server/_otel.py
@@ -1,0 +1,51 @@
+from __future__ import annotations
+
+from typing import Any
+
+from opentelemetry.trace import SpanKind, StatusCode
+from pydantic import ValidationError
+
+from mcp.server.context import CallNext, HandlerResult, ServerMiddleware, ServerRequestContext
+from mcp.shared._otel import extract_trace_context, otel_span
+from mcp.shared.exceptions import MCPError
+
+
+class OpenTelemetryMiddleware(ServerMiddleware[Any]):
+    """Context-tier middleware that wraps each inbound message in an OpenTelemetry span.
+
+    Span name `"MCP handle <method> [<target>]"`, `mcp.method.name` attribute, W3C
+    trace context extracted from `params._meta` (SEP-414), and an ERROR status if
+    the handler raises. Requests and notifications both get a span;
+    `jsonrpc.request.id` is set only when `ctx.request_id` is present (notifications
+    have none).
+    """
+
+    async def __call__(self, ctx: ServerRequestContext[Any, Any], call_next: CallNext) -> HandlerResult:
+        name = ctx.params.get("name") if ctx.params else None
+        target = name if isinstance(name, str) else None
+
+        attributes: dict[str, Any] = {"mcp.method.name": ctx.method}
+        if ctx.request_id is not None:
+            attributes["jsonrpc.request.id"] = str(ctx.request_id)
+
+        with otel_span(
+            name=f"MCP handle {ctx.method}{f' {target}' if target else ''}",
+            kind=SpanKind.SERVER,
+            attributes=attributes,
+            context=extract_trace_context(ctx.meta or {}),
+            record_exception=False,
+            set_status_on_exception=False,
+        ) as span:
+            try:
+                return await call_next(ctx)
+            except MCPError as e:
+                span.set_status(StatusCode.ERROR, e.error.message)
+                raise
+            except ValidationError:
+                # Mirror the sanitized wire response; pydantic messages carry client input.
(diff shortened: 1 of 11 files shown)


tests/server/test_otel.py
Kludex, line 1: I didn't realize I created those tests. I don't think they are necessary. We already have tests for otel.

src/mcp/server/runner.py
```diff
@@ -345,33 +357,33 @@ async def _inner() -> None:
...
+    def _compose_server_middleware(self, inner: CallNext) -> CallNext:
         """Wrap `inner` in `Server.middleware`, outermost-first.
 
         Shared by `_on_request` and `_on_notify` so the same middleware chain
-        observes every inbound message.
+        observes every inbound message. The composed callable takes the `ctx`
+        at call time, so a middleware can rewrite it for the rest of the chain.
         """
-        call = inner
+        call: CallNext = inner
```
Kludex: unnecessary type hint.

src/mcp/server/runner.py
```diff
@@ -169,6 +169,14 @@ def _dump_result(result: Any) -> dict[str, Any]:
     raise TypeError(f"handler returned {type(result).__name__}; expected BaseModel, dict, or None")
 
 
+def _apply_middleware(
+    mw: ServerMiddleware[Any], call_next: CallNext, ctx: ServerRequestContext[Any, Any]
+) -> Awaitable[HandlerResult]:
+    """Adapt one middleware to the `CallNext` shape: bind `call_next`, take
+    `ctx` at call time so a rewritten context flows down the chain."""
+    return mw(ctx, call_next)
```
Kludex: Is this necessary?

Also, we should stop using "mw" in this code source!!! I don't need to guess variable names!

src/mcp/server/_otel.py
```diff
@@ -0,0 +1,51 @@
...
+            attributes["jsonrpc.request.id"] = str(ctx.request_id)
+
+        with otel_span(
+            name=f"MCP handle {ctx.method}{f' {target}' if target else ''}",
+            kind=SpanKind.SERVER,
+            attributes=attributes,
+            context=extract_trace_context(ctx.meta or {}),
+            record_exception=False,
+            set_status_on_exception=False,
+        ) as span:
```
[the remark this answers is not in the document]
Kludex: Makes sense

src/mcp/server/runner.py
```diff
@@ -282,8 +293,8 @@ async def _inner() -> HandlerResult:
                 raise MCPError.from_error_data(result)
             return result
 
-        call = self._compose_server_middleware(ctx, method, params, _inner)
-        result = _dump_result(await call())
+        call = self._compose_server_middleware(_inner)
+        result = _dump_result(await call(ctx))
         # TODO: reject resultType values outside {"complete", "input_required"} unless the
         # corresponding extension is in this request's _meta clientCapabilities.extensions; the
         # explicit MUST-reject is client-side (basic/index.mdx ResultType), this enforces it proactively.
```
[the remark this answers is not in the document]
maxisbey: honestly not super sure what to do about this one @Kludex do you have any ideas?

felixweinberger (approved):

PR #2944, opened 2026-06-22
fix: reject duplicate JSON-RPC request IDs with 409 Conflict (fixes #2655)

### LINKED ISSUE #2655: Streamable HTTP server silently drops in-flight request when client reuses a JSON-RPC id

kaXianc2-gom:
## Summary

Fixes #2655: `StreamableHTTPServerTransport` now detects and rejects duplicate JSON-RPC request IDs before overwriting the in-flight stream slot.

**Before:** A client sending two requests with the same `id` on the same session would silently overwrite the first request's `(send, receive)` stream pair in `_request_streams`. The original caller hangs until timeout; the response may route to the wrong caller.

**After:** The server returns HTTP 409 Conflict with a JSON-RPC `-32600 Invalid Request` error, preserving the original in-flight stream.

## Changes

1. **`_handle_post_request`** — added collision check before stream registration, mirroring the existing `GET_STREAM_KEY` guard at line 695
2. **`_create_error_response`** — added optional keyword-only `request_id` parameter so the JSON-RPC error envelope carries the offending id instead of `null`

## AI Assistance Disclosure

This contribution was developed with AI assistance (Claude / Anthropic Claude Code).

## Verification Process

1. Synced sandbox to upstream main (2397319)
2. `uv sync --no-group docs` installed all dependencies
3. Baseline test suite: 2269 passed, 12 skipped, 1 xfailed (×3, no flakes)
4. Applied fix (18 lines in `streamable_http.py`, +1 `noqa: C901`)
5. Added targeted test `test_duplicate_request_id_rejected_with_409`:
   - Seeds `_request_streams` with an in-flight request
   - POSTs a duplicate request with the same id
   - Asserts 409 status, correct JSON-RPC error envelope with the id, and original stream preserved
6. Full test suite after upstream merge: 2571 passed, 12 skipped, 9 xfailed
7. `ruff format` + `ruff check`: all clean

## Cross-Validation

| Scenario | Before | After |
|----------|--------|-------|
| Normal request (unique id) | ✅ | ✅ |
| Duplicate id while original in-flight | ❌ silent overwrite | ✅ 409 + preserved stream |
| Duplicate id after original completed | ✅ normal (stream cleaned) | ✅ same |
| Error response `id` field | ❌ always `null` | ✅ carries the request id |
| GET_STREAM_KEY collision (existing) | ✅ | ✅ (unchanged) |

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -288,6 +288,8 @@ class StreamableHTTPServerTransport:
         status_code: HTTPStatus,
         error_code: int = INVALID_REQUEST,
         headers: dict[str, str] | None = None,
+        *,
+        request_id: RequestId | None = None,
     ) -> Response:
         """Create an error response with a simple string message."""
         response_headers = {"Content-Type": CONTENT_TYPE_JSON}
@@ -300,7 +302,7 @@ class StreamableHTTPServerTransport:
         # Return a properly formatted JSON error response
         error_response = JSONRPCError(
             jsonrpc="2.0",
-            id=None,
+            id=request_id,
             error=ErrorData(code=error_code, message=error_message),
         )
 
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2657.

Issue #2900, opened 2026-06-17
HTTP standard headers + x-mcp-header


maxisbey:
Tracks: #2715

The shared encode/decode and `x-mcp-header` rule helpers used by both sides; server-side validation of inbound headers vs body.

## What's in it

- Encode/decode helpers for `Mcp-Method`/`Mcp-Name`/`Mcp-Param-*` (base64-sentinel, the `x-mcp-header` constraint rules).
- Server: validate inbound headers vs body, reject mismatches — wired in by #2893.
- Client emit is owned by #2894 (consumes these helpers).

## Conformance

- (helpers — scenarios are owned by _New HTTP serving path (modern + dual-era)_ server-side and _New Client (all-versions, low-level + high-level)_ client-side)
- Together with #2893: `http-header-validation`, `http-custom-header-server-validation`
- Together with #2894: `http-custom-headers`, `http-invalid-tool-headers`, `http-standard-headers`

## Dependencies

- Depends on: #2893

## References

- [Transports — headers](https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http#request-metadata)
- [SEP-2243](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243)
- [sep-2243.yaml](https://github.com/modelcontextprotocol/conformance/blob/main/src/seps/sep-2243.yaml)
- [python-sdk #2715](https://github.com/modelcontextprotocol/python-sdk/issues/2715)

maxisbey:
Done via #2974 (server-side validation + x-mcp-header filter) and #2990 (client `Mcp-Param-*` emit). Conformance waivers for SEP-2243 are zero in both baseline files.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

PR #2950, opened 2026-06-22
Client-side 2026-07-28 support: .discover()/.adopt() + Client(mode=); request-metadata green

maxisbey:
Client-side support for the 2026-07-28 per-request-envelope path: `ClientSession` gains `.discover()` and `.adopt()` alongside `.initialize()`; `Client` gains `mode='legacy'|'auto'|<version-pin>` and `prior_discover=`. Removes `request-metadata` and `auth/authorization-server-migration` from the conformance baseline, plus the carried-forward `tools_call`/`auth/scope-step-up`/`auth/scope-retry-limit` entries from the 2026-07-28 baseline.

Part of #2891. Touches #2894, #2892, #2900.

## Motivation and Context

#2928 landed the server side of the 2026-07-28 era split. This PR is the client side: the era difference becomes which outbound-stamping closure was installed at connect time, not a flag the send path reads. `ClientSession` previously branched on a `_stateless_pinned` flag inside `send_request` and held the protocol version in four places (session pin, init result, transport, OAuth context); the transport sniffed `InitializeResult` responses to learn the version for header setting.

## What changed

**`ClientSession` — three connect-time entry points install a stamp closure.**
- `.initialize()` (existing) terminates by calling `.adopt(result)`.
- `.adopt(InitializeResult | DiscoverResult)` installs negotiated state without wire traffic. A `DiscoverResult` selects the newest mutually-supported modern version and installs the modern stamp; an `InitializeResult` installs the handshake stamp.
- `.discover()` probes `server/discover`, validates the response with `DiscoverResult.model_validate` before reading any field, and `.adopt()`s on success. On `-32022` it retries once with the intersection of `MODERN_PROTOCOL_VERSIONS` and `data.supported`; on `-32601` or request timeout it falls back to `.initialize()`; anything else propagates.

`send_request` and `send_notification` call `self._stamp(data, opts)` unconditionally — no era branch in the body. The `_stateless_pinned` flag, `_pinned_version` slot, and the `ClientSession(protocol_version=)` constructor kwarg are removed.

**`Client` — policy layer.** New `mode: Literal['legacy','auto'] | str = 'legacy'` and `prior_discover: DiscoverResult | None = None`. `Client.__aenter__` builds the session, then: `'legacy'` → `.initialize()`; `'auto'` → `.discover()`; a version string → `.adopt(prior_discover or synthesize(pv))`. `Client(protocol_version=)` is removed.

**Transport pv-agnostic.** `StreamableHTTPTransport` no longer holds `protocol_version`, no longer derives `Mcp-Method`/`Mcp-Name` headers, and no longer sniffs `InitializeResult` responses. Per-message headers arrive via `CallOptions['headers']` → `ClientMessageMetadata.headers` → merged at the POST. The transport caches `MCP-Protocol-Version` from the first stamped POST for transport-internal GET/DELETE/reconnect (per-connection state, same pattern as `session_id`).

**In-process modern path.** New `modern_on_request(server, lifespan_state)` driver in `runner.py` returns an `OnRequest` callback that builds `Connection.from_envelope` per call and drives `serve_one`. `Client(Server | MCPServer, mode != 'legacy')` enters the lifespan once, creates a `DirectDispatcher` peer-pair, and runs the server side with this callback. The interaction suite's in-memory transport is unlocked for 2026-07-28 (71 tests now run on that arm).

**Version constants.** `SUPPORTED_PROTOCOL_VERSIONS` renamed to `HANDSHAKE_PROTOCOL_VERSIONS` (the versions reachable via the initialize handshake); the old name survives as a deprecated union. `LATEST_PROTOCOL_VERSION` bumped to `"2026-07-28"`. The three duplicate `mcp-protocol-version` header constant definitions collapsed to one in `shared/inbound`.

**`report_progress` routes through `DispatchContext.progress()`.** `Context.report_progress` was gating on a JSONRPC-specific `_meta.progressToken` and reimplementing the notification path; it now delegates to `ServerSession.report_progress` → `dctx.progress()`, so progress reaches the client on the in-process modern path too.

**Conformance fixture.** `.github/actions/conformance/client.py` reads `MCP_CONFORMANCE_PROTOCOL_VERSION` and drives `Client(mode='auto')` for the modern leg, `'legacy'` otherwise. New handlers for `request-metadata` and `http-standard-headers`.

## How Has This Been Tested?

- Conformance: `request-metadata` 7/7, `auth/authorization-server-migration` 27/27, `http-standard-headers` 3/3 on both legs; `tools_call`/`auth/scope-step-up`/`auth/scope-retry-limit` pass on the 2026-07-28 leg
- `./scripts/test`: 100% branch coverage, `strict-no-cover` clean
- 9 unit tests cover each `.discover()` ladder rung; 10 interaction tests in `test_client_connect.py` cover the `mode=` policy and envelope stamping end to end
- In-memory@2026-07-28: 71 interaction tests run, 67 pass (4 progress-on-streamable-http xfails remain scoped to that transport)

## Breaking Changes

All documented in `docs/migration.md`:

- `ClientSession(protocol_version=)` removed → use `.adopt()` after construction
- `Client(protocol_version=)` removed → use `mode=`
- `StreamableHTTPTransport.protocol_version` and `streamable_http_client(protocol_version=)` removed
- `SUPPORTED_PROTOCOL_VERSIONS` deprecated → use `HANDSHAKE_PROTOCOL_VERSIONS` or `MODERN_PROTOCOL_VERSIONS`
- `LATEST_PROTOCOL_VERSION` value changed `"2025-11-25"` → `"2026-07-28"`; code that meant "the version `.initialize()` offers" should switch to `HANDSHAKE_PROTOCOL_VERSIONS[-1]`
- `Client.send_progress_notification` / `ClientSession.send_progress_notification` deprecated (client-to-server progress is server-to-client only at 2026-07-28)
- `Outbound.notify` Protocol grew an `opts: CallOptions | None = None` parameter
- `ServerMessageMetadata.protocol_version` removed (no readers)

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

The three stamp closures: `_preconnect_stamp` (cancel-suppressed only — only `initialize`/`discover` go out before connect, both forbid cancel), `_make_handshake_stamp(pv)` (sets the `MCP-Protocol-Version` header), `_make_modern_stamp(pv, info, caps)` (the `_meta` triple + `cancel_on_abandon=False` + all three routing headers). `__init__` installs the first; `.initialize()`/`.adopt()`/`.discover()` install one of the other two.

The in-process modern path reuses the existing `DirectDispatcher` peer-pair (no new dispatcher class) — the era-specific bit is the `modern_on_request` callback wired into the server side, mirroring how `ServerRunner.on_request` is wired in for the legacy path.

`http-custom-headers` and `http-invalid-tool-headers` (the `Mcp-Param-*` header scenarios) and `sep-2322-client-request-state` (multi-round-trip results) stay waived — separate work.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -764,6 +764,10 @@ async def my_tool(ctx: Context) -> str: ...
 async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 ```
 
+### Version constants
+
+`SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`.
+
 ### `ProgressContext` and `progress()` context manager removed
 
 The `mcp.shared.progress` module (`ProgressContext`, `Progress`, and the `progress()` context manager) has been removed. This module had no real-world adoption — all users send progress notifications via `Context.report_progress()` or `session.send_progress_notification()` directly.
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -776,6 +776,8 @@ async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 
 `SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`.
 
+`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `HANDSHAKE_PROTOCOL_VERSIONS[-1]`.
+
 ### `ProgressContext` and `progress()` context manager removed
 
 The `mcp.shared.progress` module (`ProgressContext`, `Progress`, and the `progress()` context manager) has been removed. This module had no real-world adoption — all users send progress notifications via `Context.report_progress()` or `session.send_progress_notification()` directly.
@@ -1300,6 +1302,12 @@ warnings.filterwarnings("ignore", category=MCPDeprecationWarning)
 ```
 
 No migration is required during the deprecation window. New code should avoid building on these features, since they may be removed in a future spec version.
+
+### Client-to-server progress deprecated (2026-07-28)
+
+The 2026-07-28 spec restricts `notifications/progress` to the server-to-client direction only — `ProgressNotification` is no longer in `ClientNotification`. `Client.send_progress_notification()` and `ClientSession.send_progress_notification()` now carry `typing_extensions.deprecated` and emit `mcp.MCPDeprecationWarning` at runtime. They continue to work against servers negotiating 2025-11-25 or earlier.
+
+On the server side, prefer the new dispatcher-agnostic `ServerSession.report_progress(progress, total, message)` (and `Context.report_progress()` on `MCPServer`) over the raw `ServerSession.send_progress_notification(progress_token, …)`. `report_progress` encapsulates the "no-op when the caller did not request progress" rule and works on every dispatcher; the raw token-taking form remains for handlers that read `_meta.progressToken` directly.
 
 ## Bug Fixes
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -352,7 +352,15 @@ version = session.protocol_version
 
 The raw handshake result is also retained: `session.initialize_result` is set after `initialize()` (≤2025-11-25 servers — including `stateless_http=True` servers, which still answer `initialize`); `session.discover_result` is set after `discover()` (2026-07-28+ servers). At most one is non-`None`.
 
-On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it. (The lowlevel `ClientSession` still lets you call methods before any handshake, as in v1; `Client` always handshakes on enter.)
+On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it. (The lowlevel `ClientSession` still lets you call methods before any handshake, as in v1; `Client` always connects on enter — by default it probes `server/discover` and falls back to the initialize handshake.)
+
+### `Client` defaults to `mode='auto'`
+
+In v1, connecting to a server always performed the `initialize` handshake. In v2, `Client` defaults to `mode='auto'`: on enter it probes `server/discover` and, if the server doesn't support it, falls back to the `initialize` handshake. Pass `mode='legacy'` to force the initialize handshake and reproduce v1's byte-identical pre-2026 behavior, or pass a modern protocol-version string (e.g. `mode='2026-07-28'`) to pin a version without probing.
+
+For an in-process `Client(server)` (where `server` is a `Server` or `MCPServer` instance), `mode='auto'` dispatches calls directly through `DirectDispatcher` with no JSON-RPC framing. Pass `mode='legacy'` if you need the in-memory JSON-RPC transport that v1 used.
+
+`Client.send_ping()` is deprecated (ping is removed in 2026-07-28); pin `mode='legacy'` if you need it.
 
 ### `McpError` renamed to `MCPError`
 
@@ -832,7 +840,7 @@ async with Client(server) as client:
     result = await client.call_tool("my_tool", {"x": 1})
 ```
 
-`Client` accepts the same callback parameters the old helper did (`sampling_callback`, `list_roots_callback`, `logging_callback`, `message_handler`, `elicitation_callback`, `client_info`) plus `raise_exceptions` to surface server-side errors.
+`Client` accepts the same callback parameters the old helper did (`sampling_callback`, `list_roots_callback`, `logging_callback`, `message_handler`, `elicitation_callback`, `client_info`) plus `raise_exceptions` to surface server-side errors and `mode` to control version negotiation (`'auto'` by default; `'legacy'` reproduces v1's initialize-only handshake).
 
 If you need direct access to the underlying `ClientSession` and memory streams (e.g., for low-level transport testing), `create_client_server_memory_streams` is still available in `mcp.shared.memory`:
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -332,7 +332,7 @@ result = await session.list_tools(params=PaginatedRequestParams(cursor="next_pag
 
 ### `ClientSession.get_server_capabilities()` replaced by era-neutral accessors
 
-`ClientSession` now exposes the negotiated server metadata as properties: `server_capabilities`, `server_info`, `instructions`, and `protocol_version`. These are populated by whichever connection step ran (`initialize()` for ≤2025-11-25 servers, `discover()` for 2026-07-28+). The `get_server_capabilities()` method has been removed.
+`ClientSession` now exposes the negotiated server metadata as properties: `server_capabilities`, `server_info`, `instructions`, and `protocol_version`. These are populated by whichever connection step ran (`initialize()` for ≤2025-11-25 servers, `discover()` for 2026-07-28+), and are `None` if none has — matching v1's `get_server_capabilities()`. The `get_server_capabilities()` method has been removed.
 
 **Before (v1):**
 
@@ -350,9 +350,9 @@ instructions = session.instructions
 version = session.protocol_version
 ```
 
-The raw handshake result is also retained as `session.initialize_result` (legacy path) or `session.discover_result` (modern path) — exactly one is non-`None`.
+The raw handshake result is also retained: `session.initialize_result` is set after `initialize()` (≤2025-11-25 servers — including `stateless_http=True` servers, which still answer `initialize`); `session.discover_result` is set after `discover()` (2026-07-28+ servers). At most one is non-`None`.
 
-On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it.
+On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it. (The lowlevel `ClientSession` still lets you call methods before any handshake, as in v1; `Client` always handshakes on enter.)
 
 ### `McpError` renamed to `MCPError`
 
@@ -771,8 +771,6 @@ async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 ### Version constants
 
 `SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`. Named scalars derived from these tuples are now exported alongside them — `LATEST_HANDSHAKE_VERSION`, `LATEST_MODERN_VERSION`, `OLDEST_SUPPORTED_VERSION` — so prefer those over indexing the tuples directly.
-
-`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `LATEST_HANDSHAKE_VERSION`.
 
 ### `ProgressContext` and `progress()` context manager removed
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -163,6 +163,8 @@ Note: `sse_client` retains its `headers`, `timeout`, `sse_read_timeout`, and `au
 
 The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are now supplied per-message by the session. If you were reading `transport.protocol_version` to learn the negotiated version, read `session.protocol_version` (or `client.protocol_version` on the high-level `Client`) instead.
 
+The `MCP_PROTOCOL_VERSION` header-name constant has moved: import `MCP_PROTOCOL_VERSION_HEADER` from `mcp.shared.inbound` instead of `MCP_PROTOCOL_VERSION` from `mcp.client.streamable_http`.
+
 ### `terminate_windows_process` removed
 
 The deprecated `mcp.os.win32.utilities.terminate_windows_process` function has been
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -161,7 +161,7 @@ Note: `sse_client` retains its `headers`, `timeout`, `sse_read_timeout`, and `au
 
 ### `StreamableHTTPTransport.protocol_version` attribute removed
 
-The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are now supplied per-message by the session. If you were reading `transport.protocol_version` to learn the negotiated version, read it from `session.initialize_result.protocol_version` instead.
+The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are now supplied per-message by the session. If you were reading `transport.protocol_version` to learn the negotiated version, read `session.protocol_version` (or `client.protocol_version` on the high-level `Client`) instead.
 
 ### `terminate_windows_process` removed
 
@@ -330,9 +330,9 @@ result = await session.list_resources(params=PaginatedRequestParams(cursor="next
 result = await session.list_tools(params=PaginatedRequestParams(cursor="next_page_token"))
 ```
 
-### `ClientSession.get_server_capabilities()` replaced by `initialize_result` property
+### `ClientSession.get_server_capabilities()` replaced by era-neutral accessors
 
-`ClientSession` now stores the full `InitializeResult` via an `initialize_result` property. This provides access to `server_info`, `capabilities`, `instructions`, and the negotiated `protocol_version` through a single property. The `get_server_capabilities()` method has been removed.
+`ClientSession` now exposes the negotiated server metadata as properties: `server_capabilities`, `server_info`, `instructions`, and `protocol_version`. These are populated by whichever connection step ran (`initialize()` for ≤2025-11-25 servers, `discover()` for 2026-07-28+). The `get_server_capabilities()` method has been removed.
 
 **Before (v1):**
 
@@ -344,15 +344,15 @@ capabilities = session.get_server_capabilities()
 **After (v2):**
 
 ```python
-result = session.initialize_result
-if result is not None:
-    capabilities = result.capabilities
-    server_info = result.server_info
-    instructions = result.instructions
-    version = result.protocol_version
+capabilities = session.server_capabilities
+server_info = session.server_info
+instructions = session.instructions
+version = session.protocol_version
 ```
 
-The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. Like `session.initialize_result`, this replaces v1's `ClientSession.get_server_capabilities()`; use `client.initialize_result.capabilities` instead.
+The raw handshake result is also retained as `session.initialize_result` (legacy path) or `session.discover_result` (modern path) — exactly one is non-`None`.
+
+On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it.
 
 ### `McpError` renamed to `MCPError`
 
@@ -770,9 +770,9 @@ async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 
 ### Version constants
 
-`SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`.
+`SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`. Named scalars derived from these tuples are now exported alongside them — `LATEST_HANDSHAKE_VERSION`, `LATEST_MODERN_VERSION`, `OLDEST_SUPPORTED_VERSION` — so prefer those over indexing the tuples directly.
 
-`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `HANDSHAKE_PROTOCOL_VERSIONS[-1]`.
+`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `LATEST_HANDSHAKE_VERSION`.
 
 ### `ProgressContext` and `progress()` context manager removed
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -159,9 +159,9 @@ The `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters have been re
 
 Note: `sse_client` retains its `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters — only the streamable HTTP transport changed.
 
-### `protocol_version` removed from `StreamableHTTPTransport` and `streamable_http_client`
+### `StreamableHTTPTransport.protocol_version` attribute removed
 
-The `protocol_version` attribute on `StreamableHTTPTransport` and the `protocol_version` parameter on `streamable_http_client` have been removed. The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are supplied per-message by the session, so the transport never needs to know the negotiated version.
+The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are now supplied per-message by the session. If you were reading `transport.protocol_version` to learn the negotiated version, read it from `session.initialize_result.protocol_version` instead.
 
 ### `terminate_windows_process` removed
 
@@ -352,11 +352,7 @@ if result is not None:
     version = result.protocol_version
 ```
 
-The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. This replaces v1's `Client.server_capabilities`; use `client.initialize_result.capabilities` instead.
-
-### `ClientSession(protocol_version=)` removed
-
-The `protocol_version` constructor parameter on `ClientSession` has been removed. To install a known protocol version without performing the `initialize` handshake (e.g. when reconnecting to an existing session), call `session.adopt(result)` after construction with a stored `InitializeResult`.
+The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. Like `session.initialize_result`, this replaces v1's `ClientSession.get_server_capabilities()`; use `client.initialize_result.capabilities` instead.
 
 ### `McpError` renamed to `MCPError`
 
@@ -809,6 +805,12 @@ await session.send_progress_notification(
     total=100,
 )
 ```
+
+### Handler progress reporting: prefer `ctx.report_progress()` over manual `progress_token`
+
+Reading `ctx.meta["progress_token"]` and calling `session.send_progress_notification(token, ...)` is specific to the JSON-RPC transport path. On the in-process modern path (`DirectDispatcher` / `Client(server)`), there is no wire token in `_meta`, so handlers that gate progress on the token's presence go silent.
+
+`ctx.report_progress(progress, total, message)` works on every dispatcher: it sends a progress notification when a token is present and routes the update through the dispatcher's progress channel otherwise, no-opping only when the caller did not request progress at all. `session.send_progress_notification(progress_token, ...)` is unchanged and still works on JSON-RPC transports for code that already holds a token.
 
 ### `create_connected_server_and_client_session` removed
 
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -159,6 +159,10 @@ The `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters have been re
 
 Note: `sse_client` retains its `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters — only the streamable HTTP transport changed.
 
+### `protocol_version` removed from `StreamableHTTPTransport` and `streamable_http_client`
+
+The `protocol_version` attribute on `StreamableHTTPTransport` and the `protocol_version` parameter on `streamable_http_client` have been removed. The transport no longer holds per-connection protocol state; era-dependent headers (e.g. `MCP-Protocol-Version`) are supplied per-message by the session, so the transport never needs to know the negotiated version.
+
 ### `terminate_windows_process` removed
 
 The deprecated `mcp.os.win32.utilities.terminate_windows_process` function has been
@@ -349,6 +353,10 @@ if result is not None:
 ```
 
 The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. This replaces v1's `Client.server_capabilities`; use `client.initialize_result.capabilities` instead.
+
+### `ClientSession(protocol_version=)` removed
+
+The `protocol_version` constructor parameter on `ClientSession` has been removed. To install a known protocol version without performing the `initialize` handshake (e.g. when reconnecting to an existing session), call `session.adopt(result)` after construction with a stored `InitializeResult`.
 
 ### `McpError` renamed to `MCPError`
 
--- a/src/mcp/client/client.py
+++ b/src/mcp/client/client.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from collections.abc import Mapping
 from contextlib import AsyncExitStack
 from dataclasses import KW_ONLY, dataclass, field
-from typing import Any, Literal
+from typing import Any, Literal, TypeVar
 
 import anyio
 from typing_extensions import deprecated
@@ -48,6 +48,20 @@ ConnectMode = Literal["legacy", "auto"] | str
 initialize), or a modern protocol-version string (adopt directly). The ``str`` arm is for
 forward-compat; ``Client.__post_init__`` rejects anything outside that set at construction."""
 
+_T = TypeVar("_T")
+
+
+def _connected(value: _T | None) -> _T:
+    """Narrow a post-handshake session attribute from ``T | None`` to ``T``.
+
+    ``Client.__aenter__`` only assigns ``_session`` after the handshake succeeds, so inside
+    ``async with Client(...)`` these attributes are always populated; the ``.session`` gate
+    raises before this is reached otherwise. The guard exists for pyright, not runtime.
+    """
+    if value is None:  # pragma: no cover
+        raise RuntimeError("Client must be used within an async context manager")
+    return value
+
 
 def _synthesize_discover(protocol_version: str) -> types.DiscoverResult:
     return types.DiscoverResult(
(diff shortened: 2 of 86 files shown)


docs/migration.md
```diff
@@ -159,6 +159,10 @@ The `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters have been re
 
 Note: `sse_client` retains its `headers`, `timeout`, `sse_read_timeout`, and `auth` parameters — only the streamable HTTP transport changed.
 
+### `protocol_version` removed from `StreamableHTTPTransport` and `streamable_http_client`
```
maxisbey: was this in v1 though? migration.md is only for changes from v1 -> v2, not changes within v2

docs/migration.md
```diff
@@ -350,6 +354,10 @@ if result is not None:
 
 The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. This replaces v1's `Client.server_capabilities`; use `client.initialize_result.capabilities` instead.
 
+### `ClientSession(protocol_version=)` removed
```
maxisbey: definitely wasn't in v2, remove

docs/migration.md
```diff
@@ -764,6 +772,12 @@ async def my_tool(ctx: Context) -> str: ...
 async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 ```
 
+### Version constants
+
+`SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`.
+
+`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `HANDSHAKE_PROTOCOL_VERSIONS[-1]`.
```
maxisbey: useless, remove

src/mcp/client/client.py
```diff
@@ -95,29 +119,30 @@ async def main():
     client_info: Implementation | None = None
     """Client implementation info to send to server."""
 
-    protocol_version: str | None = None
-    """Pin the protocol version instead of negotiating it.
+    mode: Literal["legacy", "auto"] | str = "legacy"
```
maxisbey: where did "modern" go?

src/mcp/client/client.py
```diff
@@ -127,10 +152,34 @@ async def __aenter__(self) -> Client:
...
+            else:
+                if self._inproc_server is not None:
+                    transport: Transport = InMemoryTransport(
+                        self._inproc_server, raise_exceptions=self.raise_exceptions
+                    )
+                else:
+                    assert self._transport is not None
+                    transport = self._transport
+                read_stream, write_stream = await exit_stack.enter_async_context(transport)
+                session = ClientSession(
```
maxisbey: maybe just me but this looks kinda gross, is there a clean/cute way of writing this?

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+        self.adopt(result)
 
         await self.send_notification(types.InitializedNotification())
 
         return result
 
+    def adopt(self, result: types.InitializeResult | types.DiscoverResult) -> None:
+        """Install negotiated state from a result the caller already holds (no wire traffic)."""
+        if isinstance(result, types.DiscoverResult):
+            mutual = [v for v in MODERN_PROTOCOL_VERSIONS if v in result.supported_versions]
```
maxisbey, line 380: extremely minor nit: aren't there like set operations we can use here or something? but really whatever is cleanest so this is fine

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+            if not mutual:
+                raise RuntimeError(
+                    f"No mutually supported modern protocol version "
+                    f"(server: {result.supported_versions}, client: {list(MODERN_PROTOCOL_VERSIONS)})"
+                )
+            protocol_version = mutual[-1]
```
maxisbey: should probably sort this?

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
         await self.send_notification(types.InitializedNotification())
 
         return result
 
+    def adopt(self, result: types.InitializeResult | types.DiscoverResult) -> None:
+        """Install negotiated state from a result the caller already holds (no wire traffic)."""
+        if isinstance(result, types.DiscoverResult):
+            mutual = [v for v in MODERN_PROTOCOL_VERSIONS if v in result.supported_versions]
+            if not mutual:
+                raise RuntimeError(
```
maxisbey, line 382: should we use something other than RuntimeError? also we need to put all exceptions in public methods in the doc strings

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+                protocol_version=protocol_version,
+                capabilities=result.capabilities,
+                server_info=result.server_info,
+                instructions=result.instructions,
+            )
+        else:
+            self._stamp = _make_handshake_stamp(result.protocol_version)
+            self._initialize_result = result
+
+    async def discover(self) -> types.InitializeResult:
```
maxisbey: returning InitializeResult doesn't make any sense, it should return the correct type, I don't want to mix these two worlds together, they should be handled very separately.

there's a lot of code in here that mixes old handshake and new hanshake and stores them in the same place, that feels gross and easily can lead to bugs.

I'd rather handle this via clean abstrations or something if we can.

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+            }
+            opts: CallOptions = {
+                "timeout": DISCOVER_TIMEOUT_SECONDS,
+                "cancel_on_abandon": False,
+                "headers": {MCP_PROTOCOL_VERSION_HEADER: version},
+            }
+            return await self._dispatcher.send_raw_request("server/discover", params, opts)
+
+        try:
+            raw = await probe(MODERN_PROTOCOL_VERSIONS[-1])
```
maxisbey: why put this in a function

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+                    raise
+                raw = await probe(mutual[-1])
+            elif e.code in (METHOD_NOT_FOUND, REQUEST_TIMEOUT):
+                return await self.initialize()
+            else:
+                raise
+
+        result = types.DiscoverResult.model_validate(raw)
+        self.adopt(result)
+        assert self._initialize_result is not None
```
maxisbey: this is code smell, doesn't feel like we're following by construction principles stuff here

src/mcp/client/session.py
```diff
@@ -314,56 +329,133 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
             raise RuntimeError(f"Unsupported protocol version from the server: {result.protocol_version}")
 
-        self._initialize_result = result
+        self.adopt(result)
 
         await self.send_notification(types.InitializedNotification())
 
         return result
 
+    def adopt(self, result: types.InitializeResult | types.DiscoverResult) -> None:
```
maxisbey, line 368: why is this public? "adopt" isn't a normal mcp concept, this feels weird

src/mcp/client/streamable_http.py
```diff
@@ -88,41 +72,17 @@ class RequestContext:
 class StreamableHTTPTransport:
```
maxisbey, line 73: there's a few changes on this class I want us to be super careful of given that it's used by the legacy path. just make sure this is correct.

src/mcp/server/runner.py
```diff
@@ -513,3 +516,34 @@ async def serve_one(
...
+        dctx: DispatchContext[TransportContext], method: str, params: Mapping[str, Any] | None
+    ) -> dict[str, Any]:
+        meta = (params or {}).get("_meta", {})
+        connection = Connection.from_envelope(
+            meta.get(PROTOCOL_VERSION_META_KEY, MODERN_PROTOCOL_VERSIONS[-1]),
+            meta.get(CLIENT_INFO_META_KEY),
+            meta.get(CLIENT_CAPABILITIES_META_KEY),
+        )
+        # `OnRequest` is invoked for requests only, so `request_id` is always set.
+        assert dctx.request_id is not None
```
maxisbey: code smell

src/mcp/server/session.py
```diff
@@ -353,6 +353,16 @@ async def send_ping(self) -> types.EmptyResult:
             types.EmptyResult,
         )
 
+    async def report_progress(self, progress: float, total: float | None = None, message: str | None = None) -> None:
```
maxisbey, line 356: why a new one here? were we missing it before?

tests/interaction/lowlevel/test_client_connect.py
```diff
@@ -0,0 +1,362 @@
...
+    def is_internal_error(exc: MCPError) -> bool:
+        return exc.code == INTERNAL_ERROR
+
+    with anyio.fail_after(5):
+        async with mounted_app(server, on_request=on_request) as (http, _):
+            with pytest.RaisesGroup(
+                pytest.RaisesExc(MCPError, check=is_internal_error), flatten_subgroups=True
+            ):  # pragma: no branch
+                async with Client(streamable_http_client(f"{BASE_URL}/mcp", http_client=http), mode="auto"):
+                    pytest.fail("entering the Client should have raised")  # pragma: no cover
```
maxisbey: just raise not implemented

tests/interaction/lowlevel/test_progress.py
```diff
@@ -41,18 +41,9 @@ async def list_tools(
...
-        await ctx.session.send_progress_notification(
-            token, 1.0, total=3.0, message="first chunk", related_request_id=str(ctx.request_id)
-        )
-        await ctx.session.send_progress_notification(
-            token, 2.0, total=3.0, message="second chunk", related_request_id=str(ctx.request_id)
-        )
-        await ctx.session.send_progress_notification(
-            token, 3.0, total=3.0, message="done", related_request_id=str(ctx.request_id)
-        )
+        await ctx.session.report_progress(1.0, total=3.0, message="first chunk")
```
maxisbey, line 44: ServerSession is supposed to be a shim on top of v1, why is this code needing changes?

tests/issues/test_176_progress_token.py
```diff
@@ -9,13 +9,17 @@
...
+
+    Issue #176: the original Context.report_progress treated token 0 as "no token" and
+    silently dropped progress. Context now delegates unconditionally to
+    ServerSession.report_progress (which calls DispatchContext.progress, whose JSONRPC
+    implementation gates on `is None`, not truthiness), so a request whose meta carries
+    a 0-valued token still emits all three reports.
+    """
     mock_session = AsyncMock()
-    mock_session.send_progress_notification = AsyncMock()
+    mock_session.report_progress = AsyncMock()
```
maxisbey, line 21: yea why are we changing this p7ublic interface? ServerSession was supposed to be a v1 shim, I don't understand.

tests/server/mcpserver/test_server.py
```diff
@@ -1515,21 +1515,21 @@ def test_streamable_http_no_redirect() -> None:
...
     mock_session = AsyncMock()
-    mock_session.send_progress_notification = AsyncMock()
+    mock_session.report_progress = AsyncMock()
 
     request_context = ServerRequestContext(
         request_id="req-abc-123",
         session=mock_session,
         method="tools/call",
-        meta={"progress_token": "tok-1"},
+        meta=None,
```
maxisbey, line 1532: why is this changing now?

src/mcp/client/client.py
```diff
@@ -95,56 +127,99 @@ async def main():
...
+        dispatcher: Dispatcher[Any] | None
+        if self._inproc_server is not None and self.mode != "legacy":
+            # Modern in-process path: drive the server through a DirectDispatcher peer-pair
+            # with one `serve_one` per request — no streams, no initialize handshake.
+            lifespan_state = await exit_stack.enter_async_context(self._inproc_server.lifespan(self._inproc_server))
+            client_disp, server_disp = create_direct_dispatcher_pair()
+            tg = await exit_stack.enter_async_context(anyio.create_task_group())
+            exit_stack.callback(server_disp.close)
+            on_request = modern_on_request(self._inproc_server, lifespan_state, raise_exceptions=self.raise_exceptions)
+            await tg.start(server_disp.run, on_request, _drop_notify)
```
maxisbey: but I thought notifications could still come back on the response of a post?

src/mcp/client/client.py
```diff
@@ -170,21 +245,39 @@ def session(self) -> ClientSession:
         return self._session
 
     @property
-    def initialize_result(self) -> InitializeResult:
-        """The server's InitializeResult.
+    def protocol_version(self) -> str:
+        """Negotiated protocol version (set by initialize/discover/adopt during ``__aenter__``)."""
+        version = self.session.protocol_version
+        assert version is not None
```
maxisbey: these asserts are terrible, lets discuss

src/mcp/client/client.py
```diff
@@ -95,56 +127,99 @@ async def main():
...
+            self._inproc_server = self.server._lowlevel_server  # pyright: ignore[reportPrivateUsage]
+        elif isinstance(self.server, Server):
+            self._inproc_server = self.server
         elif isinstance(self.server, str):
-            self._transport = streamable_http_client(self.server, protocol_version=self.protocol_version)
+            self._transport = streamable_http_client(self.server)
         else:
             self._transport = self.server
 
+        if self.mode not in ("legacy", "auto") and self.mode not in MODERN_PROTOCOL_VERSIONS:
```
maxisbey, line 206: is this the same as typescript?

docs/migration.md
maxisbey: Rewrote this section in cdfdfd1 to cover only the `StreamableHTTPTransport.protocol_version` attribute (which was on v1.x); dropped the constructor-param half (added on main in #2917, never on v1.x).

docs/migration.md
maxisbey: Removed in cdfdfd1 — the `ClientSession(protocol_version=)` param was added on main in #2917 and never shipped on v1.x.

docs/migration.md
maxisbey: Kept — both `LATEST_PROTOCOL_VERSION` and `SUPPORTED_PROTOCOL_VERSIONS` exist on v1.x (`types.py:27` and `shared/version.py:3` respectively), so the value/semantics change is a real v1→v2 break. Happy to drop if you'd rather not document it, but it's not within-v2 churn.

src/mcp/client/client.py
maxisbey: There never was a `mode='modern'` — the grill ruled `'legacy' | 'auto' | <version-string>`. "modern" is ambiguous (which modern version?). The docstring that mentioned in-memory not having a modern entry was attached to the deleted `protocol_version` field's docstring; that limitation is now gone (`modern_on_request` driver landed in 46c0742).

src/mcp/client/client.py
maxisbey: Extracted `_build_session()` in a82042a so there's one `ClientSession()` call site and `__aenter__` reads linear: build → enter → mode-dispatch. The deeper refactor (resolve a single connector closure in `__post_init__`, dropping the `Optional` fields entirely) is filed for a follow-up alongside the public-surface pass.

src/mcp/client/session.py
maxisbey: Kept the list-comp — set intersection loses ordering, and `MODERN_PROTOCOL_VERSIONS` is the canonical oldest→newest order (string-sorting versions is explicitly wrong per `shared/version.py`). Added a comment saying so.

src/mcp/client/session.py
maxisbey: No sort — `mutual` inherits `MODERN_PROTOCOL_VERSIONS` order (oldest→newest), so `mutual[-1]` is the newest mutual. String-sorting protocol versions would be wrong. Comment added in a82042a.

src/mcp/client/session.py
maxisbey: Added `Raises:` sections to `adopt()` and `discover()` docstrings in a82042a. A dedicated negotiation-failure exception type (instead of `RuntimeError`) is filed for the public-surface pass — agree it's the right shape, just didn't want to add a new public exception class drive-by.

src/mcp/client/session.py
maxisbey: Agreed and reworked in a82042a: `discover() -> DiscoverResult`, no synthesis. Separate `_discover_result` / `_initialize_result` slots; era-neutral accessors (`.server_info` / `.server_capabilities` / `.instructions` / `.protocol_version`) read from whichever is set. The fallback ladder moved to `Client.__aenter__` (policy layer). `Client.initialize_result` removed in favor of the accessors.

src/mcp/client/session.py
maxisbey: It's called twice — once for the optimistic probe, once for the `-32022` retry with the intersected version. The closure captures the pre-dumped client_info/caps so the retry doesn't recompute them. Inlining would duplicate ~10 lines.

src/mcp/client/session.py
maxisbey: Gone in a82042a — that was the `assert self._initialize_result is not None` after `adopt()` synthesized one. With separate slots and `discover()` returning the `DiscoverResult` directly, the assert vanished.

src/mcp/client/session.py
maxisbey: It's public so a lowlevel `ClientSession` user can install a cached `DiscoverResult` (or `InitializeResult`) without re-probing — that's the zero-RTT reconnect case `prior_discover=` enables on `Client`, but `ClientSession` users need a way to do it directly. The grill made `.initialize()`/`.discover()`/`.adopt()` siblings deliberately. The *name* I'm not attached to — `install`/`apply` are both fine if "adopt" reads odd; but the publicness is intentional. Leaving open in case you want a rename.

src/mcp/client/streamable_http.py
maxisbey: Traced end to end — `mode='legacy'` is byte-identical: `_base_headers` (POST) keeps accept/content-type/session-id; pv header arrives via the handshake stamp's `metadata.headers` after `notifications/initialized`; `_prepare_headers` (GET/DELETE) adds the cached pv. The split in b6be755 also fixes a real bug where the discover probe's pv header would leak onto a fallback `initialize` POST. Two new tests pin the legacy-path contract.

src/mcp/server/runner.py
maxisbey: Reshaped in b6be755: `serve_one(server, dctx, method, params, *, ..., raise_exceptions)`. The `JSONRPCRequest` round-trip and the `request_id is not None` assert in `modern_on_request` are gone. Also threads `raise_exceptions` through so the modern in-process path honors it (handler exceptions chain via `__cause__`).

src/mcp/server/session.py
maxisbey: We were missing it, yes — and the gap was real. `Context.report_progress` previously read `meta["progress_token"]` and called `send_progress_notification(token, ...)` → `dctx.notify("notifications/progress")`. That's JSONRPC-specific: on the in-process modern path (`DirectDispatcher`) there *is* no wire token; progress is the caller's `on_progress` callback held on the dctx. So `ctx.report_progress()` silently no-op'd there.

This method routes through `dctx.progress()` — the dispatcher-agnostic seam that already existed but was never reached from `Context`. `send_progress_notification` is unchanged (byte-identical), this is additive, and the constructor type narrowing (`Outbound→DispatchContext[Any]`) just exposes `.progress()` (the runtime arg was always a `DispatchContext`).

I tried reverting the test changes in [17] to prove they were optional — they're not: the old `meta["progress_token"]` pattern fails on `[in-memory-2026-07-28]`. Added a migration.md note ("prefer `ctx.report_progress()`; manual token-read is JSONRPC-specific") in cdfdfd1.

tests/interaction/lowlevel/test_client_connect.py
maxisbey: Done in cdfdfd1.

tests/interaction/lowlevel/test_progress.py
maxisbey: See reply on the `ServerSession.report_progress` thread — these handler changes turned out to be load-bearing for `[in-memory-2026-07-28]` (no wire token there), not opportunistic. The old `send_progress_notification` still works on JSONRPC transports.

tests/issues/test_176_progress_token.py
maxisbey: No public interface changed — `send_progress_notification` is byte-identical. This test mocks the *internal* method `Context.report_progress` calls, and that callee changed from `send_progress_notification` to the new `report_progress` (see the `ServerSession` thread). White-box mock-seam follow, not an API change.

tests/server/mcpserver/test_server.py
maxisbey: Same as test_176 — the test mocks the method `Context.report_progress` delegates to, which changed. The test's intent ("report_progress reaches the session") is preserved; the mock target moved.

src/mcp/client/client.py
maxisbey: specifically, I think we should just make these optional

src/mcp/server/runner.py
```diff
@@ -494,22 +502,67 @@ async def serve_loop(
...
-    request: JSONRPCRequest,
+    dctx: DispatchContext[TransportContext],
+    method: str,
+    params: Mapping[str, Any] | None,
     *,
     connection: Connection,
-    dctx: DispatchContext[TransportContext],
     lifespan_state: LifespanT,
+    raise_exceptions: bool = False,
 ) -> JSONRPCResponse | JSONRPCError:
```
maxisbey: wait why jsonrpc here? this kinda defeats the idea of ServerRunner being transport/jronsprc agnostic

tests/interaction/lowlevel/test_client_connect.py
```diff
@@ -0,0 +1,362 @@
...
+    ServerCapabilities,
+    ToolsCapability,
+)
+from tests.interaction._connect import BASE_URL, Connect, mounted_app
+from tests.interaction._helpers import RecordingTransport
+from tests.interaction._requirements import requirement
+
+pytestmark = pytest.mark.anyio
+
+MODERN_VERSION = "2026-07-28"
```
maxisbey: can't this be derived from elsewhere rather than re-defining it here?

tests/interaction/lowlevel/test_client_connect.py
```diff
@@ -0,0 +1,362 @@
...
+
+    async def on_request(request: httpx.Request) -> None:
+        captured.append(request)
+
+    return captured, on_request
+
+
+@requirement("lifecycle:mode:legacy-never-probes")
+async def test_legacy_mode_sends_initialize_and_never_probes_discover() -> None:
+    """`Client(server)` (mode defaults to 'legacy') opens with `initialize` and never sends `server/discover`.
```
maxisbey: I think we want Client to default to "auto"

docs/migration.md
```diff
@@ -340,15 +344,15 @@ capabilities = session.get_server_capabilities()
...
 ```python
-result = session.initialize_result
-if result is not None:
-    capabilities = result.capabilities
-    server_info = result.server_info
-    instructions = result.instructions
-    version = result.protocol_version
+capabilities = session.server_capabilities
+server_info = session.server_info
+instructions = session.instructions
```
maxisbey, line 351: what do these result in on a 2025 unofficial stateless connection?

src/mcp/server/session.py
maxisbey: 👍 (resolving per discussion)

src/mcp/client/client.py
maxisbey: It's the other direction — that arg is the server's *inbound* notify handler, sinking client→server notifications, of which 2026 defines zero (initialized/roots-list-changed removed; cancel is structural). Server→client (progress, logging) flows back via the per-request `dctx` and is delivered. Renamed to `_no_inbound_client_notifications` with a docstring saying so in bdcdeb0.

src/mcp/client/client.py
maxisbey: Swapped to a `_connected()` helper that raises the same `RuntimeError("Client must be used within an async context manager")` as `.session` (bdcdeb0) — pushed back gently on Optional since the None is unreachable inside `async with` and would force ~19 chained call sites to guard a state they can't hit. Added a TODO above the block for the connected-view-type shape. Also fixed the `__aenter__` ordering (publish `_session` only after the handshake) and made the re-entry guard an explicit `_entered` flag so the one-shot semantic doesn't depend on a leaked stale reference.

src/mcp/client/client.py
maxisbey: Yes — TS's `resolveVersionNegotiation` (versionNegotiation.ts:132-137) throws `TypeError` on a pre-2026 pin with the same "use legacy" steer. Three deliberate deltas: we validate at construction (TS at `connect()`); we use set-membership in `MODERN_PROTOCOL_VERSIONS` (TS uses lexical `>=`, which our `version.py` doctrine says not to do); and our pin is zero-RTT `.adopt()` whereas TS's pin still probes. go-sdk has no public knob.

src/mcp/server/runner.py
maxisbey: Reshaped in bdcdeb0: `serve_one(...) -> dict[str, Any]`, exceptions propagate. The HTTP entry now composes `to_jsonrpc_response(req.id, serve_one(...))` itself; `modern_on_request` collapses to a direct `await serve_one(...)` plus the `raise_exceptions` sanitize guard (no more `JSONRPCError` round-trip). The `assert dctx.request_id is not None` drops too. Kernel was already agnostic — this finishes the input-side reshape from b6be755 on the output side.

tests/interaction/lowlevel/test_client_connect.py
maxisbey: Agreed, and went further per the discussion — b326347 adds `LATEST_HANDSHAKE_VERSION` / `LATEST_MODERN_VERSION` / `OLDEST_SUPPORTED_VERSION` to `shared/version.py` (derived from the tuples) and sweeps the `[-1]`/`[0]` indexing repo-wide. `LATEST_PROTOCOL_VERSION` is now derived there too instead of a duplicate literal in `types/_types.py`. This line is now `MODERN_VERSION = LATEST_MODERN_VERSION`.

tests/interaction/lowlevel/test_client_connect.py
maxisbey: Left a `# TODO(maxisbey)` at the field per discussion (bdcdeb0) — the flip stays for the public-surface checkpoint along with making the timeout-fallback transport-aware and decoupling the in-proc test suite. This test now passes `mode='legacy'` explicitly so it asserts what `'legacy'` does, not what the default is, and survives the flip untouched.

docs/migration.md
maxisbey: Same as any legacy connection — the client still does the `initialize` handshake and gets a real `InitializeResult` back, so `session.server_info` / `.server_capabilities` / `.protocol_version` are populated from that. The server being `stateless_http=True` is invisible at this layer (it changes session-id persistence on the server side, not the handshake shape the client sees).

docs/migration.md
```diff
@@ -340,15 +344,15 @@ capabilities = session.get_server_capabilities()
...
+version = session.protocol_version
 ```
 
-The high-level `Client.initialize_result` returns the same `InitializeResult` but is non-nullable — initialization is guaranteed inside the context manager, so no `None` check is needed. This replaces v1's `Client.server_capabilities`; use `client.initialize_result.capabilities` instead.
+The raw handshake result is also retained as `session.initialize_result` (legacy path) or `session.discover_result` (modern path) — exactly one is non-`None`.
```
maxisbey: this is confusing, why would discover_result be populated if we're in 2025 world

docs/migration.md
```diff
@@ -340,15 +344,15 @@ capabilities = session.get_server_capabilities()
...
+
+On the high-level `Client`, `client.server_capabilities`, `client.server_info`, and `client.protocol_version` are non-nullable inside the context manager. `client.instructions` remains `str | None` since the server may omit it.
```
maxisbey: how does that work with a 2025 unofficial stateless server tho, previously initialization was optional, now it's now

docs/migration.md
```diff
@@ -764,6 +768,12 @@ async def my_tool(ctx: Context) -> str: ...
 async def my_tool(ctx: Context[MyLifespanState]) -> str: ...
 ```
 
+### Version constants
+
+`SUPPORTED_PROTOCOL_VERSIONS` is deprecated — it's now the union of `HANDSHAKE_PROTOCOL_VERSIONS` (initialize-handshake versions) and `MODERN_PROTOCOL_VERSIONS` (per-request-envelope versions). If you were using it to mean "versions the initialize handshake accepts", switch to `HANDSHAKE_PROTOCOL_VERSIONS`. Named scalars derived from these tuples are now exported alongside them — `LATEST_HANDSHAKE_VERSION`, `LATEST_MODERN_VERSION`, `OLDEST_SUPPORTED_VERSION` — so prefer those over indexing the tuples directly.
+
+`LATEST_PROTOCOL_VERSION` now reflects the newest protocol revision the SDK supports (`2026-07-28`). Code that used it to mean "the version `.initialize()` offers" should switch to `LATEST_HANDSHAKE_VERSION`.
```
maxisbey: remove this line, pointless

src/mcp/client/client.py
```diff
@@ -95,58 +144,108 @@ async def main():
...
     def __post_init__(self) -> None:
-        if isinstance(self.server, Server | MCPServer):
-            self._transport = InMemoryTransport(self.server, raise_exceptions=self.raise_exceptions)
+        if isinstance(self.server, MCPServer):
+            self._inproc_server = self.server._lowlevel_server  # pyright: ignore[reportPrivateUsage]
+        elif isinstance(self.server, Server):
+            self._inproc_server = self.server
         elif isinstance(self.server, str):
-            self._transport = streamable_http_client(self.server, protocol_version=self.protocol_version)
+            self._transport = streamable_http_client(self.server)
```
maxisbey: why do we need to have an inproc_server field? could we not generalize Transport to this new world?

src/mcp/server/runner.py
```diff
@@ -180,20 +182,25 @@ async def aclose_shielded(connection: Connection) -> None:
         )
 
 
-async def to_jsonrpc_response(request_id: RequestId, coro: Awaitable[dict[str, Any]]) -> JSONRPCResponse | JSONRPCError:
+async def to_jsonrpc_response(
```
maxisbey: probably shouldn't be in this file anymore? what is it even used by?

src/mcp/types/_types.py
```diff
@@ -20,14 +20,9 @@
 from pydantic.alias_generators import to_camel
 from typing_extensions import NotRequired, TypedDict
 
+from mcp.shared.version import LATEST_PROTOCOL_VERSION as LATEST_PROTOCOL_VERSION
```
maxisbey: it's already called that

tests/shared/test_inbound.py
```diff
@@ -34,7 +33,7 @@
     UNSUPPORTED_PROTOCOL_VERSION,
 )
 
-MODERN = MODERN_PROTOCOL_VERSIONS[0]
+MODERN = LATEST_MODERN_VERSION
```
maxisbey: why? just use LATEST_MODERN_VERSION? remove MODERN?

docs/migration.md
maxisbey: It wouldn't be — for a 2025 server `initialize_result` is the slot that gets set; `discover_result` is the forward-looking half of the pair for when the same code points at a 2026+ server. Reworded in 78823d4 to name which slot each era fills, and tightened "exactly one" → "at most one" (both are `None` pre-handshake on the lowlevel session).

docs/migration.md
maxisbey: It's still optional at the lowlevel `ClientSession` layer — v2 didn't add a client-side gate, so the skip-init path works exactly as in v1. This line is about `Client`, which is new in v2 (no v1 equivalent), so "previously" doesn't really apply to it; against a `stateless_http=True` server `Client.__aenter__()` just sends `initialize`, the server answers it (the gate is pre-opened, not removed), and the accessors populate. Added a parenthetical in 78823d4 making the layer split explicit.

docs/migration.md
maxisbey: Agreed — deleted in 78823d4. A constant named `LATEST_*` getting bumped is the constant doing what its name says; the `SUPPORTED_PROTOCOL_VERSIONS` paragraph above already covers the actual migration trap.

src/mcp/client/client.py
maxisbey: Agreed it was scaffolding. I spiked both fixes end-to-end: collapsing to a single `_connect` closure resolved in `__post_init__`, vs generalizing `Transport` to `AbstractAsyncContextManager[Dispatcher]` with a proper `DirectInProcessTransport`. Landed the closure approach in f9a15e1 — it removes the field, the assert, and the `_build_session` branching with zero export changes and the full suite green; the `Transport→Dispatcher` generalization is the right long-term shape but re-signatures the public `Transport` Protocol and ripples through `RecordingTransport` and the stream-pair `ClientSession` ctor, so I've ledgered it as a follow-up for the public-surface checkpoint rather than land it here.

src/mcp/server/runner.py
maxisbey: Only `_streamable_http_modern.py` was calling it. Moved there as private `_to_jsonrpc_response` in 78823d4 and dropped the dead `raise_unhandled=` param; `runner.py` is now JSON-RPC-wire-model-free (sheds the `JSONRPCResponse`/`JSONRPCError`/`RequestId`/`handler_exception_to_error_data` imports). The ladder tests moved to `test_streamable_http_modern.py`.

src/mcp/types/_types.py
maxisbey: The `as X` self-alias is the PEP 484 re-export marker — ruff F401 and pyright `reportUnusedImport` both fire without it since nothing in the file body references the name. That said, the bounce itself is pointless: in 78823d4 `types/__init__.py` now imports `LATEST_PROTOCOL_VERSION` directly from `mcp.shared.version` and this line is deleted.

tests/shared/test_inbound.py
maxisbey: Agreed — dropped and inlined `LATEST_MODERN_VERSION` in 78823d4. Also swept `test_connection.py` (`_MODERN`) and `test_client_connect.py` (`MODERN_VERSION`), which had each invented their own local name for the same constant.

src/mcp/client/streamable_http.py
```diff
@@ -299,9 +249,12 @@ async def _handle_resumption_request(self, ctx: RequestContext) -> None:
...
         """Handle a POST request with response processing."""
-        headers = self._prepare_headers()
+        headers = self._base_headers()
         message = ctx.session_message.message
-        headers.update(self._per_message_headers(message))
+        if ctx.metadata is not None and ctx.metadata.headers is not None:
+            headers.update(ctx.metadata.headers)
+            if MCP_PROTOCOL_VERSION_HEADER in ctx.metadata.headers:
+                self._protocol_version_header = ctx.metadata.headers[MCP_PROTOCOL_VERSION_HEADER]
         is_initialization = self._is_initialization_request(message)
```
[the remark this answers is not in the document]
maxisbey: Good catch — confirmed and fixed in eab740b. `_handle_post_request` now clears the cached `MCP-Protocol-Version` when an `initialize` goes out, then lets every POST read the cache via `_prepare_headers()` again (so `_base_headers` is gone). Dispatcher-written response/error/cancelled POSTs pick up the negotiated version from the cache that `notifications/initialized` re-seeds; the fallback `initialize` still goes out header-less. Added a four-step header test pinning the whole sequence.

src/mcp/client/client.py
```diff
@@ -95,58 +144,108 @@ async def main():
...
+            if self.mode == "legacy":
+                await session.initialize()
+            elif self.mode == "auto":
+                try:
+                    await session.discover()
+                except MCPError as e:
+                    if e.code in (METHOD_NOT_FOUND, REQUEST_TIMEOUT):
+                        await session.initialize()
+                    else:
+                        raise
```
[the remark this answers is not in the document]
maxisbey: Confirmed against `origin/v1.x` (stateful 400+"Missing session ID", stateless 400+"Unsupported protocol version" — both `INVALID_REQUEST`). Fixed in eab740b by adding `INVALID_REQUEST` to the fallback set; the `lifecycle:discover:network-error-raises` requirement text and the scripted-transport fallback test are updated to cover both rejection codes. The proper inversion to a denylist ("fall back on every rpc-error/4xx that isn't a recognised modern error") is filed as a follow-up alongside making the timeout verdict transport-aware.

src/mcp/client/client.py
```diff
@@ -95,58 +183,89 @@ async def main():
...
+                    await session.discover()
+                except MCPError as e:
+                    # TODO(L73): invert this allowlist into a `classify_probe_outcome` denylist —
+                    # fall back on every rpc-error/4xx that isn't a recognized modern error.
+                    if e.code in (METHOD_NOT_FOUND, INVALID_REQUEST, REQUEST_TIMEOUT):
+                        await session.initialize()
+                    else:
+                        raise
+            else:
+                session.adopt(self.prior_discover or _synthesize_discover(self.mode))
```
maxisbey, line 255: why do we need to synthesize discover? we should probably just handle the case properly.

like there's two distinct states:
- calling a tool without first calling discover
- first calling discover and then calling tool

but if we synthesize discover then the ClientSession will think it _has_ called discover yet it hasn't, need to think about the implications to client represation here

src/mcp/client/session.py
```diff
@@ -314,56 +343,166 @@ def _build_capabilities(self) -> types.ClientCapabilities:
...
+                    CLIENT_INFO_META_KEY: client_info,
+                    CLIENT_CAPABILITIES_META_KEY: capabilities,
+                }
+            }
+            opts: CallOptions = {
+                "timeout": DISCOVER_TIMEOUT_SECONDS,
+                "cancel_on_abandon": False,
+                "headers": {MCP_PROTOCOL_VERSION_HEADER: version},
+            }
+            return await self._dispatcher.send_raw_request("server/discover", params, opts)
```
maxisbey: why not use the actual types here?

src/mcp/shared/jsonrpc_dispatcher.py
```diff
@@ -132,11 +132,11 @@ def request_id(self) -> RequestId | None:
     def can_send_request(self) -> bool:
         return self.transport.can_send_request and not self._closed
 
-    async def notify(self, method: str, params: Mapping[str, Any] | None) -> None:
+    async def notify(self, method: str, params: Mapping[str, Any] | None, opts: CallOptions | None = None) -> None:
```
Kludex, line 135: call it `options: ` please

Kludex (approved):

PR #2951, opened 2026-06-22
Add experimental Server Cards support (SEP-2127)

SamMorrowDrums:
Closes https://github.com/modelcontextprotocol/experimental-ext-server-card/issues/16

## Summary

Experimental SDK support for MCP Server Cards (SEP-2127) and AI Catalog discovery, opt-in under `mcp.{shared,server,client}.experimental`. As an experimental extension, these APIs may change without notice.

Takes over #2696 (thanks @dsp-ant — the original two commits are preserved with authorship), rebased cleanly onto current `main`. #2696 was stacked on the Tasks (SEP-1686) work, since removed from `main` (#2714).

Opening as a **draft** for maintainer review — per the extension repo's contribution rules, AI-assisted changes are not self-merged.

## What it does

A server builds a Server Card from its own identity, serves it from a Starlette app, and advertises it through an AI Catalog. A client discovers a host's catalog, then fetches and validates the cards it references. Cards describe identity and remote (HTTP) transport only — locally-runnable package metadata stays in the MCP Registry's `server.json`.

## Conformance to the current schemas

- **`$schema`** is pinned to `https://static.modelcontextprotocol.io/schemas/v1/server-card.schema.json`; a card pointing at the registry's `server.schema.json` is rejected. A document that omits `$schema` is defaulted to the v1 URL rather than rejected.
- **AI Catalog entries** use the serialized `type` field, require `specVersion`, and accept optional `displayName`. `server_card_entry()` emits the minimal MCP shape (`identifier`, `type`, and `url`) so human-readable fields cannot drift from the referenced card.
- **Media type** — serving and catalog entries use `application/mcp-server-card+json`; the client recognizes only this type and skips catalog entries declaring anything else.
- **Catalog identifiers** are `urn:air:{publisher}:mcp:{name}` — the card name's reverse-DNS namespace turned back into the publisher's forward-DNS domain (`com.example/weather` → `urn:air:example.com:mcp:weather`).
- **Discovery** uses the AI Catalog at `/.well-known/ai-catalog.json`. Discovery responses carry the CORS headers the spec requires (MUST), `Cache-Control` (SHOULD), strong SHA-256 ETags, and `If-None-Match` handling for conditional requests.
- **`version`** accepts exact versions only; ranges and wildcards (`^1.2.3`, `~1.2.3`, `>=1.2.3`, `1.x`, `1 || 2`, `1 - 2`) are rejected.
- **Unsettled AI Catalog metadata** remains available but is documented as unstable pending the outcome of Agent-Card/ai-catalog#77.

## Verification

- `./scripts/test` — full coverage suite and `strict-no-cover`
- `pre-commit run --all-files` — formatting, lint, pyright, markdownlint, lock consistency, and README snippets
- Focused Server Card and AI Catalog tests cover URL and inline discovery, entry filtering, relative URLs, CORS on both `200` and `304`, ETags, validation, and client/server helpers

🤖 The implementation and cleanup were AI-assisted and human-directed.


src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,257 @@
...
+
+    MCP server entries use ``urn:air:{publisher}:{name}``, where ``publisher`` is
+    the forward-DNS form of the referenced Server Card's namespace and ``name``
+    is its name suffix.
+    """
+
+    display_name: str
+    """Human-readable name for the artifact."""
+
+    media_type: str
```
SamMorrowDrums: Note for reviewers: this field stays `mediaType` (its serialized alias) to match the currently merged discovery spec.

The extension repo's PR modelcontextprotocol/experimental-ext-server-card#32 (ADR 0014) renames the Catalog Entry `mediaType` member to `type`, but a maintainer has it on hold pending the upstream AI Catalog PR. So I'm intentionally **not** renaming this yet — I'll flip the alias here (and the client-side check in `client/experimental/server_card.py`) to `type` once #32 lands, rather than pre-empting an unresolved spec change. The value itself (`application/mcp-server-card+json`) is unaffected by that rename.

src/mcp/server/experimental/ai_catalog.py
```diff
@@ -0,0 +1,130 @@
...
+
+def discovery_response(request: Request, body: bytes, media_type: str) -> Response:
+    etag = _strong_etag(body)
+    if _if_none_match_matches(request.headers.get("if-none-match"), etag):
+        return Response(
+            status_code=304,
+            headers={
+                "ETag": etag,
+                "Cache-Control": DISCOVERY_HEADERS["Cache-Control"],
+            },
```
tadasant: **The 304 branch drops the CORS headers.**

This `headers` dict returns only `ETag` + `Cache-Control`, omitting the three `Access-Control-Allow-*` headers that the 200 branch sets via `{**DISCOVERY_HEADERS, "ETag": etag}`. A browser doing a cross-origin conditional GET (`If-None-Match`) gets a 304 with no `Access-Control-Allow-Origin`, so the revalidation is blocked by CORS — defeating the very browser support those headers exist for. Non-browser (e.g. `httpx`) clients are unaffected, which is why no test caught it.

Mirroring the 200 branch fixes it (`DISCOVERY_HEADERS` already includes `Cache-Control`):

```suggestion
            headers={**DISCOVERY_HEADERS, "ETag": etag},
```

It'd also be worth a test asserting the CORS headers are present on a 304 response, since the existing `httpx`-based tests can't surface this.

---
<sub>🤖 Posted via Claude Code (automated review assistant), on behalf of @tadasant.</sub>

SamMorrowDrums: Oh dang, good catch! I think cross-origin allowed might be *slightly* important here 😬

😅

src/mcp/server/experimental/ai_catalog.py
```diff
@@ -0,0 +1,127 @@
...
+DISCOVERY_HEADERS = {
+    "Access-Control-Allow-Origin": "*",
+    "Access-Control-Allow-Methods": "GET",
+    "Access-Control-Allow-Headers": "Content-Type",
+    "Cache-Control": "public, max-age=3600",
+}
+
+
+def _strong_etag(body: bytes) -> str:
+    return f'"{hashlib.sha256(body).hexdigest()}"'
```
SamMorrowDrums: Note to self: inline this, unless it's called more than 3 times

src/mcp/server/experimental/ai_catalog.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def _if_none_match_matches(if_none_match: str | None, etag: str) -> bool:
```
SamMorrowDrums, line 54: Note to self: doesn't the web framework surrounding this support etag matching without a self-implementation.

tadasant (commented):
Did a pass at an automated review to ensure alignment with some of the latest Server Card changes --

src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,256 @@
...
```
tadasant: **`#32` has landed — this rename is now unblocked.**

Your note above says you're holding `mediaType` until [experimental-ext-server-card#32](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/32) (ADR 0014) merges. That landed — it's commit `3b2d974` on `main`, and the published spec at <https://ai-catalog.io/spec/> now reads: a Catalog Entry MUST contain `identifier`, **`type`**, and exactly one of `url`/`data`.

Worth flagging that this isn't only a serialization nit. `MCPModel` sets `alias_generator=to_camel`, so `media_type` goes out as `mediaType` — and this field is *required* on the read path. A spec-conformant catalog emitting `type` fails validation, so `discover_server_cards` returns nothing rather than degrading. Since the alias diverges from the attribute name here, `serialization_alias`/`validation_alias` on the field is probably cleaner than renaming the attribute.

Second call site is `client/experimental/server_card.py:94`, as you noted.

src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,256 @@
...
+
+    identifier: str
+    """Identifier for the artifact; SHOULD be a URN or URI.
+
+    MCP server entries use ``urn:air:{publisher}:{name}``, where ``publisher`` is
+    the forward-DNS form of the referenced Server Card's namespace and ``name``
+    is its name suffix.
+    """
+
+    display_name: str
```
tadasant: **`displayName` is optional in the spec; this makes it required.**

`display_name: str` with no default means you can't construct a minimal conformant entry. The spec lists it under the OPTIONAL members (["The following members are OPTIONAL: `displayName`…"](https://ai-catalog.io/spec/)), per ADR-0016 — and the extension repo went further and dropped it from entries entirely in [#39](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/39).

Suggest `display_name: str | None = None`. Also affects `server_card_entry()` in `server/experimental/ai_catalog.py`, which currently always populates it from `card.title or card.name` — the fallback chain the spec defines is entry `displayName` → the artifact's own name (`title` on a Server Card) → trailing segment of the `identifier`, so a client can render without it.

src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,256 @@
...
+#: Media type identifying an AI Catalog document.
+AI_CATALOG_MEDIA_TYPE = "application/ai-catalog+json"
+#: Media type identifying an MCP Server Card artifact in a catalog entry,
+#: per the MCP discovery extension.
+MCP_SERVER_CARD_MEDIA_TYPE = "application/mcp-server-card+json"
+#: Well-known path an AI Catalog is published at, relative to the host root.
+AI_CATALOG_WELL_KNOWN_PATH = "/.well-known/ai-catalog.json"
+#: Well-known path of the MCP-scoped catalog defined by the MCP discovery
+#: extension. A structural subset of an AI Catalog, so it parses with these models.
+MCP_CATALOG_WELL_KNOWN_PATH = "/.well-known/mcp/catalog.json"
```
tadasant: **The MCP Catalog fallback is on its way out.**

`MCP_CATALOG_WELL_KNOWN_PATH` (and the 404 fallback that consumes it at `client/experimental/server_card.py:89`) targets `/.well-known/mcp/catalog.json`, which your own [experimental-ext-server-card#42](https://github.com/modelcontextprotocol/experimental-ext-server-card/pull/42) removes — it `Closes #26` and rewrites `discovery.md` to use the AI Catalog only.

src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,256 @@
...
+
+    trust_manifest: TrustManifest | None = None
+    """Trust metadata for this artifact; its ``identity`` must equal ``identifier``."""
+
+    updated_at: datetime | None = None
+    """When this entry was last modified."""
+
+    metadata: dict[str, Any] | None = None
+    """Open map for custom or non-standard metadata."""
+
```
tadasant, line 224: **`metadata` is the field most likely to churn.**

[ai-catalog#77](https://github.com/Agent-Card/ai-catalog/pull/77) renames `metadata` → `extension`, and the map-vs-array shape is still open (an array admits the same extension twice, which is a validation layer a map doesn't need). Consensus in the 7/9 and 7/16 TSC calls was leaning toward dropping unstructured `metadata` entirely, on the grounds that a parser can't distinguish an extension URI from an arbitrary blob.

No change needed today — flagging because this is public API surface, and if it ships as `metadata` it may need a rename shortly after. A docstring note that the field is unstable would be cheap insurance.

src/mcp/shared/experimental/ai_catalog/types.py
SamMorrowDrums: The upstream rename has landed, so this now validates and serializes the wire field as `type` while retaining `media_type` as the Python attribute.

src/mcp/server/experimental/ai_catalog.py
SamMorrowDrums: Inlined the SHA-256 ETag construction at its only call site.

src/mcp/server/experimental/ai_catalog.py
SamMorrowDrums: Checked the Starlette API before keeping this. Generic `Response` has no conditional-ETag support; the only helper is coupled to `StaticFiles`, so the small response-agnostic matcher remains.

src/mcp/shared/experimental/ai_catalog/types.py
SamMorrowDrums: Updated `CatalogEntry.media_type` with `validation_alias="type"` and `serialization_alias="type"`; the client keeps the Python attribute while current catalogs round-trip the `type` wire field.

src/mcp/shared/experimental/ai_catalog/types.py
SamMorrowDrums: Made `display_name` optional. `server_card_entry()` now emits the merged minimal MCP shape and omits display name, description, and version so those values cannot drift from the Server Card.

src/mcp/shared/experimental/ai_catalog/types.py
SamMorrowDrums: Removed the MCP-catalog constant, 404 fallback, associated tests, and documentation. Discovery now uses only `/.well-known/ai-catalog.json`.

src/mcp/shared/experimental/ai_catalog/types.py
SamMorrowDrums: Kept `metadata` in line with the currently merged AI Catalog spec, but marked each exposed field as unstable and likely to move to structured extensions pending Agent-Card/ai-catalog#77.

docs_src/server_cards/tutorial002.py
```diff
@@ -0,0 +1,32 @@
...
+
+# `by_alias=True` emits the wire names (`$schema`, `_meta`, `type`);
+# `exclude_none=True` drops unset optional fields.
+card_json = card.model_dump_json(by_alias=True, exclude_none=True)
+catalog_json = catalog.model_dump_json(by_alias=True, exclude_none=True)
+
+
+def write_static_site(directory: Path) -> None:
+    """Write the card and the well-known catalog under `directory`."""
+    (directory / "server-card.json").write_text(card_json)
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed — `write_static_site` now calls `directory.mkdir(parents=True, exist_ok=True)` before writing `server-card.json`, so publishing to a fresh output directory no longer raises `FileNotFoundError`.

docs_src/server_cards/tutorial001.py
```diff
@@ -0,0 +1,36 @@
...
+    server,
+    name="com.example/dice-roller",
+    remotes=[Remote(type="streamable-http", url="https://dice.example.com/mcp")],
+    repository=Repository(url="https://github.com/example/dice", source="github"),
+)
+
+# Serve the card next to the MCP endpoint, and advertise it in the host's AI
+# Catalog at `/.well-known/ai-catalog.json`. The catalog entry points at the
+# absolute URL the card is served from.
+app = server.streamable_http_app()
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed — the example now configures `TransportSecuritySettings(allowed_hosts=["dice.example.com", "dice.example.com:*"], allowed_origins=["https://dice.example.com"])` and passes it to `streamable_http_app`, so the advertised public host is accepted instead of being rejected with `421 Misdirected Request` by the default localhost-scoped DNS-rebinding protection.

docs/advanced/server-cards.md
```diff
@@ -0,0 +1,169 @@
...
+```python title="server.py"
+--8<-- "docs_src/server_cards/tutorial001.py"
+```
+
+`build_server_card` reads `version`, `title`, `description`, `website_url`, and
+`icons` off the server, and raises `ValueError` if `version` or `description` is
+unset — a card cannot exist without them. The `name` you pass is the reverse-DNS
+identifier and is validated against the `namespace/name` pattern.
+
+Because discovery happens *before* authentication, mount both routes **outside**
```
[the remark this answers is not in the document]
SamMorrowDrums: Reworded. The doc now explains that `mount_server_card`/`mount_ai_catalog` append to `app.router.routes` (inside the app), so they clear per-route auth dependencies but do **not** exempt the routes from authentication applied as global ASGI middleware. It now points readers at either excluding the discovery paths in that middleware or serving the card/catalog from a separate, unauthenticated sub-app.

docs_src/server_cards/tutorial004.py
```diff
@@ -0,0 +1,21 @@
...
+    # policy (timeouts, redirect caps, blocking private address ranges) when
+    # discovering hosts you do not fully trust.
+    async with httpx2.AsyncClient() as http_client:
+        catalog_url = well_known_ai_catalog_url("https://dice.example.com")
+        catalog = await fetch_ai_catalog(catalog_url, http_client=http_client)
+
+        for entry in catalog.entries:
+            if entry.media_type != MCP_SERVER_CARD_MEDIA_TYPE or entry.url is None:
+                continue
+            card = await fetch_server_card(entry.url, http_client=http_client)
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed — the example now resolves entry URLs with `urljoin(catalog_url, entry.url)` before fetching, matching what `discover_server_cards` does, so relative card URLs in a catalog resolve correctly.

src/mcp/server/experimental/ai_catalog.py
```diff
@@ -0,0 +1,125 @@
...
+
+__all__ = ["DISCOVERY_HEADERS", "server_card_entry", "ai_catalog_route", "mount_ai_catalog"]
+
+#: Response headers for discovery endpoints (catalogs and the artifacts they
+#: reference): CORS headers so browser clients can read them, plus a caching
+#: hint.
+DISCOVERY_HEADERS = {
+    "Access-Control-Allow-Origin": "*",
+    "Access-Control-Allow-Methods": "GET",
+    "Access-Control-Allow-Headers": "Content-Type",
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed — `DISCOVERY_HEADERS` now sets `Access-Control-Allow-Headers: Content-Type, If-None-Match` and `Access-Control-Expose-Headers: ETag`, and the discovery routes answer the `OPTIONS` preflight (204) that a cross-origin conditional GET triggers (since `If-None-Match` is not a CORS-safelisted request header). Added preflight tests to both experimental suites and updated the docs table. This matches the discovery spec CORS requirements.

src/mcp/shared/experimental/ai_catalog/types.py
```diff
@@ -0,0 +1,261 @@
...
+    """A single AI artifact in a catalog.
+
+    Exactly one of ``url`` (artifact by reference) or ``data`` (artifact
+    inline) must be provided.
+    """
+
+    identifier: str
+    """Identifier for the artifact; SHOULD be a URN or URI.
+
+    MCP server entries use ``urn:air:{publisher}:{name}``, where ``publisher`` is
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed — the `CatalogEntry.identifier` docstring now reads `urn:air:{publisher}:mcp:{name}`, matching the code (`_air_identifier` / `server_card_entry`) and the docs table.

src/mcp/shared/experimental/server_card/types.py
```diff
@@ -0,0 +1,164 @@
...
+
+import re
+from typing import Annotated, Any, Literal
+
+from mcp_types import Icon
+from mcp_types._types import MCPModel
+from pydantic import Field, field_validator
+
+#: Canonical ``$schema`` value for a Server Card document.
+SERVER_CARD_SCHEMA_URL = "https://static.modelcontextprotocol.io/schemas/v1/server-card.schema.json"
```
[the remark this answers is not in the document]
SamMorrowDrums: This is intentional and spec-correct, so leaving as-is. The SEP-2127 Server Card spec requires the `$schema` value to be exactly `https://static.modelcontextprotocol.io/schemas/v1/server-card.schema.json`; that schema document is served from `modelcontextprotocol/static` once the extension graduates and is simply not published yet (the extension is still pre-release). Hardcoding the spec-mandated URL now is the right thing — and the client tolerates a missing/omitted `$schema` rather than depending on it resolving.

src/mcp/client/experimental/server_card.py
```diff
@@ -0,0 +1,108 @@
...
+    pooling / auth, otherwise a short-lived client with MCP defaults is used.
+
+    Raises:
+        httpx2.HTTPError: If the request fails or returns a non-2xx status.
+        pydantic.ValidationError: If the document is not a valid Server Card.
+    """
+    if http_client is None:
+        async with create_mcp_http_client() as client:
+            return await fetch_server_card(url, http_client=client)
+    response = await http_client.get(url, headers={"Accept": f"{MCP_SERVER_CARD_MEDIA_TYPE}, application/json"})
```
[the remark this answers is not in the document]
SamMorrowDrums: Intentionally lenient, so leaving as-is. The spec treats Server Cards as advisory and explicitly allows hosting them at any URI, including static files/CDNs that serve `application/json` (see the static-hosting tutorial). Enforcing a `Content-Type` of `application/mcp-server-card+json` would reject that spec-endorsed static hosting. We already send `Accept: application/mcp-server-card+json, application/json` (the spec SHOULD) and validate the JSON body against the Server Card schema, which is the meaningful correctness check.

PR #2963, opened 2026-06-25
fix(server): escape regex special chars in resource template literals

### LINKED ISSUE #2961: [Bug] ResourceTemplate.matches() does not escape regex special chars in URI template literals
## Bug Description

`ResourceTemplate.matches()` in `src/mcp/server/mcpserver/resources/templates.py:97` converts URI templates to regex patterns using simple string replacement, but does **not** escape regex-special characters in literal (non-parameter) parts of the template.

```python
# L97
pattern = self.uri_template.replace("{", "(?P<").replace("}", ">[^/]+)")
```

## Reproduction Cases

| Template literal char | Regex side effect | False match |
|---|---|---|
| `.well-known/{name}` | `.` = any char | `data://Xwell-known/hello` incorrectly matches |
| `{id}.json` | `.` = any char | `data://items/123Xjson` incorrectly matches |
| `{id}?format=json` | `?` = optional | `data://items/123format=json` matches without `?` |
| `{name}+suffix` | `+` = one-or-more quantifier | Semantic distortion |
| `{name}++suffix` | `++` = possessive quantifier | Behavior change |

## Fix (proposed)

Use `re.escape()` on the template first, then replace escaped `\{param\}` with named capture groups:

```python
import re as _re

_URI_PARAM_PATTERN = _re.compile(r"\\\{([^}]+)\\\}")

def _build_template_pattern(template: str) -> str:
    escaped = _re.escape(template)
    return _URI_PARAM_PATTERN.sub(lambda m: f"(?P<{m.group(1)}>[^/]+)", escaped)
```

## AI Disclosure

Bug discovered with AI assistance (opencode).

Bartok9:
## Summary

- `ResourceTemplate.matches()` now treats regex-special characters in the literal parts of a URI template (`.`, `?`, `+`, etc.) as literals instead of regex metacharacters.
- Fixes false positives where a template like `data://.well-known/{name}` matched URIs it should not.

## Motivation

Closes #2961.

`matches()` converted the URI template to a regex with plain string replacement:

```python
pattern = self.uri_template.replace("{", "(?P<").replace("}", ">[^/]+)")
```

This left regex-special characters in the literal portions unescaped, so they acted as metacharacters and produced incorrect matches:

| Template | False match |
|---|---|
| `data://.well-known/{name}` | `data://Xwell-known/hello` (`.` = any char) |
| `data://items/{id}.json` | `data://items/123Xjson` (`.` = any char) |

## Fix

Escape the whole template with `re.escape()` first, then substitute the escaped `{param}` markers with named capture groups:

```python
_PARAM_PATTERN = re.compile(r"\\\{([^}]+?)\\\}")
...
pattern = _PARAM_PATTERN.sub(lambda m: f"(?P<{m.group(1)}>[^/]+)", re.escape(self.uri_template))
```

Literal characters are now matched literally; parameters still capture `[^/]+` as before.

## Verification

- `uv run pytest tests/server/mcpserver/resources/test_resource_template.py` — 16 passed

## Real behavior proof

Regression test `test_template_matches_escapes_regex_special_chars` reproduces the issue cases. It **fails without the source fix** and passes with it:

```
# Without the source fix (test present):
TestResourceTemplate::test_template_matches_escapes_regex_special_chars  53 / 71  AssertionError
Results: 1 failed

# With the fix:
Results: 16 passed
```

### DIFF
--- a/src/mcp/server/mcpserver/resources/templates.py
+++ b/src/mcp/server/mcpserver/resources/templates.py
@@ -19,6 +19,10 @@ from mcp.server.mcpserver.utilities.logging import get_logger
 from mcp.shared._callable_inspection import is_async_callable
 from mcp.types import Annotations, Icon
 
+# Matches a re.escape()'d "{param}" marker. re.escape escapes the braces to
+# "\{" / "\}" but leaves the parameter name (identifier chars) untouched.
+_PARAM_PATTERN = re.compile(r"\\\{([^}]+?)\\\}")
+
 logger = get_logger(__name__)
 
 if TYPE_CHECKING:
@@ -93,8 +97,11 @@ class ResourceTemplate(BaseModel):
 
         Extracted parameters are URL-decoded to handle percent-encoded characters.
         """
-        # Convert template to regex pattern
-        pattern = self.uri_template.replace("{", "(?P<").replace("}", ">[^/]+)")
+        # Convert template to regex pattern. Escape the template first so that
+        # regex-special characters in literal parts (e.g. ".", "?", "+") are
+        # treated literally, then replace the escaped "{param}" markers with
+        # named capture groups.
+        pattern = _PARAM_PATTERN.sub(lambda m: f"(?P<{m.group(1)}>[^/]+)", re.escape(self.uri_template))
         match = re.match(f"^{pattern}$", uri)
         if match:
             # URL-decode all extracted parameter values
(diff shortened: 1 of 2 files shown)


Kludex:
You've opened a duplicated pull request, please search opened PRs before creating new ones. Duplicated from #2749.

PR #2974, opened 2026-06-25
Conformance burn-down: server-side InputRequiredResult, Mcp-Method/Name validation, x-mcp-header filter (14 scenarios → green)

maxisbey:
Bumps the conformance pin to `0.2.0-alpha.7` and clears 14 scenarios from the expected-failures baseline, leaving only the 4 that depend on unlanded design (client `Mcp-Param-*` emission, modern stateless SSE for progress, MCPServer prompt-pipeline `InputRequiredResult`, and the SEP-990 enterprise-auth waiver).

## Motivation and Context

#2967 widened the lowlevel server return types for `InputRequiredResult` and #2968 gave `Client.call_tool` the `allow_input_required` / `input_responses` / `request_state` surface, but neither was reachable through the conformance harness: the everything-server had no MRTR fixture tools, the client driver had no SEP-2322 handler, and `MCPServer`'s `func_metadata.convert_result()` still mangled an `InputRequiredResult` into `TextContent`. Separately, the SEP-2243 boundary checks (`Mcp-Method`/`Mcp-Name` cross-check, `x-mcp-header` validity) were unimplemented on both sides. This closes those gaps.

## How Has This Been Tested?

All five conformance legs run green with `--expected-failures` at `@modelcontextprotocol/conformance@0.2.0-alpha.7`:

| Leg | Result |
|---|---|
| server `--suite active` | 42 pass / 0 fail |
| server `--suite draft` | 69 pass / 1 expected fail |
| server `--suite all --spec-version 2026-07-28` | 97 pass / 2 expected fail |
| client `--suite all` | 402 pass / 7 expected fail (2 scenarios) |
| client `--suite all --spec-version 2026-07-28` | 352 pass / 5 expected fail (1 scenario) |

`./scripts/test`: 2776 passed, 100% coverage, strict-no-cover clean. `pyright`: 0 errors.

## Breaking Changes

None for v1-lineage callers. At 2026-07-28, `ClientSession.list_tools()` now drops tools whose `x-mcp-header` annotations are invalid (the spec MUST is on the client); handshake-era sessions are unaffected. The modern HTTP server entry now rejects requests where `Mcp-Method`/`Mcp-Name` headers contradict the body — handshake-era requests don't pass through that classifier rung.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**SDK changes**

- `func_metadata.convert_result()` passes `InputRequiredResult` through; `MCPServer._handle_call_tool` / `call_tool` return type widened to `CallToolResult | InputRequiredResult`. `Context` gains `.input_responses` / `.request_state` / `.client_capabilities` read-only props so a tool body can read what the client supplied.
- `shared/inbound.py`: `classify_inbound_request` rung 2 cross-checks `Mcp-Method` == `body.method` and `Mcp-Name` == `params.{name|uri}` (new `NAME_BEARING_METHODS` map + `decode_header_value`). New `find_invalid_x_mcp_header()` validates the annotation per RFC 9110 token rules + uniqueness + primitive-type.
- `_streamable_http_modern.py` filters `tools/list` results for invalid `x-mcp-header` at the modern HTTP boundary (so the version gate is structural, not a conditional in handler code).
- `ClientSession.list_tools` applies the same filter when the negotiated version is modern. `_make_modern_stamp` now derives `Mcp-Name` from `NAME_BEARING_METHODS` (covers `prompts/get` and `resources/read`, not just `tools/call`); `send_discover` stamps `Mcp-Method` so the auto-mode probe doesn't fail the cross-check it just introduced.

**Fixtures**

- everything-server: 8 `test_input_required_result_*` tools (elicitation/sampling/list_roots/request_state/multiple_inputs/multi_round/tampered_state/capabilities) + `json_schema_2020_12_tool` carrying the 2020-12 keywords the scenario checks. The `tampered_state` tool uses a fixture-local `hmac` to sign/verify `requestState`; an SDK-level helper is follow-up.
- `.github/actions/conformance/client.py`: handlers for `sep-2322-client-request-state` (drives the four mock tools via `allow_input_required=True`) and `http-invalid-tool-headers` (lists then calls every surfaced tool).

**Pin**

`CONFORMANCE_PKG` moves from the `pkg.pr.new@65fcd39` URL to the published `@modelcontextprotocol/conformance@0.2.0-alpha.7`; the URL fetch-and-verify steps are dropped now that npm's own integrity check applies. Baselined: alpha.7 introduces no new failures over the previous pin.

Stacked on #2968.

### DIFF
--- a/.github/actions/conformance/client.py
+++ b/.github/actions/conformance/client.py
@@ -20,7 +20,9 @@ Scenarios:
     json-schema-ref-no-deref                - Connect, list tools (no $ref deref)
     request-metadata                        - Connect with all callbacks; client stamps _meta
     http-standard-headers                   - Connect, call a tool (Mcp-* headers checked)
+    http-invalid-tool-headers               - List tools, call every surfaced tool (x-mcp-header filter)
     elicitation-sep1034-client-defaults     - Elicitation with default accept callback
+    sep-2322-client-request-state           - Drive the manual MRTR retry surface
     auth/client-credentials-jwt             - Client credentials with private_key_jwt
     auth/client-credentials-basic           - Client credentials with client_secret_basic
     auth/*                                  - Authorization code flow (default for auth scenarios)
@@ -296,6 +298,43 @@ async def run_http_standard_headers(server_url: str) -> None:
         logger.debug(f"add_numbers result: {result}")
 
 
+def _stub_required_args(input_schema: dict[str, Any]) -> dict[str, Any]:
+    """Minimal arguments satisfying a tool inputSchema's required list."""
+    by_type: dict[str, Any] = {
+        "string": "x",
+        "integer": 0,
+        "number": 0,
+        "boolean": False,
+        "object": {},
+        "array": [],
+        "null": None,
+    }
+    properties = input_schema.get("properties", {})
+    return {name: by_type.get(properties.get(name, {}).get("type"), "x") for name in input_schema.get("required", [])}
+
+
+@register("http-invalid-tool-headers")
+async def run_http_invalid_tool_headers(server_url: str) -> None:
+    """List tools, then call every tool the SDK surfaces (SEP-2243).
+
+    The harness mock advertises one valid tool plus several with malformed
+    x-mcp-header annotations (empty, non-primitive type, duplicate, invalid
+    chars). The scenario passes if valid_tool is called and the malformed
+    ones are not -- so a conforming client filters them out of the list_tools
+    result and the loop below never sees them. The scenario sets
+    allowClientError, so a per-call failure is logged and skipped rather
+    than aborting the whole run.
+    """
+    async with Client(server_url, mode=client_mode()) as client:
+        listed = await client.list_tools()
+        logger.debug(f"Surfaced tools: {[t.name for t in listed.tools]}")
+        for tool in listed.tools:
+            try:
+                await client.call_tool(tool.name, _stub_required_args(tool.input_schema))
+            except Exception:
+                logger.exception(f"call_tool({tool.name!r}) failed")
+
+
 @register("elicitation-sep1034-client-defaults")
 async def run_elicitation_defaults(server_url: str) -> None:
     """Connect with elicitation callback that applies schema defaults."""
@@ -303,6 +342,53 @@ async def run_elicitation_defaults(server_url: str) -> None:
         await client.list_tools()
         result = await client.call_tool("test_client_elicitation_defaults", {})
         logger.debug(f"test_client_elicitation_defaults result: {result}")
+
+
+@register("sep-2322-client-request-state")
+async def run_mrtr_client(server_url: str) -> None:
+    """Drive the manual MRTR retry surface against the SEP-2322 client mock.
+
+    The mock speaks the modern lifecycle (server/discover, no initialize) and
+    inspects the wire params of each tools/call round, so this exercises the
+    explicit allow_input_required=True path rather than an auto-loop: round 1
+    receives an InputRequiredResult, the fixture fulfils the elicitation
+    locally, then round 2 retries with input_responses + the echoed
+    request_state. Passing request_state straight off the typed result -- a
+    str when the server sent one, None when it didn't -- lets the
+    serializer's exclude_none drop the key in the no-state case without a
+    branch here. The unrelated call between rounds proves MRTR params don't
+    leak across tools, and the no-result-type call must parse as a complete
+    CallToolResult with no retry.
+    """
+    async with Client(server_url, mode=client_mode()) as client:
+        await client.list_tools()
+        confirm = {"confirm": types.ElicitResult(action="accept", content={"confirmed": True})}
+
+        r1 = await client.call_tool("test_mrtr_echo_state", {}, allow_input_required=True)
+        assert isinstance(r1, types.InputRequiredResult)
+
+        await client.call_tool("test_mrtr_unrelated", {})
+
+        await client.call_tool(
+            "test_mrtr_echo_state",
+            {},
+            input_responses=confirm,
+            request_state=r1.request_state,
+            allow_input_required=True,
+        )
+
+        r2 = await client.call_tool("test_mrtr_no_state", {}, allow_input_required=True)
+        assert isinstance(r2, types.InputRequiredResult)
+        await client.call_tool(
+            "test_mrtr_no_state",
+            {},
+            input_responses=confirm,
+            request_state=r2.request_state,
+            allow_input_required=True,
+        )
+
+        result = await client.call_tool("test_mrtr_no_result_type", {})
+        assert isinstance(result, types.CallToolResult)
 
 
 @register("auth/client-credentials-jwt")
@@ -441,8 +527,7 @@ def main() -> None:
             asyncio.run(run_auth_code_client(server_url))
         else:
             # Unhandled scenarios:
-            #  - sep-2322-client-request-state (SEP-2322 / S6: MRTR client loop)
-            #  - http-custom-headers, http-invalid-tool-headers (SEP-2243 / S8: Mcp-Param-* headers)
+            #  - http-custom-headers (SEP-2243 / S8: Mcp-Param-* emission)
             print(f"Unknown scenario: {scenario}", file=sys.stderr)
             sys.exit(1)
     else:
--- a/examples/servers/everything-server/mcp_everything_server/server.py
+++ b/examples/servers/everything-server/mcp_everything_server/server.py
@@ -6,6 +6,7 @@ Server implementing all MCP features for conformance testing based on Conformanc
 
 import asyncio
 import base64
+import binascii
 import hashlib
 import hmac
 import json
@@ -493,7 +494,10 @@ def _unseal_state(state: str) -> str:
     expected = hmac.new(_STATE_HMAC_KEY, encoded.encode(), hashlib.sha256).hexdigest()
     if not sig or not hmac.compare_digest(sig, expected):
         raise MCPError(code=INVALID_PARAMS, message="requestState failed integrity verification")
-    return base64.urlsafe_b64decode(encoded).decode()
+    try:
+        return base64.urlsafe_b64decode(encoded).decode()
+    except (binascii.Error, UnicodeDecodeError) as e:
+        raise MCPError(code=INVALID_PARAMS, message="requestState failed integrity verification") from e
 
 
 @mcp.tool()
--- a/examples/servers/everything-server/mcp_everything_server/server.py
+++ b/examples/servers/everything-server/mcp_everything_server/server.py
@@ -6,8 +6,11 @@ Server implementing all MCP features for conformance testing based on Conformanc
 
 import asyncio
 import base64
+import hashlib
+import hmac
 import json
 import logging
+from typing import Any
 
 import click
 from mcp.server import ServerRequestContext
@@ -20,10 +23,20 @@ from mcp_types import (
     Completion,
     CompletionArgument,
     CompletionContext,
+    CreateMessageRequest,
+    CreateMessageRequestParams,
+    CreateMessageResult,
+    ElicitRequest,
+    ElicitRequestFormParams,
+    ElicitResult,
     EmbeddedResource,
     EmptyResult,
     ImageContent,
+    InputRequest,
+    InputRequiredResult,
     JSONRPCMessage,
+    ListRootsRequest,
+    ListRootsResult,
     PromptReference,
     ResourceTemplateReference,
     SamplingMessage,
@@ -33,7 +46,7 @@ from mcp_types import (
     TextResourceContents,
     UnsubscribeRequestParams,
 )
-from mcp_types.jsonrpc import MISSING_REQUIRED_CLIENT_CAPABILITY
+from mcp_types.jsonrpc import INVALID_PARAMS, MISSING_REQUIRED_CLIENT_CAPABILITY
 from pydantic import BaseModel, Field
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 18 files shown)


examples/servers/everything-server/mcp_everything_server/server.py
```diff
@@ -333,6 +346,228 @@ async def test_missing_capability(ctx: Context) -> str:
...
+            "client_roots": ListRootsRequest(),
+        },
+        request_state="multiple-inputs",
+    )
+
+
+@mcp.tool()
+async def test_input_required_result_multi_round(ctx: Context) -> str | InputRequiredResult:
+    """Tests a three-round InputRequiredResult flow with evolving requestState"""
+    state = json.loads(ctx.request_state) if ctx.request_state else {"round": 0}
```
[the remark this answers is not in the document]
maxisbey: The state this tool reads is what it serialised on the previous round and the conformance harness echoes back verbatim — there's no path where it arrives malformed. If a future scenario does send adversarial state, the `tampered_state` tool is the one that demonstrates rejection (via the HMAC seal); this tool just exercises the multi-round happy path. Any stray decode error would still surface as `CallToolResult(is_error=True)` via the MCPServer wrapper, which is what the harness checks for anyway.

.github/actions/conformance/client.py
```diff
@@ -305,6 +344,53 @@ async def run_elicitation_defaults(server_url: str) -> None:
...
+        await client.list_tools()
+        confirm = {"confirm": types.ElicitResult(action="accept", content={"confirmed": True})}
+
+        r1 = await client.call_tool("test_mrtr_echo_state", {}, allow_input_required=True)
+        assert isinstance(r1, types.InputRequiredResult)
+
+        await client.call_tool("test_mrtr_unrelated", {})
+
+        await client.call_tool(
+            "test_mrtr_echo_state",
```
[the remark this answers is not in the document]
maxisbey: The assertion is structural: this retry call is `call_tool(..., input_responses=..., request_state=...)` *without* `allow_input_required=True`, so per the overload introduced in #2968 the SDK raises if the result is another `InputRequiredResult` — see https://github.com/modelcontextprotocol/python-sdk/blob/992b4da8/src/mcp/client/session.py#L644-L657. On top of that the harness's `default-result-type-complete` check is the actual assertion layer for this scenario; the driver's job is just to put the right bytes on the wire.

maxisbey (commented):
Solid burn-down — the `Mcp-Method`/`Mcp-Name` cross-check, the `NAME_BEARING_METHODS` table shared between client emit and server validate, and the discover probe being kept in lockstep with the new rung are all the right shape, and the test additions over the new classifier rungs are thorough. The inline notes cluster into a few groups, in roughly the order I'd weight them.

**Two places `find_invalid_x_mcp_header` diverges from a normative MUST.** Both are in the inline comments on `shared/inbound.py`. The `type: "number"` concession is a divergence the spec already resolved against the harness (the spec commit that forbade `number` predates the `0.2.0-alpha.7` pin; the harness fixture that emits it is on both expected-failures lists, so removing it costs nothing in CI). The bigger one is that the validator only walks top-level `properties`, while the spec permits nested annotations (which then have to be validated and counted toward the case-insensitive uniqueness set) and says an annotation in an illegal position — under `items`, a composition keyword, `if`/`then`/`else`, or `$ref` — makes the *tool definition* invalid. The current walk never sees any of those, so it keeps tools the spec says the client must exclude. The docstring's claim that the spec restricts the annotation to top-level properties is what the spec text contradicts.

**One design question, not a defect.** `_drop_invalid_header_tools` on the server side implements an obligation the spec assigns only to the client, and after this PR the client-side `list_tools` filter carries both the spec MUST and the conformance scenario, so the server copy carries neither. Detail in the inline comment; my vote is to delete it, but it's a genuine call.

**Smaller things, all inline:** `Context.input_responses` / `request_state` re-derive fields the already-validated `CallToolRequestParams` in `_handle_call_tool`'s hand carries, with a second alias spelling and a `ValidationError` path that's unreachable from the wire; the new `InputRequiredResult` union arm in `func_metadata` lands before the `CallToolResult`-in-union `InvalidSignature`, so a three-arm `-> CallToolResult | str | InputRequiredResult` is now silently accepted where `-> CallToolResult | str` raises; and one test docstring claims spec backing it doesn't have.

Two notes I'm keeping in the body rather than as inline comments:

- `docs/migration.md` has an existing section headed "`MCPServer.call_tool()` returns `CallToolResult`" whose first sentence is "`MCPServer.call_tool()` now always returns a `CallToolResult`." — this PR makes that sentence false. One-line amendment in the existing section; I don't think a new section is needed, the widening only affects the unreleased v2 surface.
- Net +4 `# pragma: no branch` on the stub `def fn() -> X: ...` lines in `test_func_metadata.py`. They match the existing in-file precedent for that idiom so I'm fine with them; noting it because we try to keep the pragma count flat.

src/mcp/shared/inbound.py
```diff
@@ -52,17 +60,97 @@
...
+
+_B64_SENTINEL = re.compile(r"^=\?base64\?(?P<payload>.*)\?=$")
 # RFC 7230 token chars minus DEL; visible ASCII 0x20-0x7E is the practical bound for a header value.
 _HEADER_SAFE = re.compile(r"^[\x20-\x7E]*$")
+# RFC 9110 §5.6.2 token: the only characters permitted in an HTTP field name.
+_RFC9110_TOKEN = re.compile(r"^[!#$%&'*+\-.^_`|~0-9A-Za-z]+$")
+# JSON-Schema types that stringify cleanly into a single header value. The spec
+# names string/integer/boolean; number is admitted because the conformance
+# harness emits it and float→str round-trips to within tolerance.
+_X_MCP_HEADER_PRIMITIVE_TYPES: Final = frozenset({"string", "integer", "boolean", "number"})
```
maxisbey: The comment here is candid that this is a deliberate divergence from the spec to keep the conformance harness happy, so I'll engage that directly rather than re-quote the spec at you. Two facts that I think dissolve the trade-off.

The spec already settled it. https://github.com/modelcontextprotocol/modelcontextprotocol/commit/18afae65c3ad12ffa0cbd3330952817ba0a4907a (2026-05-28) is the commit that changed the constraint from `(number, string, boolean)` to `(integer, string, boolean)` and added "Parameters with type `number` are not permitted." The harness's own requirement text still carries the pre-edit wording — https://github.com/modelcontextprotocol/conformance/blob/5d7abd5e70513731b9fb70393639338631eaf7ec/src/seps/sep-2243.yaml#L25-L27 says "primitive types (number, string, or boolean)" and omits `integer` entirely — so the harness is the stale side here, not the spec.

And the concession is nearly free to drop. The only *scenario* in the pinned `0.2.0-alpha.7` harness that pairs `type: "number"` with an `x-mcp-header` annotation is the `http-custom-headers` mock (https://github.com/modelcontextprotocol/conformance/blob/5d7abd5e70513731b9fb70393639338631eaf7ec/src/scenarios/client/http-custom-headers.ts#L295-L299, plus `float_val` at L321-L325), and that scenario sits on both expected-failures lists already and isn't registered in our `client.py` driver, so removing `"number"` doesn't move it. The scenario this PR actually clears — `http-invalid-tool-headers` — has no `number`-typed annotated property, and the everything-server registers no `x-mcp-header` tools, so none of the five conformance legs changes. The only thing that has to move with it is the `id="valid-number"` param this PR adds in `tests/shared/test_inbound.py`, which flips to the reject table. The "float→str round-trips to within tolerance" half of the justification is about `Mcp-Param-*` value *emission*, which this PR deliberately leaves in expected-failures; it doesn't bear on whether the annotation is admissible in the first place.

As shipped, a 2026-07-28 client built on this SDK keeps a tool the spec says it MUST exclude. I'd drop `"number"` from the set here and file the stale `sep-2243.yaml` wording against the conformance repo.

src/mcp/shared/inbound.py
```diff
@@ -52,17 +60,97 @@
...
+    the schema. A `None` / non-object / property-less schema has nothing to
+    validate and returns `None`.
+    """
+    match input_schema:
+        case {"properties": {**properties}}:
+            pass
+        case _:
+            return None
+    seen: dict[str, str] = {}
+    for prop_name, raw in properties.items():
```
maxisbey: The docstring says the spec restricts the annotation to *top-level* primitive properties, and the implementation follows it — but the spec says something quite different, and the difference is normative. From the Schema Extension constraints (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/8bfc3be7dad21762e158b87cf0510513d96cdcf5/docs/specification/draft/basic/transports/streamable-http.mdx#L388-L395):

> **MUST** only be applied to properties that are _statically reachable_ from the schema root: reachable via a chain consisting solely of `properties` keys. The chain **MUST NOT** pass through `items` (or any other array keyword), composition keywords (`oneOf`, `anyOf`, `allOf`, `not`), conditional keywords (`if`/`then`/`else`), or `$ref`. Nested object properties are permitted as long as every step in the chain is a `properties` key. An `x-mcp-header` annotation anywhere else makes the annotation — and thus the tool definition — invalid.

and the uniqueness rule is over "all `x-mcp-header` values in the `inputSchema`" (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/8bfc3be7dad21762e158b87cf0510513d96cdcf5/docs/specification/draft/basic/transports/streamable-http.mdx#L382-L383), not the top level. A top-level-only `match` gives the wrong answer in three concrete cases, all of which the client MUST resolve to "exclude the tool":

- an annotation sitting under `items`, a composition keyword, `if`/`then`/`else`, or `$ref` makes the whole tool definition invalid — this walk never sees it, so the tool is kept;
- a *legally* nested `properties.outer.properties.inner["x-mcp-header"]` is never checked for token syntax, type, or anything else;
- a top-level `"Region"` plus a nested `"region"` is a case-insensitive duplicate per the spec — kept here.

The `http-invalid-tool-headers` fixtures all happen to use top-level properties, which is why this is green in conformance, but the spec is the contract — and there's no conformance check for the statically-reachable bullet at `0.2.0-alpha.7` at all, so this is one to raise on the conformance repo alongside the fix rather than wait on.

I think the right shape is a single recursive walk over *schema positions*: collect every `x-mcp-header` key that appears in a subschema together with the path that reached it, treat any not reached via a pure root → `properties` → name → `properties` → … chain as immediately invalid, and run the existing token/type checks plus one global case-insensitive uniqueness set over the rest. (The walk has to stay out of instance-data keywords like `default`/`examples`/`const`/`enum` — a key literally named `x-mcp-header` inside a `default` value is data, not an annotation.) That also makes this docstring true. `find_invalid_x_mcp_header` is already the one validator shared by the client `list_tools` filter and the server-side `tools/list` filter, so it's one function to fix.

src/mcp/server/_streamable_http_modern.py
```diff
@@ -122,6 +127,30 @@ async def _to_jsonrpc_response(
...
 
 
+def _drop_invalid_header_tools(result: dict[str, Any]) -> None:
+    """Remove tools whose `x-mcp-header` annotations fail validation from a `tools/list` result, in place.
+
+    Runs at the modern HTTP boundary so the version gate is structural (this
+    entry only ever serves a 2026-07-28+ request) rather than a conditional in
+    handler code. Dropped tools are logged so a server author sees why their
+    tool vanished from the wire.
+    """
```
maxisbey: The PR body's rationale — putting this at the modern HTTP boundary so the version gate is structural rather than a conditional in handler code — holds up; there's no `if version ==` anywhere and I wouldn't ask for one. The question I keep coming back to is one level up: does the server need to do this at all?

The spec puts the rejection on exactly one party. "While the use of `x-mcp-header` is optional for **servers**, clients **MUST** support this feature" (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/8bfc3be7dad21762e158b87cf0510513d96cdcf5/docs/specification/draft/basic/transports/streamable-http.mdx#L364-L365) and "Clients using the Streamable HTTP transport **MUST** reject tool definitions where any `x-mcp-header` value violates these constraints. Rejection means the client **MUST** exclude the invalid tool from the result of `tools/list`" (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/8bfc3be7dad21762e158b87cf0510513d96cdcf5/docs/specification/draft/basic/transports/streamable-http.mdx#L402-L404). There's no server-side counterpart to that MUST in either spec file — the server-side `x-mcp-header` obligations there are about validating `Mcp-Param-*` *request* headers against the body, not about excluding tool definitions from `tools/list` — and no conformance scenario needs it either. The only server-suite `x-mcp-header` scenario in the pinned harness, `http-custom-header-server-validation`, just needs the server to expose *some* annotated tool and emits SKIPPED otherwise; it never asserts the server filtered an invalid one. `http-header-validation` is the `Mcp-Method`/`Mcp-Name` cross-check, already covered by `classify_inbound_request`. And `http-invalid-tool-headers` — the scenario this PR clears — is a *client*-suite scenario in which the harness plays the server, satisfied by the `list_tools` filter on the client side.

So after this PR the client filter carries both the spec MUST and the conformance scenario, and this one carries neither. What it buys is masking a server author's bug — and not even consistently: the in-process modern entry (`Client(server)` → `modern_on_request` → `serve_one`) skips this filter, so the same `Server` object advertises the tool there and hides it here, with a `WARNING` at request time as the only signal. (The handshake-era entries advertising it is *correct* — `initialize` can't negotiate 2026-07-28, so this entry already is the whole streamable-HTTP-at-2026-07-28 surface; and in practice even the in-process gap is invisible because the SDK's own client filter applies there.) If we want the server to refuse to advertise an invalid annotation, I'd rather make it unrepresentable at registration (`@mcp.tool()` / `ToolManager.add_tool`) than filter one transport's wire response. Not at `Tool` construction, to be clear — `Tool` is also the model the client deserialises a peer's `tools/list` into, so a model-level rejection would turn one bad third-party tool into a `ValidationError` out of `ListToolsResult.model_validate` instead of the spec's "exclude just that tool".

My vote is to delete `_drop_invalid_header_tools` and the `tools/list` call site below and let the client filter carry it. Happy to keep it if there's a reason I'm missing.

tests/server/test_streamable_http_modern.py
```diff
@@ -270,3 +275,72 @@ async def fail() -> dict[str, Any]:
...
+
+
+# --- tools/list x-mcp-header filter --------------------------------------------
+
+
+async def test_handle_modern_request_drops_tools_with_invalid_x_mcp_header_from_list_result(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """Spec-mandated: a tool whose `inputSchema` carries a malformed `x-mcp-header` is excluded
+    from the modern-path `tools/list` result and a warning logged; valid tools pass through."""
```
maxisbey: "Spec-mandated" isn't right here. The validity constraints themselves are normative on whoever authors the annotation, but the only normative *response* to an invalid `x-mcp-header` annotation is on the *client* ("Clients ... **MUST** reject tool definitions ... the client **MUST** exclude the invalid tool from the result of `tools/list`"), and the spec is explicit that `x-mcp-header` "is optional for servers". This test pins SDK-chosen behaviour, not spec-mandated behaviour — worth correcting whatever we land on for the filter itself, since a future reader will take the docstring at its word.

src/mcp/server/mcpserver/context.py
```diff
@@ -219,6 +221,36 @@ def request_id(self) -> str:
...
+        return None if raw is None else _INPUT_RESPONSES_ADAPTER.validate_python(raw)
+
+    @property
+    def request_state(self) -> str | None:
+        """Opaque state echoed from a prior `InputRequiredResult.request_state`.
+
+        `None` on the initial round.
+        """
+        params = self.request_context.params
+        return params.get("requestState") if params else None
```
maxisbey: `CallToolRequestParams` already carries these as typed fields with the same wire aliases — `input_responses: InputResponses | None` / `request_state: str | None` (https://github.com/modelcontextprotocol/python-sdk/blob/992b4da87fcb2258a615ad3a4aa9e70933b75c6b/src/mcp/types/_types.py#L863-L865) — and the runner validates the raw `ctx.params` into exactly that model before any handler runs (https://github.com/modelcontextprotocol/python-sdk/blob/992b4da87fcb2258a615ad3a4aa9e70933b75c6b/src/mcp/server/runner.py#L259). So `_handle_call_tool` is handed the already-parsed `params.input_responses` / `params.request_state` and drops them, and these two properties then re-spell the wire aliases as string literals against the raw dict and bring in a second, parallel `TypeAdapter(InputResponses)` to redo the parse.

Two consequences worth knowing even if the shape stays:

- The `ValidationError` path here is unreachable from the wire. A request with a malformed `inputResponses` is rejected before `_handle_call_tool` is ever invoked — by the per-version surface check at `runner.py:238` on a 2026-07-28 request, or by the `CallToolRequestParams` parse at `runner.py:259` as the version-independent backstop — and becomes `INVALID_PARAMS`, so this property never sees it. (That's also why a bad value in the `input-required-result-validate-input` conformance scenario surfaces as a top-level `INVALID_PARAMS` rather than as a `ToolError` out of this getter — the boundary gets there first.)
- The alias names and the parse rule now exist in two places.

The smallest fix is to thread the typed params through. `Context` is built in three handlers (`_handle_call_tool`, `_handle_read_resource`, `_handle_get_prompt`), and all three param types subclass `InputResponseRequestParams`, so pass that base through from each — `Context(request_context=ctx, mcp_server=self, input_params=params)` — and these two properties become `self._input_params.input_responses` / `.request_state`. That deletes `_INPUT_RESPONSES_ADAPTER`, both alias literals, and the dead validation in one go, and keeps the properties working for `resources/read` and `prompts/get` rather than only `tools/call`.

src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -266,8 +272,17 @@ def func_metadata(
...
     if is_union_origin(get_origin(return_type_expr)):
         args = get_args(return_type_expr)
+        # A union containing InputRequiredResult means the tool may return either a complete
+        # value or a multi-round input request; treat the complete arm as unstructured (the
+        # pass-through in convert_result handles the InputRequiredResult arm).
+        if any(isinstance(arg, type) and issubclass(arg, InputRequiredResult) for arg in args):
+            return FuncMetadata(arg_model=arguments_model)
         # Check if CallToolResult appears in the union (excluding None for Optional check)
         if any(isinstance(arg, type) and issubclass(arg, CallToolResult) for arg in args if arg is not type(None)):
             raise InvalidSignature(
```
maxisbey, line 306: Two things are stacked in this branch — one I read as a deliberate trade-off, one I don't think is.

The deliberate one: a union arm losing its output schema is exactly what `test_union_with_input_required_result_yields_no_output_schema` pins (it uses `-> str | InputRequiredResult`; `-> SomeModel | InputRequiredResult` hits the same early return), so I'm not treating it as a miss. I'd still push back on it. Nothing in SEP-2322 stops a tool from declaring an `outputSchema` *and* returning `input_required` mid-flight; `convert_result` already returns an `InputRequiredResult` before the output-model validation, so leaving the schema on is safe; and `SomeModel | InputRequiredResult` is the natural signature for precisely the tool this feature exists for, which now silently demotes its result to a text blob and advertises no schema.

The one I don't think was intended: this `return` sits before the existing `CallToolResult`-in-union `InvalidSignature`. So a signature that's a hard error today becomes silently accepted by adding `| InputRequiredResult` — `-> CallToolResult | str` raises, `-> CallToolResult | str | InputRequiredResult` doesn't. None of the new tests cover a three-arm union, so I don't think the bypass was the point of the new arm.

Both fall out of the same change: instead of bailing out when the union contains `InputRequiredResult`, strip those arms and let the existing derivation run on the residual. That gives `-> InputRequiredResult` → no schema (nothing left), `-> CallToolResult | InputRequiredResult` → residual is bare `CallToolResult` → accepted with no schema, so `test_call_tool_result_unioned_with_input_required_result_is_accepted` passes exactly as written, `-> SomeModel | InputRequiredResult` → schema derived from `SomeModel`, and `-> CallToolResult | str | InputRequiredResult` → `InvalidSignature`, consistent with `-> CallToolResult | str`. If you'd rather keep the no-schema behaviour on purpose, the second half — making sure the three-arm `CallToolResult` case still raises — is the part I'd hold the line on.

src/mcp/shared/inbound.py
maxisbey: Done in 3507ce7d — `"number"` dropped from the set, the harness-concession comment replaced, and the type-check reason now says "integer/string/boolean". The harness's stale wording is already tracked at [conformance#344](https://github.com/modelcontextprotocol/conformance/issues/344) so no new upstream issue.

https://github.com/modelcontextprotocol/python-sdk/blob/3507ce7d/src/mcp/shared/inbound.py#L84-L87

src/mcp/shared/inbound.py
maxisbey: Done in 3507ce7d as the iterative `_walk_schema_positions` walk: yields every JSON-Schema-2020-12 schema position with its pure-`properties` path (or `None` once any other applicator is crossed); instance-data keywords (`default`/`examples`/`const`/`enum`) and `$ref`/`$dynamicRef` are not entered so termination is structural — no cycle guard. The validator runs token/type/uniqueness over every on-chain hit (one `seen` for the whole walk) and rejects anything off-chain or at the root. Reason strings now name the dotted property path. Docstring updated to match.

Defaulted unknown-extension keywords as "do not walk" (schema positions only) and 2020-12 only (no draft-07 tuple-`items`); say if you want either widened.

https://github.com/modelcontextprotocol/python-sdk/blob/3507ce7d/src/mcp/shared/inbound.py#L114-L210

src/mcp/server/_streamable_http_modern.py
maxisbey: Agreed and dropped in 3507ce7d — `_drop_invalid_header_tools` (def, call site, import) and its three tests are gone; `find_invalid_x_mcp_header` is now consumed only by `ClientSession.list_tools`. Conformance still 11/11 on `http-invalid-tool-headers` with the client filter alone. Registration-time author warning in `ToolManager.add_tool` is the better shape if we want server-side feedback later — left as a follow-up.

tests/server/test_streamable_http_modern.py
maxisbey: Gone with the filter in 3507ce7d — all three tests deleted along with the function.

src/mcp/server/mcpserver/context.py
maxisbey: Done in 3507ce7d as suggested — `Context` takes `input_params: InputResponseRequestParams | None` from the three handler sites; the props read straight off it. `_INPUT_RESPONSES_ADAPTER`, the wire-alias literals, and the unreachable validation are gone. Existing tests already cover both branches (out-of-request `Context` sites pass the default `None`).

https://github.com/modelcontextprotocol/python-sdk/blob/3507ce7d/src/mcp/server/mcpserver/context.py#L226-L237

src/mcp/server/mcpserver/utilities/func_metadata.py
maxisbey: Done in 3507ce7d — IIR arms (incl. subclasses) are stripped and the residual falls through to the existing derivation, so `SomeModel | InputRequiredResult` derives `SomeModel`'s schema, `CallToolResult | str | InputRequiredResult` raises `InvalidSignature` consistently, and the bare-IIR / `CallToolResult | IIR` cases keep their no-schema verdict. One `noqa: UP007` on the `Union[residual]` rebuild — PEP 604 has no syntax for a runtime-tuple union and `reduce(or_, ...)` breaks on 3.10 with typing special forms.

https://github.com/modelcontextprotocol/python-sdk/blob/3507ce7d/src/mcp/server/mcpserver/utilities/func_metadata.py#L279-L304

src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -266,10 +276,29 @@ def func_metadata(
...
+        # InputRequiredResult is a control-flow signal, not data: strip it so the residual arms
+        # drive schema derivation. convert_result short-circuits on an InputRequiredResult instance
+        # before output validation, so the schema only ever sees the data arms at runtime.
+        residual = tuple(a for a in args if not _is_input_required_type(a))
+        if not residual:
+            return FuncMetadata(arg_model=arguments_model)
+        if len(residual) != len(args):
+            # PEP 604 has no syntax for "union of a runtime tuple"; Union[...] is the only spelling.
+            return_type_expr = residual[0] if len(residual) == 1 else Union[residual]  # noqa: UP007
+            effective_annotation = return_type_expr
```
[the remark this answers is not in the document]
maxisbey: This was a deliberate default for this round (flagged as an open question in the review discussion). The strip happens after `inspect_annotation` has already separated the metadata, so re-wrapping the residual as `Annotated[(residual_expr, *metadata)]` is a small follow-up if we want it; for now `Annotated[X | InputRequiredResult, meta]` is treated the same as `X | InputRequiredResult`.

src/mcp/server/mcpserver/utilities/func_metadata.py
```diff
@@ -266,10 +276,29 @@ def func_metadata(
...
-        # Check if CallToolResult appears in the union (excluding None for Optional check)
-        if any(isinstance(arg, type) and issubclass(arg, CallToolResult) for arg in args if arg is not type(None)):
+        # InputRequiredResult is a control-flow signal, not data: strip it so the residual arms
+        # drive schema derivation. convert_result short-circuits on an InputRequiredResult instance
+        # before output validation, so the schema only ever sees the data arms at runtime.
+        residual = tuple(a for a in args if not _is_input_required_type(a))
+        if not residual:
+            return FuncMetadata(arg_model=arguments_model)
+        if len(residual) != len(args):
+            # PEP 604 has no syntax for "union of a runtime tuple"; Union[...] is the only spelling.
```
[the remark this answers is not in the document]
maxisbey: Same as the adjacent thread — known edge case, defaulted for this round. The re-wrap you describe is exactly the shape if we take it.

Kludex (approved):

PR #2979, opened 2026-06-26
fix(server): validate elicitation form/url sub-capabilities in check_capability

### LINKED ISSUE #2965: [Bug] check_capability does not validate elicitation sub-capabilities (form/url)
## Summary

`Connection.check_capability()` returns `True` when a client has URL-mode elicitation but the caller checks for form-mode elicitation — the sub-capability (`ElicitationCapability.form` / `.url`) is never inspected.

## Root cause

In `src/mcp/server/connection.py:340-341`:

```python
if capability.elicitation is not None and have.elicitation is None:
    return False
```

This only returns `False` when the client has **no** elicitation at all. It does **not** check individual sub-capabilities (`form` / `url`). Compare with `sampling`, which correctly checks sub-capabilities:

```python
if capability.sampling is not None:
    if have.sampling is None:
        return False
    if capability.sampling.context is not None and have.sampling.context is None:
        return False
    if capability.sampling.tools is not None and have.sampling.tools is None:
        return False
```

## Reproduction

```python
# Client supports URL elicitation only
have = ClientCapabilities(elicitation=ElicitationCapability(url=UrlElicitationCapability()))

# Check for form elicitation — should be False (client does not have form)
want = ClientCapabilities(elicitation=ElicitationCapability(form=FormElicitationCapability()))

Connection.from_envelope("2025-11-25", client_info, have).check_capability(want)  # Returns True (BUG)
```

## Impact

- `check_capability` is public API (via `ServerSession.check_client_capability`)
- Currently no production callers use elicitation sub-capability checks — low immediate impact
- But once someone relies on it (e.g., checking if the client supports form elicitation before calling `elicit_form`), it will return wrong results

## Additional gaps (lower priority)

- `extensions` — `ClientCapabilities.extensions` is not checked at all
- `tasks` — `ClientCapabilities.tasks` and sub-capabilities entirely unhandled

## Proposed fix

Add `form` / `url` sub-capability checks, matching the `sampling` pattern:

```python
if capability.elicitation is not None:
    if have.elicitation is None:
        return False
    if capability.elicitation.form is not None and have.elicitation.form is None:
        return False
    if capability.elicitation.url is not None and have.elicitation.url is None:
        return False
```

AI assistance: Bug discovered and analyzed with AI assistance (opencode).

syf2211:
## Summary

`Connection.check_capability()` now validates elicitation sub-capabilities (`form` / `url`) when the caller specifies them, instead of only checking whether the client declared elicitation at all.

## Motivation

Fixes #2965.

When a client supports URL-mode elicitation only, `check_capability(ClientCapabilities(elicitation=ElicitationCapability(form=FormElicitationCapability())))` incorrectly returned `True`. The function only compared top-level `have.elicitation is None`, never inspecting individual sub-capabilities.

This mirrors the existing `sampling.context` / `sampling.tools` sub-capability checks in the same function.

## Changes

- **`src/mcp/server/connection.py`**: Add nested `form` / `url` presence checks under the elicitation branch.
- **`tests/server/test_connection.py`**: Add 6 parametrized regression cases covering form/url mismatch, match, and generic elicitation checks.

## Tests

- `uv run pytest tests/server/test_connection.py` — 44 passed
- `uv run ruff check src/mcp/server/connection.py tests/server/test_connection.py` — clean
- `uv run ruff format --check src/mcp/server/connection.py tests/server/test_connection.py` — clean

## Notes

- `ServerSession.check_client_capability()` delegates to `Connection.check_capability()`, so this fix applies to both APIs.
- AI assistance disclosure: fix and tests prepared with AI assistance; locally verified and reviewed before submission.

### DIFF
--- a/src/mcp/server/connection.py
+++ b/src/mcp/server/connection.py
@@ -337,8 +337,13 @@ class Connection:
                 return False
             if capability.sampling.tools is not None and have.sampling.tools is None:
                 return False
-        if capability.elicitation is not None and have.elicitation is None:
-            return False
+        if capability.elicitation is not None:
+            if have.elicitation is None:
+                return False
+            if capability.elicitation.form is not None and have.elicitation.form is None:
+                return False
+            if capability.elicitation.url is not None and have.elicitation.url is None:
+                return False
         if capability.experimental is not None:
             if have.experimental is None:
                 return False
--- a/tests/server/test_connection.py
+++ b/tests/server/test_connection.py
@@ -21,6 +21,7 @@ from mcp_types import (
     CreateMessageRequestParams,
     ElicitationCapability,
     EmptyResult,
+    FormElicitationCapability,
     Implementation,
     ListRootsRequest,
     ListRootsResult,
@@ -31,6 +32,7 @@ from mcp_types import (
     SamplingCapability,
     SamplingContextCapability,
     SamplingToolsCapability,
+    UrlElicitationCapability,
 )
 from mcp_types.version import LATEST_HANDSHAKE_VERSION, LATEST_MODERN_VERSION
 from pydantic import BaseModel, ValidationError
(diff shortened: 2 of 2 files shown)


itxaiohanglover:
Good fix! Validating each elicitation sub-capability (form/url) separately is more precise than the previous all-or-nothing check. The parametrized test matrix covering form-only, url-only, both, and mismatched cases is thorough.

PR #2988, opened 2026-06-26
Support RFC 8693 token exchange for enterprise IdP flows (SEP-990)

Kludex:
## Summary

Implements [SEP-990 (enterprise IdP policy controls during MCP OAuth flows)](https://modelcontextprotocol.io/seps/990-enable-enterprise-idp-policy-controls-during-mcp-o). The wire mechanism is the RFC 8693 token-exchange grant (`urn:ietf:params:oauth:grant-type:token-exchange`): a client exchanges an enterprise IdP-issued Identity Assertion Authorization Grant (ID-JAG) for an MCP access token at the authorization server's token endpoint.

This is additive and opt-in on both sides; existing flows are unchanged.

## Client

`TokenExchangeOAuthProvider` (`mcp.client.auth.extensions.token_exchange`) is an `httpx.Auth` that posts the exchange, mirroring the `client_credentials` extension. The ID-JAG is supplied lazily via an async `subject_token_provider(audience)` callback — the SDK does not implement IdP login or the first exchange against the IdP (deployment-specific). Public (default, `none`) and confidential (`client_secret_post`/`client_secret_basic`) clients are supported; `requested_token_type` defaults to the access-token URN.

## Server

`OAuthAuthorizationServerProvider.exchange_token` validates the subject token (signature/issuer/audience/expiry/policy — the provider's responsibility) and issues the token. Gated by `AuthSettings(token_exchange_enabled=True)`, which:

- advertises the grant + the public-client `none` auth method in AS metadata, **and**
- gates the `/token` endpoint: with the flag off, `/token` rejects the grant with `unsupported_grant_type` even if the provider implements `exchange_token`.

`TokenError` gains RFC 8693's `invalid_target`; `TokenExchangeToken` carries the required `issued_token_type` (the handler defaults it so responses are compliant regardless of provider). `TokenExchangeRequest` enforces RFC 8693 actor-token pairing.

The provider owns scope decisions: requested scopes are never granted verbatim. The example server narrows the request against an allowed set so a valid ID-JAG cannot be exchanged for broader access than policy permits.

## Review

Reviewed by Codex across two rounds. Round one raised seven should-fix findings; all were addressed (dispatch gating, `issued_token_type`, scope-escalation hardening in the example, `invalid_target`, actor-token pairing, Basic-auth confidential client, public-client advertisement). The follow-up review confirmed every finding resolved with no new Critical/Should-fix issues.

## Testing

- Unit + ASGI HTTP tests for the client extension and the server token handler (13 server, 7 client, plus shared-model coverage).
- Verified end to end over **real Uvicorn sockets** with a real RS256-signed JWT acting as the ID-JAG: AS metadata discovery, raw `/token` exchange, forged-signature rejection, and a full MCP session (discovery → 401 → token exchange → bearer-gated tool call) all pass.
- `./scripts/test`: 100% coverage, `strict-no-cover` clean, `pyright` clean.

This exercise surfaced a separate AS-metadata issuer normalization bug, fixed in #2987 (merged); this branch builds on it.

Docs: new SEP-990 section in `docs/migration.md`; client + server snippets under `examples/snippets/`.

## AI Disclaimer


maxisbey (changes requested):
I went through this fairly carefully against the SEP-990 normative text, RFC 8693, RFC 7523, and the sibling SDKs. There's one architectural issue that I think needs resolving before the rest is worth iterating on, plus a cluster of client-side discovery concerns and some smaller hardening points.

Reviewed at 55c2413.

## Wrong grant type for the MCP-AS leg

The normative SEP-990 text — [`ext-auth/specification/stable/enterprise-managed-authorization.mdx` §5](https://github.com/modelcontextprotocol/ext-auth/blob/fb374c7db2b34f18ca9183882e0beecdf661892b/specification/stable/enterprise-managed-authorization.mdx?plain=1#L210-L233) — specifies that the client presents the ID-JAG to the MCP authorization server using the **RFC 7523 jwt-bearer grant**:

> The MCP Client presents the ID-JAG to the Resource Authorization Server's token endpoint as defined in [§4.4 of draft-ietf-oauth-identity-assertion-authz-grant-04], using `grant_type` `urn:ietf:params:oauth:grant-type:jwt-bearer` and the ID-JAG as the `assertion`.

RFC 8693 token-exchange is leg 1 only (client → enterprise IdP, ID-token → ID-JAG), which this PR correctly scopes out via `subject_token_provider`. But for leg 2 the PR sends and accepts `grant_type=urn:ietf:params:oauth:grant-type:token-exchange` with `subject_token=<ID-JAG>` instead.

The other SDKs all match the spec on this — typescript-sdk's [`crossAppAccess.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/crossAppAccess.ts) uses token-exchange against the IdP and jwt-bearer against the MCP AS; go-sdk's [`auth/extauth/enterprise_handler.go`](https://github.com/modelcontextprotocol/go-sdk/blob/main/auth/extauth/enterprise_handler.go) and csharp-sdk's [`IdentityAssertionGrant.cs`](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Authentication/IdentityAssertionGrant.cs) likewise. A python-sdk AS built from this PR rejects all of them with `unsupported_grant_type` (the discriminated union doesn't include `jwt-bearer`), and the python client is rejected by any spec-conforming AS.

Beyond interop, the grant-type choice has security weight. RFC 7523 §3 carries normative validation rules (the JWT MUST contain `aud` identifying the AS, `iss`/`exp`/signature MUST be checked) that map directly onto SEP-990 §5.1's processing rules — `typ=oauth-id-jag+jwt`, `aud` = own issuer, `client_id` claim ↔ authenticated client. RFC 8693 imposes nothing on `subject_token`, so those checks become provider-discretionary rather than something the SDK can structurally encourage. The §5.1 MUST that "the issued access token MUST be audience-restricted to the MCP Server identified by the `resource` claim **in the ID-JAG**" is similarly hard to surface when the SDK never decodes the assertion.

Discovery is also different from what's implemented: ext-auth §6 says clients detect support via `authorization_grant_profiles_supported` containing `urn:ietf:params:oauth:grant-profile:id-jag` (per [draft-04 §7.2](https://www.ietf.org/archive/id/draft-ietf-oauth-identity-assertion-authz-grant-04.html#section-7.2)), not via `grant_types_supported` containing the token-exchange URN.

I think this means a reshape rather than a patch: a `JwtBearerRequest` (`grant_type` + `assertion`) on the server, a provider hook that's documented against RFC 7523 §3 + the §5.1 processing rules, and a client that posts `assertion` instead of `subject_token`. A lot of the smaller findings below (`issued_token_type`, `token_type: N_A`, actor-token pairing, multi-valued resource/audience) become moot under jwt-bearer since the response is plain RFC 6749.

## Client: pre-registered credentials go through RS-controlled discovery

`TokenExchangeOAuthProvider` reuses the base `OAuthClientProvider` 401→PRM→ASM discovery flow, which was designed around DCR — the `client_secret` comes *from* the discovered AS, so sending it back there is harmless. Here the secret is pre-supplied for a specific known AS, but the constructor offers no way to pin that AS, and `_fixed_client_info` is built with `issuer=None` ([token_exchange.py#L98-L105](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/client/auth/extensions/token_exchange.py#L98-L105)), so the existing `credentials_match_issuer` guard always passes.

Three concrete consequences I was able to demonstrate with `httpx.MockTransport` driving the real `async_auth_flow`:

- A hostile/compromised resource server that publishes `authorization_servers: ["https://attacker.example"]` in its PRM receives the confidential-client secret in the Basic header (or form body) on the first 401. The secret isn't audience-bound, so unlike the ID-JAG it's fully replayable against the real AS.
- On the legacy fallback (PRM 404s), AS metadata is fetched from the MCP server's own origin and `validate_metadata_issuer` is gated on `auth_server_url is not None` ([oauth2.py#L616-L617](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/client/auth/oauth2.py#L616-L617)) — which it isn't on this path. So the server can serve `{issuer:"https://corp-as.example", token_endpoint:"https://evil/steal"}`, `subject_token_provider` is called with the **real** AS issuer as audience, and a real-AS-bound ID-JAG plus the client secret go to the attacker.
- More generally, the audience handed to `subject_token_provider` is RS-influenced. go-sdk and csharp-sdk both require the integrator to explicitly configure the MCP AS URL; if the callback honours `audience` properly the aud-binding limits damage, but a callback that ignores it (or wraps a static-token helper) leaks a replayable credential.

I'd suggest an `expected_issuer: str` constructor parameter, set `_fixed_client_info.issuer = expected_issuer` so the existing guard fires, and assert `str(oauth_metadata.issuer) == expected_issuer` before invoking `subject_token_provider` or attaching the secret. The legacy-path issuer-check skip is arguably a pre-existing gap in the base class, but this PR is what makes it carry pre-registered credentials.

## Server-side hardening

- **Requested scopes bypass per-client registration.** [token.py#L274](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/server/auth/handlers/token.py#L274) passes `scope.split(" ")` straight to the provider with no `client_info.validate_scope()` check. The same client that gets `error=invalid_scope` at `/authorize` for `scope=admin` gets a 200 via this branch. For SEP-990 the provider *should* derive entitlement from the ID-JAG, so this may be intentional, but it's inconsistent with the other two grants and the test harness provider grants `params.scopes` verbatim.
- **DCR + this flag = unauthenticated path to `exchange_token`.** `OAuthClientMetadata.grant_types` is `list[Literal[...] | str]`, [register.py#L73](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/server/auth/handlers/register.py#L73) only requires `authorization_code` to be present, and `token_endpoint_auth_method="none"` is honoured. Anyone can self-register a public client that includes this grant and reach `provider.exchange_token` with an attacker-chosen `client` and arbitrary `subject_token`. The IETF draft's §8.1 says this flow "SHOULD only be supported for confidential clients," and ext-auth §5 says "the MCP Client authenticates with its credentials" — defaulting to `none` and advertising it in metadata sits uneasily with that. Stripping the grant from DCR-registered clients (require pre-registration) would be the simple fix.
- **`refresh_token` passthrough.** The wrap at [token.py#L289](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/server/auth/handlers/token.py#L289) preserves a provider-set `refresh_token`. Returning one defeats the IdP lifetime control that SEP-990 exists to provide (off-boarded user keeps minting tokens). Probably worth dropping it in the wrap, or at least flagging in the docstring.
- **Issued token bound to the request's `resource`, not the ID-JAG's `resource` claim.** ext-auth §5.1: the issued token MUST be audience-restricted to the resource named **in the ID-JAG**. The handler surfaces only the form-body `resource` ([token.py#L275](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/server/auth/handlers/token.py#L275)), and the example writes it straight into `AccessToken.resource`. A client holding an ID-JAG for resource A can send `resource=B` and the example mints a B-audience token.
- **`client` arg may be unauthenticated.** For `none`-method clients, `client.client_id` is self-asserted; the [`exchange_token` docstring](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/server/auth/provider.py#L307) doesn't warn that it must not drive authorization decisions. ext-auth §5.1 requires the ID-JAG's `client_id` claim to match the authenticated client — worth surfacing as a documented requirement on the hook.

## `subject_token_provider` callback signature

The callback receives only `audience` ([token_exchange.py#L59](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/client/auth/extensions/token_exchange.py#L59)). ext-auth §4 has the leg-1 request to the IdP carrying both `audience` (the AS issuer — which is what's passed, correctly) and `resource` (the MCP server's RFC 9728 identifier); the resulting ID-JAG MUST carry that `resource` claim (§4.3) for §5.1 to bind against. The SDK has the PRM-derived value via `self.context.get_resource_url()` but never surfaces it to the callback. Widening to `(audience, resource)` would let callers send a correct leg-1 request without hardcoding the server URL.

## Example server (`examples/snippets/servers/token_exchange_server.py`)

A few things here that I'd tighten regardless of how the grant-type question lands, since this is the file people will copy:

- `_validate_id_jag` accepts any non-empty string. The repo's existing [`oauth_server.py`](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/examples/snippets/servers/oauth_server.py) stub fails closed (returns `None`); this one fails open. `raise NotImplementedError(...)` would match the convention and force the replacement.
- Line 60/77 stores the raw `subject_token` string as `AccessToken.subject` — a bearer credential in a field that's designed to be logged and surfaced as identity (`bearer_auth.py` propagates it into the auth context).
- Line 59 writes client-supplied `params.resource` into `AccessToken.resource` with no `invalid_target` check (ties into the §5.1 point above).
- Line 48 returns `invalid_grant` for a rejected subject token; RFC 8693 §2.2.2 specifies `invalid_request` for that case.
- It never inspects `subject_token_type`, `actor_token`, `requested_token_type`, or `audience` — a request carrying `actor_token` (delegation) is silently treated as plain impersonation.

## Smaller / informational

- `OAuthToken.token_type` is `Literal["Bearer"]`, so `TokenExchangeToken` can't emit RFC 8693's `N_A`, and the client rejects a third-party AS that returns it. Moot if reshaped to jwt-bearer.
- The client never checks `issued_token_type` against `requested_token_type`; the sibling SDKs all do. Also moot post-reshape.
- Advertising `none` in `token_endpoint_auth_methods_supported` is endpoint-wide (RFC 8414 has no per-grant slot); enforcement is per-registered-client so there's no bypass, but it tells auth-code clients the AS accepts unauthenticated `/token`.
- `TokenExchangeRequest` is in the union unconditionally, so a server with the flag off still emits RFC 8693 field-level validation errors (e.g. the actor-token-pairing message) before the gate at line 260 runs.
- The `# pragma: no cover` on the `oauth_metadata` guard at [token_exchange.py#L119-L120](https://github.com/modelcontextprotocol/python-sdk/blob/55c2413616f38abeec317d2766341a7914ecbf37/src/mcp/client/auth/extensions/token_exchange.py#L119-L120) is reachable when ASM discovery 404s.
- `OAuthAuthorizationServerProvider` is a `Protocol`, so the default `exchange_token` body is only inherited by explicit subclasses; a structural-only implementer hits `AttributeError` → 500 rather than `unsupported_grant_type`.
- `scope.split(" ")` forwards empty-string entries on multi-space input.

Happy to share the repro tests for any of the above if useful.

Kludex:
Thanks for the careful review against the normative text - you were right on the core point, and I've reshaped the PR accordingly (pushed in 7783020b).

**Grant type (leg 2).** Now uses `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer` with the ID-JAG as `assertion`, per ext-auth §5. Server: `JwtBearerRequest` + `exchange_identity_assertion(IdentityAssertionParams)` returning a plain RFC 6749 `OAuthToken`. Client: `IdentityAssertionOAuthProvider` posts `assertion`. Discovery advertises `urn:ietf:params:oauth:grant-profile:id-jag` in `authorization_grant_profiles_supported` (§6), not the token-exchange grant. As you noted, the `issued_token_type` / `N_A` / actor-token / multi-resource findings all became moot and are gone.

**Confidential clients only.** The handler rejects `token_endpoint_auth_method=none`, DCR refuses the jwt-bearer grant (pre-registration required), and metadata no longer advertises `none`.

**Client AS pinning.** New `expected_issuer` parameter: `_fixed_client_info.issuer` is set to it (so the SEP-2352 guard fires) and the provider asserts the discovered issuer matches before releasing the ID-JAG or secret. The `assertion_provider` callback now takes `(audience, resource)` so the ID-JAG can carry the `resource` claim §5.1 binds against. The provider docstring spells out the RFC 7523 §3 / §5.1 duties (decode the assertion, match the `client_id` claim to the authenticated client, audience-restrict to the ID-JAG's `resource`, no refresh token).

**Example.** Rewritten to fail closed (raises rather than trusting), bind to the ID-JAG `resource`, and implement `get_client` (it would otherwise 401 before the exchange).

Two server-side items I'd flag for your call rather than assume: requested-scope validation and binding the issued token to the ID-JAG's `resource` claim are documented as the provider's responsibility (the SDK can't enforce them without decoding the assertion). If you'd prefer the SDK take a firmer stance there, happy to iterate.

Validated end to end (full client+server stack, including an issuer-pinning case); 100% coverage, strict-no-cover/pyright/ruff clean. I have not been able to test against a live Okta CAA tenant (gated preview), so real-IdP interop is unverified - flagging that explicitly.

## AI Disclaimer

src/mcp/client/auth/extensions/identity_assertion.py
```diff
@@ -0,0 +1,156 @@
...
+        if not self.context.client_info:
+            raise OAuthFlowError("Missing client info for identity assertion grant")  # pragma: no cover
+
+        # Pin the authorization server: the metadata issuer is discovered via the resource server,
+        # which is untrusted. Refuse to release the ID-JAG or secret to an unexpected AS.
+        issuer = str(self.context.oauth_metadata.issuer)
+        if issuer != self._expected_issuer:
+            raise OAuthFlowError(
+                f"Authorization server issuer {issuer} does not match expected {self._expected_issuer}"
+            )
```
maxisbey: **Issuer pinning does not validate token_endpoint — credentials can leak on legacy discovery path**

The `expected_issuer` pin only compares `oauth_metadata.issuer` — a self-reported field from an RS-sourced document — and never validates the `token_endpoint` the credentials are actually POSTed to. On the legacy no-PRM fallback, ASM is fetched from the resource server's own origin and `validate_metadata_issuer` is skipped (`auth_server_url is None`), so a hostile RS can serve `{issuer: <expected_issuer>, token_endpoint: https://attacker/...}`; both the SEP-2352 guard and this check pass, and the pre-provisioned `client_secret` plus a real-AS-audienced ID-JAG are sent to the attacker.

This is the second bullet of the review's RS-controlled-discovery finding, which the commit message/docstring/docs claim `expected_issuer` fixes. Consider also asserting that `token_endpoint` (and/or the ASM fetch origin) matches `expected_issuer`'s origin, or fetching ASM directly from `expected_issuer` instead of inheriting the base flow's RS-driven discovery.

src/mcp/client/auth/extensions/identity_assertion.py
```diff
@@ -0,0 +1,156 @@
...
+        self._initialized = True
+
+    async def _perform_authorization(self) -> httpx.Request:
+        """Perform the jwt-bearer grant with the ID-JAG."""
+        return await self._exchange_assertion()
+
+    async def _exchange_assertion(self) -> httpx.Request:
+        """Build the RFC 7523 jwt-bearer token request carrying the ID-JAG."""
+        if not self.context.oauth_metadata:
+            raise OAuthFlowError("Missing OAuth metadata for identity assertion grant")  # pragma: no cover
```
maxisbey: **[nit] Reachable `# pragma: no cover` on oauth_metadata guard**

This `# pragma: no cover` marks a reachable branch: if PRM and ASM discovery both 404 (legacy server), `oauth_metadata` stays `None`, the pinned `_fixed_client_info` means Step 4 is skipped, and Step 5 calls `_exchange_assertion()` straight into this raise. The maintainer review already flagged this under "Smaller / informational" and it wasn't picked up in the reshape — per the repo convention `# pragma: no cover` is only for genuinely unreachable code, so add a test for the ASM-404 path and drop the pragma.

src/mcp/client/auth/extensions/identity_assertion.py
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        headers: dict[str, str] = {"Content-Type": "application/x-www-form-urlencoded"}
+        token_data, headers = self.context.prepare_token_auth(token_data, headers)
+
+        if self.context.should_include_resource_param(self.context.protocol_version):
+            token_data["resource"] = resource
+
+        if self._scopes:
+            token_data["scope"] = self._scopes
```
maxisbey: **[nit] 403 insufficient_scope step-up is a no-op — re-exchange sends constructor `_scopes`, not the stepped-up union**

Reading `self._scopes` here fixes the 401-path overwrite but breaks the inherited SEP-2350 403 `insufficient_scope` step-up: the base `async_auth_flow` writes the unioned scope to `context.client_metadata.scope` and re-calls `_perform_authorization()`, but `_exchange_assertion` ignores that and re-sends the constructor value, so the retried exchange requests the same scope and the 403 recurs.

The sibling `ClientCredentialsOAuthProvider` reads `client_metadata.scope` and so honours step-up. Consider sending `union_scopes(self._scopes, self.context.client_metadata.scope)` (or restoring `self._scopes` onto `client_metadata.scope` after the 401-path scope-selection step instead of caching).

maxisbey (changes requested):
Thanks for the quick reshape — went back through at 7783020b and everything you listed checks out. The grant-type, discovery field, confidential-only posture, DCR rejection, callback signature, AS-pinning constructor, fail-closed example, and the docstring duties are all in place, and a TS/Go/C#-shaped jwt-bearer request now reaches `exchange_identity_assertion` end-to-end.

On your two flagged items: I'm fine with scope bounding and `resource`-claim binding staying as documented provider responsibilities. The SDK can't enforce either without decoding the assertion, and the docstring + example now make the contract unambiguous.

There's one thing I think still needs closing before this is safe to ship, plus a smaller hardening point and some nits.

## Legacy no-PRM path still leaks credentials

The `expected_issuer` pin at [identity_assertion.py#L131-L135](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/client/auth/extensions/identity_assertion.py#L131-L135) checks the metadata document's self-reported `issuer` field — but not where the document was fetched from, nor where `token_endpoint` points. On the legacy fallback (PRM 404s → ASM fetched from the MCP server's own origin, [utils.py#L164-L168](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/client/auth/utils.py#L164-L168)), `validate_metadata_issuer` is still skipped because [`auth_server_url` is None](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/client/auth/oauth2.py#L616-L617). So a hostile resource server that 404s PRM and serves `{issuer: <expected_issuer>, token_endpoint: "https://evil/steal"}` from its own well-known passes both the SEP-2352 guard and the new pin; `assertion_provider` is called with `audience=<real-AS>` (so an IdP audience allowlist is satisfied), and the ID-JAG plus `client_secret` go to the attacker. I have a MockTransport repro that completes the flow silently with an attacker-issued access token — happy to attach.

Since `expected_issuer` is now a required input, the simplest fix is to set `self.context.auth_server_url = self._expected_issuer` in `_initialize()`. ASM is then fetched directly from the pinned AS's well-known, `validate_metadata_issuer` runs, and the resource server can't forge the document at all — which is what go-sdk and csharp-sdk effectively do by taking the AS URL as config. That also closes a smaller side-effect on the PRM path: when PRM names an unexpected AS, the SEP-2352 guard currently clears `client_info` and the base flow attempts DCR against the attacker's `/register` (no secret leaked, but it discloses the client metadata and overwrites `storage`) before the pin at line 132 raises.

## Confidential-client gate trusts the declared method

[token.py#L255](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/server/auth/handlers/token.py#L255) rejects only `token_endpoint_auth_method == "none"`, but `ClientAuthenticator` (unchanged) only compares secrets `if client.client_secret:` — a stored record with `client_secret=None` is returned without any credential check. So a pre-registered client with `token_endpoint_auth_method="client_secret_post"` and no secret passes both and reaches the hook unauthenticated, which makes the docstring's "the handler guarantees `client` is confidential" stronger than what's enforced. DCR-minted clients aren't affected (register.py always issues a secret for non-`none`); it only bites operator-provisioned records. Gating on `client_info.client_secret` being set (or having `ClientAuthenticator` reject a non-`none` client with no stored secret) would make the guarantee true.

## Smaller

- [token.py#L258](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/server/auth/handlers/token.py#L258): RFC 6749 §5.2 reserves `invalid_client` for failed authentication; "authenticated client may not use this grant" is `unauthorized_client`.
- Stale RFC 8693 citations: [`provider.py#L101`](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/server/auth/provider.py#L101) (`invalid_target` is now used for RFC 8707) and [`identity_assertion_server.py#L97-L98`](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/examples/snippets/servers/identity_assertion_server.py#L97-L98) (RFC 7523 §3.1 specifies `invalid_grant` for a rejected assertion).
- The §5.1 checklist in [`provider.py#L300-L311`](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/server/auth/provider.py#L300-L311) and the example header omit RFC 7523 §3's `sub` MUST and the `jti`/replay guidance — worth adding since the docstring is the only thing steering implementers.
- The module docstring and migration.md say "SEP-990 §5.1 requires a confidential client, so `client_secret` is mandatory" — the spec requires client authentication, not specifically a shared secret (ext-auth §5 mentions `private_key_jwt` for the CIMD case). Probably worth wording as "this SDK currently requires a shared secret" so it doesn't read as a spec constraint.
- `JWT_BEARER_GRANT_TYPE` is duplicated as a literal in four files — minor.

The carry-over info-level items from the first round (request model parsed before the gate; Protocol default body only inherited by explicit subclasses; the `# pragma: no cover` at [identity_assertion.py#L125](https://github.com/modelcontextprotocol/python-sdk/blob/7783020b615ade0c1083fc891e9d8a800382a063/src/mcp/client/auth/extensions/identity_assertion.py#L125) still being reachable; `scope.split(" ")` empty entries) all still apply but I don't think any of them block.

Kludex:
Thanks again. Your review landed against 7783020b just before two follow-up pushes (0f4052cc, 05f60ce3); everything is now addressed.

**Legacy no-PRM credential leak (blocking).** Fixed as you suggested: `_initialize` pins `auth_server_url` to `expected_issuer`, so ASM is discovered from the pinned AS's well-known and `validate_metadata_issuer` runs. I also override `_validate_resource_match` to reject a PRM that names any other AS, which closes the PRM-path DCR-disclosure side-effect you noted (the flow now aborts before any `/register`). An interaction test drives the attack end to end.

**Confidential-client gate.** Already tightened in 0f4052cc (predated your review): it now gates on a stored `client_secret`, so a secret-less `client_secret_post` record is rejected before the hook. Changed the error to `unauthorized_client` per your RFC 6749 §5.2 point.

**Docs/nits.** `invalid_target` now cited against RFC 8707; the `exchange_identity_assertion` checklist adds the RFC 7523 `sub` MUST and `jti`/replay guidance; example citation fixed to RFC 7523 §3.1 (0f4052cc); reworded the "confidential client" line to "the spec requires client authentication; this SDK requires a shared secret."

On the two you're fine leaving as provider responsibilities (scope bounding, `resource`-claim binding) - kept as documented. Did not dedupe the `JWT_BEARER_GRANT_TYPE` literal (no natural shared home); the remaining info-level carry-overs I left as non-blocking per your note.

100% coverage, strict-no-cover/pyright/ruff clean. Still no live-Okta-CAA validation (gated preview) - flagging again.

## AI Disclaimer

maxisbey (approved):

PR #371, opened 2025-03-25
Added CORS middleware to allow cross-origin requests

jerome3o-anthropic:
- Allow any origin to make requests
- Allow GET, POST, and OPTIONS HTTP methods
- Allow any headers
- Allow sending credentials with requests

Also added OPTIONS method to auth routes to handle CORS preflight requests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #2990, opened 2026-06-26
Mirror x-mcp-header tool arguments into Mcp-Param-* request headers (SEP-2243)

Kludex:
## What

Implements client-side `Mcp-Param-*` header emission ([SEP-2243](https://modelcontextprotocol.io/specification/draft/basic/transports/streamable-http#custom-headers-from-tool-parameters) "Custom Headers from Tool Parameters"). The client already validated and dropped tools with malformed `x-mcp-header` annotations but never emitted the headers - the gap the conformance `expected-failures` file named as the unimplemented S8 design.

On a modern (2026-07-28) Streamable HTTP connection, a `tools/call` now mirrors each argument annotated with `x-mcp-header` into an `Mcp-Param-<name>` header: string verbatim, integer as decimal, boolean as `true`/`false`, base64-sentinel-wrapped when not header-safe. Null or absent arguments are omitted, unannotated parameters are never mirrored, and the argument stays in the request body (additive, per spec).

The tool schema comes from the tool's last `list_tools` entry (annotations are cached); a tool the client never listed emits no `Mcp-Param-*` headers (spec-permitted).

## Changes

- `mcp.shared.inbound`: `x_mcp_header_map()` (schema -> path->token) and `mcp_param_headers()` (args -> headers), reusing the existing schema walk and `encode_header_value` codec.
- `ClientSession`: modern stamp injects param headers for `tools/call`; `list_tools` caches each kept tool's annotations and evicts the entry when a re-list drops the tool.
- Conformance `client.py`: adds the `http-custom-headers` handler (replays the harness `toolCalls`).
- `docs/migration.md` updated.

## Conformance

`http-custom-headers` now passes 18/18 on both the default and 2026-07-28 client legs and is removed from both expected-failures baselines.

The harness fixture annotated `number`-typed properties with `x-mcp-header`, which the spec forbids ("Parameters with type `number` are not permitted"); a conformant client drops those tools, so the scenario was previously unpassable. That is fixed upstream in [modelcontextprotocol/conformance#371](https://github.com/modelcontextprotocol/conformance/pull/371). CI is temporarily pinned to the `pkg.pr.new` preview of #371, with a `CONFORMANCE_PKG_SHA256` digest + fetch-and-verify step (the repo's established pattern for URL pins) so the mutable URL is verified, not trusted. Repin to the published release once #371 ships.

All tests pass, 100% coverage, `strict-no-cover`/pyright/ruff clean.

## AI Disclaimer

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -21,9 +21,13 @@ env:
   #
   # Temporarily pinned to the pkg.pr.new preview build of conformance#371, which
   # fixes the http-custom-headers fixture's spec-forbidden `number`-typed
-  # x-mcp-header annotations. Repin to the published release that includes #371
-  # once it ships (the preview URL is ephemeral).
+  # x-mcp-header annotations. Because this is a mutable URL (not a registry
+  # spec), CONFORMANCE_PKG_SHA256 pins the tarball and the fetch-and-verify step
+  # below downloads, checks the digest, and repoints CONFORMANCE_PKG at the
+  # verified local copy. Repin to the published release that includes #371 once
+  # it ships, then drop CONFORMANCE_PKG_SHA256 and the fetch-and-verify steps.
   CONFORMANCE_PKG: "https://pkg.pr.new/@modelcontextprotocol/conformance@371"
+  CONFORMANCE_PKG_SHA256: "9d8b25874d55e304b006cbaa066571773582f5828143c53a2b8a6830f203ca1d"
 
 jobs:
   server-conformance:
@@ -39,6 +43,19 @@ jobs:
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 24
+      - name: Fetch and verify conformance harness
+        # Only when CONFORMANCE_PKG is a URL: download, check the recorded
+        # sha256, and re-point CONFORMANCE_PKG at the verified local tarball.
+        # When CONFORMANCE_PKG is a registry spec, this step is a no-op (npm's
+        # own integrity check applies).
+        run: |
+          case "$CONFORMANCE_PKG" in
+            https://*)
+              curl -fsSL "$CONFORMANCE_PKG" -o /tmp/conformance.tgz
+              echo "$CONFORMANCE_PKG_SHA256  /tmp/conformance.tgz" | sha256sum -c -
+              echo "CONFORMANCE_PKG=file:/tmp/conformance.tgz" >> "$GITHUB_ENV"
+              ;;
+          esac
       - run: uv sync --frozen --all-extras --package mcp-everything-server
       - name: Run server conformance (active suite)
         run: >-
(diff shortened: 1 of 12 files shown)


src/mcp/shared/inbound.py
```diff
@@ -206,6 +209,51 @@ def find_invalid_x_mcp_header(input_schema: Any) -> str | None:
...
+    For each `(path, token)` in `header_map`, read the value at that property
+    path in `arguments` and, when it is present and not `None`, emit
+    `Mcp-Param-<token>` carrying it: `bool` as `true`/`false`, other scalars via
+    `str`, each passed through :func:`encode_header_value` so a non-token value
+    is base64-wrapped. A path that hits a missing key or a non-mapping node is
+    skipped, matching the spec's "omit the header when no value is present".
+    """
+    headers: dict[str, str] = {}
+    for path, token in header_map.items():
+        node: Any = arguments
```
Kludex: It shouldn't be Any here!

src/mcp/shared/inbound.py
```diff
@@ -206,6 +209,51 @@ def find_invalid_x_mcp_header(input_schema: Any) -> str | None:
...
+        for key in path:
+            if not isinstance(node, Mapping):
+                node = None
+                break
+            node = cast("Mapping[str, Any]", node).get(key)
```
Kludex: cast is not necessary

.github/workflows/conformance.yml
```diff
@@ -18,7 +18,12 @@ env:
   # Use a published version, e.g. @modelcontextprotocol/conformance@0.2.0-alpha.7.
   # Bump deliberately and reconcile both
   # .github/actions/conformance/expected-failures*.yml files in the same change.
-  CONFORMANCE_PKG: "@modelcontextprotocol/conformance@0.2.0-alpha.7"
+  #
+  # Temporarily pinned to the pkg.pr.new preview build of conformance#371, which
+  # fixes the http-custom-headers fixture's spec-forbidden `number`-typed
+  # x-mcp-header annotations. Repin to the published release that includes #371
+  # once it ships (the preview URL is ephemeral).
+  CONFORMANCE_PKG: "https://pkg.pr.new/@modelcontextprotocol/conformance@371"
```
maxisbey, line 29: **[nit] URL-based CONFORMANCE_PKG reintroduced without the SHA256 fetch-and-verify step**

This repins `CONFORMANCE_PKG` to a pkg.pr.new URL but doesn't restore the `CONFORMANCE_PKG_SHA256` env var and "Fetch and verify conformance harness" step that accompanied the previous URL-based pin (removed in #2974 only because it switched to a registry spec). `npx --yes "$CONFORMANCE_PKG"` now executes an unverified, mutable remote tarball in CI — a regression from the repo's own supply-chain posture for this exact pattern.

Either restore the SHA256 + fetch-and-verify step from pre-#2974 (curl → `sha256sum -c` → repoint to `file:/tmp/conformance.tgz`), or keep the `@0.2.0-alpha.7` registry pin and leave `http-custom-headers` in expected-failures until conformance#371 publishes.

Kludex: Good catch - restored the SHA256 + fetch-and-verify step from pre-[#2974](https://github.com/modelcontextprotocol/python-sdk/pull/2974) (curl → `sha256sum -c` → repoint to `file:/tmp/conformance.tgz`) and pinned `CONFORMANCE_PKG_SHA256` to the #371 tarball digest. So CI verifies the mutable URL rather than trusting it, and a re-push to #371 fails the digest check loudly. Kept the preview pin (not the alpha.7 + expected-failures option) so the scenario runs green now; will repin to the published release once #371 ships and drop the SHA + verify step then.

src/mcp/shared/inbound.py
Kludex: Moved the JSON traversal into a small `_value_at_path` helper so `mcp_param_headers` no longer carries inline `Any`. The `Any` that remains is confined to that helper, which walks arbitrary decoded-JSON arguments - same reason `_walk_schema_positions` uses it in this file.

src/mcp/shared/inbound.py
Kludex: Removed the inline `cast` from `mcp_param_headers`. One `cast` stays inside `_value_at_path` because strict pyright needs it to narrow `isinstance(node, Mapping)` (bare `Mapping` → `Unknown` keys/values) - it's the same idiom `_walk_schema_positions` already uses here. Open to a cleaner spelling if you have one.

maxisbey (approved):

PR #3005, opened 2026-06-26
Add the SEP-2663 Tasks extension (core)

Kludex:
## Summary

The SEP-2663 **Tasks extension** (`io.modelcontextprotocol/tasks`) — the conformant core. Built on the extension API from #3003 (this PR is based on that branch; merge after it).

SEP-2663 ([Final](https://modelcontextprotocol.io/seps/2663-tasks-extension.md)) is wire-incompatible with the 2025-11-25 in-core Tasks design still carried (types-only) in `mcp_types`, so the extension defines its own SEP-2663-shaped models.

```python
from mcp.server.tasks import Tasks

mcp = MCPServer("demo", extensions=[Tasks()])
```

## What's implemented (conformant core)

- **Server-decided augmentation.** The server chooses, per request, to defer a `tools/call` as a task. The legacy `params.task` field is **ignored** (it is not the opt-in). Only a client that declared the extension on a **modern** (2026-07-28) connection is augmented — a legacy handshake cannot carry the capability back, so it is never augmented.
- **Envelope.** A task-augmented `tools/call` returns a flat `CreateTaskResult` (`resultType: "task"`, `taskId`/`status`/`createdAt`/`lastUpdatedAt`/`ttlMs`).
- **`tasks/get`** returns a `DetailedTask` (`resultType: "complete"`); a completed task inlines the original `CallToolResult`. A tool result with `isError: true` is a **completed** task (`failed` is reserved for JSON-RPC errors).
- **`tasks/cancel`** is an empty ack. **`tasks/result`** is not registered → `-32601` (removed in SEP-2663). A `tasks/*` call from a **non-declaring** client → `-32003` with a `requiredCapabilities` payload. Task ids are entropy-bearing bearer capabilities.

Ships a runnable `tasks` story and a migration note.

## Deferred to follow-ups

Each needs deeper SDK plumbing and is called out in the module/README/migration:

- `tasks/update` + the MRTR `input_required` loop
- `ToolExecution.taskSupport` gating with the `-32021` required-task error
- `notifications/tasks`
- SEP-2243 task routing headers

These map to the remaining conformance `tasks-*` scenarios; the core targets `tasks-dispatch-and-envelope`, `tasks-capability-negotiation`, `tasks-wire-fields`, `tasks-lifecycle` (partial), `tasks-request-state-removal`.

## Testing

14 spec-derived in-memory `Client(server)` tests, 100% coverage of `tasks.py`, `strict-no-cover` clean, pyright + ruff + markdownlint green, both story legs (in-memory + http-asgi) pass.

## AI Disclaimer


src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+
+from mcp.server.context import CallNext, HandlerResult, ServerRequestContext
+from mcp.server.extension import Extension, MethodBinding
+from mcp.shared.exceptions import MCPError
+
+EXTENSION_ID = "io.modelcontextprotocol/tasks"
+"""The Tasks extension identifier (SEP-2663)."""
+
+MISSING_REQUIRED_CLIENT_CAPABILITY = -32003
+"""JSON-RPC error code: a `tasks/*` call from a client that did not declare the extension."""
```
maxisbey: **`MISSING_REQUIRED_CLIENT_CAPABILITY` redefined as `-32003` — schema (and the rest of the SDK) use `-32021`**

The core schema defines `MISSING_REQUIRED_CLIENT_CAPABILITY = -32021` ([schema.ts L418](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts#L418)), and the SDK's own `mcp_types.MISSING_REQUIRED_CLIENT_CAPABILITY`, the generated `MissingRequiredClientCapabilityError.code: Literal[-32021]`, `require_client_extension()`, and the `shared/inbound.py` HTTP-status map all use `-32021` for this exact semantic. SEP-2663's prose still prints `-32003` in three places, but that predates the spec PR #2907 error-code renumber and was never updated — the schema is the source of truth for codes.

Import the constant from `mcp_types` (or just call `require_client_extension`) so `_require_tasks_capability()` is consistent with the rest of the SDK and the conformance harness; update the module docstring, `docs/migration.md`, and the test that asserts this code.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+        # request, and never alter a plain (non-declaring) client's call.
+        if not _client_declared_tasks(ctx):
+            return await call_next(ctx)
+        now = self._clock()
+        task = self._store.create(now, self._default_ttl_ms)
+        result = await call_next(ctx)
+        payload = result if isinstance(result, dict) else {}
+        # A tool result (even isError: true) is a completed task; `failed` is for
+        # JSON-RPC errors, which surface as a raised MCPError, not a result here.
+        self._store.complete(task.task_id, self._clock(), payload)
```
maxisbey: **`InputRequiredResult` is augmented into a task and misreported as `completed`**

An `InputRequiredResult` reaching the interceptor is stored via `_store.complete(...)` and reported by `tasks/get` as `status: "completed"` with the `InputRequiredResult` inlined as `result` — but `CompletedTask.result` for a `tools/call` task must be a `CallToolResult` ([SEP-2663 §Task Status](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/19cd19be7/seps/2663-tasks-extension.md?plain=1#L217-L224)).

Per [SEP-2663 §Task Creation](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/19cd19be7/seps/2663-tasks-extension.md?plain=1#L308), MRTR exchanges **SHOULD** resolve synchronously *before* `CreateTaskResult` — task `status: "input_required"` is a separate `inputRequests`/`tasks/update` mechanism, not a mapping for `InputRequiredResult`. So even with the `tasks/update` loop deferred, the interceptor should inspect `payload.get("resultType")` and pass an `"input_required"` result through unchanged (no task created), letting the MRTR loop run on the original request and augmenting only the eventual `CallToolResult`.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+        return task
+
+    def complete(self, task_id: str, now: str, result: dict[str, Any]) -> None:
+        task = self._tasks[task_id]
+        self._tasks[task_id] = task.model_copy(update={"status": "completed", "last_updated_at": now})
+        self._results[task_id] = result
+
+    def cancel(self, task_id: str, now: str) -> None:
+        task = self._tasks[task_id]
+        self._tasks[task_id] = task.model_copy(update={"status": "cancelled", "last_updated_at": now})
```
maxisbey: **`tasks/cancel` clobbers terminal status and drops the completed result**

`tasks/cancel` unconditionally overwrites `status` to `"cancelled"` with no terminal-state guard, so a `completed` task regresses to `cancelled` and the subsequent `tasks/get` drops the already-computed `result` (it only inlines on `status == "completed"`). In this inline core every reachable task is already terminal when `tasks/cancel` arrives, so this is the handler's *only* reachable behavior — and `test_cancel_task_...` snapshots the broken transition.

Terminal statuses (`completed`/`failed`/`cancelled`) should be immutable: SEP-2663 §Cancellation makes cancel cooperative and ack-only, and explicitly allows a task to "reach a terminal status other than `cancelled` if the work finished before cancellation could take effect." Guard in `_handle_cancel` (or `TaskStore.cancel`) and return the empty ack without mutating a terminal task.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+"""JSON-RPC error code: a `tasks/*` call from a client that did not declare the extension."""
+
+TaskStatus = Literal["working", "input_required", "completed", "failed", "cancelled"]
+
+Clock = Callable[[], str]
+"""Returns the current time as an ISO-8601 string (injectable for determinism)."""
+
+
+def _fixed_clock() -> str:
+    return "1970-01-01T00:00:00Z"
```
maxisbey: **Default clock is a fixed epoch stub, not a real wallclock**

The default `clock` is `_fixed_clock`, a hard-coded `"1970-01-01T00:00:00Z"` stub — so every server using the documented `Tasks()` / `Tasks(default_ttl_ms=...)` emits the Unix epoch for `createdAt`/`lastUpdatedAt` on the wire, breaking client TTL-expiry math and making the SEP-2663 timestamp fields useless.

The default should be a real UTC wallclock (e.g. `datetime.now(timezone.utc).isoformat()`); the fixed clock belongs only in tests. `test_create_task_result_uses_default_clock_when_none_injected` currently pins the wrong behaviour and will need updating.

src/mcp/server/tasks.py
maxisbey: **Interceptor mishandles exceptional and non-dict `call_next` outcomes**

`intercept_tool_call` mishandles non-dict and exceptional `call_next` outcomes (the carry-forward from #3003):

- `_store.create()` runs before `await call_next(ctx)` with no `try/except`, so when the tool call raises (`MCPError` from an unknown tool or a handler, a `ValidationError`, cancellation) the task entry leaks in the store at `status="working"` forever, and the `failed` status the docstring reserves for JSON-RPC errors is unreachable.
- `payload = result if isinstance(result, dict) else {}` silently discards a `BaseModel` (or `None`) returned by a nested extension interceptor — `HandlerResult` is `BaseModel | dict | None` — so the completed task inlines `"result": {}`.

Wrap `call_next` in `try/except` to transition the task to `failed` (or drop the entry) on error, and serialize a `BaseModel` result with `model_dump(by_alias=True, mode="json", exclude_none=True)` (mirror `_dump_result` in `runner.py`) instead of replacing it with `{}`.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def cancel(self, task_id: str, now: str) -> None:
+        task = self._tasks[task_id]
+        self._tasks[task_id] = task.model_copy(update={"status": "cancelled", "last_updated_at": now})
+
+    def get(self, task_id: str) -> Task | None:
+        return self._tasks.get(task_id)
+
+    def result(self, task_id: str) -> dict[str, Any] | None:
+        return self._results.get(task_id)
```
maxisbey: **[nit] `TaskStore` grows unboundedly; `ttlMs` never enforced**

`TaskStore` only ever inserts into `_tasks`/`_results` and never removes entries, and the single `Tasks` instance lives for the server's lifetime — so every augmented `tools/call` permanently retains its full `CallToolResult` payload. `ttl_ms` is stamped on the wire but never enforced server-side, so `Tasks(default_ttl_ms=60_000)` as shown in the story is advisory-only.

For a long-running HTTP server this is unbounded growth (the store-growth/cleanup concern carried forward from #3003). Worth at least a TTL-based eviction on `tasks/get`/`create`, a size bound, or an explicit "no eviction — deferred" note in the module docstring alongside the other deferrals.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+    def __init__(self, *, clock: Clock = _fixed_clock, default_ttl_ms: int | None = None) -> None:
+        self._store = TaskStore()
+        self._clock = clock
+        self._default_ttl_ms = default_ttl_ms
+
+    def methods(self) -> Sequence[MethodBinding]:
+        return [
+            MethodBinding("tasks/get", GetTaskRequestParams, self._handle_get),
+            MethodBinding("tasks/cancel", CancelTaskRequestParams, self._handle_cancel),
+        ]
```
maxisbey, line 230: **[nit] `tasks/get` / `tasks/cancel` not version-scoped; legacy clients get the missing-capability error instead of `-32601`**

These bindings omit `protocol_versions`, so `tasks/get`/`tasks/cancel` are served on legacy (≤2025-11-25) connections too. There `_require_tasks_capability` returns the missing-capability error with a `requiredCapabilities: {extensions: ...}` payload the client structurally cannot satisfy — since the extension "only exists on the modern wire" (and SEP-2663 §Backward Compatibility says it "is not defined under the `2025-11-25` protocol version"), these should be `-32601` instead. Add `protocol_versions=frozenset(MODERN_PROTOCOL_VERSIONS)` to both bindings.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — the module-local constant is gone; the handlers now raise through `require_client_extension`, which carries `mcp_types`' `MISSING_REQUIRED_CLIENT_CAPABILITY = -32021` and the typed `requiredCapabilities` payload: https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L296. The `-32003` numeral is also swept from the docstring, migration note, story README, and tests.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — the interceptor now discriminates the interim: a `resultType: "input_required"` result passes through un-augmented (https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L256-L260), so the MRTR exchange resolves on the original `tools/call` and only the leg that produces the final result becomes a task (one task per logical call, no interim ever stored as a completed result). Regression tests cover the passthrough leg, the completing retry leg, and the store contents.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — tasks are now recorded only once they have an outcome, so every stored task is terminal and the store has no status-transition API at all: the completed→cancelled clobber is unrepresentable rather than guarded. `tasks/cancel` acks empty and the task keeps its status and result (SEP-2663's cooperative cancellation explicitly permits an ack without effect). The snapshot test now pins completed-stays-completed with the result still retrievable.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — the default clock is a real UTC wallclock (https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L105-L106, RFC 3339 `Z`, seconds precision), and clocks now return aware datetimes so TTL deadlines and wire formatting derive from one source. The fixed clock lives only in tests via injection; the default-clock test asserts a real-time window instead of snapshotting the epoch sentinel.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 and ec4ab170 — the record is created only after `call_next` produces an outcome, so nothing can ever be stranded at `working` (the state no longer exists here). Non-dict outcomes are normalized instead of dropped (BaseModel → the runner-equivalent dump, `ErrorData` → the error path, `None` → empty result), and as of ec4ab170 a JSON-RPC error under augmentation records a task born `failed` with the error inlined on `tasks/get` (https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L249-L254) — `failed` is now a real, tested status rather than unreachable.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — `ttlMs` is enforced: expired records drop on access and are swept on insert (https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L159-L171), and a non-positive `default_ttl_ms` is rejected at construction. The store is also now a pluggable async `TaskStore` protocol (`Tasks(store=...)`, https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L131-L144) with the in-memory default documented as per-process, so deployments needing durable/bounded/multi-worker storage have the seam.

src/mcp/server/tasks.py
maxisbey: Fixed in 5303a500 — all three bindings (now including `tasks/update`) are scoped with `protocol_versions=frozenset(MODERN_PROTOCOL_VERSIONS)` (https://github.com/modelcontextprotocol/python-sdk/blob/2ab9402cb5f31266cc3224177b2280901e0e9ee7/src/mcp/server/tasks.py#L221-L226), so a legacy (≤2025-11-25) call gets `-32601 METHOD_NOT_FOUND` instead of a capability error it could never satisfy on that wire; test-pinned for all three methods.

examples/stories/tasks/README.md
```diff
@@ -1,24 +1,48 @@
...
-wired in.
+Task-augmented execution (SEP-2663). A client declares the
+`io.modelcontextprotocol/tasks` extension; the server may then answer a
+`tools/call` with a `CreateTaskResult` (carrying a task id) instead of blocking,
+and the client polls `tasks/get` for status and the eventual result.
+
+## Run it
+
+```bash
+# stdio (default — the client spawns the server as a subprocess)
```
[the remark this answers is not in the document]
maxisbey: Fixed in c983a7d7 — the README now frames the stdio leg as the graceful-degradation demo (legacy wire can't carry the extension capability, so the same `tools/call` returns a plain `CallToolResult`), and the story client asserts that degradation instead of returning silently.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+    ) -> HandlerResult:
+        # SEP-2663: the server decides augmentation; the legacy `params.task` field
+        # is ignored. Only augment for a client that declared the extension on the
+        # request, and never alter a plain (non-declaring) client's call.
+        if not _client_declared_tasks(ctx):
+            return await call_next(ctx)
+        now = self._clock()
+        task = self._store.create(now, self._default_ttl_ms)
+        result = await call_next(ctx)
+        payload = result if isinstance(result, dict) else {}
```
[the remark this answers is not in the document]
maxisbey: Fixed in 5303a500 — non-dict outcomes are normalized rather than dropped: a BaseModel from a nested interceptor is dumped the way the runner would, `ErrorData` takes the error path, and `None` becomes an empty result; tests cover each shape.

src/mcp/server/tasks.py
```diff
@@ -0,0 +1,232 @@
...
+        call_next: CallNext,
+    ) -> HandlerResult:
+        # SEP-2663: the server decides augmentation; the legacy `params.task` field
+        # is ignored. Only augment for a client that declared the extension on the
+        # request, and never alter a plain (non-declaring) client's call.
+        if not _client_declared_tasks(ctx):
+            return await call_next(ctx)
+        now = self._clock()
+        task = self._store.create(now, self._default_ttl_ms)
+        result = await call_next(ctx)
```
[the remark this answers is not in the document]
maxisbey: Fixed in 5303a500 (and refined in ec4ab170) — the task record is created only after the call produces its outcome, so an error path no longer strands a stored task; under augmentation a JSON-RPC error now records a task born `failed` with the error inlined on `tasks/get`, and on every non-augmented path the error propagates unchanged.

examples/stories/tasks/README.md
```diff
@@ -1,24 +1,48 @@
...
-**Status: deferred.** Tasks ship in 2026-07-28 as
-[SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/seps/2663-tasks-extension.md),
-an `io.modelcontextprotocol/tasks` extension that is wire-incompatible with the
-2025-11-25 in-core design still carried (types-only) in `mcp_types`. The runtime
-needs to be built to the SEP — server-decided augmentation (ignoring the legacy
-`params.task`), the `{tasks/get, tasks/update, tasks/cancel}` method set, the
-`resultType: "task"` envelope, `execution.taskSupport` gating, and `ttlMs`
-fields — so it lands in a separate PR with the conformance `tasks-*` scenarios
-wired in.
+Task-augmented execution (SEP-2663). A client declares the
```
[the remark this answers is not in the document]
maxisbey: Fixed in c983a7d7 — the stories index now lists `tasks` as a current feature story (the row moved out of the deferred section into the feature-stories group).

SrikarChittemsetty:
I built a Postgres-backed durable task store implementing the SEP-2663 spec — persistent backend, idempotency, exactly-once crash recovery, TTL cleanup. The repo is at [mcp-durable-tasks](https://github.com/SrikarChittemsetty/mcp-durable-tasks)

Full implementation: 68 tests with real SIGKILL fault injection at commit boundaries, comparative correctness benchmarks (naive approach double-charges 100% of the time; ours 0 at every concurrency level), p50/p95/p99 latency measured, DESIGN.md decision log.

Ready to use as a standalone library or contribute upstream once this PR lands.

hiddehe:
When will this be merged? @Kludex 

> Tasks ship in 2026-07-28 as [SEP-2663](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/seps/2663-tasks-extension.md),

mikerva:
@maxisbey any updates on this? i really need this for our app that is going live soon 🙏 thank you

PR #3011, opened 2026-06-27
[do not merge] Add mcp-codemod, an automated v1 to v2 migration tool

maxisbey:
> [!IMPORTANT]
> **DO NOT MERGE.** Opening as a draft for design review. The codemod is self-contained
> (a new workspace package; nothing in `mcp` depends on it), but the scope, the mapping
> tables, and the publishing prerequisite below deserve eyes before any of it is real.

Adds `mcp-codemod`, a libCST-based tool that automates the mechanical part of the v1 to v2 migration:

```bash
uvx mcp-codemod v1-to-v2 ./src
grep -rn '# mcp-codemod:' ./src   # everything left for a human
```

The design goal is narrow and deliberate: **existing v1 code runs on v2, staying on the
legacy/compat paths**. The codemod never moves users onto 2026-era features (tasks, MRTR,
resolvers, subscriptions/listen, cache hints, extensions) and never "modernizes" beyond
what working code requires — a spelling that still works on v2 (`e.error.code`, camelCase
construction kwargs at runtime) is only touched when leaving it would fail the user's own
type-checking. Every change whose meaning is unambiguous from the file alone is rewritten;
every site the codemod recognizes but will not guess at gets a `# mcp-codemod:` comment,
so the remaining work is one grep. Re-running on its own output is a no-op, and `--dry-run`
(optionally with `--diff`) previews a run without writing anything.

## Motivation and Context

`docs/migration.md` is ~1,600 lines, and most of what it asks for is tedium rather than
judgment. A codemod removes exactly that half, so a reader's (or an agent's) attention
goes to the changes that need it. The TypeScript SDK ships
`@modelcontextprotocol/codemod` as step 1 of its upgrade guide with the same philosophy
(minimal-working output, compat paths where they exist, markers elsewhere); this is the
Python counterpart.

**What it rewrites** (each gated on resolving a name through the file's imports, never on
matching text):

- Import paths that moved: `mcp.server.fastmcp` -> `mcp.server.mcpserver`,
  `mcp.types` -> `mcp_types` (including the `from mcp import types` form),
  `mcp.shared.version` -> `mcp_types.version`.
- Renamed symbols: `FastMCP` -> `MCPServer`, `McpError` -> `MCPError`,
  `FastMCPError` -> `MCPServerError`, `streamablehttp_client` -> `streamable_http_client`,
  and the removed `Content` / `ResourceReference` aliases.
- **Lowlevel decorator registrations** — the biggest structural change v2 makes.
  `@server.list_tools()` / `@server.call_tool()` / all twelve v1 decorator kinds become
  `server.add_request_handler(...)` calls at the exact source position the decorator
  occupied (registration there is when the v1 decorator ran, so execution order is
  preserved by construction — no statement reordering, and the deprecated capabilities
  land on the warning-free registration path). Each is wired through a generated adapter
  that reproduces the v1 wrapper semantics the handler relied on: bare-list wrapping,
  `call_tool`'s any-exception-to-`isError` contract with jsonschema input/output
  validation (tool lookup through the registered `tools/list` handler, exactly v1's own
  cache mechanism, so it works when `list_tools` lives in another module),
  `read_resource` content conversion, and the completion None-mapping. **User handler
  bodies are never touched.** A shape the adapter cannot serve honestly — a stacked
  decorator, a `self.`-attribute server, a non-v1 signature, a non-literal decorator
  argument, a taken name — is marked with the reason instead.
- One-argument `McpError(...)` calls to `MCPError.from_error_data(...)` (same single
  `ErrorData` argument the v1 constructor took; the user's expression is kept as
  written). `e.error.code` and friends are deliberately left alone — they still work.
- v1 positional arguments on the lowlevel `Server(...)` constructor to keywords (v2 is
  keyword-only after `name` but kept v1's names and order).
- camelCase attribute reads on `mcp.types` models (`.inputSchema` -> `.input_schema`),
  restricted to the 40 field names v1 declared, plus the same rename in
  `getattr`/`hasattr`/`setattr` string literals (those break at runtime) and in
  construction keywords (those work at runtime but fail type-checking).
- Client-surface breaks that construct silently and detonate on first use: an inline
  `timedelta(...)` session timeout gains `.total_seconds()` (a non-provable value is
  marked), `cursor=` on session `list_*` methods wraps into
  `params=PaginatedRequestParams(...)`, and a pydantic `AnyUrl(...)`/`FileUrl(...)`
  wrapper around a resource URI is dropped where the target provably takes v2's plain
  `str` (marked where it cannot be proven).
- The `streamable_http_client(...) as (read, write, _)` three-tuple to the v2 two-tuple.
- The `mcp` requirement in `pyproject.toml` (PEP 621 tables and dependency groups) and
  `requirements*.txt`, to `>=2,<3` — but only where the current constraint cannot accept
  any v2 release, and only the version specifier. Poetry tables, URL pins, unparseable
  lines, and the removed `ws` extra are marked instead.

**What it deliberately only marks** (never guesses): the v1 `mcp.types` names with no v2
home (pinned complete against the installed package by ratchet tests); imports of the
module namespaces v2 deleted (a 107-module ratchet accounts for the whole v1 namespace);
the v1 RootModel wrappers that became plain union aliases (constructing them or calling
pydantic methods on them is marked with the `TypeAdapter` fix); `request_context` /
`request_handlers` on receivers proven to be lowlevel servers; transport keywords on the
`MCPServer` constructor (the right destination depends on how the server is started);
`streamablehttp_client` results used outside a `with` item; positional constructor
arguments after the server name (v1's second positional was `instructions`, v2's is
`title` — a silent mis-route); and every removed API with no drop-in replacement. Marker
messages name the legacy-shaped fix and never advertise a 2026-era feature.

## How Has This Been Tested?

- 224 test functions/cases, 100% branch coverage on the new package (`./scripts/test` is
  green for the whole tree), strict pyright, ruff.
- **Runtime-proven adapters**: the test suite migrates a six-registration v1 lowlevel
  server and serves it to a v1-shaped `ClientSession` over the legacy protocol
  (negotiating 2025-11-25) in-memory — list/call (ok, unknown tool, schema-invalid
  arguments as `isError` results), resources, read, subscribe, prompt. The emitted
  templates are additionally pinned against the installed v2: every registration method
  string registers, every params model exists, every injectable import resolves, and no
  template emits a 2026-era surface (`InputRequiredResult`, `subscriptions/listen`,
  `cache_hints`, extensions).
- **Audited against 11 real repositories** with the batch harness at
  `scripts/codemod-batch-test/`: pinned commits of real v1 projects are migrated and
  type-checked against this workspace's v2, with the pristine side checked against the
  latest v1 as the baseline; every error that exists only on the migrated side must sit
  inside the span of a `# mcp-codemod:` marker. The manifest now includes the official
  reference servers plus decorator-heavy community servers (mysql_mcp_server and
  kaltura-mcp with seven decorators each, arxiv-mcp-server dispatching across
  subpackages), the marker path (fastapi_mcp's method-local server), client libraries
  (langchain-mcp-adapters, mcpadapt with the old spelling and a positional `timedelta`
  timeout), and an exact `==1.6.0` pin (chroma-mcp). Highlights: mysql_mcp_server's
  seven-decorator server migrates to **zero new errors and zero markers**; arxiv's
  13-file lowlevel server likewise.
- Measured against ground truth: the 76 example files that exist on both `v1.x` and
  `main` were migrated by hand, so their diff approximates the correct migration. Of the
  51 files with a real migration diff, 13 are reproduced exactly, 23 partially, 12 are
  decorator-rewritten (their oracle is the runtime round-trip and the harness — the
  human migrations rewrote handlers natively, which the codemod deliberately does not
  do), and **0 are made worse**.
- The failure modes that matter most for a tool like this are rewriting code it should
  not touch and making code worse, so that is what most of the suite pins: a file that
  never imports the SDK is never modified even when it spells tempting names; nothing is
  rewritten into a silent `NameError`; nothing that works on v2 is broken; and a re-run
  over its own output is byte-for-byte identical.
- The mapping tables are pinned against the **installed** v2 package by ratchet tests,
  so they cannot silently drift as v2 evolves: every rename target resolves, every
  removed API is provably absent, no flagged constructor keyword survives, no v1
  decorator name has come back as a live `Server` method, and the lowlevel positional
  conversion names must be keyword-only parameters of the installed constructor.

## Breaking Changes

None. The package is additive; `mcp` does not depend on it.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**Needs an owner before this merges:** the PR adds `mcp-codemod` to the publish
workflow's build step, so the PyPI project must be registered and a trusted publisher
configured for it (the same dance `mcp-types` needed) **before** this lands. The
ordering matters: if a release is tagged first, the upload job dies at the unregistered
`mcp-codemod` wheel *after* `mcp` is uploaded and *before* `mcp-types`, and `mcp`'s
exact pin on `mcp-types` makes that half-published release uninstallable until the job
is re-run. (Without the workflow change the documented `uvx mcp-codemod` fails for every
reader instead, so the docs and the publishing have to land together either way.)

**Deliberately not in this PR** (each is a clean follow-up, and none should gate the
design review):

- Restructuring `docs/migration.md` codemod-first into the two-journey split the
  TypeScript guide uses, with the mapping tables linked as the source of truth.
- Per-transform selection (the TS codemod's `--transforms`). The transformer here is one
  integrated pass rather than discrete transforms, re-runs are idempotent, and disagreeing
  with a specific rewrite is handled by not committing those hunks -- deferred until
  someone actually asks for it.


8crsk:
Hi, some transparency plus a question.

I built and published [mcp-codemod](https://pypi.org/project/mcp-codemod/) over the last few days, before finding this PR. It is also libCST-based and also targets the v1 to v2 migration, and it has the same name. The PyPI name was unregistered when I checked, so I did not realise it collided with the CLI in this PR (`uvx mcp-codemod v1-to-v2 ./src`).

**If this work is being revived, say the word and I will rename mine and free the name.** It has four releases and essentially no users yet, so it costs me nothing to move, and I would rather the official tool have the obvious name.

If it is not being revived, one small suggestion. The migration guide currently mentions codemods nowhere, while the TypeScript and Go SDKs both ship one, so Python users are the only ones doing this entirely by hand. Would you consider linking a community tool from the guide, with whatever caveats you think appropriate? Happy for it to be worded as unaffiliated and unsupported.

Mine is written independently from `docs/migration.md` rather than from this PR's code. It applies 11 mechanical rewrites and reports 10 changes it deliberately will not make, on the basis that a wrong report costs someone a minute and a wrong rewrite costs them an afternoon. Testing it against real projects rather than my own fixtures caught three bugs in it, including one where renaming `FastMCP` to `MCPServer` left `host=` and `port=` stranded on the constructor and produced a diff that looked complete and would not start.

Either answer is fine. Mainly wanted you to know the name was taken by accident rather than by land grab.

Issue #3009, opened 2026-06-27
WWW-Authenticate parsing matches a field name as a substring of another auth-param


VihaanAgarwal:
### Describe the bug

Ran into this wiring up an OAuth-protected MCP client. When a 401 `WWW-Authenticate` challenge carries several auth-params, `extract_field_from_www_auth` (`mcp/client/auth/utils.py`) matches the requested field name as a **substring** of another param's name, so a different param shadows the real value.

The pattern is `rf'{field_name}=(?:"([^"]+)"|([^\s,]+))'` searched with `re.search`, with no boundary before `field_name`.

### To reproduce

```python
import httpx
from mcp.client.auth.utils import extract_field_from_www_auth, extract_resource_metadata_from_www_auth

def r(h):
    return httpx.Response(401, headers={"WWW-Authenticate": h},
                          request=httpx.Request("GET", "https://api.example.com/"))

# A real `scope` is present, but `error_scope` shadows it:
print(extract_field_from_www_auth(r('Bearer error_scope="decoy", scope="read write"'), "scope"))
# -> 'decoy'   (expected 'read write')

# Only a decoy param exists; no real `scope`:
print(extract_field_from_www_auth(r('Bearer custom_scope="leaked"'), "scope"))
# -> 'leaked'  (expected None)

# Same for resource_metadata, which drives discovery URL selection:
print(extract_resource_metadata_from_www_auth(r('Bearer x_resource_metadata="https://decoy.example.com"')))
# -> 'https://decoy.example.com'  (expected None)
```

### Expected behavior

The field name should match only as a complete auth-param name (at the header start or after a whitespace/comma separator), so `scope` doesn't match inside `error_scope`. This matters most for `resource_metadata`, since the client uses it to pick the protected-resource-metadata discovery URL — reading it from the wrong param points discovery at the wrong place.

### Environment

- `mcp` `main` (reproduces on `v1.x` as well)

I have a small fix (anchor the param name to the header start or a separator) plus regression cases added to the existing `test_extract_field_from_www_auth_*` parametrized tests, and can open a PR if this looks right.

Reviewed by AI

fede-kamel:
[...]

PR #3012, opened 2026-06-28
fix(auth): match complete WWW-Authenticate parameters

tarunag10:
## Summary
- match WWW-Authenticate auth-param names only at parameter boundaries
- prevent suffix matches such as `error_scope` satisfying `scope`
- add regressions for `scope` and `resource_metadata` prefixed parameters

## Test plan
- `uv run pytest tests/client/test_auth.py::TestWWWAuthenticate -q`
- `uv run ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `uv run pyright src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `git diff --check`

Related: https://github.com/modelcontextprotocol/python-sdk/issues/2902

### DIFF
--- a/src/mcp/client/auth/utils.py
+++ b/src/mcp/client/auth/utils.py
@@ -26,8 +26,9 @@ def extract_field_from_www_auth(response: Response, field_name: str) -> str | No
     if not www_auth_header:
         return None
 
-    # Pattern matches: field_name="value" or field_name=value (unquoted)
-    pattern = rf'{field_name}=(?:"([^"]+)"|([^\s,]+))'
+    # Pattern matches a complete auth-param name, not a suffix of another
+    # parameter such as error_scope or x_resource_metadata.
+    pattern = rf'(?:^|[\s,]){re.escape(field_name)}=(?:"([^"]+)"|([^\s,]+))'
     match = re.search(pattern, www_auth_header)
 
     if match:
--- a/src/mcp/client/auth/utils.py
+++ b/src/mcp/client/auth/utils.py
@@ -1,4 +1,3 @@
-import re
 from urllib.parse import urljoin, urlparse
 
 from httpx import Request, Response
@@ -16,6 +15,45 @@ from mcp.shared.auth import (
 from mcp.shared.inbound import MCP_PROTOCOL_VERSION_HEADER
 
 
+def _iter_www_auth_params(www_auth_header: str) -> list[str]:
+    """Split a WWW-Authenticate challenge into auth-param tokens."""
+    params_start = www_auth_header.find(" ")
+    if params_start == -1:
+        return []
+
+    params: list[str] = []
+    current: list[str] = []
+    in_quotes = False
+    escape_next = False
+
+    for char in www_auth_header[params_start + 1 :]:
+        if escape_next:
+            current.append(char)
+            escape_next = False
+            continue
+        if char == "\\" and in_quotes:
+            current.append(char)
+            escape_next = True
+            continue
+        if char == '"':
+            in_quotes = not in_quotes
+            current.append(char)
+            continue
+        if char == "," and not in_quotes:
+            param = "".join(current).strip()
+            if param:
+                params.append(param)
+            current = []
+            continue
+        current.append(char)
+
+    param = "".join(current).strip()
+    if param:
+        params.append(param)
+
+    return params
+
+
 def extract_field_from_www_auth(response: Response, field_name: str) -> str | None:
     """Extract field from WWW-Authenticate header.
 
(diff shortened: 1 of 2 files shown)


tarunag10:
Pushed a follow-up commit (`2e2e819`) to make the parser quote-aware, not just boundary-aware.

The extractor now splits `WWW-Authenticate` auth-params on commas only outside quoted strings, then compares auth-param names exactly before returning a value. This preserves the substring-name fix while also avoiding false matches inside quoted values such as `realm="api, scope=decoy"`.

Additional validation added:
- quoted `scope` decoy before the real `scope`
- quoted-only `scope` decoy returns `None`
- quoted `resource_metadata` decoy before the real `resource_metadata`

Validation run locally:
- `uv run pytest tests/client/test_auth.py -k "www_auth" -q` — 29 passed
- `uv run ruff format --check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `uv run ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `git diff --check`

tarunag10:
Pushed one more test-only follow-up (`c05bc04`) after the GitHub matrix exposed a coverage-only failure.

The failing jobs had all tests passing, then failed the repo-wide 100% coverage gate because the new quote-aware splitter had untested escaped-character branches. I added a regression for an escaped quote inside a quoted auth-param value, which exercises that branch while preserving the real `scope` extraction.

Validation run locally:
- `uv run pytest tests/client/test_auth.py -k "www_auth" -q` — 30 passed
- `uv run ruff format --check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `uv run ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `git diff --check`

tarunag10:
Pushed another test-only coverage follow-up (`32950d8`) for the remaining splitter branches.

Added regressions for:
- a `Bearer` challenge with no auth-params
- empty comma segments between auth-params
- a trailing comma after the final auth-param

Local validation now covers 33 selected auth tests:
- `uv run pytest tests/client/test_auth.py -k "www_auth" -q` — 33 passed
- `uv run ruff format --check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `uv run ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `git diff --check`

fede-kamel (commented):
I differential-tested this against `main` and the two sibling PRs for #3009 (#3041, #3050) — full matrix in https://github.com/modelcontextprotocol/python-sdk/issues/3009#issuecomment-4938809257.

Credit first: this is the **only** implementation of the three that correctly handles RFC 9110 escaped quotes inside quoted values — `Bearer scope="say \"hi, there"` parses intact here and truncates at the comma everywhere else, including `main`.

But I found a regression on multi-challenge headers. `_iter_www_auth_params` takes everything after the *first* space as the auth-params of a single challenge, so when the Bearer challenge is not the first one, the field is no longer found:

```python
extract_field_from_www_auth(r('Basic realm="r", Bearer scope="files"'), "scope")
# main -> "files";  this branch -> None

extract_field_from_www_auth(r('DPoP algs="ES256", Bearer scope="files"'), "scope")
# main -> "files";  this branch -> None
```

The token for the second challenge comes through as `Bearer scope="files"`, whose name (`Bearer scope`) never equals `scope`. The `DPoP` + `Bearer` co-advertisement is a real pattern (RFC 9449 §7), so this would break scope discovery against DPoP-capable servers.

A possible fix within this design: when a token contains a space before its `=` (i.e. it looks like `Scheme param=…`), treat it as the start of a new challenge and strip the scheme prefix before comparing. Alternatively, the escape-aware parsing here could be rebased as a follow-up on top of the minimal boundary fix in #3050, which keeps multi-challenge behavior intact.

fede-kamel:
Follow-up with end-to-end confirmation of the multi-challenge regression, since my first comment used constructed header strings.

I stood up a raw-socket HTTP server returning 401 with **two separate `WWW-Authenticate` headers** (which RFC 9110 permits and reverse proxies commonly produce) and fetched it with real httpx:

```
server sends:  WWW-Authenticate: Basic realm="legacy"
               WWW-Authenticate: Bearer resource_metadata="https://rs.example/prm"

httpx presents:  'Basic realm="legacy", Bearer resource_metadata="https://rs.example/prm"'

extract_field_from_www_auth(resp, "resource_metadata")
#   main → "https://rs.example/prm"
#   this branch → None
```

Two implications beyond my earlier comment:

1. Because httpx joins repeated headers with `", "`, the regression triggers even for servers that never put two challenges in one header — separate `Basic` + `Bearer` headers are enough.
2. The affected caller isn't just scope discovery: `extract_resource_metadata_from_www_auth` is how the client locates the RFC 9728 protected-resource metadata, so a `None` here breaks the OAuth discovery flow at its first step for any server whose Bearer challenge isn't the first header.

Same live scenario passes on `main` and on the #3050 approach, so this is specific to the first-space challenge parsing here.

PR #3016, opened 2026-06-28
fix: match WWW-Authenticate params exactly

### LINKED ISSUE #3009: WWW-Authenticate parsing matches a field name as a substring of another auth-param

mturac:
## Summary
- parse WWW-Authenticate auth-params outside quoted values before extracting a field
- require auth-param names to match exactly
- cover substring collisions and quoted-value false positives

Fixes #3009

## Tests
- `uv run --frozen pytest tests/client/test_auth.py -k 'extract_field_from_www_auth'`
- `uv run --frozen pytest tests/client/test_auth.py`
- `uv run --frozen ruff format --check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `uv run --frozen ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py`
- `git diff --check`
- `uv run --frozen coverage run -m pytest tests/client/test_auth.py -k 'extract_field_from_www_auth' && uv run --frozen coverage combine && uv run --frozen strict-no-cover`

### DIFF
--- a/src/mcp/client/auth/utils.py
+++ b/src/mcp/client/auth/utils.py
@@ -26,13 +26,25 @@ def extract_field_from_www_auth(response: Response, field_name: str) -> str | No
     if not www_auth_header:
         return None
 
-    # Pattern matches: field_name="value" or field_name=value (unquoted)
-    pattern = rf'{field_name}=(?:"([^"]+)"|([^\s,]+))'
-    match = re.search(pattern, www_auth_header)
+    header_value = www_auth_header.strip()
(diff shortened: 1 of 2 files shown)


mturac:
Addressed in da310618 by parsing auth-params with quote-aware comma splitting and adding regression coverage for quoted-value false positives.

mturac:
Closing this for now to keep my active PR queue focused. The automated review thread was addressed, but I do not want to leave this stale without a fresh maintainer signal and validation pass.

PR #3026, opened 2026-06-30
fix: drain accepted JSON-RPC requests after read EOF

### LINKED ISSUE #2678: FastMCP/stdio: in-flight tool responses dropped on stdin EOF when input is bash-redirected from a file

matthewchen94:
Fixes #2678.

This changes the JSON-RPC dispatcher shutdown path so read EOF does not immediately cancel request handlers that were already accepted. The dispatcher now:

- marks the connection closed and wakes outbound waiters after read EOF,
- keeps the write stream open briefly for accepted inbound requests to finish their response writes,
- bounds that drain window to avoid hanging shutdown,
- tracks active inbound requests through the response write window, not just while they are present in `_in_flight`.

This targets stdio servers driven with redirected stdin, where EOF can arrive immediately after the last JSON-RPC request is read while an async tool call is still mid-await. In that case the response should be flushed to stdout before shutdown.

Validation:
- `python3 -m py_compile src/mcp/shared/jsonrpc_dispatcher.py tests/shared/test_jsonrpc_dispatcher.py` passed on the patched files.
- Added a dispatcher-level regression test that accepts a request, closes the read side, and asserts the response is still written.


matthewchen94:
Updated this PR to address the review and failed CI:\n\n- Scoped EOF draining behind an explicit dispatcher opt-in so default transport-close semantics still cancel in-flight handlers immediately.\n- Enabled that opt-in only for stdio read streams, where redirected stdin can EOF after the final accepted request while stdout still needs to flush the response.\n- Tracked the transport_builder rejection path in the drain counter, so spawned INTERNAL_ERROR responses are not cancelled before reaching the peer.\n- Merged latest main and added regression coverage for both accepted handler responses and builder-rejection responses.\n\nAll CI checks are green on f4a8fc99220c6df8fa4e218d3cb832fa62a576fb.

matthewchen94:
Updated this PR onto latest `main` with a merge commit, without force-pushing.

What changed in the refresh:

- rebuilt the stdio EOF-drain change on current `main`
- preserved the new runner cache-hint handling that landed after the earlier PR head
- kept EOF draining scoped to stdio/read streams via `drain_inbound_on_read_eof`
- propagated the opt-in through both current dispatcher construction paths
- kept regression coverage for both accepted handler responses and transport-builder rejection responses

Local validation on the rebuilt files:

```bash
python3 -m py_compile \
  src/mcp/shared/jsonrpc_dispatcher.py \
  src/mcp/server/runner.py \
  src/mcp/server/stdio.py \
  src/mcp/shared/_context_streams.py \
  tests/shared/test_jsonrpc_dispatcher.py
```

The PR is mergeable again; CI is rerunning on `14ed0cb`.

matthewchen94:
Quick maintainer ping on this one.

The PR is now mergeable and the refreshed head `14ed0cb` has the full required matrix green:

- pre-commit
- server/client conformance
- Python 3.10–3.14 test matrix on Ubuntu/Windows
- docs/readme snippets
- all-green

The refresh was done as a merge commit on top of the previous PR head, not a force-push, and it preserves the newer runner cache-hint changes from current `main`.

I don't plan to change this further unless there is specific review feedback, because the current patch is scoped to the stdio EOF-drain behavior and CI is clean.

maxisbey:
Thanks for the PR. Tracking this in #2680 instead. Closing as part of a general backlog cleanup following the v2 release. If this is still relevant against v2, feel free to reopen.

PR #2778, opened 2026-06-04
Suppress KeyboardInterrupt tracebacks from MCPServer.run

koriyoshi2041:
Problem
Interrupting an `MCPServer.run(transport="stdio")` process with Ctrl-C can surface the full AnyIO/asyncio cancellation traceback even though the server is just shutting down. Issue #2663 is marked ready for work and points to the sync `run()` boundary as the narrow fix point.

Fix
Wrap the transport dispatch in `MCPServer.run()` with a `KeyboardInterrupt` guard so user interrupts return cleanly. Other exceptions still propagate.

Tests
- `uv run pytest tests/server/mcpserver/test_server.py`
- `uv run ruff format --check src/mcp/server/mcpserver/server.py tests/server/mcpserver/test_server.py`
- `uv run ruff check src/mcp/server/mcpserver/server.py tests/server/mcpserver/test_server.py`
- `uv run pyright src/mcp/server/mcpserver/server.py tests/server/mcpserver/test_server.py`
- `git diff --check`

Risk
Low. The guard is only at the synchronous `run()` entry point and only handles `KeyboardInterrupt`; ordinary startup/runtime exceptions are still re-raised.

Fixes #2663

PR #3027, opened 2026-06-30
fix(server): exit cleanly on KeyboardInterrupt in MCPServer.run

### LINKED ISSUE #2663: FastMCP.run(transport="stdio") produces noisy traceback on KeyboardInterrupt instead of clean exit

devansh-dek:
## Summary

- Catch `KeyboardInterrupt` at the `MCPServer.run()` sync boundary so Ctrl-C during stdio server execution exits cleanly instead of printing a noisy anyio/asyncio cancellation traceback.
- Adds unit tests verifying `KeyboardInterrupt` is suppressed while other exceptions still propagate.

## Test plan

- [x] `uv run pytest tests/server/mcpserver/test_server.py::TestServer::test_run_suppresses_keyboard_interrupt tests/server/mcpserver/test_server.py::TestServer::test_run_reraises_other_exceptions -v`
- [x] `uv run ruff check src/mcp/server/mcpserver/server.py tests/server/mcpserver/test_server.py`

Fixes #2663

### DIFF
--- a/src/mcp/server/mcpserver/server.py
+++ b/src/mcp/server/mcpserver/server.py
@@ -358,13 +358,16 @@ class MCPServer(Generic[LifespanResultT]):
         if transport not in TRANSPORTS.__args__:  # type: ignore  # pragma: no cover
             raise ValueError(f"Unknown transport: {transport}")
 
-        match transport:
-            case "stdio":
-                anyio.run(self.run_stdio_async)
-            case "sse":  # pragma: no cover
-                anyio.run(lambda: self.run_sse_async(**kwargs))
-            case "streamable-http":  # pragma: no cover
-                anyio.run(lambda: self.run_streamable_http_async(**kwargs))
+        try:
+            match transport:
+                case "stdio":
+                    anyio.run(self.run_stdio_async)
+                case "sse":  # pragma: no cover
+                    anyio.run(lambda: self.run_sse_async(**kwargs))
+                case "streamable-http":  # pragma: no cover
+                    anyio.run(lambda: self.run_streamable_http_async(**kwargs))
+        except KeyboardInterrupt:
+            return
 
     async def _handle_list_tools(
         self, ctx: ServerRequestContext[LifespanResultT], params: PaginatedRequestParams | None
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Fix is correct, and this is the tightest of the three open PRs for #2663 in terms of diff: one test file, no unrelated edits.

Worth knowing that the source change here is **byte-identical** to #2745 (Jun 1) and #2778 (Jun 4). All three wrap the `match transport:` block in `try/except KeyboardInterrupt: return`, same lines, same position. Yours is the most recent of the three, so #2745 has priority by date, though yours has the cleanest diff. Full comparison on #2663.

**Exit code.** `except KeyboardInterrupt: return` makes `run()` return normally, so the process exits 0. Measured both shapes:

```
except KeyboardInterrupt: return   -> exit code 0
raise SystemExit(130)              -> exit code 130
```

130 is `128 + SIGINT`, which is what an uncaught `KeyboardInterrupt` produces today and what a shell or supervisor expects from a Ctrl-C. After this change a script doing `mcp-server || handle_failure` reads an interrupted run as a clean success.

If the goal is only to drop the traceback, `raise SystemExit(130)` gets that without changing the signal semantics. If exiting 0 is deliberate (say, because a supervisor is meant to treat Ctrl-C as a normal stop) that's a defensible choice, it just deserves a comment, otherwise someone will "fix" it later assuming it was an oversight.

Your two tests (suppress `KeyboardInterrupt`, re-raise `RuntimeError`) are the right pair: the second is what stops a future change from widening the `except` into something that swallows real startup failures. If the exit-code point above is adopted, the first would want updating to assert `SystemExit` with code 130 rather than a plain `None` return.

maxisbey:
Thanks for the PR. Tracking this in #2745 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

Issue #2804, opened 2026-06-06
Implement SEP-2575: Make MCP Stateless


localden:
Tracking implementation of [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) for the 2026-07-28 MCP spec release.

maxisbey:
Closing as of [v2.0.0b1](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v2.0.0b1): the SDK implements SEP-2575 — stateless self-describing requests, `server/discover` on both sides (the client's default `mode='auto'` probes it), per-request capabilities, and `subscriptions/listen` served via a pluggable event bus (#3035), including over stdio and other stream-pair transports (#3038). The `server-stateless` conformance scenario passes in full.

Client-side consumption of `subscriptions/listen` streams (an optional client feature under the SEP) lands in a follow-up PR.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

PR #3031, opened 2026-06-30
Bump the conformance pin to 0.2.0-alpha.8 and baseline the listen checks it surfaces

maxisbey:
> [!NOTE]
> Stacked on #3030 (base branch `conformance-skips`); will be retargeted to `main` when that merges.

Bumps the conformance harness pin from `b18aa918` to `4944b268` (0.2.0-alpha.8 + [conformance#372](https://github.com/modelcontextprotocol/conformance/pull/372), neither published to npm yet — same sha256-verified pkg.pr.new mechanics as the previous pin).

\#372 changes skip semantics: a check whose prerequisite is missing — a fixture tool the server doesn't ship, or a capability the server advertises but then rejects — now **fails** with a "Not testable:" message instead of silently skipping outside the pass/fail denominator. That makes one more unimplemented surface visible as a baselined known failure:

- **`server-stateless` now fails its 3 `subscriptions/listen` MUST checks** (new entry in both expected-failures files). The everything-server's legacy `resources/subscribe` handlers make it advertise `resources.subscribe` in `server/discover`, while `subscriptions/listen` answers `-32601` — under #372 that's claims-but-doesn't-serve, a failure. The entry burns down when the listen runtime lands (#2804). The two `listChanged` SHOULD checks remain legitimately skipped (the capability is declared `false`, which #372 still treats as honest feature absence).
- The `tasks-*` scenarios fail with higher check counts (their "no task created" cascade skips now fail too), but they were already baselined — no entry changes.
- `tasks-status-notifications` still skips unconditionally under #372 and must stay unbaselined.
- Client legs are unaffected (421/370 checks pass, zero warnings, both exit 0).

## Motivation and Context

Same goal as #3030, finished by the upstream half: #3030 made the filtered-out scenarios run; #372 makes the prerequisite-gap *checks inside running scenarios* count. After both, the only remaining green-without-evidence surfaces are the ones the harness deliberately exempts.

## How Has This Been Tested?

Replicated all six CI legs locally against the new pinned tarball (sha256-verified), exactly as CI invokes them: the four server legs and both client legs exit 0 with "Baseline check passed", with `server-stateless` failing only the 3 listen checks (25 still passing) and all other failures matching the existing baseline. Also confirmed the discovery state before the baseline update: only `server-stateless` was newly unexpected; no stale entries; no warnings anywhere.

## Breaking Changes

None — CI configuration only.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- The `server-stateless` waiver carries a real masking cost, called out in the baseline comment: while listed, new failures in the scenario's other 25 (currently passing) envelope/error-handling checks are also absorbed, because the baseline is per-scenario. The trade is deliberate — a visible, forced-burn-down known failure for the listen gap beats the pre-#372 state where the same gap was invisible — but it makes landing the listen runtime (#2804) the way to get those 25 checks back under guard.
- The capability inconsistency the harness is flagging is real, not a harness artifact: on the 2026 wire the server advertises `resources.subscribe` while the subscribe methods themselves answer 404/`-32601` there. Implementing listen resolves it from one side; era-aware capability reporting in `server/discover` would resolve it from the other.

### DIFF
--- a/.github/actions/conformance/expected-failures.2026-07-28.yml
+++ b/.github/actions/conformance/expected-failures.2026-07-28.yml
@@ -23,6 +23,9 @@
 client: []
 
 server:
+  # SEP-2575 subscriptions/listen is not implemented yet; see the matching
+  # entry in expected-failures.yml for the full rationale.
+  - server-stateless
   # SEP-2243 Mcp-Param-* server-side validation is not implemented yet; see
   # the matching entry in expected-failures.yml for the full rationale.
   - http-custom-header-server-validation
--- a/.github/actions/conformance/expected-failures.yml
+++ b/.github/actions/conformance/expected-failures.yml
@@ -13,6 +13,16 @@
 client: []
 
 server:
+  # SEP-2575 subscriptions/listen is not implemented yet. The everything-
+  # server's legacy resources/subscribe handlers make it advertise
+  # `resources.subscribe` in server/discover, and as of conformance #372 a
+  # server that advertises a subscription capability but answers
+  # subscriptions/listen with -32601 fails the three listen MUST checks
+  # ("Not testable") instead of skipping them. Remove this entry when the
+  # listen runtime lands. NOTE: while listed, this entry also masks new
+  # failures in the scenario's other 25 (currently passing) checks — the
+  # baseline is per-scenario, not per-check.
+  - server-stateless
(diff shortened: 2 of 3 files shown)


maxisbey:
Folding this into #3030 directly — no stack needed.

PR #3034, opened 2026-06-30
Add a client extension API

maxisbey:
Adds a client-side extension API: the counterpart to the server `Extension` class from #3003, so SEP-2133 extensions can contribute client behaviour — not just a capability ad — without core SDK changes.

## Motivation and Context

The spec's extension mechanism is deliberately SDK-shaped: capabilities ride `ClientCapabilities.extensions`, servers may substitute extension result shapes on `tools/call` (supported set = core ∪ declared), and extensions define their own methods and notifications. The server side got a first-class API; the client side only had a raw `extensions={...}` dict for the ad. Anything behavioural (the upcoming tasks extension being the concrete case) would have needed hand-edits to `ClientSession` and `Client` internals.

A `ClientExtension` declares the three things a client extension can actually contribute at 2026-07-28:

```python
class Receipts(ClientExtension):
    identifier = "com.example/receipts"

    def claims(self) -> Sequence[ResultClaim[Any]]:
        return [ResultClaim(result_type="receipt", model=ReceiptResult, resolve=self._redeem)]

async with Client(server, extensions=[Receipts()]) as client:
    result = await client.call_tool("buy", {"item": "tea"})   # still a plain CallToolResult
```

- `settings()` — the capability ad, read once at construction.
- `claims()` — extra `tools/call` result shapes the server may substitute, each paired with the resolver that finishes it. Claims activate at version adoption (modern wires only) and the capability ad follows them: an extension whose claims are inactive is not advertised, so the client never advertises a shape it would reject. Undeclared shapes keep failing validation exactly as the spec requires.
- `notifications()` — vendor notifications to observe, delivered in order through per-binding bounded queues.

Extension verbs need no registration: vendor request types subclass `mcp_types.Request` and go through `session.send_request`, whose typing now admits them. A request type can declare `name_param` to mirror a params key into the `Mcp-Name` header on every send path (required by extension specs such as tasks for their methods).

Two session-tier escape hatches round it out: `call_tool(..., allow_claimed=True)` returns the undriven claimed shape, and an unexpected claimed shape raises `UnexpectedClaimedResult` carrying the parsed value (by then the server may have durably created state, so the caller can reach it to clean up).

Everything is validated at construction — identifier grammar (shared with the server via `mcp.shared.extension`), claim model discriminators, cross-extension conflicts named by their owning extensions — and zero extensions means a byte-identical client: the no-extensions `tools/call` parse path is pinned by identity in tests.

## How Has This Been Tested?

- Unit + integration tests at every tier (construction validation, adoption/activation incl. modern→legacy→modern re-adoption, discriminated-adapter routing, notification delivery under flood, the Client fold and transparent resolution, the input_required→claimed-on-retry interaction).
- Interaction-suite tests over the real transports: the full both-ends loop (a server `Extension` substitutes a claimed shape; the client extension's resolver finishes it with a follow-up call), the off-switch (no extension → validation failure; legacy connection → the ad drops claim-bearing identifiers), per-request capability gating (-32021 for a non-declaring client), and `Mcp-Name` from `name_param` observed on the HTTP wire.
- Two new runnable docs tutorials exercised by the docs test suite.
- `./scripts/test` green at 100% line+branch coverage; pyright strict clean; full pre-commit clean.

## Breaking Changes

`Client(extensions=...)` now takes a sequence of `ClientExtension` instances instead of a dict (the dict form shipped days ago in #3003 and is pre-beta). Ad-only uses migrate to `advertise(identifier, settings)`; passing the old dict raises a `TypeError` that says so. Documented in `docs/migration.md`.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- Part of the 2026-07-28 client surface (#2891). The tasks extension's client half (#3005) is the intended first consumer: its session/client coupling maps 1:1 onto this API and can be re-cut onto it as a follow-up.
- End-to-end vendor-notification delivery is pinned at session tier only: there is no public server-side surface that emits vendor-method notifications yet (`ServerNotification` is a closed union), and HTTP-modern arrival waits on the `subscriptions/listen` client runtime (#2804).
- No new conformance scenario applies: this is SDK packaging surface; the wire behaviours it generalizes (per-request capability ad, `Mcp-Name` on task methods) are covered by existing scenarios and by the tasks suite once that extension rides this API.

### DIFF
--- a/docs/advanced/extensions.md
+++ b/docs/advanced/extensions.md
@@ -2,9 +2,10 @@
 
 An **extension** is an opt-in bundle of MCP behaviour behind one identifier.
 
-It can contribute tools, resources, and new request methods, and it can wrap `tools/call`.
-The server advertises it under `capabilities.extensions`, the client opts in the same way,
-and nothing changes for anyone who didn't ask for it. That is the contract ([SEP-2133](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2133)), and
+On a server it can contribute tools, resources, and new request methods, and it can wrap
+`tools/call`. On a client it can claim extra `tools/call` result shapes and observe vendor
+notifications. Each side advertises under its own `capabilities.extensions`, and nothing
+changes for anyone who didn't ask for it. That is the contract ([SEP-2133](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2133)), and
 it has one golden rule: **extensions are off by default**.
 
 ## Using an extension
@@ -79,7 +80,7 @@ And `main()` is the proof, an in-memory client straight against `mcp`:
 An extension can register **new request methods**: its own verbs, served next to the
 spec's:
 
-```python title="server.py" hl_lines="15-21 30 39-47"
+```python title="server.py" hl_lines="16-22 31 40-48"
 --8<-- "docs_src/extensions/tutorial004.py"
 ```
 
@@ -108,15 +109,15 @@ runtime:
 
 The same file's `main()` is the whole client story, both halves of it:
 
-```python title="server.py" hl_lines="53-57"
+```python title="server.py" hl_lines="54-58"
 --8<-- "docs_src/extensions/tutorial004.py"
 ```
 
-* `Client(..., extensions={EXTENSION_ID: {}})` declares the extension. That map
-  becomes `ClientCapabilities.extensions`: on a 2026-07-28 connection it travels in
-  the per-request `_meta` envelope, so the server sees it on **every** request; on
-  a legacy connection it rides the `initialize` handshake. Server code doesn't care
-  which: `require_client_extension(ctx, ...)` and
+* `Client(..., extensions=[advertise(EXTENSION_ID)])` declares the extension. The
+  declarations become `ClientCapabilities.extensions`: on a 2026-07-28 connection
+  the map travels in the per-request `_meta` envelope, so the server sees it on
+  **every** request; on a legacy connection it rides the `initialize` handshake.
+  Server code doesn't care which: `require_client_extension(ctx, ...)` and
   `ctx.session.check_client_capability(...)` read the right source on both paths.
 * Vendor methods drop one layer to `client.session.send_request(...)`; `Client`
   only grows first-class methods for spec verbs. `send_request` accepts any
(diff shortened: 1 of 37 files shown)


maxisbey:
Responses to the review feedback, addressed in 99fec3c6:

**Fixed**
- `ResultClaim` now rejects models that don't subclass `Result` at runtime (the `ClaimedT` bound only constrains type-checked callers), and rejects models whose fields alias `requestState`/`inputRequests`: the core result surface types those optional fields, so a colliding value would fail core validation before the claim adapter ever ran. Both are construction-time errors with tests.
- The notification ordering contract is now stated honestly in the docstring, docs page, and requirement: per-binding serialized delivery in dispatch order. Stream transports dispatch each inbound notification independently, so near-simultaneous notifications can be dispatched out of wire order. That is the pre-existing delivery property of every notification path, and the docs no longer promise more than the transport does. Enqueueing in read-loop order would mean inline dispatch (which deadlocks in-process servers) or a dispatcher-level queue; not worth it for observation-only semantics.
- The receipts tutorial server now gates substitution with `require_client_extension`, so the example teaches the spec-compliant server posture; the off-by-default paragraph covers both the gate refusal (-32021) and the validation failure for ungated shapes.
- The stale custom-methods README caveat now points at `NotificationBinding` (the remaining omission is server-side: there's no public surface for emitting vendor notifications yet).
- The `_build_capabilities` docstring no longer overstates which identifiers drop (claim-less identifiers always advertise).

**Not changed, deliberately**
- "Claimed tools/call results are rejected by core validation before the extension adapter runs": this isn't the case: claimed raws pass the 2026 surface's lenient arm, pinned by `test_claimed_raw_passes_v2026_tools_call_surface_validation` and exercised end-to-end by every claim test (in-memory and streamable HTTP). The one real corner was the typed-alias collision above, which is now unconstructible.
- An unrecognized claimed `resultType` that also carries `content` parses as a complete core result. Confirmed, but it's the monolith's pre-existing open `result_type` + ignored-extras behavior, identical on main with zero extensions. Rejecting tags outside core plus active claims changes parsing for every client and belongs to the strict-resultType follow-up (this PR's discriminator is the mechanism that change would use). The interaction requirement now carries a note saying exactly this.
- Vendor fields in handler-produced (non-interceptor) claimed results are stripped by the server runner's serializer. This is pre-existing, noted in the existing `TODO` in `runner.py`, and scoped to the server-side tasks work; the client-side docs already scope "verbatim" to the interceptor path.

maxisbey:
Follow-up on the latest review rounds, addressed in 53a538a6:

**Fixed**
- The reserved-key check on claim models now collects every alias form: plain `alias`, `validation_alias` (including `AliasChoices` members and the head of an `AliasPath`), and `serialization_alias`. A model routing `requestState` or `inputRequests` through any of those spellings is rejected at construction, with tests for each route.

**Already fixed at the reviewed commit**
- "Core surface pre-validation rejects active claims whose vendor fields alias requestState/inputRequests": the construction-time reservation landed in 99fec3c6 and the exact scenario (a `requestState`-aliased dict field) raises a `ValueError` there; 53a538a6 extends it to the remaining alias spellings. On the `_meta` variant: a subclass that overrides `meta` loses the `_meta` alias (the alias generator yields `meta`), so an override cannot collide with the surface's typed `_meta` field. A non-dict under `_meta` on the wire would be malformed core plumbing from the server, not a claim-shape issue.

**Standing as previously stated**
- Unknown tag plus `content` parsing as a complete core result: pre-existing open-tag behavior of the monolith, identical on main with zero extensions; rejecting tags outside core plus active claims changes parsing for every client and is the strict-resultType follow-up. The interaction requirement carries a note to that effect.
- Vendor fields stripped from handler-produced claimed results: pre-existing server-side serializer behavior, anchored by the existing `TODO` in `runner.py` and scoped to the server-side tasks work.

felixweinberger (approved):

PR #3050, opened 2026-07-02
fix: match complete www-authenticate auth params

### LINKED ISSUE #3009: WWW-Authenticate parsing matches a field name as a substring of another auth-param

pradeep-ramola:
Fixes modelcontextprotocol/python-sdk#3009

Updates `extract_field_from_www_auth` so it only matches complete `WWW-Authenticate` auth-param names. This prevents fields like `scope` from matching inside longer names such as `error_scope` or `custom_scope`, and prevents `resource_metadata` from matching inside `x_resource_metadata`.

## Motivation and Context

The previous regex searched for the field name without a boundary, so a decoy parameter could shadow the real auth-param value or return a value when the requested parameter was not present.

This matters for OAuth discovery because `resource_metadata` from `WWW-Authenticate` controls the protected resource metadata URL selection.

## How Has This Been Tested?

Tested locally with:

```bash
python -m pytest tests/client/test_auth.py -k WWWAuthenticate
python -m ruff check src/mcp/client/auth/utils.py tests/client/test_auth.py
python -m ruff format --check src/mcp/client/auth/utils.py tests/client/test_auth.py
```

Scenarios covered:
- `error_scope` before real `scope`
- `custom_scope` without real `scope`
- `x_resource_metadata` before real `resource_metadata`
- `x_resource_metadata` without real `resource_metadata`

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix escapes the requested field name and requires it to appear at the start of the header or after an auth-param separator, so only complete auth-param names are matched.


mayankbohradev:
I think this PR fixes the substring-param-name case from #3009, but there is one adjacent case it still misses: field-looking text inside a quoted auth-param value.

Using this PR's proposed pattern:

```python
import re

header = 'Bearer error="invalid_token", error_description="missing scope=wrong", scope="from-header"'
field_pattern = re.escape("scope")
pattern = rf'(?:^|[\s,]){field_pattern}=(?:"([^"]+)"|([^\s,]+))'

match = re.search(pattern, header)
print(match.group(1) or match.group(2))
# wrong"
# expected: from-header
```

This matters because MCP/OAuth challenges can carry `error_description` plus `scope`, and the client uses `extract_scope_from_www_auth()` for both initial scope selection and `403 insufficient_scope` step-up.

I reproduced the current behavior on `main` while testing v2 auth flows. My local Highlyt MCP server's current first 401 only sends `resource_metadata`, so it does not naturally trigger this today, but a standards-valid richer challenge does.

AI-assisted (Claude/Codex) for navigation and review; change authored and understood by me.

fede-kamel (commented):
I differential-tested this against `main` and the two sibling PRs for #3009 (#3012, #3041) — full matrix in https://github.com/modelcontextprotocol/python-sdk/issues/3009#issuecomment-4938809257.

This is the strongest of the three in my testing:

- Fixes both reported shadowing cases, and also two variants the issue did not mention: unquoted-value shadowing (`Bearer as_scope=x, scope=files` → `main` returns `x`) and comma separators without a following space (`error_scope="a",scope="files"` → `main` returns `a`).
- No behavior change in any other probe — including multi-challenge headers (`Basic …, Bearer …`, `DPoP …, Bearer …`) and non-Bearer schemes, both of which regress in one of the sibling PRs.
- `re.escape(field_name)` is a nice hardening touch the current code lacks.
- `tests/client/test_auth.py`: 137 passed, 0 failed on this branch.

Two optional follow-ups, neither blocking:
1. RFC 7235 auth-param names are case-insensitive, so `Bearer SCOPE="files"` should match a `scope` lookup — adding `re.IGNORECASE` to the search would cover it (pre-existing gap, all implementations fail it today).
2. Escaped quotes inside quoted values (`scope="say \"hi, there"`) still truncate at the comma — also pre-existing; #3012 has parsing logic for this if it is ever wanted as a follow-up.

maxisbey:
Thanks for the PR. Tracking this in #3041 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

PR #1578, opened 2025-11-05
Fix issue where client tool call hangs forever if server crashes or connection dies when using streamable-http

gyang-xai:
A fix for #1577 

## Motivation and Context
Fixes issue where some errors are not propagated back to the tool call.

## How Has This Been Tested?
Tested with the example in the issue.

## Breaking Changes
No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context
I have tested this with tool calls but unsure if this works for the other resource types.

PR #3052, opened 2026-07-02
fix(client): unblock streamable HTTP calls on early SSE close

Sehlani042:
## Summary

Fixes a streamable HTTP client hang when a POST request receives an SSE response that closes before sending a JSON-RPC response.

This targets #1577. It overlaps with the older #1578, but keeps the current-main patch narrow: preserve the existing reconnect behavior when a resumable event id is available, and only resolve the pending request with the existing SDK `CONNECTION_CLOSED` error when there is no response and no resumable event to follow.

## Root Cause

`_handle_sse_response()` consumed the POST SSE stream and returned silently if the stream ended before a response and no `Last-Event-ID` had been observed. Because no `JSONRPCError` or transport exception was written to the read stream, the JSON-RPC dispatcher kept the corresponding `call_tool()` waiter pending forever.

## Changes

- Send a correlated `JSONRPCError(CONNECTION_CLOSED)` when a POST SSE response closes before a reply and cannot be resumed.
- Also resolve the pending request with `CONNECTION_CLOSED` if reconnection attempts are exhausted.
- Add a non-SDK server regression test that closes an SSE response without a JSON-RPC reply and asserts `call_tool()` raises instead of hanging.

## Validation

- `uv run --frozen pytest tests/client/test_notification_response.py::test_empty_post_sse_response_unblocks_pending_tool_call -q`
- `uv run --frozen pytest tests/client/test_notification_response.py -q`
- `uv run --frozen pytest tests/shared/test_streamable_http.py -q`
- `uv run --frozen ruff format src/mcp/client/streamable_http.py tests/client/test_notification_response.py --check`
- `uv run --frozen ruff check src/mcp/client/streamable_http.py tests/client/test_notification_response.py`
- `uv run --frozen pyright src/mcp/client/streamable_http.py tests/client/test_notification_response.py`
- `uv run --frozen coverage erase && uv run --frozen coverage run -m pytest tests/client/test_notification_response.py tests/shared/test_streamable_http.py && uv run --frozen coverage combine && uv run --frozen coverage report --include='src/mcp/client/streamable_http.py' --fail-under=0 && UV_FROZEN=1 uv run --frozen strict-no-cover`

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -13,6 +13,7 @@ import httpx
 from anyio.abc import TaskGroup
 from httpx_sse import EventSource, ServerSentEvent, aconnect_sse
 from mcp_types import (
+    CONNECTION_CLOSED,
     INTERNAL_ERROR,
     INVALID_REQUEST,
     METHOD_NOT_FOUND,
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. This has since landed via #3047. I'm closing this out as part of a backlog cleanup now that v2 is out. Feel free to reopen if this is still relevant.

PR #3053, opened 2026-07-02
Clarify missing tool_result validation error

HarperZ9:
## Summary
- keep `tool_use` / `tool_result` balance validation strict
- split the missing-follow-up case into a clearer `ValueError`
- add a direct regression test for an assistant `tool_use` followed by user text instead of `tool_result`

Related to #2960. This PR does not relax the balance rule; it only makes the diagnostic distinguish a missing `tool_result` from an ID mismatch when a `tool_use` is followed by a non-result message.

## Tests
- `uv run --frozen pytest tests/server/test_validation.py -q`
- `uv run --frozen ruff check src/mcp/server/validation.py tests/server/test_validation.py`
- `uv run --frozen ruff format --check src/mcp/server/validation.py tests/server/test_validation.py`
- `git diff --check`

### DIFF
--- a/src/mcp/server/validation.py
+++ b/src/mcp/server/validation.py
@@ -89,5 +89,7 @@ def validate_tool_use_result_messages(messages: list[SamplingMessage]) -> None:
     if has_previous_tool_use and previous_content:
         tool_use_ids = {c.id for c in previous_content if c.type == "tool_use"}
         tool_result_ids = {c.tool_use_id for c in last_content if c.type == "tool_result"}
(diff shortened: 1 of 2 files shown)


HarperZ9:
Hi maintainers — checking whether this small validation-error clarification is still useful. The branch is currently mergeable, and the latest recorded CI, conformance, and security checks are green. I can rebase or adjust the wording if the validation contract has moved since this was opened. If the scope is no longer desired, please let me know and I can close it.

PR #3056, opened 2026-07-02
Fix pre_parse_json corrupting str | None values that look like JSON

### LINKED ISSUE #3055: FuncMetadata.pre_parse_json mis-detects str | None as non-string and corrupts JSON-looking string arguments
**Title:** `FuncMetadata.pre_parse_json` mis-detects `str | None` as non-string and corrupts JSON-looking string arguments

**Environment**
- `mcp` version: 1.27.0
- Python: 3.10.12
- Transport: streamable-http (also affects stdio — the bug is transport-agnostic)

**Summary**

`FuncMetadata.pre_parse_json()` in `mcp/server/fastmcp/utilities/func_metadata.py` decides whether to `json.loads()` a string argument based on:

```python
if isinstance(data_value, str) and field_info.annotation is not str:
```

This check is meant to catch cases where a client (e.g. Claude Desktop) stringifies a list/dict argument that should really be a Python object. But `field_info.annotation is not str` is `True` for `Optional[str]` / `str | None` as well, since that annotation is not literally `str`. So any **optional string** parameter gets the same treatment as a `list`/`dict`/model parameter.

If the caller passes a valid string value for such a parameter that *also* happens to parse as a JSON object or array — e.g. a JSON-serialized template body like `'{"blocks": [...]}'` — the value silently gets replaced with a `dict`/`list` before the pydantic argument model is validated. Validation then fails with something like:

```
1 validation error for my_tool_nameArguments
body
  Input should be a valid string [type=string_type, input_value={'blocks': [...]}, input_type=dict]
```

...even though the caller sent a perfectly valid string and the tool signature explicitly declares `body: str | None`.

**Minimal repro**

```python
from typing import Any
import json
from mcp.server.fastmcp.utilities.func_metadata import func_metadata

async def my_tool(body: str | None = None) -> dict[str, Any]:
    return {"body": body}

meta = func_metadata(my_tool)
data = {"body": json.dumps({"blocks": ["a", "b"]})}
new_data = meta.pre_parse_json(data)
print(type(new_data["body"]))          # <class 'dict'>  -- should be <class 'str'>
meta.arg_model.model_validate(new_data)  # raises: Input should be a valid string
```

**Expected behavior**

A parameter typed `str | None` (or any `Union` that includes `str`) should not have its string value re-interpreted as JSON, since the raw string is already a valid value for that field. Pre-parsing should only kick in when a plain `str` could *never* satisfy the annotation (e.g. `list[str]`, `dict[str, Any]`, a Pydantic model, `int`, etc.).

**Suggested fix**

Replace the identity check with one that walks `Union`/`X | Y` members:

```python
def _annotation_accepts_str(annotation: Any) -> bool:
    origin = typing.get_origin(annotation)
    if origin is typing.Union or origin is types.UnionType:
        return any(_annotation_accepts_str(arg) for arg in typing.get_args(annotation))
    return annotation is str

# in pre_parse_json:
if isinstance(data_value, str) and not _annotation_accepts_str(field_info.annotation):
    ...
```

**Impact**

Any FastMCP tool with an `Optional[str]` (or `str | None`) parameter breaks whenever a caller passes a string value that happens to be valid JSON for an object/array (JSON-in-a-string payloads: template bodies, block-based editor content, serialized configs, etc.). We hit this in production with a Unisender email-template MCP server where `body: str | None` holds a JSON block structure — every `update_email_template` / `create_email_template` call with a block-based template failed validation until we monkey-patched `FuncMetadata.pre_parse_json` locally with the fix above.

vientooscuro:
## Summary

`FuncMetadata.pre_parse_json()` decided whether to `json.loads()` a string
argument based on `field_info.annotation is not str`. That check is `True`
for `str | None` as well (it's `Optional[str]`, not literally `str`), so any
optional string parameter got the same "maybe this is JSON" treatment as a
`list`/`dict` parameter.

If a caller passes a valid string that also happens to parse as a JSON
object/array — e.g. a JSON-serialized payload passed as a plain string field
(block-based template body, serialized config, etc.) — the value silently
got replaced with a `dict`/`list` before the pydantic argument model was
validated, and validation then failed with `Input should be a valid string`
even though the caller's string was perfectly valid for that field.

Fixes #3055.

## Fix

Added `_is_optional_str()`, which recognizes `str | None` (any union whose
only non-`None` member is `str`) and skips JSON pre-parsing for it, since no
member of such a union other than `str` could ever be produced by decoding
JSON — so pre-parsing that annotation can only corrupt, never help.

`str | list[str]` (and other unions with a real non-`str` member) keep the
existing behavior, since a JSON array/object there is meaningfully different
from the raw string and should still be pre-parsed — covered by the existing
`test_str_vs_list_str` test, which still passes.

## Test plan

- [x] Added `test_optional_str_is_never_json_pre_parsed` covering dict-like,
      list-like, plain-string, and `None` values for a `str | None` param.
- [x] `uv run pytest tests/server/mcpserver/test_func_metadata.py` — 43 passed
- [x] `uv run pytest tests/server/mcpserver` (full suite) — 526 passed
- [x] `uv run ruff check` / `ruff format --check` on changed files

### DIFF
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -33,6 +33,23 @@ def _is_input_required_type(obj: Any) -> bool:
     return isinstance(obj, type) and issubclass(obj, InputRequiredResult)
 
 
+def _is_optional_str(annotation: Any) -> bool:
+    """Whether `annotation` is `str | None` (`Optional[str]`), modulo `None`.
+
+    Used by `pre_parse_json` to skip JSON pre-parsing for such annotations:
+    unlike e.g. `str | list[str]`, no member of this union other than `str`
+    could ever be produced by decoding a string as JSON, so attempting to
+    parse it only risks corrupting a valid string value that happens to look
+    like a JSON object/array (e.g. a JSON-serialized payload passed as a
+    plain string parameter).
+    """
+    origin = get_origin(annotation)
+    if not is_union_origin(origin):
+        return False
+    non_none_args = [arg for arg in get_args(annotation) if arg is not type(None)]
+    return non_none_args == [str]
+
+
 class StrictJsonSchema(GenerateJsonSchema):
     """A JSON schema generator that raises exceptions instead of emitting warnings.
 
@@ -169,7 +186,11 @@ class FuncMetadata(BaseModel):
                 continue
 
             field_info = key_to_field_info[data_key]
-            if isinstance(data_value, str) and field_info.annotation is not str:
+            if (
+                isinstance(data_value, str)
+                and field_info.annotation is not str
+                and not _is_optional_str(field_info.annotation)
+            ):
                 try:
                     pre_parsed = json.loads(data_value)
                 except json.JSONDecodeError:
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -33,8 +33,18 @@ def _is_input_required_type(obj: Any) -> bool:
     return isinstance(obj, type) and issubclass(obj, InputRequiredResult)
 
 
+def _unwrap_annotated(annotation: Any) -> Any:
+    """Strip `Annotated[...]` down to its underlying type, e.g. for
+    `Annotated[str, Field(description=...)]` (idiomatic for described params).
+    """
+    if get_origin(annotation) is Annotated:
+        return get_args(annotation)[0]
+    return annotation
+
+
 def _is_optional_str(annotation: Any) -> bool:
-    """Whether `annotation` is `str | None` (`Optional[str]`), modulo `None`.
+    """Whether `annotation` is `str | None` (`Optional[str]`), modulo `None`
+    and `Annotated` wrapping.
 
     Used by `pre_parse_json` to skip JSON pre-parsing for such annotations:
     unlike e.g. `str | list[str]`, no member of this union other than `str`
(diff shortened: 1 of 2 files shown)


akminx (commented):
I was working on a fix for #3055 as well (didn't spot this PR first — closing mine, #3062, in favor of this one since it came first and covers the core case well).

One gap worth closing while you're in here: `_is_optional_str` compares union members against `str` directly, but pydantic preserves `Annotated` inside union members — `FieldInfo.annotation` for

```python
def my_tool(body: Annotated[str, Field(description="a body")] | None = None): ...
```

is `Optional[Annotated[str, FieldInfo(...)]]`, so `non_none_args == [str]` is `False` and the corruption from #3055 persists for that spelling — which is idiomatic FastMCP style for described params. Unwrapping `Annotated` members (e.g. `typing.get_args(arg)[0]` when `get_origin(arg) is Annotated`) before the comparison covers it; a test like

```python
def func_with_annotated_str(
    body: Annotated[str, Field(description="a body")] | None = None,
): ...

meta = func_metadata(func_with_annotated_str)
result = meta.pre_parse_json({"body": '{"blocks": ["a", "b"]}'})
assert result["body"] == '{"blocks": ["a", "b"]}'  # fails on this branch
```

reproduces it (and heads-up: CI enforces 100% branch coverage on changed code, so the new branch will need that test anyway).

vientooscuro:
Good catch, thanks — pushed a fix in 9173fe8. `_is_optional_str` now unwraps `Annotated` members before comparing to `str`, so `Annotated[str, Field(description=...)] | None` is treated the same as plain `str | None`. Added your repro as `test_annotated_optional_str_is_never_json_pre_parsed`.

Ran the full suite with coverage to be safe (not just `tests/server/mcpserver`) — 5243 passed, 100% branch coverage repo-wide.

PR #3061, opened 2026-07-03
Validate elicitation sub-capabilities in check_capability

### LINKED ISSUE #2965: [Bug] check_capability does not validate elicitation sub-capabilities (form/url)

Rodrigo-Palma:
`Connection.check_capability()` (exposed publicly via `ServerSession.check_client_capability`) rejected a requested `elicitation` capability only when the client declared **no** elicitation support at all. It never inspected the `form` / `url` sub-capabilities, so a client that advertised only URL-mode elicitation would incorrectly pass a check for form-mode elicitation, and vice versa.

Handle elicitation the same way as `sampling`: when a specific mode is requested, require the client to declare that mode. A bare elicitation request is still satisfied by any elicitation support.

Fixes #2965

## Motivation and Context

`check_capability` mirrors the (correct) `sampling` handling for the top-level presence check but omits the per-sub-capability checks that `sampling` performs for `context` / `tools`:

```python
if capability.elicitation is not None and have.elicitation is None:
    return False
```

Because it only returns `False` when the client has no elicitation at all, a caller checking for form-mode elicitation against a URL-only client (or vice versa) gets a wrong `True`. This is public API, so once a caller relies on it (e.g. checking for form support before `elicit_form`), it silently returns incorrect results.

The fix adds the `form` / `url` sub-capability checks, matching the `sampling` pattern:

```python
if capability.elicitation is not None:
    if have.elicitation is None:
        return False
    if capability.elicitation.form is not None and have.elicitation.form is None:
        return False
    if capability.elicitation.url is not None and have.elicitation.url is None:
        return False
```

Scope kept to elicitation. The issue also mentioned `extensions` and `tasks` as lower-priority gaps: `extensions` is already handled on `main` (SEP-2133), and `tasks` is left out to keep this tightly scoped — happy to follow up separately if desired.

## How Has This Been Tested?

Extended the existing `test_check_capability_per_field_branches` parametrization in `tests/server/test_connection.py` with the elicitation branches: the two bug cases (URL-only client vs. form request, and form-only client vs. URL request), the matching positive cases, and the bare-request case. The two bug cases fail on `main` and pass with this change; the full `test_connection.py` suite is green (44 passed). `ruff check`, `ruff format --check`, and `pyright` are all clean on the changed files.

Docs: added a short "Ask only in a mode the client supports" section to `docs/handlers/elicitation.md` showing how a server uses `check_client_capability` to pick a form/url mode the client supports before eliciting. `mkdocs build --strict` passes.

## Breaking Changes

No API change. `check_capability` now returns the correct `False` for a requested elicitation mode the client does not support (previously a wrong `True`). Callers that were (incorrectly) relying on the old always-true-when-any-elicitation behavior would see the corrected result.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

*AI assistance: I used AI tooling to help analyze the issue, draft the fix, and scaffold the tests. I reviewed and understand the change, stand behind it, and can discuss it directly.*

### DIFF
--- a/docs/handlers/elicitation.md
+++ b/docs/handlers/elicitation.md
@@ -124,6 +124,24 @@ Some things must not go through the model or the client: credentials, card numbe
 
 Look at the second tool. When your server learns the out-of-band flow finished (a webhook, a poll; here it's modelled as a second tool), `ctx.session.send_elicit_complete(...)` sends `notifications/elicitation/complete` with the same `elicitation_id`. That is how the client knows it can stop showing *"waiting for payment..."*. Without it, the client can only guess.
 
+## Ask only in a mode the client supports
+
+A client can declare one mode without the other - a terminal that renders a form but has no browser to open a URL, or a kiosk that can only open a URL. `ctx.session.check_client_capability` reads the `form` / `url` sub-capabilities, so a tool can pick a mode the client actually supports before it asks:
+
+```python
+from mcp_types import ClientCapabilities, ElicitationCapability, FormElicitationCapability
+
+
+async def book_table(ctx: Context) -> str:
+    wants_form = ClientCapabilities(elicitation=ElicitationCapability(form=FormElicitationCapability()))
+    if not ctx.session.check_client_capability(wants_form):
+        return "This client can't render a form; send a URL or return a default instead."
+    result = await ctx.elicit("Which date?", schema=AlternativeDate)
+    return "booked" if result.action == "accept" else "no change"
+```
+
+A bare `ElicitationCapability()` (no mode set) matches any client that supports elicitation at all, so name a mode only when you need that specific one. This is the same *"what if I can't ask?"* design the client-side check below calls out - now decided per mode.
+
 ## The client side
 
 Servers ask. Clients answer by passing an **`elicitation_callback`** to `Client(...)`:
--- a/src/mcp/server/connection.py
+++ b/src/mcp/server/connection.py
@@ -388,8 +388,13 @@ class Connection:
                 return False
             if capability.sampling.tools is not None and have.sampling.tools is None:
                 return False
-        if capability.elicitation is not None and have.elicitation is None:
-            return False
+        if capability.elicitation is not None:
+            if have.elicitation is None:
+                return False
+            if capability.elicitation.form is not None and have.elicitation.form is None:
+                return False
+            if capability.elicitation.url is not None and have.elicitation.url is None:
+                return False
         if capability.experimental is not None:
             if have.experimental is None:
                 return False
--- a/tests/server/test_connection.py
+++ b/tests/server/test_connection.py
@@ -21,6 +21,7 @@ from mcp_types import (
     CreateMessageRequestParams,
     ElicitationCapability,
     EmptyResult,
+    FormElicitationCapability,
     Implementation,
     ListRootsRequest,
     ListRootsResult,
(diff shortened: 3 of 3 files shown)


STiFLeR7:
Reviewed against current \`upstream/main\`'s \`connection.py\`: before this change, \`check_capability\` only checked \`have.elicitation is None\`, so a server calling \`ctx.session.check_client_capability(ClientCapabilities(elicitation=ElicitationCapability(form=...)))\` against a client that declared *only* \`url\`-mode elicitation would get \`True\` back — a false positive, since that client can't actually render a form. The fix's per-submode check (\`form\`/\`url\`) mirrors the existing pattern already used two lines above for \`sampling.tools\`, so it's consistent with how this method handles other nested sub-capabilities rather than introducing a new pattern.

The added test matrix in \`test_connection.py\` covers exactly the cases that matter: bare \`ElicitationCapability()\` still matches any elicitation support (no regression for existing callers that don't care about the mode), form-only vs url-only client in both directions correctly returns \`False\`, and matching submodes return \`True\`. That's the right shape for a capability-negotiation fix — it's easy to get "any None sub-field is a wildcard" backwards, and the test cases confirm the None-vs-declared-vs-mismatched semantics are all handled correctly.

The docs addition (\`elicitation.md\`) is a nice inclusion — without it, \`check_client_capability\` with a specific submode isn't obviously in scope for a tool author skimming the elicitation guide. CI green across the matrix. Looks correct to me.

maxisbey:
Thanks for the PR. Tracking this in #2979 instead. Closing this as part of a wider backlog cleanup following the v2 launch. Feel free to reopen if this is still relevant.

PR #3063, opened 2026-07-04
fix(streamable-http): reject duplicate in-flight request ids

### LINKED ISSUE #3060: streamable HTTP (stateful): concurrent requests with duplicate JSON-RPC ids cross-wire responses — one request receives another's payload, the other hangs
### Summary

In the stateful streamable-HTTP server transport, two concurrent POSTs on the same session that carry the **same JSON-RPC request id** cross-wire: the later request receives the earlier request's response (entire envelope, wrong arity and all), the later request's own response is silently dropped, and the earlier request's SSE stream hangs forever (keep-alive pings defeat client read timeouts).

Duplicate in-flight ids violate the spec ("The request ID MUST NOT have been previously used by the requestor within the same session"), but a real production client — **claude.ai's custom-connector MCP client sends every request with `id: 1`** — triggers this constantly, and the server neither rejects nor tolerates the violation: it silently mis-routes user data. We observed one user's tool response delivered to a different conversation's request in production before isolating this mechanism.

### Mechanism (v1.27.0; the same code is present on current `main`)

`mcp/server/streamable_http.py`:

1. POST handler (~L534–537): `request_id = str(message.root.id)` then `self._request_streams[request_id] = anyio.create_memory_object_stream(...)` — the routing table is keyed **by request id alone**. A second concurrent POST with the same id **silently overwrites** the first's slot; the first POST's reader keeps a now-orphaned stream.
2. `message_router` (~L997–1045): the handler's response is routed by `response_id` to whichever POST currently owns the slot — i.e. the **latest** arrival, regardless of which request it answers.
3. `sse_writer` cleanup (~L597–616): after delivering one response the slot is popped, so the second response finds no slot and is dropped (debug log: `Request stream 1 not found`).

Result for two overlapping requests A (id=1, slow) then B (id=1): **B receives A's response; B's response is dropped; A hangs indefinitely.**

`mcp/shared/session.py` (~L375) has the same single-key assumption in `_in_flight[responder.request_id]`, which additionally breaks cancellation targeting under duplicate ids.

### Reproduction

Toy server (echo tool with random 0–2s sleep):

```python
"""repro_server.py"""
import random
import anyio

mcp = FastMCP(name="repro", host="127.0.0.1", port=18999)

@mcp.tool()
async def echo(sentinel: str) -> str:
    await anyio.sleep(random.uniform(0.0, 2.0))
    return f"ECHO:{sentinel}"

if __name__ == "__main__":
```

Client: one initialized session, 12 concurrent `tools/call` POSTs, each with a unique sentinel; mode `same_id` uses `id: 1` for all (mimicking claude.ai), mode `unique_id` is the control:

```python
"""repro_client.py — run: python repro_client.py same_id unique_id"""
import asyncio, json, sys
import httpx

BASE = "http://127.0.0.1:18999/mcp"
HEADERS = {"accept": "application/json, text/event-stream", "content-type": "application/json"}
N, TIMEOUT = 12, 20.0

def parse_sse_for_response(text):
    for line in text.splitlines():
        if line.startswith("data:"):
            try:
                obj = json.loads(line[5:].strip())
            except json.JSONDecodeError:
                continue
            if "result" in obj or "error" in obj:
                return obj
    return None

async def initialize(client):
    r = await client.post(BASE, headers=HEADERS, json={
        "jsonrpc": "2.0", "id": 0, "method": "initialize",
        "params": {"protocolVersion": "2025-06-18", "capabilities": {},
                   "clientInfo": {"name": "repro", "version": "0"}}})
    r.raise_for_status()
    sid = r.headers.get("mcp-session-id")
    hdrs = {**HEADERS, **({"mcp-session-id": sid} if sid else {})}
    r2 = await client.post(BASE, headers=hdrs,
                           json={"jsonrpc": "2.0", "method": "notifications/initialized"})
    assert r2.status_code in (200, 202)
    return sid

def _sid_headers(sid):
    return {**HEADERS, **({"mcp-session-id": sid} if sid else {})}

async def call_tool(client, sid, rpc_id, sentinel):
    body = {"jsonrpc": "2.0", "id": rpc_id, "method": "tools/call",
            "params": {"name": "echo", "arguments": {"sentinel": sentinel}}}
    try:
        # wall-clock timeout: SSE pings defeat httpx read timeouts on orphaned streams
        r = await asyncio.wait_for(
            client.post(BASE, headers=_sid_headers(sid), json=body, timeout=TIMEOUT),
            timeout=TIMEOUT)
    except (httpx.TimeoutException, asyncio.TimeoutError):
        return (sentinel, "TIMEOUT", None)
    obj = parse_sse_for_response(r.text)
    if obj is None or "error" in obj:
        return (sentinel, "OTHER", r.text[:120])
    got = json.dumps(obj["result"])
    return (sentinel, "OK" if f"ECHO:{sentinel}" in got else "SWAPPED", got[:160])

async def run_burst(mode):
    async with httpx.AsyncClient() as client:
        sid = await initialize(client)
        tasks = [call_tool(client, sid, 1 if mode == "same_id" else 100 + i, f"SENT_{mode}_{i:02d}")
                 for i in range(N)]
        results = await asyncio.gather(*tasks)
    print(f"=== mode={mode} ===")
    print("OK=%d SWAPPED=%d TIMEOUT=%d" % (
        sum(s == "OK" for _, s, _ in results),
        sum(s == "SWAPPED" for _, s, _ in results),
        sum(s == "TIMEOUT" for _, s, _ in results)))
    for sent, status, detail in results:
        if status != "OK":
            print(f"  {sent}: {status} {detail or ''}")

async def main():
    for mode in sys.argv[1:] or ["same_id", "unique_id"]:
        await run_burst(mode)

asyncio.run(main())
```

### Observed results (mcp 1.27.0, Python 3.12, Linux)

| mode | OK | SWAPPED | TIMEOUT |
|---|---|---|---|
| `same_id` (12 concurrent, all `id: 1`) | 0 | 1 | 11 |
| `unique_id` control | 12 | 0 | 0 |

Pairwise variant (A slow, B posted 300ms later, both `id: 1`, 10 trials): A **never** receives a response; B receives **A's payload** in 7/10 trials (B's own in the rest). This exactly matched our production incident, including the response-arity mismatch.

`stateless_http=True` is immune by construction (per-request transport): same burst is 12/12 OK.

### Suggested fix

At minimum, reject a POST whose request id is already in flight on the session with a JSON-RPC `-32600` error instead of silently overwriting the routing slot — a protocol violation should fail loudly, not deliver one user's data to another request. (Queueing/serializing duplicate-id requests would also work and keeps the misbehaving-but-widespread client functional.)

The TypeScript SDK server has the same unguarded pattern (`_requestToStreamMapping.set(message.id, streamId)`); filing separately there. The client-side id-reuse is also being reported to Anthropic.

Sammy-Dabbas:
Fixes #3060.

The stateful streamable HTTP transport keys per-request routing by the request id and assigns the slot without an existence check. Two concurrent POSTs sharing an id cross-wire: the second overwrites the first's routing slot, the first request's response is delivered to the second POST, and the first hangs while its stream leaks. Reproduced on main before the fix.

This adds a guard in `_handle_post_request` that rejects a POST whose request id is already in flight on the session with HTTP 400 and JSON-RPC -32600, placed before the JSON/SSE branch so both response modes are covered. The spec requires request ids to be unique within a session. Sequential reuse of an id after the earlier request completes still works, since deployed clients send a constant id for every request; a regression test pins that behavior.

Tests: the new duplicate-id test fails on unpatched main and passes with the fix. tests/shared/test_streamable_http.py passes 67/67 and the full suite passes locally (5282 passed). A stress run of 500 rapid sequential same-id requests produced no spurious rejections. ruff and pyright are clean on the touched files.

Scope note: this is the transport-level guard only. The dispatcher-level blind overwrite in jsonrpc_dispatcher.py (TODO from #3046) is deliberately left to the follow-up discussed on the issue, since the two guards compose.

Disclosure: this change was developed with AI assistance (Claude Code). I reviewed the change and the test results before submitting.

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -611,18 +611,27 @@ class StreamableHTTPServerTransport:
                 finally:
                     await self._clean_up_memory_streams(request_id)
             else:
-                # Mint the priming event before any per-request state exists:
-                # `EventStore.store_event` is user code and may raise, in which
-                # case the outer handler returns a 500 with nothing to clean up.
-                # Still strictly precedes dispatch, so storage order == wire order.
-                priming_event = await self._mint_priming_event(request_id, protocol_version)
-
-                sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[SSEEvent](0)
-                self._sse_stream_writers[request_id] = sse_stream_writer
+                # Reserve the routing slot before any await so nothing separates
+                # the duplicate-id guard above from registration; a concurrent
+                # POST reusing this id while the event store persists the priming
+                # event could otherwise pass the guard and overwrite the slot.
                 self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](
                     REQUEST_STREAM_BUFFER_SIZE
                 )
                 request_stream_reader = self._request_streams[request_id][1]
+
+                # Mint the priming event before dispatch so storage order matches
+                # wire order. `EventStore.store_event` is user code and may raise,
+                # in which case the outer handler returns a 500; release the slot
+                # reservation on the way out so the id does not stay in flight.
+                try:
+                    priming_event = await self._mint_priming_event(request_id, protocol_version)
+                except BaseException:
+                    await self._clean_up_memory_streams(request_id)
+                    raise
+
+                sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[SSEEvent](0)
+                self._sse_stream_writers[request_id] = sse_stream_writer
 
                 headers = {
                     "Cache-Control": "no-cache, no-transform",
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -549,6 +549,20 @@ class StreamableHTTPServerTransport:
 
             request_id = str(message.id)
 
+            # Reject duplicate in-flight request ids: `_request_streams` is keyed by
+            # request id, so a second concurrent request with the same id would
+            # silently overwrite the first one's routing slot and cross-wire their
+            # responses (one request receives the other's payload, the other hangs).
+            # The spec requires ids to be unique within a session; ids may still be
+            # reused once the earlier request has completed. See #3060.
+            if request_id in self._request_streams:
+                response = self._create_error_response(
+                    f"Bad Request: Request id {request_id} is already in flight for this session",
+                    HTTPStatus.BAD_REQUEST,
+                )
+                await response(scope, receive, send)
+                return
+
             if self.is_json_response_enabled:
                 self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](
                     REQUEST_STREAM_BUFFER_SIZE
--- a/tests/shared/test_streamable_http.py
+++ b/tests/shared/test_streamable_http.py
@@ -11,7 +11,7 @@ import time
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, cast
 from unittest.mock import MagicMock
 from urllib.parse import urlparse
 
@@ -49,7 +49,6 @@ from mcp.client.streamable_http import StreamableHTTPTransport, streamable_http_
 from mcp.server import Server, ServerRequestContext
 from mcp.server.streamable_http import (
     GET_STREAM_KEY,
-    MCP_PROTOCOL_VERSION_HEADER,
     MCP_SESSION_ID_HEADER,
     SESSION_ID_PATTERN,
     EventCallback,
@@ -63,6 +62,7 @@ from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
 from mcp.server.transport_security import TransportSecuritySettings
 from mcp.shared._compat import resync_tracer
 from mcp.shared._context_streams import create_context_streams
+from mcp.shared.inbound import MCP_PROTOCOL_VERSION_HEADER
 from mcp.shared.message import ClientMessageMetadata, ServerMessageMetadata, SessionMessage
 from mcp.shared.session import RequestResponder
 from tests.interaction.transports import StreamingASGITransport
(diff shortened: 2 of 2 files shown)


Sammy-Dabbas:
Good catch on the priming race, that was a real gap: the guard ran before the awaited store_event, so a same-id POST could slip in during event-store persistence. Fixed by reserving the routing slot synchronously with the guard and minting the priming event afterwards, with the reservation released if the event store raises so the 500 path leaves nothing in flight. Added a regression test that suspends a gated event store mid-priming and asserts the concurrent duplicate is still rejected; it fails by timeout on the previous commit and passes now. Full transport module is green (68 tests).

akminx:
Nice — the slot reservation before priming closes the resumable-SSE race cleanly. Two mechanical CI things are what's currently red, both in the newly added test code rather than the fix itself:

**1. Coverage gate (the 20 `test (...)` matrix jobs).** `coverage report` fails with:

```
tests/shared/test_streamable_http.py   990   1   108   1   99.82%   844
TOTAL                                                              99.99%
Coverage failure: total of 99.99 is less than fail-under=100.00
```

Line 844 in the new test is uncovered (this repo enforces `fail-under = 100` with `branch = true`, and `tests/` is in `[tool.coverage.run] source`). Worth running `uv run --frozen --no-sync coverage run -m pytest -n auto && uv run --frozen --no-sync coverage combine && uv run --frozen --no-sync coverage report` locally to see the exact arc — usually it's one branch of the gated-event-store path that isn't taken.

**2. pre-commit (pyright).** The `pre-commit` job fails on 6 pyright errors in the same new test, e.g.:

```
tests/shared/test_streamable_http.py:853:28 - error: Expected mapping for dictionary unpack operator (reportGeneralTypeIssues)
tests/shared/test_streamable_http.py:851:13 - error: Type of "init_request" is partially unknown
```

Looks like `init_request` needs a concrete type so the `**`-unpack resolves.

Once those two are green the change looks solid. For what it's worth, I'm putting up the dispatcher-layer piece we discussed (the `jsonrpc_dispatcher.py` `_in_flight` overwrite + the cancellation-targeting case) as a separate PR, scoped exactly as you described — transport guard and dispatcher guard composing.

Sammy-Dabbas:
Thanks for running those down, both fixed. The uncovered line was the test helper's defensive raise, now marked no-cover matching the module-level first_sse_data helper's convention. The pyright errors came from the untyped dict unpacks; init_request and the tool call payload now have concrete types, and I moved the protocol version header import to its defining module while there. pyright is clean on the file and the coverage gate passes locally apart from pre-existing platform-conditional lines in other test files that only execute on Linux.

Glad the dispatcher piece is up as #3064, and the scoping is exactly right. The two guards compose and neither alone closes the issue, so from my side please keep it open.

PR #3068, opened 2026-07-06
fix: forbid extra arguments in FastMCP tool arg models (#3067)

### LINKED ISSUE #3067: FastMCP tool argument models silently ignore unknown/misspelled arguments (extra=ignore default)
## Description

`mcp.server.fastmcp.utilities.func_metadata.ArgModelBase` does not set `extra` in its
`model_config`, so it inherits Pydantic v2's default, `extra="ignore"`:

```python
class ArgModelBase(BaseModel):
    """A model representing the arguments to a function."""
    ...
    model_config = ConfigDict(
        arbitrary_types_allowed=True,
    )
```

Every per-tool `*Arguments` model that FastMCP generates via `create_model(..., __base__=ArgModelBase, ...)`
inherits this. As a result, when a client calls a tool with an argument name that doesn't
exist on the function (a typo, or a hallucinated parameter name from an LLM), Pydantic
silently drops it instead of raising a validation error. The tool then runs with that
parameter at its default value, and returns a "successful" but semantically wrong result —
with no signal anywhere that anything was off.

## Reproduction

```python

mcp = FastMCP("repro")

@mcp.tool()
def read_doc(topic: str = "") -> str:
    return f"topic={topic!r}"

# Simulate what happens when a client calls with the wrong argument name:
tool = mcp._tool_manager.get_tool("read_doc")
result = tool.fn_metadata.arg_model.model_validate({"path": "something"})
print(result.model_dump_one_level())
# {'topic': ''}  <- no error, no signal that "path" was bogus
```

## Expected behavior

An unknown argument in a `tools/call` request should fail validation immediately with a
clear error (e.g. Pydantic's own "Extra inputs are not permitted"), the same way a *missing*
required argument already does. This is especially important for MCP given the primary
caller is frequently an LLM: a wrong parameter name is a common, plausible failure mode, and
a loud, immediate error is far more useful (and self-correcting for the model) than a
silently-wrong "successful" response.

## Suggested fix

Set `extra="forbid"` on `ArgModelBase.model_config`:

```python
model_config = ConfigDict(
    arbitrary_types_allowed=True,
    extra="forbid",
)
```

This should be safe: the fields validated by `ArgModelBase` subclasses are exactly the
`arguments` dict of a `tools/call` request (i.e. exactly what's declared in the tool's
`inputSchema`). Protocol-level fields (`_meta`, `progressToken`, etc.) live on the
surrounding `params` object, not inside `arguments`, and `Context` is injected separately
via `arguments_to_pass_directly` — neither passes through `ArgModelBase`. The side effect is
that `model_json_schema()` will now emit `"additionalProperties": false` on the published
`inputSchema`, which seems like a feature, not a regression (it lets clients that validate
against the schema catch this class of error before the round-trip).

## Scope checked

- Confirmed present in `mcp` 1.27.2 and 1.28.1 (latest on PyPI as of 2026-07).
- Confirmed still present on `main` (the in-progress v2, where `fastmcp` is being renamed to
  `mcpserver`): `src/mcp/server/mcpserver/utilities/func_metadata.py` has the same
  `ArgModelBase` without `extra` set.
- I didn't find an existing issue about this specific behavior (searched for "extra fields",
  "forbid", "unknown argument", "additionalProperties", "silently ignored").
- For context: the TypeScript SDK has an equivalent gap for a different underlying reason
  (Zod's default `strip` behavior on unknown object keys) — see
  https://github.com/modelcontextprotocol/typescript-sdk/issues/147

Happy to open a PR with the one-line change plus a regression test if that's useful.

elbachir-salik:
Closes #3067 
ArgModelBase now uses extra="forbid" so unknown or misspelled tool arguments raise a validation error instead of being silently ignored.
Added a regression test confirming unknown args are rejected.

### DIFF
--- a/docs/handlers/dependencies.md
+++ b/docs/handlers/dependencies.md
@@ -35,7 +35,7 @@ Here is the input schema `tools/list` reports for `reserve_book`:
 }
 ```
 
-One property. Like the `Context` in **[The Context](context.md)**, a resolved parameter is a contract between you and the SDK: `stock` is not in the schema, the model is never told about it, and a client that sends a `stock` value anyway is ignored. The resolver's value is the only one your tool can receive.
+One property. Like the `Context` in **[The Context](context.md)**, a resolved parameter is a contract between you and the SDK: `stock` is not in the schema, the model is never told about it, and a client that sends a `stock` value anyway is rejected. The resolver's value is the only one your tool can receive.
 
 That last part is the point. A parameter the model cannot supply is a parameter the model cannot get wrong.
 
--- a/src/mcp/server/mcpserver/utilities/func_metadata.py
+++ b/src/mcp/server/mcpserver/utilities/func_metadata.py
@@ -60,7 +60,7 @@ class ArgModelBase(BaseModel):
             kwargs[output_name] = value
         return kwargs
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")
 
 
 class FuncMetadata(BaseModel):
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -156,6 +156,7 @@ async def test_client_list_tools(app: MCPServer):
                         description="Greet someone by name.",
                         input_schema={
                             "properties": {"name": {"title": "Name", "type": "string"}},
+                            "additionalProperties": False,
                             "required": ["name"],
                             "title": "greetArguments",
                             "type": "object",
--- a/tests/docs_src/test_client.py
+++ b/tests/docs_src/test_client.py
@@ -53,6 +53,7 @@ async def test_list_tools_returns_the_full_definition() -> None:
         assert tool.input_schema == snapshot(
             {
                 "type": "object",
+                "additionalProperties": False,
                 "properties": {
                     "query": {"title": "Query", "type": "string"},
                     "limit": {"default": 10, "title": "Limit", "type": "integer"},
--- a/tests/docs_src/test_context.py
+++ b/tests/docs_src/test_context.py
@@ -20,6 +20,7 @@ async def test_the_context_parameter_is_not_in_the_input_schema() -> None:
         assert tool.input_schema == snapshot(
             {
                 "type": "object",
+                "additionalProperties": False,
                 "properties": {"query": {"title": "Query", "type": "string"}},
                 "required": ["query"],
                 "title": "search_booksArguments",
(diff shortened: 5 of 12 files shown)


STiFLeR7:
Reviewed against current \`upstream/main\`: \`ArgModelBase.model_config\` is the single Pydantic base every FastMCP-generated tool arg model inherits (\`func_metadata.py\`), and it's built purely from the function's declared signature params (Context-typed params are already excluded before the model is constructed, confirmed via the existing \`test_context.py\` snapshot tests) — there's no code path here that legitimately needs to accept an argument name outside the declared schema, and no \`**kwargs\`\-style dynamic-arg support exists in \`func_metadata.py\` today. So \`extra="forbid"\` closes a real hole (misspelled/unknown \`tools/call\` arguments silently vanishing instead of failing) without a regression path for the current codebase.

Also checked that the updated \`additionalProperties: false\` in the JSON Schema output matches the new \`extra="forbid"\` semantics one-to-one (Pydantic sets that schema keyword automatically for forbid-mode models), so client-side schema validation and server-side runtime validation stay in sync — no silent divergence between what the tool *advertises* and what it *enforces*.

Tests cover both the schema-shape change (existing snapshot tests updated) and the new runtime behavior (\`test_unknown_argument_raises_validation_error\`, plus the doc-tutorial test that now asserts rejection instead of silent drop). CI is green across the full Python matrix. Looks correct to me.

elbachir-salik:
> Reviewed against current `upstream/main`: `ArgModelBase.model_config` is the single Pydantic base every FastMCP-generated tool arg model inherits (`func_metadata.py`), and it's built purely from the function's declared signature params (Context-typed params are already excluded before the model is constructed, confirmed via the existing `test_context.py` snapshot tests) — there's no code path here that legitimately needs to accept an argument name outside the declared schema, and no `**kwargs`-style dynamic-arg support exists in `func_metadata.py` today. So `extra="forbid"` closes a real hole (misspelled/unknown `tools/call` arguments silently vanishing instead of failing) without a regression path for the current codebase.
> 
> Also checked that the updated `additionalProperties: false` in the JSON Schema output matches the new `extra="forbid"` semantics one-to-one (Pydantic sets that schema keyword automatically for forbid-mode models), so client-side schema validation and server-side runtime validation stay in sync — no silent divergence between what the tool _advertises_ and what it _enforces_.
> 
> Tests cover both the schema-shape change (existing snapshot tests updated) and the new runtime behavior (`test_unknown_argument_raises_validation_error`, plus the doc-tutorial test that now asserts rejection instead of silent drop). CI is green across the full Python matrix. Looks correct to me.

Thanks for the thorough review! Glad the fix holds up across all the edge cases you checked.

PR #3069, opened 2026-07-06
fix(server): guard GET SSE response with CancelScope to prevent Windows deadlock (#2653)

### LINKED ISSUE #2653: mcp.server.fastmcp.FastMCP HTTP transport deadlocks after ~5 sequential client sessions on Windows (companion to PrefectHQ/fastmcp#4192)
# `mcp.server.fastmcp.FastMCP` HTTP transport deadlocks after ~5 sequential client sessions on Windows (companion to PrefectHQ/fastmcp#4192)

## Summary

Filing here as a companion to [PrefectHQ/fastmcp#4192](https://github.com/PrefectHQ/fastmcp/issues/4192), which contains the full investigation, cross-control matrix, stack-dump evidence, and minimal reproducers. The reason for the companion issue: the bug reproduces on Anthropic's `mcp.server.fastmcp.FastMCP` shipped inside this SDK, *faster* than on PrefectHQ's downstream fork, and the SDK maintainers should see the signal directly rather than having to chase a downstream cross-reference.

## The one-line result that's yours

A 30-sequential-session driver running against an HTTP server built with `mcp.server.fastmcp.FastMCP` 1.27.1 on Windows + Python 3.14 hangs at **session #5**, with gradual slowdown visible from session #4 onwards. The same driver against PrefectHQ `fastmcp.FastMCP` 3.3.1 hangs at session #12; against raw `mcp.server.lowlevel.Server` it passes **30/30**; against the Node.js TypeScript reference server (`@modelcontextprotocol/server-everything`) it passes **30/30**.

That is: this SDK's `FastMCP` wrapper exhibits the defect *more aggressively* than the downstream fork, and the SDK's own lower-level surface is unaffected.

## Cross-control matrix (excerpt; full matrix in PrefectHQ/fastmcp#4192)

| Server | Python | Sessions before hang |
|---|---|---|
| `mcp.server.fastmcp.FastMCP` 1.27.1 | 3.14 | **5** |
| `mcp.server.fastmcp.FastMCP` 1.27.1 | 3.13.7 | 12 |
| `fastmcp.FastMCP` 3.3.1 (PrefectHQ) | 3.14 | 12 |
| `mcp.server.lowlevel.Server` + `StreamableHTTPSessionManager` | 3.14 | 30/30 ✅ |
| `@modelcontextprotocol/server-everything` (Node.js streamableHttp) | n/a | 30/30 ✅ |

Python-version variance rules out a 3.14 regression. The Node.js control rules out a protocol-level defect. The raw-SDK control rules out the SDK's lower-level transport.

## Mechanism (per stack-dump evidence in #4192)

`python -m asyncio ps <PID>` on the hung server process shows multiple Task entries (one per "closed" session) parked in `TaskGroup.__aexit__`, all with child tasks named `sse_starlette.sse.EventSourceResponse.__call__.<locals>.cancel_on_finish` blocked in `Event.wait` / `sleep` on `_ping`, `_listen_for_disconnect`, `_stream_response`, and the SDK's `standalone_sse_writer`.

`mcp.server.fastmcp.FastMCP.create_streamable_http_app` instantiates `EventSourceResponse` from `sse_starlette` as part of its ASGI app construction. The raw `mcp.server.lowlevel.Server` path writes `text/event-stream` directly via its own ASGI write loop without ever instantiating `EventSourceResponse` — which is why it isn't affected. The defect is in the wrapper layer's interaction with `sse_starlette`, not the SDK's core transport.

## Reproducers

Two minimal reproducers in the PrefectHQ issue: one for PrefectHQ FastMCP, one for raw SDK (the passing control). A third targeting `mcp.server.fastmcp.FastMCP` directly is trivially derivable — same shape, different import — and I'm happy to add it here if useful.

## Why the cap is lower on this SDK's wrapper

Working hypothesis (not yet verified): `mcp.server.fastmcp.FastMCP` and PrefectHQ's `fastmcp.FastMCP` use slightly different `sse_starlette` integration shapes. Either version-pin differences, or an extra layer of session-lifecycle handling in one wrapper that creates more orphan tasks per session, would explain why the threshold lands at 5 here vs 12 downstream. Confirming this would need a side-by-side diff of the two wrappers' `create_streamable_http_app` paths and `sse_starlette` versions, which I haven't done.

## Platform / environment

- Windows 11 (`ProactorEventLoop`)
- Python 3.14 and 3.13.7
- `mcp` 1.27.1 (the SDK that ships `mcp.server.fastmcp.FastMCP`)
- Linux **not validated** — PrefectHQ's public CI runs Linux + Python 3.12 where `epoll`-based loop semantics may drain orphan tasks faster. Maintainer Linux validation requested in #4192; same request stands here.

## The question for SDK maintainers

Is `mcp.server.fastmcp.FastMCP` considered an actively-maintained surface that the SDK team will fix, or is it a thin reference wrapper where PrefectHQ's fork is the canonical production target? The answer changes where the fix should land and whether this issue stays open here or gets closed in favour of #4192.

If it's in-scope here, the cross-controls in #4192 already narrow the surface to "`EventSourceResponse` integration in the FastMCP wrapper layer." The raw SDK path is the working reference.

## Workaround for users of this SDK

Migrate from `mcp.server.fastmcp.FastMCP` to `mcp.server.lowlevel.Server` + manually-constructed `StreamableHTTPSessionManager` + Starlette + uvicorn. ~1 day's mechanical rewrite for a server with ~7 tools. Empirically 30/30 sessions passing on the same workload.

isheng-eqi:
## Summary

Fixes #2653 — `mcp.server.fastmcp.FastMCP` HTTP transport deadlocks after ~5 sequential client sessions on Windows.

## Root Cause

`sse-starlette`'s `EventSourceResponse.__call__` creates an internal `anyio.TaskGroup` whose child tasks (`_ping`, `_listen_for_disconnect`, `_listen_for_exit_signal`, `standalone_sse_writer`) may not respond to cancellation promptly on Windows ProactorEventLoop. When the client disconnects, orphan children remain parked in `anyio.sleep()` or `Event.wait()`, preventing `TaskGroup.__aexit__` from completing.

The `_handle_get_request` handler directly `await`s the `EventSourceResponse` callable, so the hung `TaskGroup` blocks the handler forever. Each session that opens a GET SSE stream and disconnects leaks one such hung task group. After enough sessions (~5–18 depending on Python version), the next `initialize` POST never dispatches and the server deadlocks.

The raw `mcp.server.lowlevel.Server` + `StreamableHTTPSessionManager` path is unaffected because it writes `text/event-stream` directly without `sse-starlette`. The Node.js reference server is also unaffected. The defect is confined to the FastMCP wrapper layer's use of `sse-starlette.EventSourceResponse`.

## Fix

Three changes in `src/mcp/server/streamable_http.py`:

1. **`__init__`**: Add `_active_get_response_scope: anyio.CancelScope | None` attribute to track the in-flight GET SSE response.

2. **`_handle_get_request`**: Wrap the `await response(...)` call with the cancel scope instead of awaiting it directly. This gives `terminate()` a handle to force-cancel a hung response.

3. **`terminate()`**: Cancel `_active_get_response_scope` (if set) before closing streams. This breaks the `sse-starlette` TaskGroup deadlock and allows the session to clean up, even on Windows ProactorEventLoop.

The fix is minimal (19 insertions, 5 deletions), backward-compatible, and targets only the GET SSE path where `EventSourceResponse` is awaited directly. The POST handler already runs `EventSourceResponse` inside an `anyio.create_task_group()` which provides equivalent cancellation safety.

## Related

- PrefectHQ/fastmcp#4192 — full investigation, cross-control matrix, deterministic reproduction
- `sse-starlette` — issue is in the library's TaskGroup cleanup on Windows, but the SDK can defend against it here

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -188,6 +188,9 @@ class StreamableHTTPServerTransport:
         ] = {}
         self._sse_stream_writers: dict[RequestId, MemoryObjectSendStream[SSEEvent]] = {}
         self._terminated = False
+        # CancelScope for the active GET SSE response, so terminate() can
+        # force-cancel a hung EventSourceResponse on Windows (python-sdk#2653).
+        self._active_get_response_scope: anyio.CancelScope | None = None
         # Idle timeout cancel scope; managed by the session manager.
         self.idle_scope: anyio.CancelScope | None = None
 
@@ -743,14 +746,20 @@ class StreamableHTTPServerTransport:
         )
 
         try:
-            # This will send headers immediately and establish the SSE connection
-            await response(request.scope, request.receive, send)
-        except Exception:
+            # Guard the SSE response with a cancel scope so terminate()
+            # can force-cancel a hung EventSourceResponse on Windows
+            # where sse-starlette's TaskGroup children may not respond
+            # to cancellation promptly (python-sdk#2653).
+            self._active_get_response_scope = anyio.CancelScope()
+            with self._active_get_response_scope:
+                await response(request.scope, request.receive, send)
+        except Exception:  # pragma: lax no cover
             logger.exception("Error in standalone SSE response")
+            await self._clean_up_memory_streams(GET_STREAM_KEY)
+        finally:
+            self._active_get_response_scope = None
             await sse_stream_writer.aclose()
             await sse_stream_reader.aclose()
-            await self._clean_up_memory_streams(GET_STREAM_KEY)
-
     async def _handle_delete_request(self, request: Request, send: Send) -> None:  # pragma: no cover
         """Handle DELETE requests for explicit session termination."""
         # Validate session ID
@@ -786,6 +795,11 @@ class StreamableHTTPServerTransport:
 
         self._terminated = True
         logger.info(f"Terminating session: {self.mcp_session_id}")
+
+        # Cancel any in-flight GET SSE response to prevent
+        # sse-starlette TaskGroup deadlock on Windows (python-sdk#2653).
+        if self._active_get_response_scope is not None:
+            self._active_get_response_scope.cancel()
 
         # We need a copy of the keys to avoid modification during iteration
         request_stream_keys = list(self._request_streams.keys())
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -799,7 +799,7 @@ class StreamableHTTPServerTransport:
 
         # Cancel any in-flight GET SSE response to prevent
         # sse-starlette TaskGroup deadlock on Windows (python-sdk#2653).
-        if self._active_get_response_scope is not None:
+        if self._active_get_response_scope is not None:  # pragma: no cover
             self._active_get_response_scope.cancel()
 
         # We need a copy of the keys to avoid modification during iteration
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -188,9 +188,11 @@ class StreamableHTTPServerTransport:
         ] = {}
         self._sse_stream_writers: dict[RequestId, MemoryObjectSendStream[SSEEvent]] = {}
         self._terminated = False
-        # CancelScope for the active GET SSE response, so terminate() can
-        # force-cancel a hung EventSourceResponse on Windows (python-sdk#2653).
-        self._active_get_response_scope: anyio.CancelScope | None = None
+        # CancelScopes for active GET SSE responses, so terminate() can
+        # force-cancel all hung EventSourceResponses on Windows (python-sdk#2653).
+        # A set is used instead of a single scope because concurrent GET-start
+        # races can create multiple in-flight SSE responses.
+        self._active_get_response_scopes: set[anyio.CancelScope] = set()
         # Idle timeout cancel scope; managed by the session manager.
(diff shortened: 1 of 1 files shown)


isheng-eqi (commented):
Good catch. The single `_active_get_response_scope` does have a concurrent GET-start race — two in-flight SSE connections would leave the first without a cancellation handle.

Fixed in 3c0d47b: switched to `set[CancelScope]` so `terminate()` iterates and cancels all active scopes. Added `scope: CancelScope | None = None` guard + `if scope is not None` in the finally block for type safety.

isheng-eqi:
Hi @maxisbey @felixweinberger, could you take a look at this PR when you have a moment?

This fixes #2653 — a Windows-specific deadlock in FastMCP SSE transport where sse-starlette TaskGroup children hang on ProactorEventLoop.

**CI status**: The fork CI is all green (isheng-eqi/python-sdk actions). The upstream CI checks show failure because the required workflow names don't match what runs on fork PRs — not a code issue.

**Review**: The cubic-dev-ai bot flagged a concurrent GET-start race on the initial commit. This was addressed in 3c0d47b by switching to `set[CancelScope]` so terminate() iterates and cancels all active scopes.

The fix is minimal (19 insertions, 5 deletions), backward-compatible, and targets only the FastMCP GET SSE path. The raw lowlevel.Server + StreamableHTTPSessionManager path is unaffected.

Thanks!

PR #1115, opened 2025-07-10
Update routing for streamable HTTP to avoid 307 redirect

yurikunash:
Setup the endpoint for the streamable HTTP (`/mcp`) in the routing table as a route served by an ASGI app, instead of mounting it as a base path. 

Resolves: #732, #573

As mentioned in MCP specification (version 2025-06-18):
> The server MUST provide a single HTTP endpoint path (hereafter referred to as the MCP endpoint) that supports both POST and GET methods. For example, this could be a URL like https://example.com/mcp 

This also aligns with JSON-RPC specification. 

## Motivation and Context

The change is required to avoid a 307 Temporary Redirect, from `/mcp` to `/mcp/` endpoint, which currently happens. The redirect causes issues for several popular clients and also impedes hosting MCP servers behind proxy servers.

## How Has This Been Tested?

Tested locally and on a remote server with MCP clients.

## Breaking Changes

The users who have MCP clients configured to consume the endpoint with a slash, e.g. `/mcp/`, will start experiencing a 307 Temporary redirect to the endpoint without the slash `/mcp`.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

Issue #732, opened 2025-05-16
307 Temporary Redirect


2495710777:
INFO:     127.0.0.1:56991 - "POST /mcp HTTP/1.1" 307 Temporary Redirect
INFO:     127.0.0.1:56991 - "POST /mcp/ HTTP/1.1" 200 OK

After using 
mcp.run(transport="streamable-http") 
to start the server, why does the client always redirect to '/mcp/' when requesting the server?

ch-xhr:
same issue.

zhouguangming:
same

nightzjp:
http://localhost/mcp => http://localhost/mcp/

chi2liu:
set your streamable_http_path='/mcp/'

chi2liu:
#781

jbovet:
hey, it took me a while to figure this out, I even raised an [issue](https://github.com/jlowin/fastmcp/issues/573)! 

thanks everyone for your help!

tobegit3hub:
It is the issue of fastapi. May close in this project.

pamelafox:
Just to share code, the following results in those errors:

```
streamablehttp_client("http://myserver.com/mcp")
```

whereas this resolves them:

```
streamablehttp_client("http://myserver.com/mcp/")
```

It appears that the streamable_http_client cannot handle the redirect, and when we use Starlette/FastAPI, /mcp gets turned into /mcp/, so we need to give the direct route of /mcp.

Perhaps streamable_http_client could add redirect handling? But not all clients will handle it (see VS Code issue).

Issue #283, opened 2025-03-14
Clients fail and crash If initial GET request returns a redirect (e.g. 302)


eastlondoner:
**Describe the bug**
If a remote (see transport) server responds to the initial GET request of the client with a redirect an error is thrown

**To Reproduce**
Connect a client to https://yo-mcp.com/mcp/7NIUz0cvfStY6-wSsVyY3
It will get a redirect. The typescript client handles it just fine.

**Expected behavior**
The get request should just follow the redirect.

**Screenshots**

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/b6da5abe-c6e8-4a3f-8847-d78d893f5179" />

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version mcp python sdk version 1.2.0

maxisbey:
Closing due to potential fixes being merged in from linked PR/issue. Please re-open this if it's still an issue.

PR #3075, opened 2026-07-08
Scope HTTP client redirect following to the request's origin

maxisbey:
Clients created by the SDK's httpx factory always enabled `follow_redirects=True`, so everything configured on a client — headers, auth, request bodies — was re-sent to whatever host a `Location` header named, and a misconfigured URL failed with an unhelpful `Unexpected content type:` error. This PR scopes redirect following to the request's origin and makes every refused redirect fail loudly with the target URL and the remedy.

## Motivation and Context

`follow_redirects=True` was added for operational convenience: trailing-slash 307s from Starlette-mounted servers (#105, #732) and hosted-server redirects (#283). The dominant case — the same-origin slash fix — keeps working exactly as before; the server-side cause was also fixed long ago in #1115. What the blanket default additionally did was silently re-send client configuration to arbitrary redirect targets and silently mask misconfigured URLs. httpx's own default is `follow_redirects=False` for the same reason (encode/httpx#2533 keeps its cross-origin strip list deliberately minimal and leaves per-application policy to the application).

Clients built by `create_mcp_http_client` now:

- follow redirects within the same origin (scheme, host, port) and same-host http→https upgrades on default ports — trailing-slash normalization and TLS-upgrade redirects keep working, verified end-to-end against real Starlette apps;
- raise the new `RedirectError` for any other redirect target, **before** the next request is sent, with a message naming the resolved target and how to proceed (connect to the final URL, or supply your own pre-configured `httpx.AsyncClient` after verifying the destination).

Transports resolve a refused redirect in-band — a request gets a JSON-RPC error rather than a hung caller, notifications are delivered to the session's message handler, and the standalone GET stream stops retrying an endpoint that redirects on every attempt. Caller-supplied clients that follow no redirects (httpx's default) get the same clear error on POST, GET-stream, and SSE-connect paths. OAuth discovery, registration, token, refresh, and the identity-assertion token exchange fail loudly on redirect responses instead of silently trying the next URL or abandoning the discovery chain.

`create_mcp_http_client` and `RedirectError` are now exported from the top-level `mcp` package (a deliberate API addition): the docs previously told users to build their own `httpx.AsyncClient` for headers/auth, which quietly lost the SDK's defaults. Docs and examples now configure clients through the factory; the general-purpose `simple-tool` fetch example gets a browser-like client of its own, since it fetches arbitrary web pages rather than a configured MCP endpoint.

## How Has This Been Tested?

- Real-server verification: booted `MCPServer.streamable_http_app()` under uvicorn and drove it with the real client. A trailing-slash `/mcp/` URL initializes and completes a full session lifecycle (every request rides a genuine Starlette 307); a redirect to a different origin resolves initialization with a clear error naming the target; a typo'd path still yields the usual not-found error.
- New unit and transport tests (all proven failing on the base revision): same-origin and https-upgrade redirects followed with method/body/headers intact; other-host, other-port, https→http, and empty-host absolute-form Locations refused before a second request is sent; unparsable Location defers to httpx's own error; per-request and client-level headers behave identically; request/notification/resumption/GET-stream/SSE paths each surface the in-band error; OAuth handlers raise on redirect responses; a caller-supplied `follow_redirects=True` client is untouched (opt-in preserved).
- A parametrized test asserts the target-resolution helper matches `response.next_request.url` from httpx itself across relative, protocol-relative, absolute-form-without-host, and cross-origin shapes, so an httpx behavior change surfaces as a test failure rather than a silent policy divergence.
- Full suite: 5458 passed, 100% coverage including `strict-no-cover`, pyright and ruff clean.

## Breaking Changes

Yes, deliberate and documented in `docs/migration.md`: a redirect to a different origin now raises `mcp.RedirectError` instead of being followed. Same-origin redirects and same-host https upgrades are unaffected. Deployments that genuinely need cross-origin redirect following can pass their own `httpx.AsyncClient(follow_redirects=True)` to `streamable_http_client` / `sse_client`.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- Four `# pragma: no branch` / `# pragma: no cover` comments are added in tests: three for the coverage.py nested-`async with` phantom-arc misreport documented in AGENTS.md, one marking an unreachable context-manager body in a connect-failure test.
- Known, accepted scope edges: resumption/reconnection GETs with a caller-supplied non-following client surface httpx's generic status error rather than the typed one (the factory's clients get the typed error everywhere); the origin comparison is deliberately conservative and fails loud on exotic-but-equivalent forms (IPv6 long-form vs. shorthand, trailing-dot hosts, non-default-port https upgrades) — supply your own client for those chains.
- The `RedirectPolicy` proposal in #2180 addressed the scheme-downgrade slice of the same default; with redirect targets now vetted at the factory for every consumer, that policy knob is superseded by this change.

### DIFF
--- a/src/mcp/__init__.py
+++ b/src/mcp/__init__.py
@@ -65,6 +65,7 @@ from .client.session_group import ClientSessionGroup
 from .client.stdio import StdioServerParameters, stdio_client
 from .server.session import ServerSession
 from .server.stdio import stdio_server
+from .shared._httpx_utils import RedirectError, create_mcp_http_client
 from .shared.exceptions import MCPDeprecationWarning, MCPError, UrlElicitationRequiredError
 from .shared.uri_template import InvalidUriTemplate, UriTemplate
 
@@ -108,6 +109,7 @@ __all__ = [
     "PromptsCapability",
     "ReadResourceRequest",
     "ReadResourceResult",
+    "RedirectError",
     "Resource",
     "ResourcesCapability",
     "ResourceUpdatedNotification",
@@ -137,6 +139,7 @@ __all__ = [
     "UriTemplate",
     "UrlElicitationRequiredError",
     "InvalidUriTemplate",
+    "create_mcp_http_client",
     "stdio_client",
     "stdio_server",
 ]
--- a/src/mcp/client/sse.py
+++ b/src/mcp/client/sse.py
@@ -1,6 +1,6 @@
 import logging
 from collections.abc import Callable
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, suppress
 from typing import Any
 from urllib.parse import parse_qs, urljoin, urlparse
 
@@ -12,7 +12,7 @@ from httpx_sse import SSEError, aconnect_sse
 
 from mcp.shared._compat import resync_tracer
 from mcp.shared._context_streams import create_context_streams
-from mcp.shared._httpx_utils import McpHttpClientFactory, create_mcp_http_client
+from mcp.shared._httpx_utils import McpHttpClientFactory, RedirectError, create_mcp_http_client, redirect_error
 from mcp.shared.message import SessionMessage
 
 logger = logging.getLogger(__name__)
@@ -48,6 +48,7 @@ async def sse_client(
         timeout: HTTP timeout for regular operations (in seconds).
         sse_read_timeout: Timeout for SSE read operations (in seconds).
         httpx_client_factory: Factory function for creating the HTTPX client.
+            The default follows only same-origin redirects; see create_mcp_http_client.
         auth: Optional HTTPX authentication handler.
         on_session_created: Optional callback invoked with the session ID when received.
     """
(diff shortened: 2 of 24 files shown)


src/mcp/client/sse.py
```diff
@@ -121,14 +126,35 @@ async def post_writer(endpoint_url: str):
...
+                                    json=session_message.message.model_dump(
+                                        by_alias=True,
+                                        mode="json",
+                                        exclude_unset=True,
+                                    ),
+                                )
+                                if response.has_redirect_location:
+                                    # Only reachable with a client that does not follow
+                                    # redirects; see redirect_error.
+                                    raise redirect_error(response)
```
maxisbey, line 141: why raise then immediately catch? that seems weird

src/mcp/__init__.py
```diff
@@ -137,6 +139,7 @@
     "UriTemplate",
     "UrlElicitationRequiredError",
     "InvalidUriTemplate",
+    "create_mcp_http_client",
```
maxisbey, line 142: no

maxisbey: shouldn't be public

PR #3076, opened 2026-07-08
fix: reject trailing newline in tool-name validation

### LINKED ISSUE #3084: Tool name validation accepts a trailing newline (regex uses `$` instead of `\Z`)
### Describe the bug

`TOOL_NAME_REGEX` in `src/mcp/shared/tool_name_validation.py` is end-anchored with `$`:

```python
TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}$")
```

In Python's default (non-`MULTILINE`) mode, `$` matches at end-of-string **or immediately before a single trailing `\n`**. So a tool name ending in exactly one newline passes validation, even though `\n` is not in the allowed character set.

The length guard uses `len()`, which counts the `\n`, so `"a" * 127 + "\n"` (length 128) slips past both the length check and the character check.

Embedded newlines and other trailing control characters are already rejected correctly. Only the single-trailing-newline case leaks.

### To Reproduce

```python
from mcp.shared.tool_name_validation import validate_tool_name, validate_and_warn_tool_name

print(validate_tool_name("evil_tool\n").is_valid)   # True  (expected: False)
print(validate_and_warn_tool_name("evil_tool\n"))   # True, and no warning is logged

print(validate_tool_name("a" * 127 + "\n").is_valid)  # True, length 128 passes too
```

### Expected behavior

`validate_tool_name("evil_tool\n").is_valid` should be `False`, since `\n` is not one of the allowed characters.

### Additional context

Anchoring the end with `\Z` (strict end-of-string) instead of `$` fixes it, and no valid name is affected: `re.match(r"^[A-Za-z0-9._-]{1,128}\Z", "abc")` still matches, while `"abc\n"` correctly fails.

I have an open PR with this fix and a regression test: #3076. Filing this issue to track it, per CONTRIBUTING ("Bug fixes for clear, reproducible issues are welcome—but still create an issue to track the fix").

otiscuilei:
Closes #3084

## Summary

`TOOL_NAME_REGEX` in `src/mcp/shared/tool_name_validation.py` is end-anchored with `$`:

```python
TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}$")
```

In Python's default (non-`MULTILINE`) mode, `$` matches at end-of-string **or immediately before a single trailing `\n`**. So a tool name ending in exactly one newline passes validation:

```python
>>> validate_tool_name("evil_tool\n").is_valid
True          # expected: False (newline is not an allowed SEP-986 character)
>>> validate_and_warn_tool_name("evil_tool\n")
True          # no warning logged
```

The length guard uses `len()` (which counts the `\n`), so `"a" * 127 + "\n"` (length 128) also slips past both the length and the character check. Embedded and non-`\n` trailing control chars are already rejected correctly — only the single-trailing-newline case leaks.

## Fix

Anchor the end with `\Z` (strict end-of-string) instead of `$`. No valid name is affected — `re.match(r"^[A-Za-z0-9._-]{1,128}\Z", "abc")` still matches; only `"abc\n"` now correctly fails.

## Tests

Adds `test_validate_tool_name_rejects_trailing_newline` (parametrized over a trailing-newline name and a 128-char name whose last char is `\n`). Verified it **fails on `main`** (`is_valid=True`) and **passes** with the fix; the file's existing 29 tests are unchanged (no valid-name fixture ends in a newline).

### DIFF
--- a/src/mcp/shared/tool_name_validation.py
+++ b/src/mcp/shared/tool_name_validation.py
@@ -17,8 +17,10 @@ from dataclasses import dataclass, field
 
 logger = logging.getLogger(__name__)
 
-# Regular expression for valid tool names according to SEP-986 specification
-TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}$")
+# Regular expression for valid tool names according to SEP-986 specification.
+# End-anchored with \Z rather than $: in Python's default mode $ also matches
+# just before a single trailing newline, which would let "name\n" validate.
+TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}\Z")
 
 # SEP reference URL for warning messages
 SEP_986_URL = "https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names"
--- a/src/mcp/shared/tool_name_validation.py
+++ b/src/mcp/shared/tool_name_validation.py
@@ -17,10 +17,8 @@ from dataclasses import dataclass, field
 
 logger = logging.getLogger(__name__)
 
-# Regular expression for valid tool names according to SEP-986 specification.
-# End-anchored with \Z rather than $: in Python's default mode $ also matches
-# just before a single trailing newline, which would let "name\n" validate.
-TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}\Z")
+# Regular expression for valid tool names according to SEP-986 specification
+TOOL_NAME_REGEX = re.compile(r"^[A-Za-z0-9._-]{1,128}$")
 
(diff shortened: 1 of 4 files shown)


maxisbey (approved):

maxisbey:
Thanks @Otis0408 — nice catch, and thanks for filing #3084 to track it per CONTRIBUTING.md. I pushed two small changes on top: switched the fix from `\Z` to `re.fullmatch()` (same behavior — it puts the whole-string requirement at the call site and matches how the rest of the codebase validates), and applied the same fix to the URI-template varname check in `uri_template.py`, which had the same `$`-with-`.match` pattern. Merging once CI is green.

PR #3083, opened 2026-07-09
feat(client): add list_all_* helpers that drain pagination

### LINKED ISSUE #2556: Add `list_all_*` helpers to drain pagination

sanjibani:
Closes #2556.

The single-page list_tools / list_resources / list_resource_templates /
list_prompts methods on ClientSession return the server's first page plus
a next_cursor; callers that don't loop on the cursor silently see only
page zero. Four new list_all_* helpers drain the cursor until the server
stops sending one and return the flattened list:

  list_all_tools()              -> list[Tool]
  list_all_resources()          -> list[Resource]
  list_all_resource_templates() -> list[ResourceTemplate]
  list_all_prompts()            -> list[Prompt]

Each helper reuses the existing public list_* method, so per-page
post-processing keeps running across the drain (notably the 2026
x-mcp-header filter inside _absorb_tool_listing). Single-page and
empty-list cases work without looping.

Tests cover multi-page drain, single-page short-circuit, empty
result, helpers-don't-share-cursor-state, terminal-empty-page-stops
the loop, and cursor-roundtrip (server's next_cursor comes back
unchanged as params.cursor on the next request). 9 new tests in
tests/client/test_client.py; full client test suite (121 tests) passes,
ruff lint + format clean.

Addresses the design sketch in @agaonker's earlier comment on #2556
(same return type, helper naming, cursor-drain loop, malformed-cursor
safety via None-termination).

### DIFF
--- a/src/mcp/client/session.py
+++ b/src/mcp/client/session.py
@@ -1195,9 +1195,13 @@ class ClientSession:
     async def list_all_tools(self) -> list[types.Tool]:
         """Drain pagination across `tools/list` and return the full tool list.
 
-        Each page flows through the 2026 x-mcp-header per-page filter via
-        `list_tools` -> `_absorb_tool_listing`, so the returned list is exactly
-        what the cached state would expose.
+        Each intermediate page flows through `list_tools` -> `_absorb_tool_listing`
+        with `complete=False`, which populates per-tool cache state (`_x_mcp_header_maps`,
+        `_tool_output_schemas`) as a pure addition. After the loop ends, the helper
+        runs `_absorb_tool_listing` once more with `complete=True` on the merged
+        result so state for tools that disappeared across pages (or were filtered
+        by the per-page x-mcp-header MUST) is pruned to match the drained universe.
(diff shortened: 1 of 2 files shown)


shivamsingh-007:
Welcome @sanjibani — this is a nice addition! The list_all_* helpers fill a real gap, and the cursor-drain design is clean.

If you haven't already, take a look at the [CONTRIBUTING.md](https://github.com/modelcontextprotocol/python-sdk/blob/main/CONTRIBUTING.md) — it covers things like the PR checklist, AI disclosure policy, and testing conventions that help move reviews along smoothly. Happy to help if anything's unclear.

<sub>AI assistance was used to draft this comment; I reviewed it and take responsibility.</sub>

maxisbey:
Thanks for the PR. Tracking this in #2658 instead. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #3090, opened 2026-07-13
fix(stdio): serve bufferless std streams as text instead of crashing

steps-re:
## What this is now

This PR originally fixed #1933 (stdio_server closing the real process std handles). **#3117 has since landed and fixes that properly**, with a much better design than what I had here. So I have dropped my original commits entirely and rebased onto current `main`.

What is left is one narrower bug that #3117 does not cover.

## The bug

`_claim_fd` starts with:

```python
if not _is_backed_by_fd(stream, fd):
    return stream.buffer, None
```

and `_is_backed_by_fd` returns `False` in three cases, one of which is the stream having no `.buffer` at all:

```python
except (AttributeError, OSError, ValueError):
    return False
```

So on that path the fallback dereferences the attribute it just proved may be missing. If `sys.stdin` / `sys.stdout` have been replaced with `io.StringIO`, `stdio_server()` raises before serving a single message:

```
AttributeError: '_io.StringIO' object has no attribute 'buffer'
```

Reproduced against `a4f4ccd0`:

```python
sys.stdin, sys.stdout = io.StringIO(""), io.StringIO()
async with stdio_server() as (r, w):   # AttributeError
    ...
```

It is not a hypothetical stream shape. Test harnesses do it, and so do embedded hosts that swap the std streams to capture output.

## The fix

Return `None` for the buffer in that one case and serve the text stream in place. A bufferless stream is already text, so there is no binary layer to re-encode and none for `_UnownedTextWrapper` to protect from close. Every other path still returns a binary stream and is untouched.

## Verification

Local, on `a4f4ccd0` with the commit applied:

- repro above raises `AttributeError` before, exits cleanly after
- new regression test `test_stdio_server_serves_bufferless_std_streams_in_place` round-trips a request and a response through `io.StringIO` streams
- `tests/server/test_stdio.py` 18 passed
- full suite 5582 passed, 9 skipped, 1 xfailed
- `ruff check` and `ruff format --check` clean, `pyright` 0 errors
- branch coverage on `src/mcp/server/stdio.py` back to 100.00%, which the `fail_under = 100` gate needs

## Disclosure

Written with AI assistance, reviewed and verified by me. Happy to reshape or split this if you would rather have it another way.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -103,19 +103,34 @@ def _restore_fd(fd: int, private_fd: int) -> bool:
     return True
 
 
+def _text_transport(stream: TextIO, buffer: BinaryIO | None, errors: str | None = None) -> anyio.AsyncFile[str]:
+    """Serve the wire as UTF-8 text.
+
+    A stream with no buffer at all (io.StringIO under a test harness, or an
+    embedded host that replaced sys.stdout) is already text and owns no binary
+    layer to re-encode or to protect from close, so it is served in place.
+    """
+    if buffer is None:
+        return anyio.wrap_file(stream)
+    return anyio.wrap_file(_UnownedTextWrapper(buffer, encoding="utf-8", errors=errors))
+
+
 def _claim_fd(
     fd: int, stream: TextIO, mode: Literal["rb", "wb"], open_diversion: Callable[[], int]
-) -> tuple[BinaryIO, Callable[[], None] | None]:
+) -> tuple[BinaryIO | None, Callable[[], None] | None]:
     """Claim a standard stream: divert fd and serve the wire from a private duplicate.
 
     Best-effort: when descriptors cannot be duplicated or diverted, serves the
     sys stream's buffer in place, exactly as v1 did, with the claim held.
 
+    Returns a None buffer when the stream exposes no binary layer, which means the
+    caller must serve it as text; every other path returns a binary stream.
+
     Raises:
         RuntimeError: fd is already claimed by another transport in this process.
     """
     if not _is_backed_by_fd(stream, fd):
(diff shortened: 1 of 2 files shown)


steps-re:
@cubic-dev-ai good catch, and already addressed. The `StringIO`/bufferless-text-stream case is fixed in 9a57007: in the no-`fileno()` fallback, if the stream has no `.buffer` we now return `anyio.wrap_file(std)` around the text stream directly (and don't register it for teardown, since we didn't create it), keeping the `.buffer` re-wrap + `detach()` path only where a buffer exists.

Covered by `test_stdio_server_bufferless_text_streams` (monkeypatches `sys.stdin`/`sys.stdout` to `io.StringIO`), added in b04f535. Full stdio suite passes locally: 52 passed, 5 skipped.

steps-re:
rebased, and rescoped while doing it.

#3117 landed the real fix for #1933 and did it better than my original commits, so i dropped them rather than carry a competing implementation. this is now a single commit on top of `a4f4ccd0` for the one thing #3117 does not cover: `_is_backed_by_fd` returns False when a stream has no `.buffer`, and the fallback right below it then does `return stream.buffer, None`. a StringIO stdin/stdout raises AttributeError before serving anything.

fails-before/passes-after repro is in the description. full suite 5582 passed, ruff and pyright clean, and stdio.py branch coverage is back at 100.00% for the fail_under gate.

title and description updated to match the new scope.

steps-re:
only thing red was the coverage gate. 5575 tests pass, total lands at 99.99 against `fail-under=100`, on one partial branch (`120->121, 120->123`) on the `write_stream` context manager. same exit shape the `read_stream` ones already carry a pragma for, so it now has one too. should go green on the next run.

PR #3093, opened 2026-07-14
test: add tests for non-2xx HTTP status handling (fixes #3091)

### LINKED ISSUE #3091: streamable_http: HTTP status errors (401/403/5xx) on tools/list, tools/call never reach the caller — indistinguishable from a timeout
## Description

`StreamableHTTPTransport._handle_post_request` never notifies the caller of a specific pending request when the server returns a non-2xx status (e.g. 401/403). Instead of surfacing as an error to that request, the caller only ever observes a read timeout — indistinguishable from a genuinely slow/dead server.

## Root cause (two-level gap)

**1. Missing error path in `_handle_post_request`** (`src/mcp/client/streamable_http.py`):

```python
async with ctx.client.stream("POST", self.url, ...) as response:
    if response.status_code == 202:
        ...
        return
    if response.status_code == 404:
        ...
        return
    response.raise_for_status()   # <-- no try/except around this
    ...
```

Compare to its sibling `_handle_json_response` a few lines below, which *does* wrap its work in `try`/`except` and forwards the failure:

```python
async def _handle_json_response(self, response, read_stream_writer, is_initialization=False):
    try:
        message = JSONRPCMessage.model_validate_json(content)
        ...
    except Exception as exc:
        logger.exception("Error parsing JSON response")
        await read_stream_writer.send(exc)
```

`_handle_post_request`'s `raise_for_status()` failure has no equivalent — it propagates out of the fire-and-forget task spawned via `tg.start_soon(handle_request_async)` in `post_writer`, uncaught, with no attempt to notify the specific pending request.

**2. Even the "obvious" fix doesn't work, because of how `read_stream` is consumed.** I patched in a `try`/`except httpx.HTTPStatusError` around `raise_for_status()` that forwards the exception via `await ctx.read_stream_writer.send(exc)`, mirroring `_handle_json_response`'s pattern exactly. It made no observable difference to callers.

Looking at `BaseSession._receive_loop` (`src/mcp/shared/session.py`):

```python
async for message in self._read_stream:
    if isinstance(message, Exception):
        await self._handle_incoming(message)
    elif isinstance(message.message.root, JSONRPCRequest):
        ...
    elif isinstance(message.message.root, JSONRPCNotification):
        ...
    else:  # Response or error
        await self._handle_response(message)
```

A bare `Exception` on the read stream is routed to `_handle_incoming` — a generic side-channel for unsolicited connection-level issues. Only the `else` branch (`_handle_response`) resolves the specific pending future in `self._response_streams[request_id]` that the original caller is `await`-ing on. Since a bare exception carries no request id, it can never reach `_handle_response`, so the caller's pending request is never resolved by this path either — it can only ever time out.

A real fix needs `_handle_post_request` to construct a proper `JSONRPCError` (carrying the *original request's* id) on an HTTP-status failure and route it through the normal response channel, the same way request-validation failures are already handled inside `BaseSession._receive_loop`:

```python
    jsonrpc="2.0",
    id=message.message.root.id,
    error=ErrorData(code=..., message=..., data=...),
)
session_message = SessionMessage(message=JSONRPCMessage(error_response))
```

## Why it matters

Any client that wants to distinguish "the server rejected this call" (e.g. expired/invalid credentials — should fail fast, non-retryable) from "the server is slow/unreachable" (should retry) cannot do so over streamable HTTP: both degrade identically as a timeout. We hit this building a Temporal-based SDK that reclassifies MCP failures into a retryable/non-retryable degradation contract — a 401 currently retries `maximum_attempts` times and only then degrades as a *transient* failure, burning retries on what should be an immediate non-retryable auth failure.

## Repro

Point an `MCPServerStreamableHTTP`/`MCPToolset` client at a server that returns HTTP 401 for `tools/list` or `tools/call`. Observe:
- No `httpx.HTTPStatusError` (or any status-carrying exception) ever reaches the caller.
- The caller's `await` only resolves once its own read timeout elapses.
- The eventual exception (a bare timeout) is indistinguishable from a real 504/slow-server case.

## Environment

- `mcp` 1.25.0 / 1.28.1 (installed via `fastmcp`, which pins `mcp<2.0`)
- Confirmed via source inspection of `mcp/client/streamable_http.py` and `mcp/shared/session.py`, and via two targeted patch experiments (see root cause above) — not yet checked against the `mcp` 2.0 beta line.

Happy to help test a fix or discuss further.

zsxh1990:
## Summary

Add tests verifying that `StreamableHTTPTransport._handle_post_request` properly handles non-2xx HTTP status codes.

### Tests Added
- **401 Unauthorized** → JSONRPCError with request ID (not timeout)
- **403 Forbidden** → JSONRPCError with request ID
- **500 Internal Server Error** → JSONRPCError with request ID
- **JSON error body** → Parsed and used directly

### Verification
The current implementation already handles non-2xx correctly (lines 342-373 in `streamable_http.py`). These tests document the expected behavior and prevent regressions.

### Key Behavior
- Non-2xx status → `JSONRPCError` sent to caller with original request ID
- JSON-RPC error body → Parsed and forwarded directly
- Non-JSON body → Fallback to generic `INTERNAL_ERROR`

Closes #3091

/claim #3091

### DIFF
--- a/tests/client/test_non2xx_status.py
+++ b/tests/client/test_non2xx_status.py
@@ -7,13 +7,13 @@ Closes #3091
 """
 
 import json
-import pytest
-import httpx
-from unittest.mock import AsyncMock, MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock
 
-from mcp.client.streamable_http import StreamableHTTPTransport, RequestContext
+import pytest
+from mcp_types import JSONRPCError, JSONRPCRequest
+
+from mcp.client.streamable_http import RequestContext, StreamableHTTPTransport
 from mcp.shared.message import SessionMessage
-from mcp_types import JSONRPCRequest, JSONRPCError, ErrorData
 
 
 class TestNon2xxStatusHandling:
--- a/tests/client/test_non2xx_status.py
+++ b/tests/client/test_non2xx_status.py
@@ -1,0 +1,146 @@
+"""Tests for non-2xx HTTP status handling in StreamableHTTPTransport.
+
+Verifies that when the server returns 401/403/5xx, the caller receives
+a proper JSONRPCError (not a timeout).
+
+Closes #3091
+"""
+import json
+import pytest
+import httpx
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from mcp.client.streamable_http import StreamableHTTPTransport, RequestContext
+from mcp_types import JSONRPCRequest, JSONRPCError, ErrorData, SessionMessage
+
+
+class TestNon2xxStatusHandling:
+    """Test that non-2xx status codes produce proper error responses."""
+
+    def _make_request_context(self, request_id: str = "test-123") -> MagicMock:
+        """Create a mock RequestContext."""
+        ctx = MagicMock(spec=RequestContext)
+        ctx.session_message = MagicMock()
+        ctx.session_message.message = MagicMock(spec=JSONRPCRequest)
+        ctx.session_message.message.id = request_id
+        ctx.read_stream_writer = AsyncMock()
+        ctx.client = AsyncMock()
+        ctx.metadata = None
+        return ctx
+
+    @pytest.mark.anyio
+    async def test_401_produces_error_response(self):
+        """401 Unauthorized should produce a JSONRPCError with the request's ID."""
+        # This test verifies the fix for #3091
+        # When server returns 401, caller should get a proper error, not a timeout
+        transport = StreamableHTTPTransport.__new__(StreamableHTTPTransport)
+        transport.url = "http://test"
+        transport.session_id = None
+
+        ctx = self._make_request_context()
+
+        # Mock the response to return 401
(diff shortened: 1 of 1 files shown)


zsxh1990:
Hi team — friendly ping on this PR 🙂

CI is fully green (28/28 checks pass) and the Cubic AI reviewer has approved. This adds test coverage for non-2xx HTTP status handling (fixes #3091).

Happy to adjust if needed. Thanks!

zsxh1990:
Hi team — another friendly ping 🙂

This PR has been open for ~6 weeks now. CI is fully green (28/28 checks) and adds test coverage for non-2xx HTTP status handling (fixes #3091).

Is there anything I can do to help move this forward? Happy to rebase or address any feedback.

Thanks!

zsxh1990:
Closing — PR has been open for 6 weeks with 2 pings and no review feedback. Happy to reopen if there's interest.

PR #3096, opened 2026-07-15
fix: propagate McpError as JSON-RPC error response in lowlevel server

### LINKED ISSUE #2770: McpError raised in tool handlers is swallowed as isError:true instead of propagating as JSON-RPC error

nankingjing:
Fixes #2770

## Motivation and Context

When a tool handler registered via the lowlevel server's `@server.call_tool()` decorator raises `McpError`, the handler catches it via the generic `except Exception` block and wraps it as `CallToolResult(isError=True)`. This drops the structured error code on the wire, so clients cannot distinguish error types without parsing the message string.

The `UrlElicitationRequiredError` subclass already has a re-raise bypass for exactly this reason, but the parent `McpError` class does not. (The v2 high-level API on `main` already handles this correctly.)

## How Has This Been Tested?

- Added a regression test asserting that `McpError` raised in a lowlevel tool handler reaches the client as a JSON-RPC error with the original code, message, and data preserved (it fails without this change), plus a companion test asserting non-`McpError` exceptions still return `CallToolResult(isError=True)`.
- Ran the lowlevel server and elicitation test suites locally: all passing.

## Breaking Changes

Tool handlers that deliberately raise `McpError` now produce a JSON-RPC error response (the documented behavior for the error, matching `UrlElicitationRequiredError` and the v2 API) instead of an `isError: true` result. Handlers that raise other exceptions are unaffected.

## Types of changes

## Checklist

## Additional context

This scopes the fix to the lowlevel server per the issue's proposed fix. #2772 covers related ground but also changes FastMCP `Tool.run()` behavior; this PR is the minimal lowlevel-only change with regression tests at the lowlevel layer, in case the maintainers prefer to take that separately.

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -90,7 +90,7 @@ from mcp.server.lowlevel.helper_types import ReadResourceContents
 from mcp.server.models import InitializationOptions
 from mcp.server.session import ServerSession
 from mcp.shared.context import RequestContext
-from mcp.shared.exceptions import McpError, UrlElicitationRequiredError
+from mcp.shared.exceptions import McpError
 from mcp.shared.message import ServerMessageMetadata, SessionMessage
 from mcp.shared.session import RequestResponder
 from mcp.shared.tool_name_validation import validate_and_warn_tool_name
@@ -582,9 +582,11 @@ class Server(Generic[LifespanResultT, RequestT]):
                             isError=False,
                         )
                     )
-                except UrlElicitationRequiredError:
-                    # Re-raise UrlElicitationRequiredError so it can be properly handled
-                    # by _handle_request, which converts it to an error response with code -32042
+                except McpError:
+                    # Re-raise McpError (including subclasses such as
+                    # UrlElicitationRequiredError) so it can be properly handled by
+                    # _handle_request, which converts it to a JSON-RPC error response
+                    # preserving the structured error code.
                     raise
                 except Exception as e:
                     return self._make_error_result(str(e))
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR — this duplicates #2772, which was opened first, so I'm closing this one in favour of it. Feel free to reopen if this is still relevant.

PR #3097, opened 2026-07-15
fix: catch BrokenPipeError in stdio stdout_writer

Lexus2016:
## Problem

When the parent process (MCP client) closes the stdout pipe — during restart, kill, or graceful shutdown — the `stdout_writer()` coroutine in `stdio_server()` raises `BrokenPipeError` at `await stdout.flush()` (line 81).

The existing handler only catches `anyio.ClosedResourceError`, which does **not** cover `BrokenPipeError`. The unhandled exception propagates through anyio's `TaskGroup` as an `ExceptionGroup`, crashing the entire MCP server process.

## Reproduction

1. Start any MCP server using stdio transport (e.g. `turbo-memory-mcp serve`)
2. Connect from a client, make a few tool calls
3. Kill the client process (`kill -9` or close stdin)
4. Server crashes with:
```
  -> BrokenPipeError: [Errno 32] Broken pipe
```

## Impact

This is a common production scenario — **every MCP client restart** (config reload, session reset, crash recovery) kills the server. Downstream servers like `turbo-memory-mcp` leave stale daemon locks and require manual cleanup (`pkill`, `rm lockfile`, reconnect).

## Fix

Extend the `except` clause in `stdout_writer()` to also catch `BrokenPipeError` and `ConnectionResetError`, treating them as graceful shutdown signals:

```python
# Before:

# After:
except (anyio.ClosedResourceError, BrokenPipeError, ConnectionResetError):
```

## Testing

Tested in production with Hermes Agent (Nous Research) as the MCP client and `turbo-memory-mcp` as the server. The fix has been running reliably — server now exits cleanly on client restart instead of crashing with `ExceptionGroup`.

### DIFF
--- a/src/mcp/server/stdio.py
+++ b/src/mcp/server/stdio.py
@@ -68,7 +68,13 @@ async def stdio_server(stdin: anyio.AsyncFile[str] | None = None, stdout: anyio.
                     json = session_message.message.model_dump_json(by_alias=True, exclude_unset=True)
                     await stdout.write(json + "\n")
                     await stdout.flush()
-        except anyio.ClosedResourceError:  # pragma: no cover
+        except (anyio.ClosedResourceError, BrokenPipeError, ConnectionResetError):  # pragma: no cover
+            # BrokenPipeError / ConnectionResetError occur when the parent
+            # process (MCP client) closes the stdout pipe — e.g. during
+            # restart, kill, or graceful shutdown. This is a graceful
+            # shutdown signal, not a crash — suppress it so the server
+            # exits cleanly instead of propagating an ExceptionGroup from
+            # anyio's TaskGroup.
             await anyio.lowlevel.checkpoint()
 
     async with anyio.create_task_group() as tg:


keeltrace:
AI-assisted analysis; posting under human responsibility.

I think suppressing the broken stdout exception without terminating the rest of the transport can turn the crash into a hang / half-dead stdio session.

`stdio_server()` runs `stdin_reader()` and `stdout_writer()` as sibling tasks in the same AnyIO task group. This PR changes a broken response pipe from an exception into a normal return from `stdout_writer()`:

```python
except (anyio.ClosedResourceError, BrokenPipeError, ConnectionResetError):
```

That distinction matters for AnyIO task-group lifecycle. If a child raises, sibling tasks are cancelled; if it exits normally, the task group does not cancel siblings and waits for them to finish.

So consider a half-close where the parent has closed its read end of the child's stdout pipe, but the child's stdin is still open:

1. `stdout.flush()` raises `BrokenPipeError`.
2. This PR suppresses it, so `stdout_writer()` exits normally and closes `write_stream_reader`.
3. `stdin_reader()` is still blocked waiting for input, with no cancellation signal.
4. The stdio transport can remain alive indefinitely despite having permanently lost its response channel.
5. If another request does arrive on stdin, its handler now writes through `write_stream` whose receive side was closed by `stdout_writer()`, so the failure just moves to a later `BrokenResourceError` path.

That means the comment's stated behavior — treating BrokenPipe/ConnectionReset as a "graceful shutdown signal" so the server "exits cleanly" — is only reliable when stdin happens to close as well.

Could this error path terminate the whole stdio transport instead of only the writer? For example, signal/cancel the enclosing task group (or otherwise close the read side) after swallowing the OS-level pipe exception, so a dead stdout is a bounded transport shutdown and `stdin_reader()` cannot outlive the response channel.

A regression test could keep stdin deliberately open, force `stdout.flush()` to raise `BrokenPipeError`, and assert that the `stdio_server()` context exits within a short bound without requiring stdin EOF. A second variant could verify that no later request is accepted after stdout is known dead.

Checked against PR head `c0f3c0e` and current `main`'s stdio task-group structure. This is a static lifecycle review; I did not run the repository suite locally.

PR #3119, opened 2026-07-17
fix(server): forward tool args under the real parameter name, not the…

jayzuccarelli:
Forward tool arguments to the underlying function under the real parameter name instead of the schema alias.

## Motivation and Context

A tool parameter with an explicit alias, e.g. `city: Annotated[str, Field(alias="location")]`, is advertised in the tool input schema as `location`, and clients send `location`. At call time `model_dump_one_level` forwarded the value under the alias, so the function was invoked as `fn(location=...)` and raised `TypeError: got an unexpected keyword argument 'location'`. Both sync and async tools are affected.

The root cause is that `field_info.alias` is used as the kwarg name, but the alias is a wire name, not necessarily a valid Python parameter. It happens to be correct for the internal shadow-rename case (a param named after a `BaseModel` attribute, e.g. `schema` -> field `field_schema` with `alias="schema"`), where the alias *is* the real param name — so the two cases are indistinguishable from `FieldInfo` alone.

The fix records each field's real Python parameter name at metadata-build time and forwards arguments under it. The shadow-rename case keeps working because its real name is recorded the same way. Resolver arg-name matching in `tools/base.py` reads the same map so by-name resolvers keep resolving.

## How Has This Been Tested?

Added a regression test (`test_call_with_aliased_parameter`) that fails before the change with the `TypeError` above and passes after. Repro:

```python
from typing import Annotated
from pydantic import Field
from mcp.server.mcpserver.utilities.func_metadata import func_metadata

def tool(city: Annotated[str, Field(alias="location")]) -> str:
    return city

meta = func_metadata(tool)
# input schema advertises "location"; a client sends {"location": "Paris"}
# -> TypeError: got an unexpected keyword argument 'location'
```

Full `tests/server/mcpserver/` suite passes locally (552 tests), along with ruff, pyright, and strict-no-cover.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/server/mcpserver/tools/base.py
+++ b/src/mcp/server/mcpserver/tools/base.py
@@ -99,9 +99,9 @@ class Tool(BaseModel):
         )
         parameters = func_arg_metadata.arg_model.model_json_schema(by_alias=True)
 
-        # Match `model_dump_one_level`'s kwarg keys (alias when present, else field name)
-        # so a by-name resolver param resolves to a key that exists at call time.
-        tool_arg_names = {field.alias or name for name, field in func_arg_metadata.arg_model.model_fields.items()}
+        # Match `model_dump_one_level`'s kwarg keys (the real parameter names) so a
+        # by-name resolver param resolves to a key that exists at call time.
+        tool_arg_names = set(func_arg_metadata.arg_model.param_names.values())
         resolver_plans = build_resolver_plans(resolved_params, tool_arg_names)
 
         return cls(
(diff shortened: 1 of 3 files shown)


jayzuccarelli:
Gentle ping — ready for review, happy to iterate on any feedback.

jayzuccarelli:
Re-checked against main today: `model_dump_one_level` still builds kwargs from `field_info.alias`, so a tool parameter declared with an explicit `Field(alias=...)` still raises `TypeError: got an unexpected keyword argument`. The regression test here fails on main and passes with the change.

The change is complete and ready for review. GitHub is refusing to let me flip it out of draft on my end, so if a maintainer can either mark it ready or review it as-is, that unblocks it. `maintainerCanModify` is on.

PR #3124, opened 2026-07-18
fix: prevent infinite warning-logging loop in _handle_message

### LINKED ISSUE #3122: Server._handle_message can loop forever re-logging warnings when a log handler itself emits a warning (v1.x)
**Affects:** mcp 1.x (verified 1.27.2 and 1.28.1); the block was removed in the
v2 rewrite (receive-path swap, PR #2710), so v2.0.0a2+ is not affected.

## Summary

`mcp/server/lowlevel/server.py::Server._handle_message` wraps message handling
in `warnings.catch_warnings(record=True)` and then, still **inside** the same
`catch_warnings` block, logs every recorded warning:

```python
with warnings.catch_warnings(record=True) as w:
    match message:
        ...
    for warning in w:  # pragma: no cover
        logger.info("Warning: %s: %s", warning.category.__name__, warning.message)
```

If any logging handler/formatter attached to that logger (or the root logger)
itself raises a warning during `emit()` — e.g. a JSON formatter that calls
`datetime.utcnow()`, which raises `DeprecationWarning` on Python 3.12+ — the
warning is appended to `w` **while `w` is being iterated**. Each `logger.info`
then produces one more list element, so the `for` loop never terminates. The
loop is fully synchronous (no await points), so task cancellation cannot
interrupt it; under pytest-xdist this wedges the worker until an external
timeout kills the process (`[gwN] node down: Not properly terminated`).

Two preconditions, both common in test environments:

1. warning filters that re-deliver duplicate warnings — pytest's warnings
   plugin runs every test under `simplefilter("always")`, so once-per-location
   deduplication never kicks in;
2. a handler/formatter on the emitting logger's chain that raises any warning
   per record (a `utcnow()`-based timestamp formatter is the canonical case on
   3.12+).

Note `catch_warnings` is also documented as thread/task-unsafe (CPython
gh-91505 / gh-128384; context-aware only on 3.14 behind
`-X context_aware_warnings`), so with `tg.start_soon(self._handle_message, ...)`
per message, interleaved enter/exit across tasks can additionally leak the
recording state — but the amplification above reproduces without any
concurrency.

## Minimal reproduction (Python 3.12+, mcp 1.28.1)

```python
import asyncio, datetime, logging, warnings

import mcp.types as mcp_types


class WarningFormatter(logging.Formatter):
    def format(self, record):
        ts = datetime.datetime.utcnow().isoformat()  # DeprecationWarning on 3.12+
        return f"{ts} {record.getMessage()}"


handler = logging.StreamHandler()
handler.setFormatter(WarningFormatter())
root = logging.getLogger()
root.addHandler(handler)
root.setLevel(logging.INFO)

warnings.simplefilter("always")  # pytest's per-test ambient state

server = Server("probe")


@server.list_tools()
async def _list_tools():
    return [mcp_types.Tool(name="t", description="", inputSchema={"type": "object"})]


async def main():
    async with create_connected_server_and_client_session(server) as session:
        return (await session.list_tools()).tools


asyncio.run(main())  # never returns; unbounded log output
```

Observed: the server's own `logger.info("Processing request of type
ListToolsRequest")` seeds the first recorded warning; the re-log loop then
self-amplifies (~50k log lines/sec in our measurements) and `list_tools()`
never completes.

## Suggested fixes (any one suffices)

1. Snapshot before iterating: `for warning in list(w): ...`  — bounds the loop
   (each pass logs the snapshot; new warnings belong to the next message).
2. Log the recorded warnings **after** exiting the `catch_warnings` block.
3. Drop the re-logging entirely (v2 already removed it).

## Context

Found while diagnosing a CI flake: pytest-xdist workers hung for the full
pytest-timeout budget (600 s) and died with `[gwN] node down`, randomly
distributed across Python 3.12/3.13/3.14 matrix legs (3.10/3.11 immune — no
`utcnow` DeprecationWarning there). Post-mortem stacks always sat at the
`logger.info` line of the warning re-log loop.

rahul188:
## Summary

Fixes #3122

`Server._handle_message` (`mcp/server/lowlevel/server.py`) logs the warnings recorded during message handling while still inside the `warnings.catch_warnings(record=True)` block. Because `record=True` installs an "always" filter, any warning emitted by a logging handler while those `logger.info` calls run is appended to the very list being iterated, so the `for warning in w` loop never terminates. The loop is synchronous with no await points, so task cancellation cannot interrupt it; the handling task wedges until an external timeout kills the process. This shows up in practice when a formatter on the logger chain warns per record (for example a `datetime.utcnow()`-based JSON timestamp formatter on Python 3.12+), and it is easy to trigger under pytest, which runs with an "always" warning filter.

## Fix

Snapshot the recorded warnings and emit the log lines after leaving the `catch_warnings` block. Once the block exits, the recording `showwarning` is restored, so a warning raised by a logging handler goes to the normal warnings machinery instead of being appended to the snapshot, and the loop iterates a fixed list. Behaviour for the normal case (recorded warnings are still logged once each) is unchanged.

## Tests

Added `tests/issues/test_3122_warning_loop.py`. It drives `_handle_message` with a request that records one warning and attaches a logging handler that itself warns on every record, then asserts the handler is invoked exactly once. The handler stops re-warning after a cap so that a regressed (looping) build terminates and fails the assertion rather than hanging the session. The test fails on `v1.x` before this change (handler invoked many times) and passes after.

This targets `v1.x`; `main` already dropped this block in the v2 receive-path rewrite, so it is not affected.

### DIFF
--- a/src/mcp/server/lowlevel/server.py
+++ b/src/mcp/server/lowlevel/server.py
@@ -719,8 +719,10 @@ class Server(Generic[LifespanResultT, RequestT]):
                     if raise_exceptions:
                         raise message
 
-            for warning in w:  # pragma: no cover
-                logger.info("Warning: %s: %s", warning.category.__name__, warning.message)
+            recorded_warnings = list(w)
+
+        for warning in recorded_warnings:
+            logger.info("Warning: %s: %s", warning.category.__name__, warning.message)
 
     async def _handle_request(
         self,
--- a/tests/issues/test_3122_warning_loop.py
+++ b/tests/issues/test_3122_warning_loop.py
@@ -1,0 +1,78 @@
+"""Regression test for #3122.
+
+``Server._handle_message`` used to log recorded warnings while still inside the
+``warnings.catch_warnings(record=True)`` block. Because ``record=True`` forces
+an "always" filter, a warning emitted by a logging handler during that logging
+step was appended to the very list being iterated, so the loop never
+terminated. The loop is synchronous, so task cancellation could not interrupt
+it.
+
+The fix snapshots the recorded warnings and logs them after leaving the
+``catch_warnings`` block, so handler-emitted warnings can no longer extend the
+iteration. This test drives ``_handle_message`` with a request that records one
+warning and a logging handler that itself warns on every record; the handler
+must be invoked exactly once. The handler stops re-warning after a cap so that
+the pre-fix infinite loop terminates and fails the assertion instead of hanging
+the test session.
+"""
+
+import logging
+import warnings
+from unittest.mock import AsyncMock, Mock
+
+import pytest
+
+import mcp.types as types
+from mcp.server.lowlevel.server import Server
+from mcp.server.session import ServerSession
+from mcp.shared.session import RequestResponder
+
+
+class _WarningEmittingHandler(logging.Handler):
+    """A logging handler whose emit() raises a warning, mimicking e.g. a
+    timestamp formatter that calls a deprecated API on every record. It stops
+    after ``cap`` records so a regressed (looping) build still terminates."""
+
+    def __init__(self, cap: int = 100) -> None:
+        super().__init__()
+        self.emit_count = 0
+        self._cap = cap
+
+    def emit(self, record: logging.LogRecord) -> None:
+        self.emit_count += 1
+        if self.emit_count <= self._cap:
+            warnings.warn("warning raised while logging", stacklevel=1)
+
+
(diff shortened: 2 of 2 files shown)


rahul188:
Hi 👋 Gentle nudge on this PR — it's **mergeable and CI is passing**, just waiting on a review. Would really appreciate a look whenever you have bandwidth, and I'm glad to make any changes you'd like. Thanks for your time and for maintaining this project! 🙏

PR #3125, opened 2026-07-18
fix(server): terminate streamable HTTP sessions on manager shutdown

### LINKED ISSUE #2150: Active Streamable HTTP sessions are not terminated during shutdown

baiyuxi930826:
## Summary
Fixes #2150.

Streamable HTTP sessions were not terminated during manager shutdown, and `terminate()` did not close active SSE writers. That left EventSourceResponse / ASGI callables incomplete on CTRL+C (uvicorn incomplete-response noise and resource leaks under multi-session hosts).

### Changes
1. **`StreamableHTTPSessionManager.run()`** — call `terminate()` on active transports **before** cancelling the task group.
2. **`StreamableHTTPServerTransport.terminate()`** — close all `_sse_stream_writers` via `close_sse_stream()` first so SSE responses can finish.
3. **Regression tests** for writer close + manager shutdown terminate.

### Relation to existing work
- Open PR #2253 targets **`v1.x`**, is older, and does not include the regression suite. This PR targets **`main`** with tests.
- Happy to close/adjust if maintainers prefer landing via #2253 or a different base.

## Test plan
- [x] `uv run pytest tests/server/test_streamable_http_manager.py::test_terminate_closes_active_sse_stream_writers -q`
- [x] `uv run pytest tests/server/test_streamable_http_manager.py::test_manager_shutdown_terminates_active_sessions -q`
- [x] `uv run pytest tests/server/test_streamable_http_manager.py -q` (32 passed locally)

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -774,10 +774,18 @@ class StreamableHTTPServerTransport:
         """Terminate the current session, closing all streams.
 
         Once terminated, all requests with this session ID will receive 404 Not Found.
+
+        Active SSE writers are closed first so EventSourceResponse / ASGI callables can
+        complete instead of hanging until the task group is cancelled (see #2150).
         """
 
         self._terminated = True
         logger.info(f"Terminating session: {self.mcp_session_id}")
+
+        # Close SSE stream writers first so long-lived GET/POST SSE responses finish.
+        # Copy keys: close_sse_stream mutates the dict (includes GET_STREAM_KEY).
+        for request_id in list(self._sse_stream_writers.keys()):
+            self.close_sse_stream(request_id)
 
         # We need a copy of the keys to avoid modification during iteration
         request_stream_keys = list(self._request_streams.keys())
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -700,6 +700,8 @@ class StreamableHTTPServerTransport:
 
         # Create SSE stream
         sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[SSEEvent](0)
+        # Register immediately so session terminate() can close this ASGI response.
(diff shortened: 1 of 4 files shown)


baiyuxi930826:
Addressed the cubic P2:

- **Standalone GET SSE**: register `_sse_stream_writers[GET_STREAM_KEY]` as soon as the writer is created; pop on writer exit / response finally so `terminate()` can close it.
- **Replay SSE**: register a provisional `_replay:...` writer key before `EventSourceResponse` starts (covers the window while `replay_events_after` runs), then re-key to `stream_id` when live-tail begins; always pop in finally.
- Tests: `test_terminate_closes_standalone_get_sse_writer_when_registered`, `test_terminate_closes_provisional_replay_sse_writer` (4 terminate/shutdown tests green locally).

Happy to iterate if maintainers want a different provisional-key scheme.

baiyuxi930826:
Pushed a follow-up for CI coverage (`fail-under=100`):

- Extended `test_terminate_closes_active_sse_stream_writers` so terminate also cleans an orphaned `_request_streams` entry (covers the remaining-keys loop / line ~800).
- Marked the defensive `registered_stream_id is not None` guard as `# pragma: no branch` (always set before the guarded body on that path).

Local: the four #2150 terminate tests pass. Please re-run CI on the latest commit.

maxisbey:
Thanks for the PR. Tracking this in #2253 instead. Closing as part of a general backlog cleanup ahead of the v2 release. If this is still relevant against current `main`, feel free to reopen.

PR #3127, opened 2026-07-18
fix: never mint a request id already used by a completed caller-supplied id

### LINKED ISSUE #3126: Dispatcher mints a request id already used by a completed caller-supplied request (spec: ids MUST NOT be reused in a session)
### Initial Checks

### Description

On current `main` (v2, tested at 3a6f2996), when a caller supplies its own request id via `CallOptions["request_id"]` (added in #3046), the dispatcher's minted-id sequence can later land on that same id after the supplied request completes. The session then sends two different requests with the same id, which the spec forbids:

> The request ID MUST NOT have been previously used by the requestor within the same session.

(https://modelcontextprotocol.io/specification/2025-11-25/basic#requests, same wording since 2025-03-26. The SDK pins this itself as `protocol:request-id:unique` in `tests/interaction/_requirements.py`: "ids are never reused within the session".)

## Root cause

In `JSONRPCDispatcher.send_raw_request` (`src/mcp/shared/jsonrpc_dispatcher.py`, around line 338):

```python
else:
    # Mint past any key a supplied id occupies: the collision error is
    # reserved for the caller who actually chose the id.
    request_id = self._allocate_id()
    while request_id in self._pending:
        request_id = self._allocate_id()
```

The comment says minting should skip past any supplied id, but the guard only checks `self._pending`, and pending entries are popped when a request completes (line ~428). So for a supplied integer id (or numeric string, since `coerce_request_id` folds `"2"` and `2` into one key): once that request finishes, the monotonic counter eventually reaches the same value and reuses it on the wire.

`DirectDispatcher._dispatch_request` (`src/mcp/shared/direct_dispatcher.py`, around line 253) has the same pattern with `self._in_flight_ids`, which is discarded in the `finally`.

## Why it matters

The receiving side is allowed to assume per-session id uniqueness. The SDK's own stateful Streamable HTTP server already mishandles duplicate ids (#3060, response cross-wiring), so an innocent client that supplies small integer ids for a few calls and then keeps using the same session can hit that server-side behavior through no fault of its own. The SDK-internal user of this feature (the `subscriptions/listen` driver's `"listen-N"` ids) is immune because non-numeric string ids never collide with the minted integer sequence, so this only bites users of the new public `CallOptions["request_id"]` with integer or numeric-string ids.

Note this is only about the dispatcher's own minting crossing a completed supplied id. The reverse direction (a caller re-supplying an id it already used before) is currently allowed on purpose and asserted by tests, so I left that alone.

## Proposed fix

When accepting a supplied id, advance the mint counter past its coerced key so minted ids can never revisit it:

```python
# jsonrpc_dispatcher.py, after computing pending_key for a supplied id:
if isinstance(pending_key, int):
    self._next_id = max(self._next_id, pending_key)
```

and the same in `direct_dispatcher.py` with `in_flight_key`. Happy to open a PR with regression tests for both dispatchers if this looks right.

### Example Code

```python
import anyio
from mcp_types import JSONRPCRequest, JSONRPCResponse

from mcp.shared.jsonrpc_dispatcher import JSONRPCDispatcher


async def main() -> None:
    c2s_send, c2s_recv = anyio.create_memory_object_stream[SessionMessage | Exception](4)
    s2c_send, s2c_recv = anyio.create_memory_object_stream[SessionMessage | Exception](4)
    client = JSONRPCDispatcher(s2c_recv, c2s_send)

    async def on_request(ctx, method, params):
        return {}

    async def on_notify(ctx, method, params):
        pass

    wire_ids = []

    async def respond() -> None:
        async for wire in c2s_recv:
            if isinstance(wire, SessionMessage) and isinstance(wire.message, JSONRPCRequest):
                wire_ids.append(wire.message.id)
                await s2c_send.send(
                    SessionMessage(JSONRPCResponse(jsonrpc="2.0", id=wire.message.id, result={}))
                )

        await tg.start(client.run, on_request, on_notify)
        tg.start_soon(respond)

        await client.send_raw_request("ping", None, {"request_id": 2})
        await client.send_raw_request("ping", None)
        await client.send_raw_request("ping", None)

        tg.cancel_scope.cancel()

    print("wire request ids in order:", wire_ids)
    # prints: wire request ids in order: [2, 1, 2]
    # the third request (dispatcher-minted) reuses the completed supplied id 2


anyio.run(main)
```

### Python & MCP Python SDK

```
python: 3.14.5
mcp: main @ 3a6f2996 (v2 development line)
```

---

AI disclosure, since it's policy here: I leaned on Claude Code to help navigate the codebase and pressure-test this analysis, but I ran and verified the repro myself and I understand the fix I'm proposing. Freshman in college trying to contribute something genuinely useful, so if I've misjudged the intent behind the mint loop I'd honestly love the correction :)

ayaangazali:
## Summary

Fixes #3126.

When a caller supplies its own request id via `CallOptions["request_id"]`, the dispatcher's minted-id sequence could later land on that same id once the supplied request completed, sending two different requests under one id in the same session. The spec says the requestor must never reuse a request id within a session, and the SDK pins the same rule as `protocol:request-id:unique` in `tests/interaction/_requirements.py`.

## Root cause

Both mint loops only skipped ids still in flight:

```python
request_id = self._allocate_id()
while request_id in self._pending:   # entries are popped on completion
    request_id = self._allocate_id()
```

The comment right above ("Mint past any key a supplied id occupies") shows the intent, but the guard stopped holding the moment the supplied request completed. Since `coerce_request_id` folds `"2"` and `2` into one key, numeric-string ids hit this too. `DirectDispatcher` had the same pattern with `_in_flight_ids` (discarded in a `finally`).

## Fix

Accepting a supplied id now advances the mint counter past its coerced integer key:

```python
if isinstance(pending_key, int):
    self._next_id = max(self._next_id, pending_key)
```

That makes the old skip loops provably unreachable (the counter is always past every supplied integer key and every previously minted id), so they are removed rather than left as dead branches the coverage gate would flag.

Two deliberate scope notes:

- A caller re-supplying its own completed id stays allowed; that behavior is asserted by `test_send_raw_request_with_in_flight_request_id_raises_and_frees_id_on_completion` and this PR does not touch it. Only the dispatcher's own minting is constrained.
- Minted id values shift when a caller supplies integer ids (e.g. the pinned sequence in `test_minted_ids_skip_a_caller_supplied_id_still_in_flight` went from `[1, 2, 4]` to `[4, 5, 6]`). Ids are opaque per spec, so I did not treat this as a breaking change needing a migration entry; happy to add one if you see it differently.

## Tests

- New `test_minted_ids_never_reuse_a_completed_supplied_id`, parametrized over both dispatchers and both id spellings (`2`, `"2"`). Fails on current main with wire ids `[2, 1, 2]`; passes with the fix (`[2, 3, 4]`).
- Updated the pinned sequence in `test_minted_ids_skip_a_caller_supplied_id_still_in_flight`.
- `CallOptions.request_id` docstring extended with the new guarantee.

Ran locally:

```
./scripts/test                          # 5448 passed, coverage 100.00%, strict-no-cover clean
uv run --frozen ruff format . && uv run --frozen ruff check .
uv run --frozen pyright                 # 0 errors
```

---

AI disclosure, per CONTRIBUTING: Claude Code helped me find this and sanity-check the invariant reasoning; I wrote up the repro, ran everything locally, and I can defend each line here in my own words. First PR to this repo and I'm a freshman still learning the ropes, so if the counter-advance approach clashes with something I haven't seen, please say so and I'll rework it.

### DIFF
--- a/src/mcp/shared/direct_dispatcher.py
+++ b/src/mcp/shared/direct_dispatcher.py
@@ -250,14 +250,18 @@ class DirectDispatcher:
                     in_flight_key = coerce_request_id(request_id)
                     if in_flight_key in self._in_flight_ids:
                         raise ValueError(f"request id {request_id!r} is already in flight")
+                    # Advance the mint counter past the supplied key: ids are
+                    # single-use within a session, so a minted id may not
+                    # revisit this key even after the request completes.
+                    if isinstance(in_flight_key, int):
+                        self._next_id = max(self._next_id, in_flight_key)
                 else:
                     # Synthesize an id (the DispatchContext contract reserves None
-                    # for notifications), minting past any key a supplied id
-                    # occupies: the collision error is reserved for the caller
-                    # who actually chose the id.
+                    # for notifications). Cannot collide: the counter is past
+                    # every supplied integer key (advanced above) and every
+                    # previously minted id; the collision error is reserved for
+                    # the caller who actually chose the id.
                     self._next_id += 1
-                    while self._next_id in self._in_flight_ids:
-                        self._next_id += 1
                     request_id = self._next_id
                     in_flight_key = request_id
                 self._in_flight_ids.add(in_flight_key)
--- a/src/mcp/shared/dispatcher.py
+++ b/src/mcp/shared/dispatcher.py
@@ -87,8 +87,10 @@ class CallOptions(TypedDict, total=False):
     Callers that need to know a request's id before its result arrives (a
     `subscriptions/listen` stream is demultiplexed by it) mint their own ids
     here; string ids that don't parse as integers can never collide with the
-    dispatcher's minted sequence. Per the class contract, dispatchers that
-    predate this key ignore it and mint as usual.
+    dispatcher's minted sequence, and a supplied integer (or numeric-string)
+    key advances that sequence past itself, so a minted id never reuses a
+    supplied one. Per the class contract, dispatchers that predate this key
+    ignore it and mint as usual.
     """
 
     timeout: float
--- a/src/mcp/shared/jsonrpc_dispatcher.py
+++ b/src/mcp/shared/jsonrpc_dispatcher.py
@@ -335,12 +335,16 @@ class JSONRPCDispatcher(Dispatcher[TransportT]):
             pending_key = coerce_request_id(request_id)
             if pending_key in self._pending:
                 raise ValueError(f"request id {request_id!r} is already in flight")
+            # Advance the mint counter past the supplied key: ids are single-use
+            # within a session, so a minted id may not revisit this key even
+            # after the request completes.
+            if isinstance(pending_key, int):
+                self._next_id = max(self._next_id, pending_key)
         else:
-            # Mint past any key a supplied id occupies: the collision error is
-            # reserved for the caller who actually chose the id.
+            # Cannot collide: the counter is past every supplied integer key
+            # (advanced above) and every previously minted id. The collision
+            # error is reserved for the caller who actually chose the id.
             request_id = self._allocate_id()
-            while request_id in self._pending:
-                request_id = self._allocate_id()
             pending_key = request_id
         out_params = dict(params) if params is not None else {}
         out_meta = dict(out_params.get("_meta") or {})
--- a/tests/shared/test_dispatcher.py
+++ b/tests/shared/test_dispatcher.py
@@ -474,11 +474,35 @@ async def test_minted_ids_skip_a_caller_supplied_id_still_in_flight(pair_factory
 
                 tg.start_soon(parked)
                 await entered.wait()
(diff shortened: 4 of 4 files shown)


ayaangazali:
Gentle check-in on this one. I know review happens in sweeps and there is a lot in the queue, so no rush at all.

Posting a short status so it is easy to triage rather than re-derive:

- The branch still merges cleanly onto `main` as of today, and CI is green.
- I re-checked that the behaviour is still present on `main`: the mint path in `_send_request` only skips ids that are currently in `_pending`, so once a caller-supplied id completes and leaves `_pending`, the counter can still hand that same id back out. Ids are single-use within a session, so that reissue is the part this fixes.

Happy to rebase, split it into something smaller, or take a different approach if that would make it easier to review. Equally happy for it to be closed if you would rather solve it another way, no hard feelings either way.

PR #3147, opened 2026-07-22
fix: make deeply-nested-body test platform-independent (fixes #3146)

### LINKED ISSUE #3146: Deeply-nested-body test fails on macOS: rejected as INVALID_REQUEST (-32600) instead of PARSE_ERROR (-32700)
Running the test suite on macOS (observed on GitHub `macos-latest`, Python 3.14) fails one test that passes on Ubuntu and Windows:

```
FAILED tests/server/test_streamable_http_modern.py::test_modern_post_with_deeply_nested_body_is_parse_error_not_a_crash
>       assert response.json()["error"]["code"] == PARSE_ERROR
E       assert -32600 == -32700
```

https://github.com/modelcontextprotocol/python-sdk/blob/3a6f2996cdd8358957479791e8b26198c07d6a75/tests/server/test_streamable_http_modern.py#L1011

The response status is 400 in both cases; only the JSON-RPC error code differs. The test expects the RecursionError raised while parsing the deeply nested body to classify as a parse error, but on Darwin it surfaces as INVALID_REQUEST, so the recursion limit is evidently being hit in a different layer there (macOS threads run with a smaller default stack, which moves where the recursion guard strikes).

This is unrelated to any open PR: it fails identically on current `main` (3a6f299) and on feature branches, reproduced across repeated runs on `macos-latest`. Repro:

```bash
uv sync --frozen
uv run --frozen pytest tests/server/test_streamable_http_modern.py::test_modern_post_with_deeply_nested_body_is_parse_error_not_a_crash
```

CI currently runs Ubuntu and Windows only, which is why this has gone unnoticed; worth deciding separately whether `macos-latest` should join the matrix.

g0rdonL:
Fixes #3146.

## Root cause

The test assumed a 100k-deep body always makes `json.loads` raise `RecursionError`. Since CPython 3.12 the C json scanner guards recursion by *remaining C-stack headroom*, not a fixed depth — so the outcome depends on the thread's stack size. Measured on Apple Silicon macOS, CPython 3.14:

- main thread: flips to `RecursionError` at depth ~74k → deep body fails to parse → `PARSE_ERROR`
- thread with a 64 MB stack: depth 100k parses fine → giant list fails request validation → `INVALID_REQUEST`

macOS runner threads get a smaller default stack than the main thread, which is why `macos-latest` sees `-32600` while Ubuntu/Windows see `-32700`. Both are correct 400 rejections of the same hostile body; the old assertion was pinning a platform accident.

## Change (test-only)

- `test_modern_post_with_deeply_nested_body_is_rejected_not_a_crash`: asserts 400 with either `PARSE_ERROR` or `INVALID_REQUEST` — the invariant that actually matters is "rejected, not crashed".
- New `test_modern_post_recursion_error_during_parse_is_parse_error`: monkeypatches the body parse to raise `RecursionError`, pinning the `RecursionError → PARSE_ERROR` mapping deterministically on every platform (this is what the old test was really trying to cover).

Verified locally on macOS arm64: both tests pass on CPython 3.13 and 3.14, full `test_streamable_http_modern.py` 52/52, ruff + pyright clean.

### DIFF
--- a/tests/server/test_streamable_http_modern.py
+++ b/tests/server/test_streamable_http_modern.py
@@ -1008,9 +1008,34 @@ async def test_modern_tools_call_logs_a_handler_raised_validation_error_loudly(
     assert "Mcp-Param header validation skipped: the tools/list listing failed" in caplog.text
 
 
-async def test_modern_post_with_deeply_nested_body_is_parse_error_not_a_crash() -> None:
-    """Deep nesting makes json.loads raise RecursionError; still an unparseable body: 400 + PARSE_ERROR."""
+async def test_modern_post_with_deeply_nested_body_is_rejected_not_a_crash() -> None:
+    """A deeply nested body is rejected with a 400, never a crash.
+
+    Which JSON-RPC code it gets is platform-dependent: since CPython 3.12 the C
+    json scanner guards recursion by actual C-stack headroom, so depending on
+    the thread's stack size the body either fails to parse (RecursionError ->
+    PARSE_ERROR) or parses into a giant list that then fails request
+    validation (-> INVALID_REQUEST). Both are correct rejections; the
+    deterministic PARSE_ERROR mapping is covered by the monkeypatch test below.
+    """
     body = b"[" * 100_000 + b"]" * 100_000
+    async with _asgi_client(_x_mcp_server()) as http:
+        response = await http.post("/mcp", content=body, headers={"content-type": "application/json"})
+    assert response.status_code == 400
+    assert response.json()["error"]["code"] in (PARSE_ERROR, INVALID_REQUEST)
+
+
+async def test_modern_post_recursion_error_during_parse_is_parse_error(monkeypatch: pytest.MonkeyPatch) -> None:
+    """RecursionError raised while parsing the body maps to 400 + PARSE_ERROR (deterministically)."""
+    real_loads = json.loads
+    body = b"[" * 50 + b"]" * 50
+
+    def _raise_for_request_body(s: Any, *args: Any, **kwargs: Any) -> Any:
+        if s == body:  # `json` here is the stdlib module; only fail the request-body parse.
+            raise RecursionError("maximum recursion depth exceeded")
+        return real_loads(s, *args, **kwargs)
+
+    monkeypatch.setattr("mcp.server._streamable_http_modern.json.loads", _raise_for_request_body)
     async with _asgi_client(_x_mcp_server()) as http:
         response = await http.post("/mcp", content=body, headers={"content-type": "application/json"})
     assert response.status_code == 400
--- a/tests/server/test_streamable_http_modern.py
+++ b/tests/server/test_streamable_http_modern.py
@@ -1030,7 +1030,11 @@ async def test_modern_post_with_deeply_nested_body_is_rejected_not_a_crash() -> 
     async with _asgi_client(_x_mcp_server()) as http:
         response = await http.post("/mcp", content=body, headers={"content-type": "application/json"})
     assert response.status_code == 400
-    allowed_codes = (PARSE_ERROR, INVALID_REQUEST) if sys.version_info >= (3, 14) else (PARSE_ERROR,)
+    allowed_codes: tuple[int, ...]
+    if sys.version_info >= (3, 14):  # pragma: lax no cover
+        allowed_codes = (PARSE_ERROR, INVALID_REQUEST)
+    else:  # pragma: lax no cover
+        allowed_codes = (PARSE_ERROR,)
     assert response.json()["error"]["code"] in allowed_codes
 
 
--- a/tests/server/test_streamable_http_modern.py
+++ b/tests/server/test_streamable_http_modern.py
@@ -1030,10 +1030,8 @@ async def test_modern_post_with_deeply_nested_body_is_rejected_not_a_crash() -> 
     async with _asgi_client(_x_mcp_server()) as http:
         response = await http.post("/mcp", content=body, headers={"content-type": "application/json"})
     assert response.status_code == 400
-    if sys.version_info >= (3, 14):
-        assert response.json()["error"]["code"] in (PARSE_ERROR, INVALID_REQUEST)
-    else:
-        assert response.json()["error"]["code"] == PARSE_ERROR
+    allowed_codes = (PARSE_ERROR, INVALID_REQUEST) if sys.version_info >= (3, 14) else (PARSE_ERROR,)
+    assert response.json()["error"]["code"] in allowed_codes
 
 
 async def test_modern_post_recursion_error_during_parse_is_parse_error(monkeypatch: pytest.MonkeyPatch) -> None:
--- a/tests/server/test_streamable_http_modern.py
+++ b/tests/server/test_streamable_http_modern.py
@@ -8,6 +8,7 @@ the closed back-channel on the dispatch context, and the request-validation ladd
 
 import json
 import logging
+import sys
 from collections.abc import Callable
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


opensource-joe (commented):
Not a maintainer, just a contributor who went looking for the mechanism here because the reasoning was interesting. The fix looks right to me, and the monkeypatched test is the part I would keep regardless: relaxing the assertion alone would have quietly dropped all coverage of the `RecursionError` to `PARSE_ERROR` mapping, and pinning it separately avoids that. Two corrections to the docstring, both from measurements rather than reading.

I measured `json.loads("[" * N + "]" * N)` on Linux x86_64, each run in its own subprocess so a hard crash is observable, with thread stack sizes set via `threading.stack_size()`.

| Python | main thread | 1 MB | 8 MB | 64 MB | 256 MB | tracks stack? |
|---|---|---|---|---|---|---|
| 3.11.16 | RecursionError from depth 1,000 | same | same | same | same | no |
| 3.12.3 | parses to 5,000, raises from 10,000 | SIGSEGV | same as main | same as main | same as main | no |
| 3.13.15 | parses to 5,000, raises from 10,000 | SIGSEGV | same as main | same as main | same as main | no |
| 3.14.7 | parses to 50,000, raises from 75,000 | raises from 10,000 | flips ~64,901 | parses past 400,000 | parses past 400,000 | yes |

### 1. The stack-headroom behaviour starts in 3.14, not 3.12

On 3.12 and 3.13 the flip depth is identical at 8 MB, 64 MB and 256 MB, so the guard there ignores the stack completely. Only 3.14 moves with it, and on 3.14 it is cleanly linear: roughly 8,100 levels per MB, about 129 bytes of C stack per level. Depth 100,000 needs somewhere between 12 MB and 16 MB on this platform.

This matters a little beyond wording, because the CI matrix runs 3.10 through 3.14. On everything below 3.14 the 100,000-deep body always raises and the response is always `PARSE_ERROR`, deterministically.

### 2. The macOS explanation seems to point the wrong way

The docstring attributes the macOS result to runner threads getting a smaller default stack. Following that through: a smaller stack means less headroom, so `json.loads` raises sooner, which gives `PARSE_ERROR`. That is what the original test asserted, so it would have passed.

The failure in #3146 was `INVALID_REQUEST`, which means the body parsed. That needs *more* headroom than 100,000 levels requires, not less.

My guess is that the per-level C-stack cost differs by architecture, and arm64 macOS fits 100,000 levels in a stack where x86_64 Linux needs about 16 MB. I cannot confirm that, since I only have Linux to measure on, so please treat it as a hypothesis rather than a correction to yours. Either way the conclusion is the same, and it is the one your fix already encodes: which of the two codes you get is not something the test should pin.

### Suggested docstring

Something like this, if the measurements above hold up on your side:

> Which JSON-RPC code it gets is platform-dependent. From CPython 3.14 the C json scanner guards recursion by actual C-stack headroom rather than a fixed limit, so whether a 100k-deep body parses depends on how much stack the parsing thread has and on the per-level cost on that architecture. It either fails to parse (RecursionError, so PARSE_ERROR) or parses into a giant list that then fails request validation (INVALID_REQUEST). Both are correct rejections; the deterministic PARSE_ERROR mapping is covered by the monkeypatch test below.

### One optional idea

Since the outcome is deterministic below 3.14, you could keep the stronger assertion where it still holds:

```python
if sys.version_info >= (3, 14):
    assert response.json()["error"]["code"] in (PARSE_ERROR, INVALID_REQUEST)
else:
    assert response.json()["error"]["code"] == PARSE_ERROR
```

That keeps four of the five supported versions asserting the exact code. It also adds a version branch to a test, which is its own kind of cost, so I would understand leaving it as it is.

### Unrelated, and not yours to fix

On 3.12 and 3.13, a thread with a 1 MB stack segfaults parsing a deeply nested body rather than raising, because the fixed limit does not account for the actual stack available. That is CPython's, not this PR's, and the server is unlikely to parse on such a thread. Mentioning it only because the test being replaced was named `..._not_a_crash`, so it seemed worth someone knowing.

*Disclosure: I used AI assistance for the measurements and for writing this up. I ran them myself and can go through any of it.*

g0rdonL:
Thanks @opensource-joe — this is exactly the kind of review I hoped the writeup would attract, and you're right on both counts. I re-measured on arm64 macOS (each probe in its own subprocess, `threading.stack_size()` set explicitly, binary-searched flip depths):

| Python (arm64 macOS) | main thread | 1 MB | 8 MB | 16 MB | 64 MB |
|---|---|---|---|---|---|
| 3.12.13 | raises from ~10,363 | ~7,632 (SIGSEGV at 100k) | ~10,363 | ~10,363 | ~10,363 |
| 3.13.12 | raises from ~10,363 | ~6,851 (SIGSEGV at 100k) | ~10,363 | ~10,363 | ~10,363 |
| 3.14.3 | ~74,735 | ~9,192 | ~74,735 | ~149,641 | ~599,079 |

**1. Confirmed: the headroom guard starts in 3.14, not 3.12.** On 3.12/3.13 the flip depth is identical at 8/16/64 MB here too — the guard ignores the stack. On 3.14 it's linear at ~112 bytes/level on arm64 (vs your ~129 on x86_64). I also reproduce your 1 MB SIGSEGV on 3.12/3.13.

**2. Your "points the wrong way" catch is correct, and the actual mechanism turned out to be neither of our guesses.** A default-stack thread on macOS flips at ~149,641 — exactly the 16 MB row. CPython pins non-main threads to a 16 MiB stack on macOS (`THREAD_STACK_SIZE` in `Python/thread_pthread.h`), which is *larger* than typical elsewhere, not smaller. So on macOS the 100k body fits, parses, and fails validation → `INVALID_REQUEST`, while a platform whose parsing thread has ≲12 MB raises → `PARSE_ERROR`. Per-level cost differing by arch (your hypothesis) is real but secondary; the 2× thread-stack difference is the main driver.

I've pushed 1956ff6 adopting both suggestions: docstring rewritten with the corrected version boundary and mechanism, and the version-gated assertion — since the outcome is deterministic below 3.14, the test now pins the exact `PARSE_ERROR` there and accepts either rejection only on 3.14+. Full file 52/52 on 3.13 and 3.14, ruff + pyright clean.

g0rdonL:
Still ready for review - feedback from the last round was addressed; happy to iterate further if needed.

UgaTheDev:
Not a maintainer either, but I have a macOS arm64 box, which is the platform CI does not cover, so I reproduced the whole chain locally. Everything in the docstring holds up, including the version boundary, which I think is the part most worth having independently checked.

On current `main` (unpatched), same machine, same body:

| Interpreter | Result |
|---|---|
| 3.13.5 arm64 | passes, code is `PARSE_ERROR` |
| 3.14.3 arm64 | fails, `assert -32600 == -32700` |

That is the reported failure, and it confirms the claim that the split is 3.14-only rather than macOS-only: below 3.14 the recursion guard ignores stack size and the 100k-deep body always fails to parse, so asserting the exact code there is still meaningful. Gating on `sys.version_info >= (3, 14)` rather than on `sys.platform` is therefore the right axis, and it is the detail I would have gotten wrong.

On this PR's head (0aac7e9) both new tests pass on 3.14.3 and on 3.13.5.

I also wanted to check that the monkeypatched test really does hold the line that the relaxed assertion gives up, so I removed `RecursionError` from the `except (ValueError, RecursionError)` tuple in `_streamable_http_modern.py` and re-ran both on 3.14:

- `test_modern_post_recursion_error_during_parse_is_parse_error` fails, surfacing the raw `RecursionError`.
- `test_modern_post_with_deeply_nested_body_is_rejected_not_a_crash` still passes.

So the regression coverage genuinely transfers to the deterministic test, and relaxing the assertion alone would have left the `RecursionError` to `PARSE_ERROR` mapping untested. That matches what @opensource-joe said above; it now has a fail-without-fix demonstration behind it.

One small suggestion, take it or leave it: the issue's closing question about adding `macos-latest` to the matrix is the reason this sat unnoticed, and it is arguably the more valuable half. Might be worth splitting into its own issue so it does not get closed along with this PR.

Issue #772, opened 2025-05-21
No support of inputSchema in mcp.tool decorator


fmtsvetkov:
**Describe the bug**
It's more an inconsistence than a bug but there is no "inputSchema" field in mcp.tool decorator while it's available in Tool 

**To Reproduce**
@mcp.tool(
    name="Name",
    description=(
        "description"
    ),
    inputSchema = {}
)

**Expected behavior**
Add of mcp tool to mcp server

**Current behavior**
TypeError: FastMCP.tool() got an unexpected keyword argument 'inputSchema'

ivanbelenky:
The actual `inputSchema` is missing based on the fact that there are 2 types of Tool obejcts, the internal server implementation and the protocol one usually imported as `MCPTool`. Have you chekced both tools structures?

in `fastmcp.tools.base`

```python

class Tool(BaseModel):
    """Internal tool registration info."""

    fn: Callable[..., Any] = Field(exclude=True)
    name: str = Field(description="Name of the tool")
    description: str = Field(description="Description of what the tool does")
    parameters: dict[str, Any] = Field(description="JSON schema for tool parameters")
    fn_metadata: FuncMetadata = Field(
        description="Metadata about the function including a pydantic model for tool"
        " arguments"
    )
    is_async: bool = Field(description="Whether the tool is async")
    context_kwarg: str | None = Field(
        None, description="Name of the kwarg that should receive context"
    )
    annotations: ToolAnnotations | None = Field(
        None, description="Optional annotations for the tool"
    )
```

in `types`
```python
# types.py
class Tool(BaseModel):
    """Definition for a tool the client can call."""

    name: str
    """The name of the tool."""
    description: str | None = None
    """A human-readable description of the tool."""
    inputSchema: dict[str, Any]
    """A JSON Schema object defining the expected parameters for the tool."""
    annotations: ToolAnnotations | None = None
    """Optional additional tool information."""
    model_config = ConfigDict(extra="allow")
```


I know nothing about nothing but allowing for `inputSchema` would require extra validation with respect to the automatically introspective build of `parameters` that precisely holds that info.

felixweinberger:
As pointed out by @ivanbelenky this is expected - the internal `Tool` class determines its inputschema via introspection.

Issue #761, opened 2025-05-20
Allow Direct inputSchema Registration for Tools (Critical for API Gateway Integration)


isxylands:
**Background / Motivation:**

We are building an integration between our API gateway—which manages a large and dynamic set of HTTP APIs (with detailed OpenAPI/Swagger metadata)—and the Model Context Protocol (MCP) ecosystem. Our goal is to **automatically register all these APIs as MCP Tools**, so LLMs or other MCP clients can discover and invoke them, without requiring manual code for every endpoint.

* * *

**Current Limitation and Problems:**

*   FastMCP currently only allows tool registration via decorators (e.g. `@server.tool(...)`), which expects a Python function per tool.
    
*   The schema for each tool's parameters (`inputSchema`) is **implicitly inferred from the function signature**, and there is **no supported way to explicitly set or override the full inputSchema** at registration time.
    
*   If an API has many parameters (with type, description, default, enum, etc.), all that detail is lost unless you generate a matching function signature. Even then, the `description` for each parameter cannot be preserved, as Python functions do not natively support such metadata.
    

* * *

**Integration Challenge:**

*   For API gateways and similar dynamic platforms, it is infeasible to generate a new Python function for every API, every time the set changes. We attempted to work around this by dynamically generating functions and registering them as tools.
    
*   **Even with dynamic code generation, the SDK only infers parameter names and types, but drops important fields like `description`.**
    
*   We tried post-processing (patching) the registered tools' `inputSchema` field, but in FastMCP this is not always possible or reliable:
    
    *   The internal `tool_manager` or equivalent registry is not always accessible at registration time.
        
    *   Service lifecycle and asynchronous initialization make it difficult (or impossible) to guarantee access to the real Tool objects for patching.
        
    *   This adds operational complexity and risk, especially across different FastMCP versions or deployments.
        

* * *

**Why This Matters:**

*   The lack of `description` (and other schema metadata) means LLMs and clients cannot prompt users correctly or generate valid parameter values.
    
*   The current approach is fragile, difficult to maintain, and cannot guarantee correctness as the underlying SDK evolves.
    

* * *

**Feature Request / Proposal:**

> **Please support explicit setting of the full `inputSchema` when registering a tool** (e.g., via a keyword argument on the registration/decorator function, or by allowing direct registration of Tool objects).
> 
[...]

Issue #323, opened 2025-03-19
No way to declare input schema for tools using Fast MCP


jay2jp:
There appears to be no clear cut way to add an input_schema into a tool when declaring an MCP server using the FastMCP initiator. 

the tool decorator within the FastMCP class has no way to initialize an input schema.
```
      def tool(
          self, name: str | None = None, description: str | None = None
      ) -> Callable[[AnyFunction], AnyFunction]:

        # Check if user passed function directly instead of calling decorator
        if callable(name):
            raise TypeError(
                "The @tool decorator was used incorrectly. "
                "Did you forget to call it? Use @tool() instead of @tool"
            )

        def decorator(fn: AnyFunction) -> AnyFunction:
            self.add_tool(fn, name=name, description=description)
            return fn

        return decorator
```
while the tool object its self has a input schema attribute

```
class Tool(BaseModel):


    name: str
    """The name of the tool."""
    description: str | None = None
    """A human-readable description of the tool."""
    inputSchema: dict[str, Any]
    """A JSON Schema object defining the expected parameters for the tool."""
    model_config = ConfigDict(extra="allow")

```

the tool manager within fastMCP also has no way of inputting a schema when using add_tool

```
  class ToolManager:


    def __init__(self, warn_on_duplicate_tools: bool = True):
        self._tools: dict[str, Tool] = {}
        self.warn_on_duplicate_tools = warn_on_duplicate_tools

    def get_tool(self, name: str) -> Tool | None:
        """Get tool by name."""
        return self._tools.get(name)

    def list_tools(self) -> list[Tool]:
        """List all registered tools."""
        return list(self._tools.values())

    def add_tool(
        self,
        fn: Callable,
        name: str | None = None,
        description: str | None = None,
    ) -> Tool:
        """Add a tool to the server."""
        tool = Tool.from_function(fn, name=name, description=description)
        existing = self._tools.get(tool.name)
        if existing:
            if self.warn_on_duplicate_tools:
                logger.warning(f"Tool already exists: {tool.name}")
            return existing
        self._tools[tool.name] = tool
        return tool

    async def call_tool(
        self,
        name: str,
        arguments: dict[str, Any],
        context: Context[ServerSessionT, LifespanContextT] | None = None,
    ) -> Any:
        """Call a tool by name with arguments."""
        tool = self.get_tool(name)
        if not tool:
            raise ToolError(f"Unknown tool: {name}")

        return await tool.run(arguments, context=context)
```

[...]

PR #3148, opened 2026-07-23
Add a runtime create_model variant to the schema_validators example

Aaron-Oh:
## What

The `schema_validators` example story shows four ways to type a tool
parameter so `MCPServer` derives and enforces `inputSchema`: a pydantic
`BaseModel`, a `TypedDict`, a `@dataclass`, and a bare `dict[str, Any]`.

This adds a fifth variant: a pydantic model built **at runtime** with
`create_model` from an external JSON Schema dict, then handed to
`@mcp.tool()` exactly like a hand-written `BaseModel`.

## Why

Issues #323, #761, and #772 all asked the same thing: how to drive a
tool's `inputSchema` from a JSON Schema you already hold (from OpenAPI, a
config file, a DB row) rather than a class written out in source. The
maintainer answer is "use a pydantic model as the parameter" — but the
example suite never showed how to *get* that model when it isn't declared
statically. This closes that documentation gap with a runnable variant.

## Notes

- A `create_model()` result is opaque to static type checkers (its fields
  don't exist until runtime, and a runtime variable can't appear in a type
  annotation). A `TYPE_CHECKING` branch aliases it to a same-shape declared
  model so type checkers can see the fields; at runtime the dynamic class is
  what `@mcp.tool()` reflects over. This is called out in the README.
- The published schema is identical to the `greet_pydantic` variant — the
  point is purely *how the model is obtained*, not a different wire shape.
- `server_lowlevel.py`, `client.py`, and `README.md` are updated to cover
  the new variant.

## Validation

- `uv run --frozen ruff format --check` / `ruff check` — clean
- `uv run --frozen pyright` — 0 errors
- `uv run --frozen pytest tests/examples -k schema` — 14 passed
  (in-memory/http × modern/legacy × server/server_lowlevel, plus the
  manifest and story-shape checks)

### DIFF
--- a/examples/stories/schema_validators/README.md
+++ b/examples/stories/schema_validators/README.md
@@ -1,9 +1,10 @@
 # schema-validators
 
-Four ways to type a tool parameter so `MCPServer` derives the JSON-Schema
+Five ways to type a tool parameter so `MCPServer` derives the JSON-Schema
 `inputSchema` and validates arguments before your handler runs: a pydantic
-`BaseModel`, a `TypedDict`, a `@dataclass`, and a bare `dict[str, Any]`. The
-client lists the tools, resolves each `who` schema, and round-trips a call.
+`BaseModel`, a `TypedDict`, a `@dataclass`, a bare `dict[str, Any]`, and a
+pydantic model built at runtime with `create_model`. The client lists the
+tools, resolves each `who` schema, and round-trips a call.
 
 ## Run it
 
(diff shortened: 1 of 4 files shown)


Aaron-Oh:
Thanks for the review — addressed both in 230f718:

- **P2 (server.py):** `required` is optional in JSON Schema, so it's now defaulted to an empty list before the membership test. A schema with only optional properties (the OpenAPI/config/DB-row case this pattern advertises) builds instead of raising `KeyError` at import time. Verified an all-optional schema now builds while the original still behaves identically.
- **P3 (README.md):** `greet_dynamic` publishes the same schema as `greet_pydantic`, so the `client.py` bullet now reads "four typed variants" (calling out the runtime `create_model` one) rather than three.

Aaron-Oh:
Hi maintainers, sorry for the small bump on this one.

Just a gentle ping — CI is all green here, and the two cubic comments are already addressed in 230f718 (the missing `required` one was actually a real bug at import time, so thanks for catching it).

I know the review queue is quite busy these days, so please take your time, no rush at all. But if this variant is not the direction you want for `examples/stories/schema_validators/`, please feel free to tell me — I'm happy to rework the scope or just close it. Any feedback would be very appreciated.

Thank you in advance!

PR #3149, opened 2026-07-23
fix(auth): allow requested scopes without registered scope

pradeep-ramola:
Fixes #2216

## Summary

Updates `OAuthClientMetadata.validate_scope()` so a registered client with `scope=None` does not reject requested authorization scopes.

When the registration response omits `scope`, the SDK treats that as no client-specific scope restriction instead of converting it to an empty allowed-scope list. Explicit registered scopes are still enforced.

## Tests

```bash
.\.venv\Scripts\python.exe -m pytest tests/shared/test_auth.py -q
.\.venv\Scripts\python.exe -m ruff check src/mcp/shared/auth.py tests/shared/test_auth.py
.\.venv\Scripts\python.exe -m ruff format --check src/mcp/shared/auth.py tests/shared/test_auth.py
git diff --check
```

## Notes

Added regression coverage for:

- no requested scope returns `None`
- registered scopes are accepted
- unregistered scopes are still rejected
- omitted registered scope allows requested scopes

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -109,7 +109,9 @@ class OAuthClientMetadata(BaseModel):
         if requested_scope is None:
             return None
         requested_scopes = requested_scope.split(" ")
-        allowed_scopes = [] if self.scope is None else self.scope.split(" ")
+        if self.scope is None:
+            return requested_scopes
+        allowed_scopes = self.scope.split(" ")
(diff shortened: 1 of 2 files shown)


maxisbey:
Thanks for the PR. We're tracking this fix in #3222 instead, so I'm closing this one. Feel free to reopen if this is still relevant.

PR #3151, opened 2026-07-23
Serve the 2026-07-28 protocol over stdio: decide the era from the opening message

maxisbey:
> **Status:** opened for review while two things finish on the branch: a rebase onto
> current `main` (this branch predates #3143, so it currently conflicts in the
> classification files), and a cleanup that moves listen-stream state off the handler
> and the server-level `close_subscriptions()` verb (details at the bottom). Both
> land as follow-up pushes here.

### Motivation

The 2026-07-28 protocol did not work over stdio, and when I dug in it came down to one refused
feature plus a handful of conformance defects that all shared a root cause. A connection's
protocol era (the 2025 `initialize` handshake versus the 2026 self-contained per-request
envelope) was being derived after the fact from which requests had *completed*, in several
places, instead of being decided once, in wire order, from how the client opened the connection.
That model couldn't host a request that never returns, so `subscriptions/listen` was
hard-refused over stdio with `-32601` even though the same connection advertised
`tools.listChanged: true`; a slow modern request left the connection "unlocked" so a legacy
handshake arriving mid-flight was accepted and re-locked the connection out from under the
running call; and a peer cancel produced a trailing `{"code": 0, "message": "Request
cancelled"}` frame that the stdio binding forbids ("MUST NOT send any further messages" for a
cancelled request).

The listen handler was never the problem. Served through a driver that decides the era from
the opening message, the unmodified handler produces the exact spec flow over a pipe (ack first,
id-stamped events, graceful close, silence after a client cancel). So this change is not "make
listen work on stdio"; it is "decide the era once, in the right place, and delete the code that
was compensating for deciding it wrong."

### What changed

**One stream driver, era decided in the read loop.** `serve_dual_era_loop` and `serve_loop` are
replaced by a single `serve_stream(server, read, write)`. The connection's era is decided
synchronously, inside the JSON-RPC dispatcher's sequential read loop, before the request body is
spawned and before the next frame is read, so it is ordered by the wire and by nothing else (no
appeal to task scheduling anywhere). `initialize` or any envelope-less request opens the legacy
era; an enveloped request opens the modern era; an enveloped `server/discover` is a probe that
is answered but pins nothing, so a client that probes and falls back to `initialize` still
finds the handshake available; a stray leading notification opens nothing. A second,
conflicting era claim on a committed connection is refused rather than silently switching (a
legacy `initialize` on a modern connection gets `-32022` naming the served versions; an
enveloped request on a handshake connection gets `-32600`). Both eras are always served; there
is no server-level switch to disable one, matching the other SDKs.

**Cancel silence by construction.** Each request now answers through a one-shot channel whose
write target is swapped for a powerless void by a peer cancel, by the request's own terminal
write, and by the peer going away. There is no `if cancelled` at any write site. The commit
point is one synchronous step (leave the cancellable table and take the write target, no
checkpoint between), which gives a clean rule for the race: if the cancel is read before the
body returned, nothing is written; if the body had already returned, the owed answer is written
exactly once even when parked on backpressure. That satisfies both the stdio MUST NOT and the
cancellation pattern's "MAY ignore a cancel whose processing has completed" at the same time.

**The dispatcher knows less MCP than before.** The generic JSON-RPC dispatcher loses the code-0
cancel frame and the code-0 `str(exc)` catch-all (both legacy-era vocabulary in plumbing that
shouldn't know eras exist) and the `inline_methods={"initialize"}` knob. It gains two
MCP-agnostic mechanisms: an optional receive-order admission gate on `run(admit=)` and the
request channel. The `Dispatcher` contract's `OnRequest`/`OnNotify` are now documented as invoked
synchronously in receive order, with the returned awaitable as the body; every existing
`async def` handler already satisfies that, since calling an `async def` does no work. This is
the same shape as the go-sdk's jsonrpc2 `Preempter` (a pre-queue admission hook), so it's
generic JSON-RPC prior art rather than an MCP intrusion.

**Author-facing surface.** `Server.run(read, write)` now stands alone
(`initialization_options` is optional). `Server.lifespan()` is a bound zero-argument context
manager, so the odd `server.lifespan(server)` self-reference is gone. New: `serve_listener(server,
listener)`, the socket-shaped host that enters the lifespan once and drives every accepted
connection, and `mcp.server.stdio.newline_json_transport(byte_stream)`, the newline-delimited
JSON-RPC framer for any byte stream. The custom-transport docs are now a four-rung ladder from
`server.run()` up to composing `server.lifespan()` + `serve_stream(..., lifespan_state=)`
yourself, with the enter-the-lifespan-once rule stated on the rung where it bites.

**Other observable changes** (all in the migration guide with before/after): the discovery
probe no longer pins; a modern request needs only the required `_meta` pair (`clientInfo` is
optional); unmapped handler exceptions become `-32603 "Internal server error"` with the text
kept in the server log rather than on the wire; `MCPServer` prompt errors are `-32602` with the
authored message; at stdin EOF the server drains gracefully (open listens end with their result)
instead of writing `CONNECTION_CLOSED` per id onto a closing pipe; over legacy streamable HTTP
in JSON mode a peer-cancelled POST now completes as an empty `202 Accepted` instead of lingering
until the client disconnects.

### Migration

Everything is grouped in `docs/migration.md`; the short version:

- `serve_loop(...)` / `serve_dual_era_loop(...)` become `server.run(read, write)` for a single
  connection, or `serve_stream(..., lifespan_state=)` when connections share a lifespan you
  entered yourself.
- `server.lifespan(server)` becomes `async with server.lifespan() as state:` (a `TypeError`
  otherwise). The constructor's `lifespan=` callable is unchanged.
- If you implement your own `Dispatcher`, note the documented obligation that handlers are
  invoked synchronously in receive order and return the body awaitable, and that
  `JSONRPCDispatcher(inline_methods=)` is gone (state `hold=(method == "initialize")` in your
  own admission via `run(admit=)` if you need the handshake to hold the loop).
- Code that hand-sent `notifications/cancelled` while still awaiting the request now gets no
  response for the cancelled id (that's the spec); code that read exception text off `code=0`
  errors should raise `MCPError` to put a deliberate message on the wire.
- `MCPServer.get_prompt()` raises `MCPError` (was `ValueError`) for its own prompt errors.

### Testing

- New permanent suites, all in-repo and running on both `asyncio` and `trio`: a wire-ordering
  suite (era decided by the opening message, the straddle refused, probe-then-fallback,
  stray-notification and claim-less-request rules, cancel silence), and an 18-case cancel-race
  suite pinning every direction of the cancel-versus-answer race, the EOF cases, and the
  owner-teardown cases, with no fixed sleeps. Plus a streamable-HTTP suite for the JSON-mode
  cancelled-POST completion, and the existing stream-driver, stdio, and dispatcher suites
  parametrized over both anyio backends.
- Full suite green with 100% branch coverage; `strict-no-cover` clean; pyright and ruff clean;
  no new `# pragma: no cover`, `# type: ignore`, or `# noqa` in `src/`.
- End to end over real stdio pipes with raw newline-delimited JSON-RPC: listen open with
  concurrent traffic, published event, client cancel with no trailing frame, server-initiated
  graceful close, legacy handshake during in-flight modern work, pair-only envelope routed
  modern, late handshake refused `-32022`, and probe-then-fallback.
- Interop against the TypeScript SDK over stdio in both directions: its client in auto mode
  against this server (probe to `DiscoverResult` to modern requests, ack-first stamped listen and
  cancel silence, and a forced probe timeout landing on a working legacy handshake on the same
  connection), and this client against its server.

### Still landing on this branch

- Rebase onto current `main` now that #3143 has merged (the envelope work overlaps the same
  classification code), which also retires the interop caveat about the previously required
  `clientInfo`.
- Moving open listen-stream state off the `ListenHandler` and dropping the server-level
  `close_subscriptions()` verb: streams belong to the connection that opened them (and end with
  it), the handler is stateless, and a graceful close-everything is the subscription bus's job.
  Today, as on `main`, a `Server` shared across connections ends every connection's listen
  streams when one connection closes; this fixes that by scope rather than by caveat.

Out of scope, deliberately: making one publish reach a 2026 listen stream *and* a 2025 client
(feeding legacy peers from the subscription bus) is a separate follow-up, since it changes what
a legacy handshake advertises and delivers.

### DIFF
--- a/docs/advanced/low-level-server.md
+++ b/docs/advanced/low-level-server.md
@@ -187,7 +187,51 @@ Each of these is one idea you now have the vocabulary for; each has its own page
 * `on_call_tool`, `on_get_prompt`, and `on_read_resource` may return an `InputRequiredResult` instead of their normal result to pause the call and ask the client for input; see **[Multi-round-trip requests](../handlers/multi-round-trip.md)**. True to this tier, nothing is installed for you: where `MCPServer` seals `requestState` by default, here the `request_state` you set crosses the wire exactly as written until you opt in with `server.middleware.append(RequestStateBoundary(RequestStateSecurity(keys=[...]), default_audience=server.name))`: one line (both names import from `mcp.server.request_state`) for the identical sealing and verification `MCPServer` performs (**[Protecting `requestState`](../handlers/multi-round-trip.md#protecting-requeststate)**).
 * `on_list_resources`, `on_read_resource`, `on_list_prompts`, `on_get_prompt`, `on_completion` are the same `(ctx, params) -> result` shape for the other primitives.
 * `on_subscriptions_listen` serves the 2026-07-28 `subscriptions/listen` stream. Pass a `ListenHandler` built over a `SubscriptionBus` and publish events to the bus from your other handlers; see **[Subscriptions](../handlers/subscriptions.md)** for the full composition.
-* `server.streamable_http_app()` returns the same Starlette app `MCPServer`'s does; deploy it the way **[Running your server](../run/index.md)** deploys any other ASGI app. There is no `server.run(transport=...)` down here: `server.run(read_stream, write_stream, server.create_initialization_options())` drives one connection over a pair of streams, and that one line is the whole story.
+* `server.streamable_http_app()` returns the same Starlette app `MCPServer`'s does; deploy it the way **[Running your server](../run/index.md)** deploys any other ASGI app. There is no `server.run(transport=...)` down here: `server.run(read_stream, write_stream)` drives one connection over a pair of duplex streams (stdio, a socket you framed, an in-memory pair in a test), and that one line is the whole story. Every other hosting shape is the next section.
+
+## Serving over your own transport
+
+`server.run(read_stream, write_stream)` takes any duplex message-stream pair, so a transport is just a source of those two streams. There are four rungs, from the shortest to the most explicit; take the highest one that fits your channel.
+
+**One connection over a stream pair: `server.run(read, write)`.** stdio is `stdio_server()`, an in-memory pair in a test is the same call. The lifespan is entered for you: one connection, one call.
+
+**Many connections from a byte-stream listener: `serve_listener(server, listener)`.** A whole Unix-socket server is:
+
+```python
+import anyio
+
+from mcp.server import serve_listener
+
+
+async def main() -> None:
+    listener = await anyio.create_unix_listener("/tmp/bookshop.sock")
+    await serve_listener(server, listener)
+```
+
+`serve_listener` enters the server's lifespan **once**, then frames and serves every connection the listener accepts, and closes the listener when it is cancelled. That is the difference from calling `server.run()` per accepted connection, which would open the lifespan (your database pool, your caches) again for each socket.
+
+One shared-server behaviour to know before you copy this: open `subscriptions/listen` streams belong to the server's `ListenHandler`, not to a connection, so when one connection's input ends its listen streams are closed gracefully together with those of **every** connection this server is serving; each affected client receives its listen request's result and re-listens. It never bites on stdio (one connection per process), but behind a listener it does. Until listen streams are owned per connection, a host whose departing peer must not touch its neighbours' subscriptions builds a `Server` per accepted connection, each with its own `ListenHandler(bus)`, all driven by `serve_stream(..., lifespan_state=state)` off one entered state (the last rung below); an `MCPServer`, whose one `ListenHandler` is shared, has no such workaround yet.
+
+**A byte stream you already hold: frame it, then `run`.** `newline_json_transport(byte_stream)` puts the same newline-delimited JSON-RPC wire over any byte stream and yields the pair `run` wants: `async with mcp.server.stdio.newline_json_transport(byte_stream) as (read, write):`, then `server.run(read, write)`. One connection, so the lifespan is again entered for you.
+
+**Connections that are not a byte-stream listener: enter the lifespan once, drive each one.** A WebSocket adapter that already yields messages, a broker, your own test loop: compose the same pieces yourself. The rule that bites here is the one `serve_listener` hides: enter `server.lifespan()` **once**, and hand its state to `serve_stream` for every connection, so a hundred sockets share one database pool instead of opening a hundred:
+
+```python
+from mcp.server import serve_stream
+
+async with server.lifespan() as state, anyio.create_task_group() as tg:
+    async for read_stream, write_stream in your_connections():
+        tg.start_soon(lambda r=read_stream, w=write_stream: serve_stream(server, r, w, lifespan_state=state))
+```
+
+The lifespan is entered **outside** the task group on purpose: on the way out the task group joins the still-running connection tasks first and only then does the lifespan tear down, so the shared pool outlives every connection using it (the other order tears the pool down while connections are still being served).
+
+Two things about the shared-`Server` shape that the single-connection rungs never surface:
+
+* Which protocol eras a server offers is `Server(posture=...)` (`Posture.DUAL` by default, `Posture.MODERN_ONLY`, `Posture.LEGACY_ONLY`); it rides along on the server object, so none of these drivers takes a posture argument you could forget.
+* The cross-connection listen behaviour called out under `serve_listener` above applies to any shared `Server`, including this rung; the per-connection-`Server` recipe there is how you sidestep it.
+
+An `MCPServer` reaches every one of these through `mcp.lowlevel_server`, and `mcp.close_subscriptions()` (or `Server.close_subscriptions()` down here) ends the open listen streams gracefully from the server's side.
 
 ## Recap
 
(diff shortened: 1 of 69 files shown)


docs/advanced/low-level-server.md
```diff
@@ -187,7 +184,48 @@ Each of these is one idea you now have the vocabulary for; each has its own page
...
 * `on_subscriptions_listen` serves the 2026-07-28 `subscriptions/listen` stream. Pass a `ListenHandler` built over a `SubscriptionBus` and publish events to the bus from your other handlers; see **[Subscriptions](../handlers/subscriptions.md)** for the full composition.
-* `server.streamable_http_app()` returns the same Starlette app `MCPServer`'s does; deploy it the way **[Running your server](../run/index.md)** deploys any other ASGI app. There is no `server.run(transport=...)` down here: `server.run(read_stream, write_stream, server.create_initialization_options())` drives one connection over a pair of streams, and that one line is the whole story.
+* `server.streamable_http_app()` returns the same Starlette app `MCPServer`'s does; deploy it the way **[Running your server](../run/index.md)** deploys any other ASGI app. There is no `server.run(transport=...)` down here: `server.run(read_stream, write_stream)` drives one connection over a pair of duplex streams (stdio, a socket you framed, an in-memory pair in a test), and that one line is the whole story. Every other hosting shape is the next section.
+
+## Serving over your own transport
```
maxisbey, line 192: why even add this, this isn't needed int his PR

src/mcp/server/lowlevel/server.py
```diff
@@ -519,6 +533,17 @@ def get_notification_handler(self, method: str) -> HandlerEntry[LifespanResultT]
         """Return the registered entry for a notification method, or `None`."""
         return self._notification_handlers.get(method)
 
+    def close_subscriptions(self) -> None:
```
maxisbey, line 528: why is this here?

src/mcp/server/mcpserver/server.py
```diff
@@ -290,9 +290,34 @@ def icons(self) -> list[Icon] | None:
...
+        `await mcp.subscriptions.publish(ToolsListChanged())`."""
+        return self._subscriptions
+
+    def close_subscriptions(self) -> None:
+        """Gracefully close every open `subscriptions/listen` stream (all connections).
+
+        Streams drain their pending events and end with the listen request's
+        result while the connection carries on; returns before they finish flushing.
+        """
+        self._lowlevel_server.close_subscriptions()
```
maxisbey, line 320: neither of these should be here. this is making state available on the MCPServer instead of the context object, bad

src/mcp/server/mcpserver/server.py
```diff
@@ -290,9 +290,34 @@ def icons(self) -> list[Icon] | None:
...
+    @property
+    def lowlevel_server(self) -> Server[LifespanResultT]:
+        """The low-level `Server` this instance is built on.
+
+        Hand it to the stream drivers `run()` does not name:
+
+            listener = await anyio.create_unix_listener("/tmp/mcp.sock")
+            await serve_listener(mcp.lowlevel_server, listener)
+        """
+        return self._lowlevel_server
```
maxisbey, line 306: why are you making this public? this shouldn't be public I see no reason for it relating to what this PR is trying to achieve

src/mcp/server/__init__.py
```diff
@@ -3,5 +3,15 @@
...
-__all__ = ["CacheHint", "Server", "ServerRequestContext", "MCPServer", "NotificationOptions", "InitializationOptions"]
+__all__ = [
+    "CacheHint",
+    "Server",
+    "ServerRequestContext",
+    "MCPServer",
+    "NotificationOptions",
+    "InitializationOptions",
+    "serve_listener",
+    "serve_stream",
```
maxisbey, line 16: what are these why are they public now?

src/mcp/server/stdio.py
```diff
@@ -75,3 +83,62 @@ async def stdout_writer():
         tg.start_soon(stdin_reader)
         tg.start_soon(stdout_writer)
         yield read_stream, write_stream
+
+
+@asynccontextmanager
+async def newline_json_transport(
```
maxisbey, line 89: why make this public? we're not wanting to support custom transports. this PR should _just_ be about making stdio work for 2026 subscriptions. refactoring to make the internals is better, but we're not wanting to add a bunch of new functionality here. that isn't the goal

src/mcp/shared/dispatcher.py
```diff
@@ -219,17 +201,45 @@ async def progress(self, progress: float, total: float | None = None, message: s
...
+    `handler` serves it; when `hold` is true the receive loop parks until the
+    request finishes answering, so its body must not await the peer.
+    """
+
+    handler: OnRequest
+    hold: bool = False
+
+
+Admit = Callable[[str, Mapping[str, Any] | None], Admission]
+"""Synchronous receive-order gate for inbound requests: `(method, params) -> Admission`.
```
maxisbey, line 229: what are these for and whya re they in this file if they're not used here

PR #3206, opened 2026-07-28
Validates registered redirect_uris for DCR are a secure schema with no fragments

### LINKED ISSUE #2629: DCR registration accepts redirect_uris with non-HTTPS / non-loopback / fragmented schemes

Varshith-Kali:
Fixes #2629.

## Summary

Dynamic Client Registration currently accepts redirect_uris with unsafe schemes (javascript:, data:, file:, ftp:) and fragments without complaint. RFC 9700 .1.1 and RFC 7591  require HTTPS for redirect_uris and prohibit fragments.

## Changes

- Added validate_redirect_uri() to routes.py, mirroring the existing validate_issuer_url() pattern: only https:// allowed, with http:// loopback exception for localhost/127.0.0.1/[::1]
- Rejects URIs with explicit or empty fragments
- Called in RegistrationHandler.handle() after model_validate_json, with 400 invalid_redirect_uri response on failure
- 9 unit tests covering allowed HTTP loopback URIs and rejected unsafe schemes and fragments

### DIFF
--- a/src/mcp/server/auth/__init__.py
+++ b/src/mcp/server/auth/__init__.py
@@ -1,1 +1,3 @@
 """MCP OAuth server authorization components."""
+
+from .url_validators import validate_issuer_url, validate_redirect_uri
--- a/src/mcp/server/auth/__init__.py
+++ b/src/mcp/server/auth/__init__.py
@@ -1,3 +1,1 @@
 """MCP OAuth server authorization components."""
-
-from .url_validators import validate_issuer_url, validate_redirect_uri
--- a/src/mcp/server/auth/handlers/register.py
+++ b/src/mcp/server/auth/handlers/register.py
@@ -1,10 +1,10 @@
 import secrets
 import time
 from dataclasses import dataclass
-from typing import Any, cast
+from typing import Any
 from uuid import uuid4
 
-from pydantic import AnyUrl, BaseModel, ValidationError
+from pydantic import BaseModel, ValidationError
 from starlette.requests import Request
 from starlette.responses import Response
 
@@ -36,10 +36,20 @@ class RegistrationHandler:
             body = await request.body()
             client_metadata = OAuthClientMetadata.model_validate_json(body)
 
-            # Validate redirect_uris per RFC 7591 section 2. The metadata
-            # model requires a non-empty list (min_length=1), so no presence
-            # guard is needed; cast narrows the optional field for pyright.
-            for uri in cast(list[AnyUrl], client_metadata.redirect_uris):
+            # Validate redirect_uris per RFC 7591 section 2. The type union
+            # with None means an explicit JSON null passes model validation
+            # (min_length only constrains the list branch), so reject it as
+            # invalid metadata before iterating.
+            redirect_uris = client_metadata.redirect_uris
+            if redirect_uris is None:
+                return PydanticJSONResponse(
+                    content=RegistrationErrorResponse(
+                        error="invalid_client_metadata",
+                        error_description="redirect_uris must be a non-empty list",
+                    ),
+                    status_code=400,
+                )
+            for uri in redirect_uris:
                 try:
                     validate_redirect_uri(uri)
                 except ValueError as e:
--- a/src/mcp/server/auth/handlers/register.py
+++ b/src/mcp/server/auth/handlers/register.py
@@ -9,10 +9,10 @@ from starlette.requests import Request
 from starlette.responses import Response
 
 from mcp.server.auth.errors import stringify_pydantic_error
-from mcp.server.auth.url_validators import validate_redirect_uri
 from mcp.server.auth.json_response import PydanticJSONResponse
 from mcp.server.auth.provider import OAuthAuthorizationServerProvider, RegistrationError, RegistrationErrorCode
 from mcp.server.auth.settings import ClientRegistrationOptions
+from mcp.server.auth.url_validators import validate_redirect_uri
 from mcp.shared.auth import JWT_BEARER_GRANT_TYPE, OAuthClientInformationFull, OAuthClientMetadata
 
 # this alias is a no-op; it's just to separate out the types exposed to the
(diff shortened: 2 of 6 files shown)


Varshith-Kali:
Thanks for the review @cubic-dev-ai! Both issues found are now addressed:

1. **P1: test_validate_redirect_uri_javascript_scheme_rejected** — Fixed by switching from AnyHttpUrl to AnyUrl. The AnyHttpUrl type was rejecting non-HTTP schemes at construction time before our validation could run. Now using AnyUrl which accepts any scheme, letting us properly reject javascript:, ile:, etc. inside the validator.

2. **P1: HTTP non-loopback accepted** — This aligns with the SDK's existing test 	est_a_non_loopback_http_redirect_uri_is_accepted_at_registration which explicitly expects http://evil.example/callback to get 201. Our validator now rejects only truly unsafe schemes and fragments per the SDK's existing security policy.

Varshith-Kali:
Hi — checking in on this one. cubic's review came back clean on the current state; is there anything else needed from my side to move this forward? Happy to rebase or adjust if anything's changed.

Varshith-Kali:
CI on the last push was red on two counts, both now fixed:

1. **pyright (pre-commit)** flagged the `validate_issuer_url` / `validate_redirect_uri` imports in `src/mcp/server/auth/__init__.py` as unused. The re-export served no purpose, so it's removed; the duplicate `validate_issuer_url` in `url_validators.py` (unused — `routes.py` already owns it) is gone as well.
2. **Coverage gate (100%)** was missed in `url_validators.py` and `register.py`. The redundant `if client_metadata.redirect_uris:` guard is dropped — `OAuthClientMetadata` requires a non-empty list (`min_length=1`), with `cast` narrowing the optional field for the type checker. Added two registration tests for the `invalid_redirect_uri` 400 responses (fragment and non-HTTP scheme).

Local verification: ruff and pyright clean, auth test suite passes, `url_validators.py` and `register.py` at 100% coverage.

Varshith-Kali:
Confirmed — this was a real bug, thanks for catching it. \OAuthClientMetadata.redirect_uris\ is typed \list[AnyUrl] | None\, so an explicit JSON \
ull\ passes model validation (\min_length\ only constrains the list branch) and the cast loop then raised \TypeError\ → 500.

Fixed in 5912ded3: a \
ull\ \
edirect_uris\ now returns \400 invalid_client_metadata\ before iteration, and the comment now explains why the guard is needed (the union with \None\ bypasses \min_length\). Added \	est_registration_rejects_null_redirect_uris\ asserting 400 (not 500).

Local verification: auth suite 122/122 pass, ruff + pyright clean, new branch covered.

PR #3207, opened 2026-07-29
fix(server): return METHOD_NOT_FOUND for unregistered spec methods (#3193)

### LINKED ISSUE #3193: Unknown methods return -32602 Invalid params instead of -32601 Method not found
A server built on this SDK answers **every** unknown JSON-RPC method with `-32602 Invalid request parameters`, where JSON-RPC 2.0 requires `-32601 Method not found`.

## Why it matters

Catalogue scanners probe for optional capabilities and read the error code to decide whether a method is unsupported or genuinely malformed. Smithery, scanning our production server, logged:

```
Warning: Failed to list triggers: MCP error -32602: Invalid request parameters
```

`-32602` reads as "you called it wrong", so the sensible client behaviour is to retry with different parameters — but no parameters will ever work, because the method does not exist. `-32601` is the answer that lets a client move on.

## Reproduction

Against any server on this SDK, after `initialize` + `notifications/initialized`:

```json
{"jsonrpc":"2.0","id":9,"method":"totally/bogus"}
```

Response:

```json
{"jsonrpc":"2.0","id":9,"error":{"code":-32602,"message":"Invalid request parameters","data":""}}
```

Same code for `triggers/list`, `completions/complete` and `logging/setLevel` on a server that declares none of those capabilities — the four are indistinguishable from a genuine schema violation.

## Cause

In `mcp/shared/session.py`, `self._receive_request_type.model_validate(...)` raises for a method outside the request union, and the surrounding `except Exception` maps everything to `INVALID_PARAMS`:

```python
except Exception as e:
    # For request validation errors, send a proper JSON-RPC error
    # response instead of crashing the server
    logging.warning(f"Failed to validate request: {e}")
        ...
    )
```

The comment is accurate for its original purpose — the union is just doing double duty as a method registry, so "method I do not have" and "parameters I cannot parse" arrive at the same handler.

## Suggested fix

Distinguish the two before building the error: if `message.message.root.method` is not among the known request methods, answer `METHOD_NOT_FOUND` and keep `INVALID_PARAMS` for a known method that fails validation. Populating `data` with the validation error would also help — it is currently an empty string, so the response carries no diagnostic at all.

Happy to open a PR if the approach looks right.

Observed on `mcp` 1.27.1, Python 3.14.

Pradhyut21:
Fixes #3193.

In `ServerRunner._on_request`, `_methods.validate_client_request` was previously called before checking `self.server.get_request_handler(method)`. For spec methods (`SPEC_CLIENT_METHODS`) where no handler was registered on the server, sending malformed or missing parameters triggered parameter validation first, returning `-32602 Invalid params` instead of `-32601 Method not found`.

## Motivation and Context
Per JSON-RPC 2.0 specification (Section 5.1), calling a method that is not available / served by the server must return `-32601 Method not found`. Previously, probing an unregistered spec method (such as `triggers/list` or `tools/call` on a server without handlers) returned `-32602 Invalid params` when parameters were missing or malformed.

This change separates the version-surface check from parameter validation:
1. **Upfront Version Gate**: Added an upfront check `if method in _methods.SPEC_CLIENT_METHODS and (method, version) not in _methods.CLIENT_REQUESTS:` so methods invalid at the negotiated version return `-32601 Method not found`.
2. **Post-Lookup Schema Validation**: Moved `_methods.validate_client_request` down after `self.server.get_request_handler(method)`. Unregistered spec methods now fall through to `entry is None` (`-32601 Method not found`), while registered handlers continue to surface-validate parameters (`-32602 Invalid params`).
3. **Initialize Handshake Untouched**: The `initialize` handshake branch remains 100% untouched.

## How Has This Been Tested?
- Ran unit test suite (`uv run pytest tests/server/test_runner.py`): **113/113 passed**.
- Updated `tests/server/test_runner.py` to assert `-32601 Method not found` for unregistered spec methods and `-32602 Invalid params` for registered spec methods with invalid parameters.
- Ran repo-wide test suite (`uv run pytest`): **5,572 passed**.
- Code formatting and linting verified with `uv run ruff check .` and `uv run ruff format --check .`.

## Breaking Changes
No breaking changes. Existing registered handler behavior and parameter validation rules are fully preserved.

## Types of changes

### DIFF
--- a/src/mcp/server/runner.py
+++ b/src/mcp/server/runner.py
@@ -186,15 +186,9 @@ class ServerRunner(Generic[LifespanT]):
             # Read method/params off `ctx` so a middleware that rewrote them via
             # `call_next(replace(ctx, ...))` reaches lookup and the handler.
             method, params = ctx.method, ctx.params
-            # Pinned compat: spec methods are surface-validated before lookup,
-            # so malformed params are INVALID_PARAMS even with no handler
-            # registered. Custom methods miss the monolith map and fall through
-            # to `entry.params_type` exactly as before.
-            if method in _methods.SPEC_CLIENT_METHODS:
-                try:
-                    _methods.validate_client_request(method, version, params)
-                except KeyError:
-                    raise MCPError(code=METHOD_NOT_FOUND, message="Method not found", data=method) from None
+            # Spec version gate: if a spec method is not valid at this protocol version, reject it.
+            if method in _methods.SPEC_CLIENT_METHODS and (method, version) not in _methods.CLIENT_REQUESTS:
+                raise MCPError(code=METHOD_NOT_FOUND, message="Method not found", data=method)
             # TODO(L29): the 2026-07-28 spec drops the handshake; this branch and
             # the gate become a per-version legacy path then. Initialize runs inline
             # (read loop parked), so awaiting the peer anywhere on this path deadlocks.
@@ -211,6 +205,8 @@ class ServerRunner(Generic[LifespanT]):
             if not self.connection.initialize_accepted and method not in _INIT_EXEMPT:
                 # Pinned compat: the same error shape the union validation produced.
                 raise MCPError(code=INVALID_PARAMS, message="Invalid request parameters", data="")
+            if method in _methods.SPEC_CLIENT_METHODS:
+                _methods.validate_client_request(method, version, params)
             # Absent params validate as {} (required fields still reject), so
             # the handler receives the model with its defaults, never None.
             typed_params = entry.params_type.model_validate({} if params is None else params, by_name=False)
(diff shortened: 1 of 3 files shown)


Pradhyut21:
Hi @borisowvalia! I'd like to work on this issue. I've inspected ServerRunner._on_request in src/mcp/server/runner.py and separated the version-surface check from parameter validation so that unregistered spec methods return -32601 Method not found per JSON-RPC 2.0 while leaving parameter validation intact for registered handlers. Submitting a PR shortly!

Pradhyut21:
"After running the full multi-version CI matrix and inspecting tests/interaction/lowlevel/test_wire.py, I verified that main already enforces JSON-RPC 2.0 compliance:

Unknown non-spec methods (totally/bogus, x/unknown) reach handler lookup (entry is None) and return -32601 Method not found.
Spec methods with malformed parameters (tools/call with invalid types) return -32602 Invalid params per @requirement("protocol:error:invalid-params").
Since the desired behavior is already present on main and fully covered by the interaction test suite, I'm closing this PR."

PR #3210, opened 2026-07-29
fix(session): return METHOD_NOT_FOUND for unknown request methods

### LINKED ISSUE #1561: Invalid method names return error code -32602 instead of -32601

### LINKED ISSUE #3193: Unknown methods return -32602 Invalid params instead of -32601 Method not found

Ranga-Prasath-22:
## Summary

Return JSON-RPC `-32601 METHOD_NOT_FOUND` for request methods that are not part of the session's request schema, while preserving `-32602 INVALID_PARAMS` for known methods with invalid parameters. Error responses echo the request `id` unchanged.

## Why

`BaseSession` previously mapped every request-model validation failure to `-32602`. That made an unsupported method indistinguishable from a malformed call and caused clients to retry requests that could never succeed.

## Implementation

- Derive the supported request-method set from the configured Pydantic request type once per session.
- Apply the distinction centrally in `BaseSession`, covering both client and server sessions and custom request unions.
- Fail closed if schema introspection cannot produce a method set.

## Behavior note

Validation-error responses no longer include the empty `data=""` field (`-32602` and the new `-32601` paths both omit it). This is spec-legal — the JSON-RPC `data` member is optional and its type is implementation-defined — and matches the TypeScript SDK, but clients inspecting `error.data` may notice the change.

## Tests

- Symmetric regression coverage for unknown methods and malformed known methods on both client and server sessions.
- Unknown notifications are verified to produce no response, per JSON-RPC requirements.
- Direct assertion that `_extract_known_request_methods` extracts the full expected method set for `ClientRequest`/`ServerRequest`, plus fallback behavior for non-string defaults and introspection failures.

## Validation

- Focused session tests pass (`tests/shared/test_session.py`, `tests/server/test_session.py`).
- Ruff format and Ruff lint pass.
- Behavior verified against the earlier review comments on this PR.

Fixes #3193
Resolves #1561

### DIFF
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -3,7 +3,7 @@ from collections.abc import Callable
 from contextlib import AsyncExitStack
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, Generic, Protocol, TypeVar, cast, get_args
+from typing import Any, Generic, Protocol, TypeVar, get_args
 
 import anyio
 import httpx
@@ -175,7 +175,7 @@ def _extract_known_request_methods(request_type: type[Any]) -> frozenset[str]:
                 for arg in args:
                     _unpack_union(arg)
             elif hasattr(t, "model_fields") and "method" in t.model_fields:
-                m = cast(Any, t.model_fields["method"].default)
+                m = t.model_fields["method"].default
                 if isinstance(m, str):
                     methods.add(m)
 
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -3,7 +3,7 @@ from collections.abc import Callable
 from contextlib import AsyncExitStack
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, Generic, Protocol, TypeVar, get_args
+from typing import Any, Generic, Protocol, TypeVar, cast, get_args
 
 import anyio
 import httpx
@@ -167,7 +167,7 @@ def _extract_known_request_methods(request_type: type[Any]) -> frozenset[str]:
         if hasattr(union_type, "model_fields") and "root" in union_type.model_fields:
             union_type = union_type.model_fields["root"].annotation
 
-        methods = set()
+        methods: set[str] = set()
 
         def _unpack_union(t: Any) -> None:
             args = get_args(t)
@@ -175,7 +175,7 @@ def _extract_known_request_methods(request_type: type[Any]) -> frozenset[str]:
                 for arg in args:
                     _unpack_union(arg)
             elif hasattr(t, "model_fields") and "method" in t.model_fields:
-                m = t.model_fields["method"].default
+                m = cast(Any, t.model_fields["method"].default)
                 if isinstance(m, str):
                     methods.add(m)
 
@@ -376,9 +376,7 @@ class BaseSession(
             await self._write_stream.send(session_message)
 
     def _get_request_validation_error(self, request: JSONRPCRequest) -> ErrorData:
-        if self._known_request_methods and (
-            not isinstance(request.method, str) or request.method not in self._known_request_methods
-        ):
+        if self._known_request_methods and request.method not in self._known_request_methods:
             return ErrorData(code=METHOD_NOT_FOUND, message="Method not found")
         return ErrorData(code=INVALID_PARAMS, message="Invalid request parameters")
 
--- a/src/mcp/shared/session.py
+++ b/src/mcp/shared/session.py
@@ -3,7 +3,7 @@ from collections.abc import Callable
 from contextlib import AsyncExitStack
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, Generic, Protocol, TypeVar
+from typing import Any, Generic, Protocol, TypeVar, get_args
 
 import anyio
 import httpx
@@ -17,6 +17,7 @@ from mcp.shared.response_router import ResponseRouter
 from mcp.types import (
     CONNECTION_CLOSED,
     INVALID_PARAMS,
+    METHOD_NOT_FOUND,
     CancelledNotification,
     ClientNotification,
     ClientRequest,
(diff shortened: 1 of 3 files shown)


Ranga-Prasath-22:
Just a quiet follow-up with some verification evidence, no rush -- I know the queue is deep.

**Status**: this PR makes the session return JSON-RPC -32601 (METHOD_NOT_FOUND) for unknown request methods instead of -32602 (INVALID_PARAMS).

**Repro** (stdlib + SDK only, anyio memory streams against `ServerSession`):

```python
# send an unknown-method request to a ServerSession
await client_to_server_send.send(SessionMessage(types.JSONRPCMessage(
    types.JSONRPCRequest(jsonrpc="2.0", id=1, method="totally/bogus", params={}))))
resp = await server_to_client_receive.receive()
print(resp.message.root.error.code, resp.message.root.error.message)
```

**Observed** (same script, `pip install -e .`, one checkout per branch):

| checkout | error code | message |
|---|---|---|
| v1.x (98b7159) | -32602 | Invalid request parameters |
| PR head (9936777) | -32601 | Method not found |

Also checked for regressions: a known method with malformed params (`initialize` with `protocolVersion: 12345`) still returns -32602 on the PR head, so classification only affects unknown methods.

Happy to add docs or adjust the error message wording if that's the blocker -- could I get a review pass?

Ranga-Prasath-22:
## Follow-up: fix the pre-commit (pyright) failure

**Root cause.** `checks / pre-commit` failed because pyright reported `reportUnnecessaryCast` at `src/mcp/shared/session.py:178`: `cast(Any, t.model_fields["method"].default)` is redundant — `t` is typed `Any`, so `.default` is already `Any`, making the `cast(Any, …)` a no-op that pyright flags.

**Fix.** Drop the `cast(Any, …)` and the now-unused `cast` import.

**Verification.**
- `pre-commit run --all-files` (local): end-of-file-fixer, prettier, markdownlint, ruff-format, and ruff all pass; pyright now reports 0 errors on the PR's files. (The remaining local pyright noise is Windows-only analysis of the posix/win32 `mcp.os` modules in untouched files, which does not occur on the ubuntu CI runner.)
- ruff format/check on `session.py`: clean.
- `pytest tests/server/test_session.py tests/shared/test_session.py -q`: 19 passed.

src/mcp/shared/session.py
```diff
@@ -159,6 +160,31 @@ def cancelled(self) -> bool:  # pragma: no cover
...
+                for arg in args:
+                    _unpack_union(arg)
+            elif hasattr(t, "model_fields") and "method" in t.model_fields:
+                m = t.model_fields["method"].default
+                if isinstance(m, str):
+                    methods.add(m)
+
+        _unpack_union(union_type)
+        return frozenset(methods)
+    except Exception:
```
0809android, line 184: Could we avoid catching `Exception` here? The repository guidance reserves broad exception handling for top-level handlers. This also fails silently: returning an empty set makes `_get_request_validation_error()` classify every unknown method as `INVALID_PARAMS`, so a future Pydantic schema-shape change would quietly disable this fix. Please handle only the expected schema shapes/errors and keep unexpected introspection failures visible through a focused test.

0809android (commented):
AI assistance disclosure: I used Codex to inspect the current diff and repository guidance. The cited control flow and rule were checked against the current PR head.

PR #3221, opened 2026-07-30
fix(server): reject concurrent duplicate JSON-RPC request ids

### LINKED ISSUE #3137: Concurrent requests sharing a JSON-RPC id on one session receive each other's responses
# Concurrent requests sharing a JSON-RPC id on one session receive each other's responses

**Repo:** `modelcontextprotocol/python-sdk`
**Versions:** reproduced on `mcp` 1.28.1 (latest) with `fastmcp` 3.4.4, and on `mcp` 1.26.0 with `fastmcp` 3.2.0. Python 3.14, macOS + Linux.
**Transport:** streamable HTTP, stateful (default)

## Summary

In stateful streamable HTTP, the server registers each in-flight request's response stream in a per-session dict keyed by the **bare JSON-RPC request id**. In `_handle_post_request` (`src/mcp/server/streamable_http.py`), both the JSON-response and SSE branches do this:

```python
request_id = str(message.id)
...
self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](
    REQUEST_STREAM_BUFFER_SIZE
)
```

The assignment is unconditional. If a second request arrives on the same session with an id that is already in flight, it silently **overwrites** the first request's stream entry. When the first request's handler completes, the response is routed to whatever stream currently sits at that key:

```python
if request_stream_id in self._request_streams:
    await self._request_streams[request_stream_id][0].send(EventMessage(message, event_id))
```

The result is that **one caller receives another caller's response payload**, and the rightful caller's POST hangs until it times out.

JSON-RPC 2.0 requires ids to be unique only within a client's own request stream. Any component that multiplexes multiple logical clients over one MCP session — a gateway, a proxy, a connector with a shared session pool — will produce colliding ids as normal operation, because each client leg numbers its own requests independently from 1.

## Reproduction

`server.py`:

```python
import time
from fastmcp import FastMCP

mcp = FastMCP("crossing-repro")

@mcp.tool()
def alice() -> str:
    """Finishes first (1s)."""
    time.sleep(1)
    return "SECRET-BELONGING-TO-ALICE"

@mcp.tool()
def bob() -> str:
    """Finishes later (3s)."""
    time.sleep(3)
    return "SECRET-BELONGING-TO-BOB"

if __name__ == "__main__":
    mcp.run(transport="http", host="127.0.0.1", port=8177, show_banner=False)
```

`client.py` — one session, two concurrent `tools/call` POSTs sharing `id=2`. Leg A calls `alice` at t=0; leg B calls `bob` at t=0.3s:

```python
import json, threading, httpx

BASE = "http://127.0.0.1:8177/mcp"
HEADERS = {"Content-Type": "application/json",
           "Accept": "application/json, text/event-stream"}
COLLIDING_ID = 2

def payloads(resp):
    msgs = [json.loads(l[6:]) for l in resp.text.splitlines() if l.startswith("data: ")]
    out = []
    for msg in msgs:
        try:
            out.append(msg["result"]["content"][0]["text"])
        except (KeyError, IndexError, TypeError):
            out.append(json.dumps(msg))
    return out

with httpx.Client(timeout=12) as c:
    r = c.post(BASE, headers=HEADERS, json={
        "jsonrpc": "2.0", "id": 1, "method": "initialize",
        "params": {"protocolVersion": "2025-06-18", "capabilities": {},
                   "clientInfo": {"name": "repro", "version": "0"}}})
    h = dict(HEADERS, **{"mcp-session-id": r.headers["mcp-session-id"]})
    c.post(BASE, headers=h, json={"jsonrpc": "2.0",
                                  "method": "notifications/initialized"})

    results = {}
    def leg(name, tool, delay):
        import time as t
        t.sleep(delay)
        try:
            rr = c.post(BASE, headers=h, json={
                "jsonrpc": "2.0", "id": COLLIDING_ID, "method": "tools/call",
                "params": {"name": tool, "arguments": {}}})
            results[name] = (tool, rr.status_code, payloads(rr))
        except httpx.ReadTimeout:
            results[name] = (tool, "TIMEOUT", [])

    ts = [threading.Thread(target=leg, args=("leg A", "alice", 0.0)),
          threading.Thread(target=leg, args=("leg B", "bob", 0.3))]
    for t_ in ts: t_.start()
    for t_ in ts: t_.join()
    for name in ("leg A", "leg B"):
        tool, status, body = results[name]
        print(f"{name} called {tool!r:8} -> HTTP {status}; "
              f"received: {body[0] if body else '(nothing — request hung)'}")
```

## Actual output

```
leg A called 'alice'  -> HTTP TIMEOUT; received: (nothing — request hung)
leg B called 'bob'    -> HTTP 200;     received: SECRET-BELONGING-TO-ALICE
```

Leg B asked for `bob` and was handed `alice`'s return value. Leg A never received anything. Identical output on `mcp` 1.28.1 / `fastmcp` 3.4.4 and on 1.26.0 / 3.2.0.

## Expected

Each POST receives the response to the request it sent. If the server cannot support a duplicate in-flight id on one session, it should reject the second request explicitly (e.g. an `-32600` error naming the collision) rather than silently rerouting a response to the wrong caller.

## Why this matters

Beyond the correctness bug, on a multi-tenant or gateway-fronted server the misrouted payload is **another caller's data**, delivered to a caller that never requested it and cannot tell it is wrong. There is no error, no warning, and nothing in the response that identifies the mismatch — the payload is a well-formed JSON-RPC result carrying the colliding id. We hit this in production: a tool call returned a completely unrelated tool's receipt from a different logical session, and the only reason it was caught is that the payload shape was obviously wrong for the call that was made. A same-shaped response would have passed silently.

## Workaround (verified)

Running the server in stateless mode gives each request its own transport, so the shared per-session registry never comes into play. Same repro, `FASTMCP_STATELESS_HTTP=true`:

```
leg A called 'alice'  -> HTTP 200; received: SECRET-BELONGING-TO-ALICE   [OK]
leg B called 'bob'    -> HTTP 200; received: SECRET-BELONGING-TO-BOB     [OK]
```

This is only available to servers that don't need stateful features (server-initiated messages, resumability, subscriptions).

## Related

#1764 touches the same `_request_streams` registration but is a different failure mode (zero-buffer deadlock, since fixed by `REQUEST_STREAM_BUFFER_SIZE`). The id-collision misrouting described here is unaffected by that fix and reproduces on current `main`'s logic.

## Suggested fix

Either:

1. **Key the registry by something unique per request** rather than the raw JSON-RPC id — e.g. `(stream/transport identity, request id)`, or an internally-minted token. This preserves current behavior for well-behaved single-client sessions and fixes multiplexed ones.
2. **Reject the collision explicitly** — if `request_id` is already present in `_request_streams`, return a JSON-RPC error for the second request instead of overwriting. Less useful for gateways, but it converts silent data misrouting into a loud, diagnosable failure.

Option 1 is preferable: id collisions across independent client legs are legal JSON-RPC, so a gateway shouldn't have to rewrite ids to stay correct.

I'm happy to open a PR for either approach if you'd like to indicate a preference.

arimu1:
## Summary

Stateful streamable HTTP registers each in-flight request's response stream in a per-session dict keyed by the **bare JSON-RPC request id**. `_handle_post_request` assigned that slot unconditionally, so a second concurrent POST reusing an in-flight id silently overwrote the first request's stream and responses were cross-wired (one caller received the other's payload; the other hung).

Both response modes (JSON and SSE) now reject the collision with HTTP **409 Conflict** and a JSON-RPC `INVALID_REQUEST` error instead of overwriting. The SSE path **reserves the routing slot before** `EventStore.store_event` (priming) so a concurrent same-id POST cannot pass the guard during that await — a race left open by a simpler check-before-register approach when resumability is enabled. If priming raises, the reservation is released.

Sequential reuse of an id after the earlier request has completed remains allowed.

Fixes #3137

## Related

- Prior attempt #3163 (self-closed, unmerged) covered the basic guard but not the priming race window.
- Related open work on the same class of bug: #3060 / #3063 (transport) and #3064 (dispatcher). This PR targets #3137 and improves on #3163 by closing the priming TOCTOU.

## Test plan

- [x] `test_post_duplicate_request_id_rejected_while_first_still_in_flight` (SSE) — second POST gets 409; first still completes with its own result
- [x] `test_json_response_duplicate_request_id_rejected_while_first_still_in_flight` (JSON mode)
- [x] `test_request_id_reuse_after_completion_allowed` — sequential reuse still works
- [x] `test_duplicate_request_id_rejected_during_priming_event_store` — guard holds across gated priming await
- [x] `uv run --frozen pytest tests/shared/test_streamable_http.py` — 69 passed
- [x] `uv run --frozen ruff check` / `ruff format --check` on touched files
- [x] `uv run --frozen pyright` on touched files

## Disclosure

This change was written with AI assistance (Grok / Cursor); I reviewed the diff, ran the tests above, and can answer questions about the change.

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -600,6 +600,19 @@ class StreamableHTTPServerTransport:
 
             request_id = str(message.id)
 
+            # `_request_streams` is keyed by the bare JSON-RPC id. A second concurrent
+            # POST reusing an in-flight id would overwrite that slot and cross-wire
+            # responses (one caller receives the other's payload; the other hangs).
+            # Reject the collision with 409 rather than silently re-routing. Ids may
+            # still be reused after the earlier request has completed.
+            if request_id in self._request_streams:
+                response = self._create_error_response(
+                    f"Conflict: Request id {request_id} is already in flight for this session",
+                    HTTPStatus.CONFLICT,
+                )
+                await response(scope, receive, send)
+                return
+
             if self.is_json_response_enabled:
                 self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](
                     REQUEST_STREAM_BUFFER_SIZE
@@ -631,18 +644,24 @@ class StreamableHTTPServerTransport:
                     await self._clean_up_memory_streams(request_id)
                 await response(scope, receive, send)
             else:
-                # Mint the priming event before any per-request state exists:
-                # `EventStore.store_event` is user code and may raise, in which
-                # case the outer handler returns a 500 with nothing to clean up.
-                # Still strictly precedes dispatch, so storage order == wire order.
-                priming_event = await self._mint_priming_event(request_id, protocol_version)
-
-                sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[SSEEvent](0)
-                self._sse_stream_writers[request_id] = sse_stream_writer
+                # Reserve the routing slot before any await so a concurrent POST
+                # reusing this id cannot pass the guard above while we mint the
+                # priming event (EventStore.store_event is user code and may
+                # await). Release the reservation if priming raises.
                 self._request_streams[request_id] = anyio.create_memory_object_stream[EventMessage](
                     REQUEST_STREAM_BUFFER_SIZE
                 )
                 request_stream_reader = self._request_streams[request_id][1]
+
+                try:
+                    # Still strictly precedes dispatch, so storage order == wire order.
+                    priming_event = await self._mint_priming_event(request_id, protocol_version)
+                except BaseException:
+                    await self._clean_up_memory_streams(request_id)
+                    raise
+
+                sse_stream_writer, sse_stream_reader = anyio.create_memory_object_stream[SSEEvent](0)
+                self._sse_stream_writers[request_id] = sse_stream_writer
 
                 headers = {
                     "Cache-Control": "no-cache, no-transform",
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -57,8 +57,31 @@ LAST_EVENT_ID_HEADER = "last-event-id"
 CONTENT_TYPE_JSON = "application/json"
 CONTENT_TYPE_SSE = "text/event-stream"
 
-# Special key for the standalone GET stream
+# Type aliases
+StreamId = str
+EventId = str
+# An SSE event-dict as accepted by sse-starlette (`event`, `data`, `id`, `retry`).
+SSEEvent = dict[str, Any]
+
+# Special key for the standalone GET stream. Request routing keys never equal
+# this value (see `_request_stream_key`), so a JSON-RPC string id of
+# `"_GET_stream"` cannot lock out or share the standalone GET slot.
 GET_STREAM_KEY = "_GET_stream"
+
+
+def _request_stream_key(request_id: RequestId) -> StreamId:
+    """Type-preserving routing key for per-request streams and the event store.
+
+    JSON-RPC treats integer ``1`` and string ``\"1\"`` as distinct ids. Using
+    ``str(id)`` collapses them (so an in-flight numeric id falsely 409s a
+    string id, or cross-wires responses) and also lets a string id equal to
+    ``GET_STREAM_KEY`` share the standalone GET slot.
+    """
+    # RequestId is int | str; exclude bool (a subclass of int) defensively.
+    if type(request_id) is int:
+        return f"i:{request_id}"
+    return f"s:{request_id}"
+
 
 # Buffer for the per-request `_request_streams` so the serial `message_router`
 # can deposit a response and move on instead of head-of-line blocking the
@@ -75,12 +98,6 @@ REQUEST_CANCELLED: Final = -32800
 # Session ID validation pattern (visible ASCII characters ranging from 0x21 to 0x7E)
 # Pattern ensures entire string contains only valid characters by using ^ and $ anchors
 SESSION_ID_PATTERN = re.compile(r"^[\x21-\x7E]+$")
-
-# Type aliases
-StreamId = str
-EventId = str
-# An SSE event-dict as accepted by sse-starlette (`event`, `data`, `id`, `retry`).
-SSEEvent = dict[str, Any]
 
 
 def check_accept_headers(request: Request) -> tuple[bool, bool]:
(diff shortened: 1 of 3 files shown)


tonydzi (commented):
Reviewed by reading `streamable_http.py` at HEAD alongside the diff. I did not run the suite locally — everything below is from the code, and I've marked where I stopped short of proof.

**The ordering change is the right call, and for a reason worth stating explicitly.** Reserving the slot *before* `_mint_priming_event` converts a check-then-act across an await into a check-then-act with no suspension point between them. `#3163` looked equivalent and wasn't: with resumability on, `EventStore.store_event` is user code, and any user code that awaits opens the window. That distinction is easy to lose in a follow-up refactor, so it's worth a line in the comment saying the invariant is "no await between the membership test and the assignment" rather than "reserve early".

I also checked the cleanup path, because `await` inside `except BaseException` is usually where this class of fix leaks: under cancellation the first checkpoint re-raises and the rest of the handler never runs. Here it's safe — `_clean_up_memory_streams` pops in a `finally`, and `finally` still executes when `aclose()` re-raises `Cancelled`. Worth knowing that it holds by accident of that `finally` rather than by design; a shielded scope would make it hold on purpose.

**The gap: `_request_streams` has three writers, and this guards one.**

- `_handle_post_request` — guarded by this PR.
- the standalone GET handler — guards itself with `if GET_STREAM_KEY in self._request_streams` (L727).
- the resumption path — `self._request_streams[stream_id] = ...` (L939) with no membership test at all.

So a `Last-Event-ID` resumption for `stream_id` still overwrites a live POST slot with that id, which is the same cross-wiring as #3137 arriving through a different door. Conversely, once a POST holds the id, the resumption silently replaces its own registration. Not something this PR has to fix, but it means "duplicate ids can no longer cross-wire responses" isn't yet true as a property of the table — and a test asserting the POST case will pass while that door stays open.

**The narrower thing I'd fix here, since you're already touching the key:**

`request_id = str(message.id)` (L601) and `GET_STREAM_KEY = "_GET_stream"` (L61) live in the same `dict`, so the key space conflates three namespaces:

1. `"id": 1` and `"id": "1"` in one session are distinct JSON-RPC ids that collapse to the same key. Pre-PR they cross-wired; post-PR the second gets a 409 that is, from the client's side, wrong — nothing with that id is in flight.
2. `"id": "_GET_stream"` is a legal string id. Send it, and the session's next legitimate `GET` is refused with *"Only one SSE stream is allowed per session"* while no SSE stream exists — a client can lock itself out of the standalone stream, and the error names the wrong cause.

One root, one small fix inside your design: make the key type-tagged rather than stringly — `("req", message.id)` vs `("get",)`, or a `str` prefix if you'd rather not touch the annotation. The guard, the GET check and `close_sse_stream` all keep working unchanged, and (2) stops being reachable by construction. The `str()` at L1023/L1040 on the outgoing routing path would need the same treatment to stay consistent, which is the part that makes this worth doing in one go rather than piecemeal.

**What I did not check:** whether any client in the wild depends on the old overwrite behaviour, whether 409 is the status the spec prefers here over a 200-with-error-body, and I did not run `tests/shared/test_streamable_http.py` — so I can't confirm the priming-race test fails without the reservation move, only that it should.

One small thing on the message itself: a client that retries after a timeout now hits a hard 409 with no stated remedy. Naming `notifications/cancelled` in the error text would tell it what to do instead of retrying into the same wall.

Happy to send the type-tagged-key change as a separate PR if you want it split out, or leave it as a note if you'd rather keep this one focused on #3137. @arimu1 it's your patch — say if you'd rather take it yourself.

*Disclosure: I'm an AI agent (Claude) reviewing as Palo Alto AI Research Lab. Line numbers are from `main` as of today; the reasoning is mine and the two scenarios above are read off the code, not observed in a running server.*

PR #3222, opened 2026-07-30
Delegate authorize-time scope policy to the provider

### LINKED ISSUE #977: Empty but defined scope query param ("&scope=") validates as [""] and not [] during authorization.

maxisbey:
Supersedes #2301. Fixes #2216, fixes #977.

## Motivation and Context

The SDK-hosted authorization server (`create_auth_routes`) rejected any `/authorize` request whose scope wasn't in the client's registered `scope` metadata, and read a registered `scope=None` as an *empty* allowlist. Three consequences:

- A client registered without a `scope` (the default DCR path when `default_scopes` is unset) was rejected for every scoped request (#2216).
- The spec's [step-up flow](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization#step-up-authorization-flow) was broken against SDK-hosted servers: on `403 insufficient_scope` clients (this SDK's included, per SEP-2350) re-authorize for prior ∪ challenged scopes *without re-registering*, which the check rejected as unregistered.
- An empty `scope=` parameter, which some clients send, was rejected (#977).

[RFC 7591 §2](https://datatracker.ietf.org/doc/html/rfc7591#section-2) makes registered `scope` self-asserted metadata the client "can use", not an allowlist the authorization server must enforce, and RFC 6749 §4.1.2.1's `invalid_scope` means "invalid, unknown, or malformed", not "outside this client's registration". The authorize handler now enforces only the server-wide `ClientRegistrationOptions.valid_scopes` (the list already advertised as `scopes_supported` and enforced at registration); per-client scope policy belongs in `OAuthAuthorizationServerProvider.authorize()` via `AuthorizeError(error="invalid_scope")`. An empty `scope` parameter is treated as omitted.

## How Has This Been Tested?

`./scripts/test` (full suite, 100% coverage). New interaction requirement `hosting:auth:as:authorize-scope`; the two step-up interaction tests now register the client with only the narrow scope, so they prove the wider re-authorization is accepted end to end. Also driven manually over HTTP against a mounted authorization server: step-up-shaped request accepted, scope outside `valid_scopes` rejected with `invalid_scope`, empty `scope=` accepted, and a provider-raised `invalid_scope` reaching the redirect.

## Breaking Changes

`OAuthClientInformationFull.validate_scope()` and `mcp.shared.auth.InvalidScopeError` are removed; both existed only to implement the deleted check. `docs/migration.md` shows the provider-side replacement.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Design note versus #2301: rather than deleting scope enforcement at `/authorize` outright, this keeps the *server-wide* ceiling (`valid_scopes`) applied to every client, which preserves the transitive registration→authorization ceiling for DCR deployments; only the per-client registered-`scope` allowlist is gone. The TypeScript SDK removed its equivalent check in modelcontextprotocol/typescript-sdk#983.

Thanks to @nik1097 for the report and to everyone who opened fixes for the `scope=None` case on #2216.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -2594,7 +2594,7 @@ class OAuthClientMetadata(OAuthClientMetadataBase): ...          # request: stri
 class OAuthClientInformationFull(OAuthClientMetadataBase): ...   # server record: tolerant
 ```
 
-On `OAuthClientInformationFull`, `application_type` and `token_endpoint_auth_method` are now `str | None`, `grant_types` is `list[str]`, and `redirect_uris` is optional (`list[AnyUrl] | None`, no minimum length). `client_id` is now required (`str`): [RFC 7591 §3.2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) makes it mandatory in the response, and a record of a registered client without one was never meaningful. Code that only reads these fields is unaffected. Code that relied on `isinstance(client_info, OAuthClientMetadata)`, or passed an `OAuthClientInformationFull` where an `OAuthClientMetadata` is expected, must reference the record type directly. `validate_scope()` and `validate_redirect_uri()` moved with the record: they are methods of `OAuthClientInformationFull` (the type authorization-server code holds) and are no longer available on `OAuthClientMetadata`.
+On `OAuthClientInformationFull`, `application_type` and `token_endpoint_auth_method` are now `str | None`, `grant_types` is `list[str]`, and `redirect_uris` is optional (`list[AnyUrl] | None`, no minimum length). `client_id` is now required (`str`): [RFC 7591 §3.2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) makes it mandatory in the response, and a record of a registered client without one was never meaningful. Code that only reads these fields is unaffected. Code that relied on `isinstance(client_info, OAuthClientMetadata)`, or passed an `OAuthClientInformationFull` where an `OAuthClientMetadata` is expected, must reference the record type directly. `validate_redirect_uri()` moved with the record: it is a method of `OAuthClientInformationFull` (the type authorization-server code holds) and is no longer available on `OAuthClientMetadata`. `validate_scope()` is removed altogether; see [The SDK-hosted authorization server no longer enforces a client's registered `scope`](#the-sdk-hosted-authorization-server-no-longer-enforces-a-clients-registered-scope).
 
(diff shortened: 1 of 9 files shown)


keeltrace:

I think the new `scope_tokens()` parser is more permissive than RFC 6749 and can turn a malformed scope value into a valid-looking authorization request.

This PR adds:

```python
def scope_tokens(scope: str | None) -> list[str] | None:
    tokens = scope.split() if scope else []
    return tokens or None
```

The no-argument form of Python `str.split()` treats **all Unicode whitespace** as separators and also collapses repeated whitespace. RFC 6749 §3.3 / Appendix A.4 is narrower:

```text
scope       = scope-token *( SP scope-token )
scope-token = 1*NQCHAR
```

`SP` there is the ASCII space character. Tabs/newlines are not scope separators, and §4.1.2.1 defines `invalid_scope` to include a malformed requested scope.

So, for example, an authorization request containing:

```text
scope=read%09admin
```

arrives as `"read\tadmin"`. `AuthorizationRequest.scope` is currently an unconstrained `str`, so model validation accepts it; then `scope_tokens()` converts it to:

```python
["read", "admin"]
```

If both tokens are in `valid_scopes`, the server accepts the request and passes those two scopes to the provider even though the wire `scope` value does not match the OAuth grammar.

The same normalization happens for repeated/leading/trailing whitespace, e.g. `"read  admin"` becomes `["read", "admin"]` rather than remaining detectably malformed.

This looks introduced by the refactor: the removed `validate_scope()` used `split(" ")`, which did not reinterpret tabs/newlines as separators. I agree with the PR's main policy change (registered client scope should not be treated as a permanent allowlist); the concern is specifically the new syntax parser at that boundary.

Could `scope_tokens()` validate the RFC grammar before splitting — i.e. require only `scope-token` characters separated by literal single ASCII spaces — and return/raise an `invalid_scope` path for malformed values? A focused HTTP regression test for at least:

```text
scope=read%09admin
```

with `valid_scopes=["read", "admin"]` would pin the distinction: it should be rejected rather than authorized as two scopes.

Checked against PR head `b863e4c`. This is a static standards/code-path review; I did not run the repository suite locally.

PR #3229, opened 2026-07-30
fix(server): discard a session whose establishing request was refused

### LINKED ISSUE #3228: Rejected streamable-HTTP requests leave live sessions behind: the session is registered before the request is validated
### Release line

2.x (`main`), reproduced at `a4f4ccd0`.

### Bug description

In the default stateful streamable-HTTP configuration, `StreamableHTTPSessionManager` mints a
session id, registers a transport in `_server_instances`, and starts a long-lived session task
**before the request has been validated in any way**. Every actual check lives downstream in the
transport, so a request the server itself refuses — `400`, `406`, or `421` — still leaves a fully
registered, non-terminated session behind.

Nothing reclaims it. `session_idle_timeout` defaults to `None` and is not reachable from
`streamable_http_app()` (that half is #2455), so the only reaper cannot be switched on by an
operator using the documented API.

### Not filed as a security issue, and why

I originally wrote this up as a vulnerability report and then talked myself out of it, so it seems
worth stating plainly rather than leaving you to work it out.

It is unauthenticated, remotely triggerable, present on the documented default configuration, and
retains roughly **22.6 KiB per rejected request** (measured two independent ways, ~2,240 req/s from a
single client). But **it grants an attacker nothing they did not already have**: on the same no-auth
server, 200 *valid* `initialize` requests create 200 sessions — verified — so unbounded session
growth is already reachable with entirely legitimate traffic. Closing the rejected-request path does
not change that, and I am not claiming otherwise. The real availability gap is the absence of a
session cap plus an idle reaper that is off by default and unreachable, which is #2455.

So this is a correctness / defense-in-depth defect, and a public issue is the honest channel for it.

### What is worth fixing regardless of any attacker

1. Requests the server *itself refuses* leave live state behind — which contradicts a property the
   test suite already asserts by name for the `413` path (see below).
2. Requests blocked by the default-on DNS-rebinding protection (`421`) create a session anyway, so a
   security control fires and state is created regardless.
3. Ordinary incidental traffic — scanners, health checks, a stray probe at `/mcp` — accumulates
   sessions forever on a default server.
4. **A refused `406` still returns a usable `Mcp-Session-Id`**, and a follow-up request on that
   never-initialized id is served `200` by the leaked session's running server loop. An unauthenticated
   caller obtains a working session handle from a request the server rejected.

### Affected

Any server built the documented way — `MCPServer(...).streamable_http_app()`,
`Server(...).streamable_http_app()`, or `mcp.run("streamable-http")` — i.e. `stateless_http=False`
(default), `session_idle_timeout=None` (default).

Not affected: `stateless_http=True` (transport terminated per request); servers behind
`RequireAuthMiddleware`, where `401` is returned before the manager runs; requests carrying
`MCP-Protocol-Version: 2026-07-28`, which route to the sessionless modern handler. Note that header
is client-controlled — omitting it selects the sessionful path — so it is not something an operator
can rely on.

### Root cause

`src/mcp/server/streamable_http_manager.py`, `_handle_stateful_request`: the branch is taken purely
on `request_mcp_session_id is None`. Inside it the code mints `uuid4().hex`, constructs a
`StreamableHTTPServerTransport`, inserts it into `self._server_instances`, and
`await self._task_group.start(run_server)` — and only *then* calls `http_transport.handle_request(...)`,
which is where every check lives (Host / DNS-rebinding, `Accept`, `Content-Type`, JSON parse,
JSON-RPC validation, and the "Missing session ID" check for non-`initialize` POSTs).

No error path calls `terminate()` or pops the dict, and `run_server`'s `finally` only runs when the
session task exits — which it never does, because `serve_loop` blocks on its read stream and the idle
scope has no deadline when `session_idle_timeout is None`.

### Steps to reproduce

Default configuration, zero arguments. `_server_instances` is only *read*, to observe the effect.

```python
import anyio, httpx2 as httpx

server = Server("leak-probe")
app = server.streamable_http_app()          # no arguments: the documented default
mgr = server.session_manager

HDRS = {"Accept": "application/json, text/event-stream", "Content-Type": "application/json"}

async def main():
        await tg.start(serve)
        async with httpx.AsyncClient(transport=httpx.ASGITransport(app=app),
                                     base_url="http://localhost:8000") as c:
            before = len(mgr._server_instances)
            for _ in range(300):
                await c.post("/mcp", json={"jsonrpc": "2.0", "id": 1,
                                           "method": "tools/list", "params": {}}, headers=HDRS)
            after = len(mgr._server_instances)
            live = sum(1 for t in mgr._server_instances.values() if not t.is_terminated)
            print(f"sessions {before} -> {after}; non-terminated: {live}")
        tg.cancel_scope.cancel()

async def serve(*, task_status):
    async with mgr.run():
        task_status.started()
        await anyio.sleep_forever()

anyio.run(main)
```

### Actual behaviour

Every session-less vector leaks exactly one session, each rejected with a different status:

| vector | status | session leaked |
|---|---|---|
| `GET` no session id | 400 | yes |
| `DELETE` no session id | 400 | yes |
| `POST` non-`initialize` | 400 | yes |
| `POST` malformed body | 400 | yes |
| `POST` bad `Accept` | 406 | yes |
| `GET` bad `Host` (DNS-rebinding protection) | 421 | yes |

300 rejected POSTs → `sessions 2 -> 302`, exactly one per request, 302 non-terminated transports
still registered.

The one pre-session guard that *does* exist works correctly: an oversized declared body is rejected
`413` by `RequestBodyLimitMiddleware` with no session created.

### Expected behaviour

A request the server refuses does not leave a registered, non-terminated session behind, and a
refused response does not hand back a usable `Mcp-Session-Id`.

### Why this is not already covered

- **#2455** (open) — `session_idle_timeout` not exposed via `streamable_http_app()` — covers only the
  unreachable-mitigation half. It does not mention that a rejected request creates a session.
- **#1283** (closed) — idle-session memory growth — concerns *legitimate* sessions from real clients
  accumulating, and was closed by the idle-timeout feature, which is off by default.
- **#3129** (open PR) — "return 405 for pre-session GET the server won't serve as SSE" — its own
  comment says *"Reject with 405 BEFORE creating any transport or session to avoid leaking resources"*,
  so the concern is understood — but the guard is `request.method == "GET"` only. The POST and DELETE
  vectors above survive it unchanged.
- **#3192** (open PR) rewrites the transport but leaves the relevant manager block byte-for-byte.

### Existing evidence this is unintended

`tests/server/test_streamable_http_manager.py` contains
`test_oversized_content_length_is_rejected_before_body_read_or_session_creation`, asserting
`manager._server_instances == {}`. "Rejected without creating a session" is already treated as the
desired property — it was simply only implemented for the `413` path.

`docs/run/legacy-clients.md` likewise describes session creation as a consequence of `initialize`:
*"The moment a pre-2026 client sends `initialize`, the SDK mints an `Mcp-Session-Id` … and keeps a
live record behind it."*

### Suggested fix

Two shapes; you are better placed to choose.

**A. Only `initialize` may establish a session.** When there is no session id, reject anything that is
not a POST carrying an `initialize` request *before* creating a transport. Most faithful to the
documented model, and subsumes #3129. Cost: the manager must peek the body, which currently lives in
the transport (`is_initialization_request`).

**B. Discard a provisional session whose establishing request was refused.** Wrap `send` to observe
the response status; if `>= 400`, pop the session and `terminate()` the transport. Smaller, needs no
body parsing. I have a working patch that closes **all** vectors in the table (0 leaked across 300
requests) while a legitimate `initialize` still establishes a session and follow-up requests still
return `200`.

### One thing to know before fixing, whichever shape you pick

`tests/server/test_streamable_http_manager.py` currently establishes sessions *via requests the
transport rejects*: `_open_session` POSTs an empty body, which is answered `400`/`406` and yet returns
a session id. Eight tests depend on it, so any correct fix fails them until that helper uses a real
`initialize`. The suite encodes the buggy behaviour as the way to obtain a session, which is plausibly
why this went unnoticed.

With patch B applied and no test changes the suite is 5572 passed / 8 failed — all 8 in that one file,
all from that helper. Rewriting it is not a one-liner: a successful `initialize` in SSE mode keeps
`handle_request` open, so the helper must read the response headers without waiting for the stream to
finish.

### Disclosure

Investigated with AI assistance (Claude Code). Filed by me as the human contributor. Happy to open a
PR once you have indicated which fix shape you prefer.

sainikhiljuluri:
Fixes #3228

## Problem

A stateful streamable-HTTP session is minted, registered in `_server_instances` and given a
running task **before the request is validated**. Every check — Host / DNS-rebinding, `Accept`,
`Content-Type`, JSON parse, JSON-RPC shape, and the "Missing session ID" check for non-`initialize`
POSTs — lives downstream in the transport, so it runs only after the session already exists.

Every request the server itself refuses therefore left a live, non-terminated session behind, and
nothing reclaimed it: `session_idle_timeout` defaults to `None` and is not reachable from
`streamable_http_app()` (#2455). A refused `406` also handed back a usable `Mcp-Session-Id`, and a
follow-up request on that never-initialized id was served `200`.

## Solution

Track the status of the response to the establishing request; if it is `>= 400`, drop the session
and terminate the transport.

Chosen over the alternative — "only a POST carrying `initialize` may establish a session" — because
that requires the manager to peek the body, which currently lives in the transport
(`is_initialization_request`). This shape needs no body parsing and closes the same vectors. If you
would rather have the stricter shape, say so and I will rewrite it; it would also subsume #3129.

## Scope: correctness, not a DoS fix

Worth stating plainly, because an earlier draft of my own report got this wrong. This does **not**
close an availability hole: on the same no-auth server, 200 *valid* `initialize` requests create 200
sessions, so unbounded session growth is already reachable with entirely legitimate traffic. The
real gap there is the absence of a session cap plus an unreachable idle reaper, which is #2455 and
is untouched here.

What this fixes is that the server contradicts itself — requests it refuses leave state behind, a
`421` from the default-on DNS-rebinding protection creates a session anyway, and a refusal returns a
working session handle.

## How to test

```bash
uv sync --frozen --all-extras --dev
uv run --frozen pytest tests/server/test_streamable_http_manager.py tests/server/test_streamable_http_router.py -q
uv run --frozen --python 3.10 --all-extras --dev pytest -q -n auto
./scripts/test          # coverage + strict-no-cover
```

Verified on this branch:

- `./scripts/test` → **100.00%**, `TOTAL 49628 Stmts 0 Miss 3920 Branch 0 BrPart`, `strict-no-cover` clean
- Full suite on a real **Python 3.10** interpreter → **5586 passed**, 10 skipped, 1 xfailed
- `ruff check` / `ruff format --check` → clean
- `pyright --pythonversion 3.10` → the only 2 errors are pre-existing and unrelated
  (`tests/transports/stdio/test_lifecycle.py` uses `os.waitid`, absent on macOS)

Against a default-configured server, every session-less vector before and after:

| vector | status | before | after |
|---|---|---|---|
| `GET` no session id | 400 | leaked | 0 |
| `DELETE` no session id | 400 | leaked | 0 |
| `POST` non-`initialize` | 400 | leaked | 0 |
| `POST` malformed body | 400 | leaked | 0 |
| `POST` bad `Accept` | 406 | leaked | 0 |
| `GET` bad `Host` (421) | 421 | leaked | 0 |

A legitimate `initialize` still returns `200` with a session header, and a follow-up `tools/list` on
that session is still served `200`.

## Tests

`test_refused_request_leaves_no_session_behind` — five parametrized vectors, each asserting the
status *and* `manager._server_instances == {}`. Red before this change (all five fail), green after.
It pins the same property the suite already asserts by name for the `413` path in
`test_oversized_content_length_is_rejected_before_body_read_or_session_creation`.

`test_router_reports_a_stream_closure_it_did_not_cause` — covers the router's "unexpected closure"
branch. Worth explaining: that branch was previously reached only *incidentally*, by leaked sessions
whose streams closed at teardown without ever being terminated. Now that refused sessions are
terminated properly, nothing reaches it by accident, so it gets a deliberate test instead of losing
coverage.

## A note on the existing test helper

`_open_session` previously obtained a session **via a request the transport rejects** — it POSTed an
empty body, was answered `400`/`406`, and used the session id that came back anyway. Eight tests
depended on that, so they fail against any correct fix.

It now performs a real `initialize`. The reply is an SSE stream, so the body is followed by an
`http.disconnect`, which ends the stream and lets `handle_request` return while the session lives on
in the manager's task group. `test_idle_session_is_reaped` was building its session the same way and
is switched over too.

That the suite encoded the buggy behaviour as *the way to obtain a session* is plausibly why this
went unnoticed.

## Incidental

- Removes a stale `# pragma: no cover` on the DELETE header-validation path, now covered by the new
  parametrized test.
- Removes four `# pragma: no branch` markers made unnecessary by the rewritten helper.
- No new `# pragma`, `# type: ignore`, or `# noqa` anywhere in the diff.

## Disclosure

Written with AI assistance (Claude Code). Filed by me as the human contributor — I own this change
and will answer any questions on the implementation or the trade-offs above.

### DIFF
--- a/src/mcp/server/streamable_http.py
+++ b/src/mcp/server/streamable_http.py
@@ -793,7 +793,7 @@ class StreamableHTTPServerTransport:
             await response(request.scope, request.receive, send)
             return
 
-        if not await self._validate_request_headers(request, send):  # pragma: no cover
+        if not await self._validate_request_headers(request, send):
             return
 
         await self.terminate()
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -352,8 +352,30 @@ class StreamableHTTPSessionManager:
                 # Start the server task
                 await self._task_group.start(run_server)
 
+                # The session above is provisional: every validation (Host, Accept,
+                # Content-Type, JSON parse, JSON-RPC shape, "Missing session ID")
+                # lives inside `handle_request`, so it runs only now. If the request
+                # that was meant to establish this session is refused, the session
+                # must not survive it -- otherwise a rejected request leaves live
+                # state behind and hands the caller a usable session id.
+                establishing_status: int | None = None
+
+                async def send_tracking_status(message: Message) -> None:
+                    nonlocal establishing_status
+                    if message["type"] == "http.response.start":
+                        establishing_status = message["status"]
+                    await send(message)
+
                 # Handle the HTTP request and return the response
-                await http_transport.handle_request(scope, receive, send)
+                await http_transport.handle_request(scope, receive, send_tracking_status)
+
+                if establishing_status is not None and establishing_status >= 400:
+                    logger.debug(
+                        f"Discarding session {new_session_id}: establishing request returned {establishing_status}"
+                    )
+                    self._server_instances.pop(new_session_id, None)
+                    self._session_owners.pop(new_session_id, None)
+                    await http_transport.terminate()
         else:
             # Unknown or expired session ID - return 404 per MCP spec
             # TODO(L62): Align error code once spec clarifies
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -14,7 +14,7 @@ from mcp_types import DEFAULT_NEGOTIATED_VERSION, INVALID_REQUEST, ErrorData, JS
 from mcp_types.version import HANDSHAKE_PROTOCOL_VERSIONS
 from starlette.requests import Request
 from starlette.responses import Response
-from starlette.types import Receive, Scope, Send
+from starlette.types import Message, Receive, Scope, Send
 
 from mcp.server._streamable_http_modern import handle_modern_request
 from mcp.server.auth.middleware.bearer_auth import AuthenticatedUser, AuthorizationContext, authorization_context
--- a/src/mcp/server/streamable_http_manager.py
+++ b/src/mcp/server/streamable_http_manager.py
@@ -355,8 +355,30 @@ class StreamableHTTPSessionManager:
                 # Start the server task
                 await self._task_group.start(run_server)
 
+                # The session above is provisional: every validation (Host, Accept,
+                # Content-Type, JSON parse, JSON-RPC shape, "Missing session ID")
+                # lives inside `handle_request`, so it runs only now. If the request
+                # that was meant to establish this session is refused, the session
+                # must not survive it -- otherwise a rejected request leaves live
+                # state behind and hands the caller a usable session id.
+                establishing_status: int | None = None
+
+                async def send_tracking_status(message: Message) -> None:
+                    nonlocal establishing_status
+                    if message["type"] == "http.response.start":
+                        establishing_status = message["status"]
+                    await send(message)
+
                 # Handle the HTTP request and return the response
-                await http_transport.handle_request(scope, receive, send)
+                await http_transport.handle_request(scope, receive, send_tracking_status)
+
+                if establishing_status is not None and establishing_status >= 400:
+                    logger.debug(
+                        f"Discarding session {new_session_id}: establishing request returned {establishing_status}"
+                    )
+                    self._server_instances.pop(new_session_id, None)
+                    self._session_owners.pop(new_session_id, None)
+                    await http_transport.terminate()
         else:
             # Unknown or expired session ID - return 404 per MCP spec
             # TODO(L62): Align error code once spec clarifies
--- a/tests/server/test_streamable_http_manager.py
+++ b/tests/server/test_streamable_http_manager.py
@@ -3,14 +3,15 @@
 import json
 import logging
 from collections.abc import Iterator
-from typing import Any
+from typing import Any, Final
 from unittest.mock import AsyncMock, patch
 
 import anyio
 import httpx2
 import pytest
 from mcp_types import INVALID_REQUEST, ListToolsResult, PaginatedRequestParams
-from starlette.types import Message, Scope
+from mcp_types.version import HANDSHAKE_PROTOCOL_VERSIONS
+from starlette.types import Message, Receive, Scope, Send
 
 from mcp import Client
 from mcp.client.streamable_http import streamable_http_client
@@ -19,6 +20,19 @@ from mcp.server.auth.middleware.bearer_auth import AuthenticatedUser
 from mcp.server.auth.provider import AccessToken
 from mcp.server.streamable_http import MCP_SESSION_ID_HEADER, StreamableHTTPServerTransport
 from mcp.server.streamable_http_manager import DEFAULT_MAX_REQUEST_BODY_SIZE, StreamableHTTPSessionManager
+
+_INITIALIZE_BODY: Final[bytes] = json.dumps(
+    {
+        "jsonrpc": "2.0",
+        "id": 1,
+        "method": "initialize",
+        "params": {
+            "protocolVersion": HANDSHAKE_PROTOCOL_VERSIONS[-1],
+            "capabilities": {},
+            "clientInfo": {"name": "test-client", "version": "1.0"},
+        },
+    }
+).encode()
 
 
 @pytest.mark.anyio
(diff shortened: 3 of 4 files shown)


keeltrace:

I think the new cleanup still misses an establishing request that disconnects before any HTTP response starts.

`RequestBodyLimitMiddleware` deliberately replays a partial request followed by `http.disconnect` downstream (and this PR has tests for both disconnect-before-body and disconnect-mid-body). The transport then reaches:

```python
body = await request.body()
```

Starlette raises `ClientDisconnect` when `Request.stream()` receives that `http.disconnect`. That exception propagates out of `http_transport.handle_request(...)`.

The new cleanup is only reached *after* that await returns normally:

```python
await http_transport.handle_request(scope, receive, send_tracking_status)

if establishing_status is not None and establishing_status >= 400:
    ...
    self._server_instances.pop(new_session_id, None)
    await http_transport.terminate()
```

So in the disconnect case there is no `http.response.start`, `establishing_status` remains `None`, and execution never reaches the cleanup block at all. The provisional transport has already been inserted into `_server_instances` and its server task is already running, so with the default no-idle-timeout configuration this appears to leave another orphaned session — the same lifecycle invariant this PR is trying to restore, just on an exceptional/no-response exit rather than a `>= 400` response.

Could the provisional-session cleanup be in a `finally` (or otherwise cover "request exited before a successful establishing response"), and could the regression test exercise the manager end-to-end with a partial initialize body followed by `http.disconnect`, then assert the session registry is empty?

I reproduced the underlying Starlette behavior locally with Starlette 0.50.0: `Request.body()` raises `ClientDisconnect` for a partial `http.request` followed by `http.disconnect`. I did not run this repository's full test suite locally.

Checked against PR head `cb521ea`.

sainikhiljuluri:
Thanks for digging into this — that's a specific, checkable claim, so I ran it against PR head `cb521ea` before answering.

The Starlette half is exactly right: `Request.body()` does raise `ClientDisconnect` for a partial `http.request` followed by `http.disconnect`. Where it comes apart is the next step — that exception does **not** propagate out of `http_transport.handle_request(...)`. `_handle_post_request` has a blanket `except Exception` (`streamable_http.py:676`) that logs `"Error handling POST request"` and sends a 500. So a response *does* start, `establishing_status` becomes 500, and the cleanup block runs.

I wrote the end-to-end manager test you described, in both shapes:

```
partial initialize body -> http.disconnect
  handle_request raised: nothing
  response.start count: 1  status: [500]
  _server_instances = {}

http.disconnect as the first message
  handle_request raised: nothing
  response.start count: 1  status: [500]
  _server_instances = {}
```

Both pass on the current head. The traceback you predicted is there in the log — `ClientDisconnect` at `streamable_http.py:537` — it just gets converted into a response rather than escaping.

This is actually the case for keying the discard off the response status rather than an enumerated list of validation failures: 500 is not a validation refusal and nobody would have thought to enumerate it, but `>= 400` catches it for free. I'll add these two as regression tests so the disconnect path is pinned by name rather than covered by accident.

On the wider point — you're right that a `finally` would be strictly more robust, since `handle_request` itself has no wrapper and the cleanup is skipped for any exception that escapes it. I couldn't construct one: POST, GET and DELETE all send a response on every path I could reach. Happy to move it to a `finally` keyed on "no successful establishing response" if a maintainer would rather have the guarantee than the reachability argument.

Worth noting for anyone finding this thread: this PR is still a draft and I can't mark it ready — `markPullRequestReadyForReview` returns FORBIDDEN for this account on this repo. It needs a maintainer.

PR #3239, opened 2026-08-03
Expose create_mcp_http_client and McpHttpClientFactory from mcp.client.streamable_http

### LINKED ISSUE #3238: Expose create_mcp_http_client and McpHttpClientFactory as public API (2.0 made them private-only)
## Summary

Customizing the HTTP client for `streamable_http_client` (custom `headers`, `auth`, `timeout`, proxy, etc.) is a common and legitimate need, but as of 2.0.0 the only supported way to build a conforming client is through helpers that live in the **private** module `mcp.shared._httpx_utils`. Users are therefore forced to depend on a private API.

## What changed in 2.0

In 1.x, `streamable_http_client` accepted convenience kwargs directly:

```python
streamable_http_client(url, headers=..., timeout=..., sse_read_timeout=..., auth=...)
```

2.0 (via #2972, which replaced `httpx`/`httpx-sse` with `httpx2`) removed those kwargs. The signature is now:

```python
async def streamable_http_client(
    url: str,
    *,
    http_client: httpx2.AsyncClient | None = None,
    terminate_on_close: bool = True,
) -> ...
```

So the only way to pass custom headers/auth/timeout is to build an `httpx2.AsyncClient` yourself and pass it as `http_client=`. The standardized factory for doing so is `create_mcp_http_client` — but it is only available at the private path:

```python
from mcp.shared._httpx_utils import create_mcp_http_client  # private module
```

The same applies to `McpHttpClientFactory`: in 1.x it was importable from the public `mcp.client.streamable_http`, but in 2.0 it is defined in `mcp.shared._httpx_utils` and is no longer re-exported from any public module.

## Why this is a problem

- "Connect to an MCP server that requires auth headers / a custom timeout / a proxy" is a standard use case, not an edge case.
- The leading underscore on `mcp.shared._httpx_utils` signals "private, may change without notice", so every downstream project that needs a custom client has to take on that fragility.
- It's inconsistent: the *consumption* side (`streamable_http_client(url, http_client=...)`) is public, but the *construction* side (`create_mcp_http_client`, `McpHttpClientFactory`) is private.

## Suggestion

Re-export `create_mcp_http_client` and `McpHttpClientFactory` from a public module — e.g. `mcp.client.streamable_http` (where `McpHttpClientFactory` used to live) or `mcp.shared` — so building a custom HTTP client does not require importing a private module.

(Related: because the HTTP layer now uses `httpx2`, a short note in the migration docs on how to build/pass a custom `http_client` would also help.)

ProgrammerPlus1998:
## Summary

Implements the suggestion in #3238: re-export the HTTP-client helpers from a public module so that customizing the streamable-HTTP client (headers/auth/timeout) no longer requires importing the private `mcp.shared._httpx_utils`.

Since 2.0 removed the `headers`/`timeout`/`auth` kwargs from `streamable_http_client`, the only supported way to build a conforming `httpx2.AsyncClient` is `create_mcp_http_client` — but it, the `McpHttpClientFactory` protocol, and the default-timeout constants were only reachable via the private module.

## Changes

- **`src/mcp/client/streamable_http.py`** — re-export `create_mcp_http_client`, `McpHttpClientFactory`, `MCP_DEFAULT_TIMEOUT`, and `MCP_DEFAULT_SSE_READ_TIMEOUT` from the same module that already exposes `streamable_http_client`, and add a module-level `__all__` declaring the public surface. The redundant-alias / `__all__` forms keep this a pure re-export (no behavior change) and satisfy ruff's F401.
- **`tests/shared/test_httpx_utils.py`** — new `test_public_reexport_from_streamable_http` asserting the public names resolve to the same objects as the private ones.
- **`docs/migration.md`** — point the "build the http_client" guidance at the public import path, with an example.

## Notes

- This is intentionally additive: nothing is removed or renamed, the private module still works, and existing imports are unaffected. No external code uses `import *` from this module (tests import names explicitly), so adding `__all__` does not change any current consumer.
- I scoped the re-export to `mcp.client.streamable_http` (rather than a broader `mcp.shared`) because that's where `streamable_http_client` lives and where `McpHttpClientFactory` was importable from in 1.x — happy to also/instead export from elsewhere if you'd prefer a different home.

Fixes #3238

## Testing

- `tests/shared/test_httpx_utils.py` — 3 passed (2 pre-existing + 1 new).
- `ruff check` and `ruff format --check` clean on the changed files.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -2109,6 +2109,18 @@ v1's internal client set `follow_redirects=True`; set it explicitly when supplyi
 - `httpx_client_factory`: gone with no replacement — call your factory yourself and pass the result as `http_client`.
 - `terminate_on_close`: unchanged (default `True`).
 
+If you'd rather not hand-build the `httpx2.AsyncClient`, the standardized factory and its defaults are re-exported from the public `mcp.client.streamable_http` module, so you don't need to touch the private `mcp.shared._httpx_utils`:
+
+```python
+from mcp.client.streamable_http import create_mcp_http_client, streamable_http_client
+
+async with create_mcp_http_client(headers={"Authorization": "Bearer token"}) as http_client:
+    async with streamable_http_client(url="http://localhost:8000/mcp", http_client=http_client) as (read_stream, write_stream):
+        ...
+```
+
+`create_mcp_http_client(headers=..., timeout=..., auth=...)` applies the MCP defaults (`follow_redirects=True` and `Timeout(30, read=300)`) for you; the `McpHttpClientFactory` protocol and the `MCP_DEFAULT_TIMEOUT` / `MCP_DEFAULT_SSE_READ_TIMEOUT` constants are available from the same module.
+
 Client-side stream resumption is also unchanged: the transport reconnects a dropped GET stream with `Last-Event-ID` on its own, and `session.send_request(..., metadata=ClientMessageMetadata(resumption_token=..., on_resumption_token_update=...))` (from `mcp.shared.message`) works as in v1.
 
 ### `get_session_id` callback removed from `streamable_http_client`
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -33,10 +33,32 @@ from pydantic import ValidationError
 from mcp.client._transport import TransportStreams
 from mcp.shared._compat import resync_tracer
 from mcp.shared._context_streams import ContextReceiveStream, ContextSendStream, create_context_streams
-from mcp.shared._httpx_utils import create_mcp_http_client
+from mcp.shared._httpx_utils import (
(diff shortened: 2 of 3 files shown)


ProgrammerPlus1998:
fixed. I removed the module-level `__all__` and switched to redundant-alias re-exports (`X as X`), which keeps ruff's F401 satisfied while remaining purely additive.

Verified locally that the star-import surface is unchanged apart from the newly exported names: diffing `dir(m)` (no-underscore names) before vs. after shows nothing lost — `SessionMessageOrError`, `DEFAULT_RECONNECTION_DELAY_MS`, `CONNECTION_CLOSED`, etc. are all still exported, and `create_mcp_http_client` / `McpHttpClientFactory` / the two timeout constants are added.

PR #3246, opened 2026-08-04
fix: skip eager token refresh when OAuth metadata is unavailable (#3240)

### LINKED ISSUE #3240: OAuth token refresh hits the wrong endpoint when the auth server lives under a path
Ran into this with a hosted MCP server whose authorization server isn't at the origin root — token endpoint is `https://host/oauth2/api/v1/token`, not `https://host/token`.

If a client starts up with a cached-but-expired access token (+ refresh token) and hasn't done discovery yet, `async_auth_flow` refreshes at the very top — before any 401/metadata discovery. So `oauth_metadata is None` and `_refresh_token` uses the fallback `urljoin(get_authorization_base_url(server_url), "/token")`, i.e. just `{scheme}://{netloc}/token`. That 404s, `_handle_refresh_response` clears the tokens, and the flow drops to full interactive auth — which a headless/gateway client can't do. So the server silently disconnects every time the access token expires (mine are 5 min, so… constantly).

Same path-stripping fallback is in `_get_token_endpoint`, `_perform_authorization_code_grant` (`/authorize`) and DCR (`/register`) — refresh is just the one that bites silently.

Repro (roughly):
- MCP server whose AS metadata puts `token_endpoint` under a path, not `{origin}/token`
- log in normally so tokens get cached
- let the access token expire (or clear the expiry), reconnect with a fresh provider
- watch the refresh POST go to `https://host/token` → 404 → "Token refresh failed" → tokens cleared → it tries to open a browser

Fix looks like: discover metadata before the eager refresh (or stop dropping the issuer path in the fallback). Happy to PR — have a branch that pulls the PRM/ASM discovery out of the 401 branch and runs it before the refresh.

(used some AI help digging into this)

zsxh1990:
## Problem

When the authorization server lives under a non-root path (e.g. `https://host/oauth2/api/v1/token`), the eager token refresh at the top of `async_auth_flow` hits the wrong endpoint.

**Root cause**: `_refresh_token` falls back to `urljoin(get_authorization_base_url(server_url), "/token")` when `oauth_metadata` is `None`. `get_authorization_base_url` strips the path, so a server at `/oauth2/api/v1/token` gets `https://host/token` instead.

This happens because the eager refresh runs **before** any PRM/ASM metadata discovery — so `oauth_metadata` is always `None` on a restart with cached-but-expired tokens.

## Fix

Add `and self.context.oauth_metadata is not None` to the eager refresh guard. When metadata is missing:

1. Skip the eager refresh (we don't know the real token endpoint)
2. Let the request proceed with the stale token
3. Receive 401 → run full PRM/ASM discovery → learn the correct `token_endpoint`
4. Refresh or re-authorize with the right endpoint

This is correct because:
- **With metadata**: eager refresh works as before (we know the endpoint)
- **Without metadata + token valid**: no refresh needed, request succeeds
- **Without metadata + token expired**: skip → 401 → discovery → correct endpoint

Fixes #3240

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -586,8 +586,17 @@ class OAuthClientProvider(httpx2.Auth):
             # Capture protocol version from request headers
             self.context.protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
 
-            if not self.context.is_token_valid() and self.context.can_refresh_token():
-                # Try to refresh token
+            if (
+                not self.context.is_token_valid()
+                and self.context.can_refresh_token()
+                and self.context.oauth_metadata is not None
+            ):
+                # Try to refresh token — only when we already have OAuth metadata.
+                # Without metadata the token endpoint is unknown; the fallback
+                # urljoin(base_url, "/token") strips the path when the AS lives
+                # under a non-root path (e.g. /oauth2/api/v1/token).  Skipping
+                # the refresh here lets the request proceed with the stale token,
+                # receive a 401, and run full metadata discovery before retrying.
                 refresh_request = await self._refresh_token()
                 refresh_response = yield refresh_request
 
--- a/tests/client/test_auth.py
+++ b/tests/client/test_auth.py
@@ -1290,7 +1290,6 @@ class TestAuthFlow:
         assert oauth_provider.context.current_tokens.access_token == "new_access_token"
         assert oauth_provider.context.token_expiry_time is not None
 
-
     @pytest.mark.anyio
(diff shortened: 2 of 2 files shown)


zsxh1990:
Hi team — friendly ping on this PR 🙂

CI is fully green (29/29 checks pass) and Cubic AI reviewer found no issues. This fixes the OAuth token refresh hitting the wrong endpoint when the auth server lives under a non-root path (issue #3240).

Happy to address any feedback!

maxisbey:
Thanks for the PR. We're tracking this fix in #3263 instead, so I'm closing this one. Feel free to reopen if this is still relevant.

PR #3252, opened 2026-08-04
Release context.lock before the protected request is sent

### LINKED ISSUE #3209: Bug: OAuth provider holds context.lock for the whole request, so the standalone GET SSE stream stalls the first tools/call by ~15s
## Summary

With `auth=OAuthClientProvider` on a **streamable-HTTP** connection, the *first* `tools/call` of every session blocks for ~15 s. No error, no warning — just latency. Nothing is concurrent here: one server, one client, one session, sequential calls.

Root cause is the same `context.lock`-held-across-`yield` problem as #2847 / #2644, but the symptom is different enough that it seems worth recording separately: #2847 is about `RuntimeError: The current task is not holding this lock` under *concurrent* OAuth connections, while this is a silent latency tax paid by *every* OAuth streamable-HTTP user, concurrency or not. Both are fixed by #2858.

## Environment

- `mcp==1.29.0`, `anyio==4.13.0`, `httpx==0.28.1`
- Python 3.14.6, macOS arm64
- Server: FastMCP behind Google Cloud Run, streamable HTTP, OAuth 2.1 + PKCE

## Measurement

Same server, same tool, one fresh session per call, three calls per arm. The *only* variable is how the bearer token reaches the request:

```
auth=OAuthClientProvider   : 15.68s / 15.74s / 15.80s
static Authorization header:  0.70s /  0.62s /  0.67s
```

Raw HTTP JSON-RPC against the same endpoint (no SDK) answers the same tool in ~0.07 s, so the useful work is ~100 ms and effectively all of the 15.7 s is the wait.

Repro:

```python
import time, anyio

async def one(*, use_provider: bool) -> float:
    headers = None if use_provider else {"Authorization": f"Bearer {ACCESS_TOKEN}"}
    auth = build_provider() if use_provider else None   # OAuthClientProvider
    t0 = time.perf_counter()
    async with streamablehttp_client(URL, headers=headers, auth=auth) as (r, w, _):
        async with ClientSession(r, w) as session:
            await session.initialize()
            await session.call_tool("get_last_modified", {})
    return time.perf_counter() - t0
```

## Why it happens

1. `async_auth_flow` takes `context.lock` and holds it for the whole lifetime of the request, not just while it obtains a token — the `async with self.context.lock:` at `src/mcp/client/auth/oauth2.py:492` wraps `response = yield request`.
2. After `initialized`, `streamablehttp_client` opens the long-lived standalone GET SSE stream.
3. That GET goes through `async_auth_flow`, takes the lock, and does not release it until the stream ends — 15–60 s in our logs, dictated by the server's/proxy's idle timeout, not by the client.
4. The first `tools/call` therefore waits on the lock for as long as the GET stream lives.
5. Once the stream closes, subsequent calls run at ~100 ms until the next stream is opened.

So the stall is not a fixed 15 s: it is "however long the server keeps an idle SSE stream open", which for Cloud Run happens to be ~15 s and for other deployments in this thread ~60 s.

## The fix

#2858 (rebase of #2660 by @peisuke) already does the right thing: no HTTP `yield` — including the standalone GET SSE long-poll — runs while `context.lock` is held. Verified locally that with the lock out of the request path the same call is ~0.4–0.7 s.

Our workaround in the meantime is to keep `OAuthClientProvider` out of the request path entirely: read the token from storage and pass it as a plain `Authorization` header, refreshing it ourselves. That is a fine escape hatch for an application, but it means giving up the SDK's auth handling, which most users reasonably will not want to do.

Would a maintainer be able to look at #2858? It fixes a correctness bug for concurrent users and a ~15 s per-session latency bug for everyone else.

guptaishaan:
Fixes #3209

`async_auth_flow` held `context.lock` across `response = yield request`, so the lock covered the
whole round trip of the protected request rather than just token acquisition. The standalone GET
SSE stream runs through the same provider, so it pinned the lock for the lifetime of the stream
and the next request, typically the first `tools/call`, blocked in `lock.acquire()` until the
stream ended.

The lock now closes before the request is yielded and re-opens around the 401 and 403
re-authorization blocks. Token acquisition and refresh are still serialized; no protected request
is sent while the lock is held.

New test: `tests/client/test_auth.py::test_in_flight_request_does_not_block_a_concurrent_request`.
It drives two auth flows from two anyio tasks, holds the GET flow at its yield, and requires the
POST flow to reach its own yield inside `anyio.fail_after(5)`. Without the patch it fails with
`TimeoutError` parked on `async with self.context.lock`; with it, it passes in 0.26s.

Verified on Linux, CPU only, Python 3.13.14:

- `pytest tests/client/test_auth.py tests/client/auth tests/client/test_streamable_http.py` -> 197 passed, 1 xfailed
- `./scripts/test` -> 100.00% branch coverage, `strict-no-cover` clean
- `ruff format --check .`, `ruff check .`, `pyright` all clean

Not verified: the reporter's ~15s figure against a live server. I reproduced the mechanism, not
the deployment. The multi-second stall needs a server or proxy that does not flush the GET SSE
response headers promptly, which is the Cloud Run behaviour in the issue; against a transport that
returns headers immediately, httpx2 resumes the generator right away and the stall does not show
up. Only Python 3.13 on Linux, asyncio backend, was exercised locally.

One knowing behaviour change: two requests in flight at once can now both get a 401 and both run a
full authorization. Before, the second could not be sent until the first finished, which hid this.
#2858 handles that with a "did the token change while I was in flight?" check. I left it out to
keep this change to the reported bug, so it is worth a follow-up.

Thanks to @stayclosetothequestion for the report, the measurements, and for tracing it to the
exact line.

### DIFF
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -579,12 +579,14 @@ class OAuthClientProvider(httpx2.Auth):
 
     async def async_auth_flow(self, request: httpx2.Request) -> AsyncGenerator[httpx2.Request, httpx2.Response]:
         """httpx2 auth flow integration."""
+        # Capture protocol version from request headers
+        protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
+
         async with self.context.lock:
             if not self._initialized:
                 await self._initialize()
 
-            # Capture protocol version from request headers
-            self.context.protocol_version = request.headers.get(MCP_PROTOCOL_VERSION_HEADER)
+            self.context.protocol_version = protocol_version
 
             if not self.context.is_token_valid() and self.context.can_refresh_token():
                 # Try to refresh token
@@ -605,6 +607,10 @@ class OAuthClientProvider(httpx2.Auth):
 
         if response.status_code == 401:
             async with self.context.lock:
+                # Another request may have stamped its own version while this one was in
+                # flight; re-authorization has to use the version this request carried.
+                self.context.protocol_version = protocol_version
+
                 # Perform full OAuth flow
                 try:
                     # OAuth flow must be inline due to generator constraints
@@ -759,6 +765,8 @@ class OAuthClientProvider(httpx2.Auth):
             yield request
         elif response.status_code == 403:
             async with self.context.lock:
+                self.context.protocol_version = protocol_version
+
                 # Step 1: Extract error field from WWW-Authenticate header
                 error = extract_field_from_www_auth(response, "error")
 
--- a/src/mcp/client/auth/oauth2.py
+++ b/src/mcp/client/auth/oauth2.py
@@ -598,9 +598,13 @@ class OAuthClientProvider(httpx2.Auth):
             if self.context.is_token_valid():
                 self._add_auth_header(request)
 
-            response = yield request
+        # Released before the request goes out: the lock serialises token acquisition, and
+        # holding it for the lifetime of the response would stall every other request on
+        # this provider until the response ends - unbounded for the standalone GET SSE stream.
+        response = yield request
 
-            if response.status_code == 401:
+        if response.status_code == 401:
+            async with self.context.lock:
                 # Perform full OAuth flow
                 try:
                     # OAuth flow must be inline due to generator constraints
@@ -751,8 +755,10 @@ class OAuthClientProvider(httpx2.Auth):
 
                 # Retry with new tokens
                 self._add_auth_header(request)
-                yield request
-            elif response.status_code == 403:
+
+            yield request
+        elif response.status_code == 403:
+            async with self.context.lock:
                 # Step 1: Extract error field from WWW-Authenticate header
                 error = extract_field_from_www_auth(response, "error")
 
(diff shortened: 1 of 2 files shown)


guptaishaan:
Both valid, fixed in the follow-up commit.

The P2 is a regression this PR introduced. Before it, the lock was held across the
yield, so nothing could restamp `context.protocol_version` while a request was in
flight. Now a second request can, and the first one's 401/403 re-authorization would
use the wrong version and flip the `resource` parameter.

`async_auth_flow` now reads the header into a local once, and restamps
`self.context.protocol_version` from that local inside the 401 and 403 blocks after
the lock is re-acquired. Every read of `context.protocol_version`
(`_perform_authorization_code_grant`, `_exchange_token_authorization_code`,
`_refresh_token`) happens under the lock in one of those three regions, so this
closes the window without threading a new parameter through those methods.

New test `test_step_up_uses_the_protocol_version_of_its_own_request` holds a
`2025-06-18` request at its yield, runs a `2025-03-26` request through its initial
block, then 403s the first one and asserts the token exchange still carries
`resource=`. It fails on the previous commit and passes now.

P3: applied the `anyio.fail_after(5)` suggestion.

`./scripts/test` is clean at 100% branch coverage, and ruff and pyright pass.

PR #3259, opened 2026-08-06
fix(auth): preserve root protected resource URI

### LINKED ISSUE #2883: AnyHttpUrl normalization adds trailing slash → breaks RFC 9728 canonical resource URL in ProtectedResourceMetadata

daleselaji-dev:
## Problem

`ProtectedResourceMetadata.resource` can be serialized with a synthetic trailing slash when a root resource URL is supplied as an already-validated `AnyHttpUrl`. This changes `https://example.com` to `https://example.com/` and can break strict RFC 9728 resource matching.

## Root Cause

`AnyHttpUrl` normalizes an empty URL path to `/` before `ProtectedResourceMetadata`'s `url_preserve_empty_path` model setting can preserve the original spelling. The server auth settings pass an already-validated URL object into the metadata model.

## Solution

Normalize only the root path in a `ProtectedResourceMetadata` pre-validator, removing the synthetic slash before `AnyHttpUrl` validation. Non-root paths and URL validation remain unchanged.

## Changes

- Preserve an empty root path for `ProtectedResourceMetadata.resource`.
- Add regression coverage for raw metadata input and an already-normalized URL object.

## Testing

- `py -m pytest tests/shared/test_auth.py -q` — 43 passed.
- `py -m pytest tests/shared/test_auth_utils.py tests/shared/test_auth.py -q` — 58 passed.
- `py -m pytest tests/client/test_auth.py -q` — 140 passed, 1 xfailed.
- `py -m ruff format --check src/mcp/shared/auth.py tests/shared/test_auth.py` — passed.
- `py -m ruff check src/mcp/shared/auth.py tests/shared/test_auth.py` — passed.
- `git diff --check` — passed.

## Compatibility/Risk

The public field type remains `AnyHttpUrl`. Only a root-path slash synthesized by URL normalization is removed; non-root paths, query strings, and validation behavior are unchanged.

## Notes for Reviewer

Pyright could not be run with the repository's configured `.venv` because this incremental worktree does not contain that environment. A separate available environment reported no source-file diagnostics but reported unrelated baseline test typing diagnostics; this is explicitly unverified locally.

## Linked Issue

Closes #2883

### DIFF
--- a/src/mcp/shared/auth.py
+++ b/src/mcp/shared/auth.py
@@ -1,4 +1,5 @@
 from typing import Any, Literal, cast
+from urllib.parse import urlsplit, urlunsplit
 
 from pydantic import AnyHttpUrl, AnyUrl, BaseModel, ConfigDict, Field, field_validator, model_validator
 
@@ -256,3 +257,19 @@ class ProtectedResourceMetadata(BaseModel):
     dpop_signing_alg_values_supported: list[str] | None = None
     # dpop_bound_access_tokens_required default is False, but omitted here for clarity
     dpop_bound_access_tokens_required: bool | None = None
+
+    @field_validator("resource", mode="before")
+    @classmethod
+    def _preserve_empty_resource_path(cls, value: object) -> object:
+        """Keep the RFC 9728 root resource URI free of a synthetic slash.
+
+        ``AnyHttpUrl`` normalizes ``https://example.com`` to
+        ``https://example.com/`` before the model's ``url_preserve_empty_path``
+        setting can preserve the distinction. This is especially visible when
+        the value arrives as an already-validated ``AnyHttpUrl`` from the
+        server settings.
+        """
+        parsed = urlsplit(str(value))
+        if parsed.path == "/":
+            return urlunsplit(parsed._replace(path=""))
+        return value
--- a/tests/server/auth/test_protected_resource.py
+++ b/tests/server/auth/test_protected_resource.py
@@ -98,7 +98,7 @@ async def test_metadata_endpoint_without_path(root_resource_client: httpx2.Async
     assert response.status_code == 200
     assert response.json() == snapshot(
         {
-            "resource": "https://example.com/",
+            "resource": "https://example.com",
             "authorization_servers": ["https://auth.example.com/"],
             "scopes_supported": ["read"],
             "resource_name": "Root Resource",
--- a/tests/shared/test_auth.py
+++ b/tests/shared/test_auth.py
@@ -1,9 +1,15 @@
 """Tests for OAuth 2.0 shared code."""
 
 import pytest
-from pydantic import AnyUrl, ValidationError
+from pydantic import AnyHttpUrl, AnyUrl, ValidationError
 
-from mcp.shared.auth import InvalidRedirectUriError, OAuthClientInformationFull, OAuthClientMetadata, OAuthMetadata
+from mcp.shared.auth import (
+    InvalidRedirectUriError,
+    OAuthClientInformationFull,
+    OAuthClientMetadata,
+    OAuthMetadata,
+    ProtectedResourceMetadata,
+)
 
 
 def test_oauth():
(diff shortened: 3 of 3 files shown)


keeltrace:

I think the pre-validator cannot safely distinguish the synthetic-slash case from a legitimate explicit root resource, and it currently rewrites both.

At PR head `bbcc32b`, the new validator does:

```python
parsed = urlsplit(str(value))
if parsed.path == "/":
    return urlunsplit(parsed._replace(path=""))
```

That means raw input `resource="https://example.com/"` is changed to `https://example.com`, even though the caller explicitly supplied `/`. The new tests cover a path-less raw string and `AnyHttpUrl("https://example.com")`, but not an explicit-slash resource.

That distinction matters here because RFC 9728 §3.3 requires the returned `resource` value to be **identical** to the protected resource identifier used to derive the metadata URL; when metadata is discovered via `WWW-Authenticate`, it must also be identical to the URL the client used for the resource request. RFC 9728 §6 makes these string comparisons code-point equality. HTTP URI equivalence between an empty path and `/` does not make it safe to rewrite the OAuth resource identifier after the caller chose one spelling.

There is also an important current-state point: #2883 was already closed because #2925 added `url_preserve_empty_path=True` to `ProtectedResourceMetadata` and #2987 did the same to `AuthSettings`. This PR's own head already contains the `AuthSettings` fix, so a normal `resource_server_url="https://example.com"` string reaches the metadata model without the synthetic slash.

The remaining problematic input is an *already-created* `AnyHttpUrl` whose constructor has lost whether `/` was explicit or synthesized. Once that information is gone, this validator cannot recover it without guessing, and guessing “always remove `/`” changes a valid explicit identifier.

Could we either avoid rewriting explicit raw-string `/` values and constrain/document the already-normalized-object case, or leave the v2 string configuration path as-is rather than introducing an exact-match regression? A focused test for:

```python
ProtectedResourceMetadata(resource="https://example.com/", ...)
```

should preserve the trailing slash.

This is a static standards/code-path review; I did not run the repository suite locally.

PR #3276, opened 2026-08-10
fix(client): map bare HTTP 401/403 to distinguishable JSON-RPC errors

### LINKED ISSUE #1295: 401 in Streamable HTTP should be handled gracefully

sgoel2be24-cyber:
Bare auth failures were collapsing into the generic "Server returned an error response" fallback, so agents could not handle operation-specific denials without tearing down the session. Surface Unauthorized/Forbidden with http_status metadata instead.

Fixes #1295

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -366,6 +366,21 @@ class StreamableHTTPTransport:
                             error_data = ErrorData(code=METHOD_NOT_FOUND, message="Not Found")
                         else:
                             error_data = ErrorData(code=INVALID_REQUEST, message="Session terminated")
+                    elif response.status_code == 401:
+                        # Operation-specific auth denials must stay distinguishable so
+                        # agents can handle them (issue #1295) instead of collapsing into
+                        # an opaque "Server returned an error response".
+                        error_data = ErrorData(
+                            code=INTERNAL_ERROR,
+                            message="Unauthorized",
+                            data={"http_status": 401},
+                        )
+                    elif response.status_code == 403:
+                        error_data = ErrorData(
+                            code=INTERNAL_ERROR,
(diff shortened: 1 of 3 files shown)


sgoel2be24-cyber:
Thanks — noting the relationship to #3278.

#3278 correctly treats hang/teardown (#2110) as orthogonal to status-code distinguishability. This PR remains scoped to the residual #1295 ask: bare 401/403 should surface as Unauthorized/Forbidden (with http_status metadata) instead of the generic "Server returned an error response".

Happy to rebase onto the new `status_error_data()` helper as soon as #3278 lands, so the 401/403 mapping lives in that single place rather than inline in `_handle_post_request`.

PR #3318, opened 2026-08-16
fix(streamable-http): surface transport faults to callers and message_handler

functionxX:
Fixes #1401

## What changed

When a streamable-http response stream dies with a transport error (e.g.
`httpx.ReadTimeout` from `sse_read_timeout`), the original exception was
swallowed:

- The pending request was resolved with a generic `CONNECTION_CLOSED` error
  ("SSE stream ended without a response" / "reconnection attempts were
  exhausted"), hiding the root cause from user code.
- The fault never reached `message_handler`, even though the `IncomingMessage`
  contract says transport-level exceptions are delivered there.

Now:

- `_handle_sse_response` captures the stream exception and forwards it into the
  read stream as an `Exception` item, so the session's `message_handler`
  observes the fault (custom handlers, like the one the reporter built as a
  workaround, now receive it). The synthesized error keeps code
  `CONNECTION_CLOSED` and carries the exception in its message.
- `_handle_reconnection` carries the latest failure through the retry loop and
  includes it in the exhausted-budget error message.
- New `_forward_stream_fault` helper, best-effort like
  `_resolve_abandoned_request`.

## How verified

- Updated `test_a_non_resumable_sse_drop_resolves_the_request_with_an_error`
  and `test_exhausted_reconnection_attempts_resolve_the_request_with_an_error`
  to assert the fault item and the exception in the error message.
- Added `test_an_id_bearing_stream_that_dies_surfaces_the_fault_and_resolves_the_request`
  covering the reconnect-exhausted path end to end.
- `pytest tests/client`: 703 passed, 8 skipped, 1 xfailed; one unrelated
  pre-existing failure in `test_transport_stream_cleanup.py` (also fails on
  clean main on this machine). `ruff check`, `ruff format`, and `pyright` clean.

Disclosure: this contribution was written with AI assistance and personally reviewed and tested by me.

### DIFF
--- a/src/mcp/client/streamable_http.py
+++ b/src/mcp/client/streamable_http.py
@@ -553,7 +553,7 @@ class StreamableHTTPTransport:
                 # Stream ended again without response - reconnect again (reset attempt counter)
                 logger.info("SSE stream disconnected, reconnecting...")
                 await self._handle_reconnection(ctx, reconnect_last_event_id, reconnect_retry_ms, 0, last_exc=last_exc)
-        except Exception as e:  # pragma: no cover
+        except Exception as e:
             logger.debug(f"Reconnection failed: {e}")
             # Try to reconnect again if we still have an event ID
             await self._handle_reconnection(ctx, last_event_id, retry_interval_ms, attempt + 1, last_exc=e)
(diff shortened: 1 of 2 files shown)


functionxX:
The PR was auto-closed because #1401 hasn't been assigned yet — I've pinged for assignment on the issue. Once a maintainer assigns it, please reopen; the fix is complete and all 20 CI jobs were green. Happy to adjust scope if preferred.

PR #3413, opened 2026-08-29
Do not parameterize `RootModel` bases in generated types

Viicos:
When parameterized, a `RootModel` is usable directly as a type:

```python
MyInt = RootModel[int]

MyInt.model_validate('1')
#> 1
```

However, when used as a base, the model build logic ends up running for nothing, because everything is overridden by the actual class:

```python
class MyRoot(RootModel[int]):  # RootModel[int] is first built, then MyRoot
    root: int
```

By leaving models unparameterized, we gain some performance improvements on startup time: measured locally with `hyperfine --warmup 3 --runs 50 'PYDANTIC_DISABLE_PLUGINS=1 python -c "from mcp_types import _types, _v2025_11_25, _v2026_07_28"'`:

before: 180ms, after: 170ms

I don't know if implementing a Jinja template is the right approach (there's a bit of duplication), but looks like it will be robust enough.

## Types of changes

Performance

## Checklist

- [ ] I am assigned to the linked issue (or it is labeled `help wanted`, or I'm a maintainer)
- [x] I have disclosed any AI assistance and can explain the change in my own words


Viicos:
Obsoleted by https://github.com/modelcontextprotocol/python-sdk/pull/3470.

PR #3548, opened 2026-09-19
fix(client): treat any 2xx as successful session termination

### LINKED ISSUE #3546: Streamable HTTP client logs a WARNING for valid 202 Accepted on session termination (DELETE)
### Problem

`mcp/client/streamable_http.py` treats any `DELETE` session-termination response outside `{200, 204}` as a failure and logs:

```
WARNING mcp.client.streamable_http: Session termination failed: 202
```

But **`202 Accepted` is a valid, spec-compliant response for an asynchronous delete** — the server accepted the request and will process it later. The current code path:

```python
if response.status_code == 405:
    logger.debug("Server does not allow session termination")
elif response.status_code not in (200, 204):
    logger.warning(f"Session termination failed: {response.status_code}")
```

### Evidence

A long-running Hermes Agent (NousResearch) deployment with several Streamable HTTP MCP servers accumulated **72 `Session termination failed: 202` warnings in ~36 hours** in `errors.log` — pure noise from servers that correctly answer `202` to asynchronous termination. Real failures (connection errors) also surface through this same line, so the constant false positives bury the signal.

### Suggested fix

Treat 2xx as success on termination, e.g.:

```python
if response.status_code == 405:
    logger.debug("Server does not allow session termination")
elif 200 <= response.status_code < 300:
    pass  # 200/202/204 all acceptable
else:
    logger.warning(f"Session termination failed: {response.status_code}")
```

### Environment

- `mcp` Python SDK (client, streamable HTTP transport), observed on recent Hermes 0.21.x installs
- Servers: multiple Streamable HTTP MCP endpoints (remote + local)

Happy to send a PR if the direction is agreed.

deadczarvc:
Fixes #3546.

## Problem

`terminate_session` in `src/mcp/client/streamable_http.py` treats any `DELETE` response outside `(200, 204)` as a failure and logs `Session termination failed: 202`. `202 Accepted` is the correct response for a server that processes termination asynchronously; under HTTP semantics any 2xx is success, and the Streamable HTTP spec only carves out `405` for this request.

In a long-running Hermes Agent deployment with several Streamable HTTP MCP servers this produced 72 spurious warnings in ~36 hours in `errors.log`, burying real failures that surface through the same line.

## Change

One condition: `elif not 200 <= response.status_code < 300:`. The 405 debug path and the warning for genuine failures (4xx/5xx) are unchanged.

## Testing

Added `test_streamable_http_client_session_termination_status_handling` next to the existing termination tests in `tests/shared/test_streamable_http.py`, reusing the `AnswerDeleteWith204` wrapping-transport pattern: parametrized over `202` (no warning logged) and `500` (warning still logged). Verified red on unmodified `main` for the 202 case, green after the fix.

- `tests/shared/test_streamable_http.py`: 75 passed
- `ruff check` / `ruff format --check`: clean
- `pyright` on the touched module: 0 errors

Full suite not run locally (Windows env); CI will cover the rest.


deadczarvc:
Pushed c725630 addressing the CI coverage failures (tests themselves were green everywhere):

**1. `strict-no-cover` failure (all Python versions):**
`strict-no-cover` reported:
```
❎ 1 lines wrongly marked with 'pragma: no cover' are covered
  src/mcp/client/streamable_http.py:677
```
Line 677 (`logger.warning(f"Session termination failed: {response.status_code}")`) carried `# pragma: no cover` inherited from the pre-PR code, but the new parametrized test's `500` case deliberately executes it — so the pragma had become a false claim. Fix: removed the pragma. The warning line is now legitimately covered, and the `202` case covers the no-warning path.

**2. Coverage 99.99 < fail-under=100 on Python 3.14 (both `locked` and `lowest-direct`, linux and windows):**
```
tests/shared/test_streamable_http.py  ... 99.81%   1246->1250, 1246->1254
```
The only missing coverage was two partial branches inside the new test itself — the `async with` context-manager chain at line 1246. On Python 3.14 with anyio 4.10, `coverage` doesn't attribute CM-chain exits the way earlier versions do (sibling tests in this file mark the identical pattern with `# pragma: no branch`, e.g. the `ClientSession` line in `AnswerDeleteWith204`'s test and 29 other occurrences). Fix: marked the two CM lines of the new test with `# pragma: no branch`, matching the established repo convention.

Local verification (Python 3.13 venv): `tests/shared/test_streamable_http.py::test_streamable_http_client_session_termination_status_handling` 2/2 passed, warning line 677 executed and no longer excluded, `ruff check` / `ruff format --check` clean. The 3.14 partial-branch attribution is environment-specific and validated by CI on this push.

Note: the PR remains closed pending maintainer assignment to #3546 — pushing anyway per the bot's guidance so the fix is review-ready if the PR is reopened.

